Issue #13, opened 2021-06-09
MEASUREMENT response in SPDM 1.0 and 1.1 always includes a Nonce


jcahill0:
In the MEASUREMENT response in SPDM the structure always includes a Nonce for SPDM 1.0, 1.0.1, 1.1, and 1.1.1.  The requirement comes from the table "Successful MEASUREMENTS response message":
"8 + L Nonce 32 The Responder should choose a random value."

Yet the SPDM Requestor code that parses the MEASUREMENT response does not account for the Nonce in the non-GENERATE_SIGNATURE case.  Specifically in try_spdm_get_measurement() the if-block for generating a signature does contain code to account for the Nonce.

        	if (request_attribute ==
        	    SPDM_GET_MEASUREMENTS_REQUEST_ATTRIBUTES_GENERATE_SIGNATURE) {

But then the complementary else-statement both fails to account for the None and has a comment saying the Nonce is not there.

	} else {
        		//
        		// nonce is absent if there is not signature
        		//
        		if (spdm_response_size &lt;
        		    sizeof(spdm_measurements_response_t) +
        			    measurement_record_data_length + sizeof(uint16)) {
        			return RETURN_DEVICE_ERROR;
        		}
        		ptr = measurement_record_data + measurement_record_data_length;
        		opaque_length = *(uint16 *)ptr;
        		if (opaque_length &gt; MAX_SPDM_OPAQUE_DATA_SIZE) {
        			return RETURN_SECURITY_VIOLATION;
        		}
        		ptr += <span class="pl-k">sizeof</span>(uint16);

The statement about losing a Nonce if a signature is not being generated in SPDM 1.0, 1.0.1, 1.1, and 1.1.1 is only true in the requester structure's table "GET_MEASUREMENTS request message":
"4 Nonce 32 The Requester should choose a random value. This field is only present if a signature is required
on the response. See Table 23."

I haven't fully analyzed the SPDM Responder code for the MEASUREMENT response to see if it produces the Nonce field in both cases.

Unfortunately, I'm in a position where I won't be able to provide a code contribution for an unknown amount of time.  This is a pretty fundamental piece, so it may need to be fixed sooner than I can contribute.

The current draft of SPDM (maybe 1.2) does contain text to remove the Nonce field from the MEASUREMENT response if no signature is being generated.

jyao1:
We also need fix spdm-dump tool (https://github.com/DMTF/spdm-dump/tree/main/spdm_dump)

PR #42, opened 2021-07-15
Add Nonce field in the non-GENERATE_SIGNATURE case

liyi77:
Be tested at spdm-emu & spdm-dump


jyao1:
https://github.com/DMTF/libspdm/issues/13

liyi77:
sync unit-test code, all case passed

jyao1:
approved.

Issue #16, opened 2021-06-14
Calculate shared key according to RFC8446


jyao1:
Responder authentication multiple key exchange example
The DHE secrets are computed as specified in clause 7.4 of RFC 8446.

in RFC 8446, 7.4:
The negotiated key (Z) is converted to a
byte string by encoding in big-endian form and left-padded with zeros
up to the size of the prime.

steven-bellock:
Related issue in specification: https://github.com/DMTF/Security-TF/issues/1351

PR #45, opened 2021-07-18
enforce zero padding for shared key.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/16


jyao1:
Approved.

Issue #37, opened 2021-07-08
A dead code in the function spdm_check_response_flag_compability


yzhu52:
In the function spdm_check_response_flag_compability(), the if statement of Case3 in below is a dead code because it will be blocked by the Case1 (key_ex_cap != 0) and Case2 (key_ex_cap == 0),  we cannot enter into the return statement of Case3.

	case SPDM_MESSAGE_VERSION_11: {
                ......
		//Case1: Key_ex_cap set and encrypt_cap+mac_cap cleared
		if (key_ex_cap != 0 && (encrypt_cap == 0 && mac_cap == 0)) {
			return FALSE;
		}
                ......
		//Case2: Handshake_in_the_clear_cap set and key_ex_cap cleared
		if (handshake_in_the_clear_cap != 0 && key_ex_cap == 0) {
			return FALSE;
		}

		//Case3: Handshake_in_the_clear_cap set and encrypt_cap+mac_cap cleared
		if ((encrypt_cap == 0 && mac_cap == 0) &&
		    handshake_in_the_clear_cap != 0) {
			return FALSE;
		}
		......

yzhu52:
same issue also in spdm_check_request_flag_compability function.

PR #47, opened 2021-07-20
adjust the if statement order to resolve the dead code

yzhu52:
The details in https://github.com/DMTF/libspdm/issues/37

Signed-off-by: Yonghong Zhu <yonghong.zhu@intel.com>


jyao1:
I dont understand the fix. If it is dead code, why not remove it?

yzhu52:
I will update the patch to directly remove those dead code.

PR #52, opened 2021-07-23
Adjust message appending flow

### LINKED ISSUE #24: Error handling for CHALLENGE_AUTH signature generation
From https://github.com/jyao1/openspdm/issues/182

The spec requires the message concatenation performed only after a successful completion response.

258 Concatenate() is the standard concatenation function that is performed only after a successful
completion response on the entire request and response contents.

265 ◦ If a response contains an ErrorCode other than ResponseNotReady :
266 No concatenation function is performed on the contents of both the original request and
response.

But the behavior of the current Responder is typically:

SpdmGetResponseFoo()
{
1. validate request
2. append request into managed buffer
3. generate response
4. append response into managed buffer
}
This model didn't handle the scenario when some errors occur after 2), the request shouldn't be appended into the buffer.
It might be better to change the behavior as:

SpdmGetResponseFoo()
{
1. validate request
2. generate response
3. if everything is okay, concatenate request and response and append it into the managed buffer together.
}
(Note that lower layer failures, e.g. errors in MCTP or SMBus causing the response not sent to the requester, are ignored here.)

The current responder code for MEASUREMENT signature generation in the other hand implemented the similar requirement.

SpdmGetResponseMeasurement()
{
1. validate request
2. append request into managed buffer
3. generate response
3-1. reset the buffer if error X occurs
3-2. reset the buffer if error Y occurs
... and so on.
}
But it should be simpler to also apply the proposed model to simplify the flow.

liyi77:
Fix #24
And fix derivation issues of bug24:
1. Remove useless shrink function using.
2. Unit-test should not assume there is any data in meesage buffer if response fail.

### DIFF
--- a/library/spdm_requester_lib/challenge.c
+++ b/library/spdm_requester_lib/challenge.c
@@ -99,15 +99,6 @@ return_status try_spdm_challenge(IN void *context, IN uint8 slot_id,
 		return RETURN_DEVICE_ERROR;
 	}
 
-	//
-	// Cache data
-	//
-	status = spdm_append_message_c(spdm_context, &spdm_request,
-				       sizeof(spdm_request));
-	if (RETURN_ERROR(status)) {
-		return RETURN_SECURITY_VIOLATION;
-	}
-
 	spdm_response_size = sizeof(spdm_response);
 	zero_mem(&spdm_response, sizeof(spdm_response));
 	status = spdm_receive_spdm_response(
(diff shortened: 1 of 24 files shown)


jyao1:
reviewed-by: Jiewen Yao <jiewen.yao@intel.com>

steven-bellock:
Looks good to me.

PR #60, opened 2021-07-27
Correcting argument description on spdm_bin_concat.

### LINKED ISSUE #57: Fix spdm_bin_concat description
Currently, the function description for spdm_bin_concat in library/spdm_secured_message_lib/session.c mentions that the parameter label_size should include the NULL string terminator, when describing the size of the string pointed to by "label".

However, the implementation of the function does not expect a NULL terminator, and usage of this function shows that the NULL terminator is not included in label_size arguments.

This description should be updated, or the function itself updated.

taprinz:
Fixes #57

Updates comment to provide accurate description on size of string expected for spdm_bin_concat.


steven-bellock:
Approved.

PR #61, opened 2021-07-29
Clear up for GET_VERSION/VERSION

### LINKED ISSUE #28: Need more clear up for GET_VERSION/VERSION
From https://github.com/jyao1/openspdm/issues/170

For example: all session states, negotiated algo, etc,

liyi77:
Fix #28

### DIFF
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -272,6 +272,15 @@ void spdm_set_last_spdm_error_struct(IN void *spdm_context,
 void spdm_init_context(IN void *spdm_context);
 
 /**
+  Reset an SPDM context.
+
+  The size in bytes of the spdm_context can be returned by spdm_get_context_size.
+
+  @param  spdm_context                  A pointer to the SPDM context.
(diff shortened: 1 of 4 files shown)


jyao1:
why not call spdm_init_context()?

liyi77:
> spdm_init_context

I don’t want to clear information that is irrelevant to last connection, such like transport layer func, local version, local capability, etc. 
Especially local cert chain provision, I think that it is hard to retrieve if be cleaned up in get_version phase.

jyao1:
I see. Agree. Then how about we create another function - spdm_reset_context()  and move all code there ?

I really dont want to add so many code in two places.

liyi77:
> I see. Agree. Then how about we create another function - spdm_reset_context() and move all code there ?
> 
> I really dont want to add so many code in two places.

Reasonable, agree

jyao1:
reviewed-by: Jiewen Yao <jiewen.yao@intel.com>

steven-bellock:
Looks good to me.

Issue #22, opened 2021-07-07
record transcript hash, instead of transcript itself 


jyao1:
From https://github.com/jyao1/openspdm/issues/184

Current code keeps the whole transcript and calculate hash finally. That means we need big memory to hold the transcript.
The better idea is to just keep calculating the hash. Then we just need memory to hold the hash context.

jyao1:
@steven-bellock : may use a compiler option to decide to transcript or hash. It also helps to debug.

Agree.

PR #72, opened 2021-08-05
Add transcript hash feature

liyi77:
Just send for main structure review.
Just test at emu tool currently because of most related unittest case will be effected by #52.
I will sync unittest and re-send request after #52 be fixed if main structure of this is fine.


liyi77:
#22

liyi77:
Dump tool cannot decode securedSPDM if message_f and message_k use transcript hash:
...
88 (1628146535) PCI_DOE(1, 2) RSP->REQ SecuredSPDM(0xffffffff) SPDM(11, 0x67) SPDM_PSK_FINISH_RSP () 
89 (1628146535) PCI_DOE(1, 2) (?)->(?) SecuredSPDM(0xffffffff) <Unknown> 
90 (1628146535) PCI_DOE(1, 2) (?)->(?) SecuredSPDM(0xffffffff) <Unknown> 
91 (1628146535) PCI_DOE(1, 2) (?)->(?) SecuredSPDM(0xffffffff) <Unknown> 
...

 I don’t have any inspiration for this question at the moment.

jyao1:
Can we move append_hash_managed_buffer() inside of append_managed_buffer() ?

=================
	return append_managed_buffer(&spdm_context->transcript.message_mut_c,
				     message, message_size);
	#ifdef USE_TRANSCRIPT_HASH
		return append_hash_managed_buffer(spdm_context,
							&spdm_context->transcript.message_mut_c,
							message, message_size);
	#else
		return append_managed_buffer(&spdm_context->transcript.message_mut_c,
				     	message, message_size);
	#endif
=================

liyi77:
> Can we move append_hash_managed_buffer() inside of append_managed_buffer() ?

Another buffer that dont store hash will use append_managed_buffer(), such like messageA, certchain verify, hmac ...
I think that can be distinguished by buffer size(if the size of these two kinds buffer are always different in future),
or use a flag in buffer type.

This makes me think about that should managed buffers used in calculation of certchain verify, hmac... store hash also?
If sure, only the case of messageA needs to be distinguished.

liyi77:
Optimized issue-solving method

PR #77, opened 2021-08-10
fix get_digests verification

### LINKED ISSUE #64: question of get_digests verification
In spec, responder will return digests of its local certchains one by one when received a valid get_digests request,
and requester should compare cached digests to digests in response(spec 214 and 234).
image
In libspdm, this implemented by compare digest of local peer_cert_chain_provision to digests in response.
But now it looks like that treated all digests as a whole and finally only compared first digest in response.
Is it a bug?
https://github.com/DMTF/libspdm/blob/ad088c17ca2d7aa4ef36b07be7e805dd13f7a793/library/spdm_requester_lib/get_digests.c#L150-L157

https://github.com/DMTF/libspdm/blob/ad088c17ca2d7aa4ef36b07be7e805dd13f7a793/library/spdm_common_lib/crypto_service.c#L361-L378

liyi77:
Fix #64
Requester should compare digests in response to local cached digests one by one.

### DIFF
--- a/library/spdm_common_lib/crypto_service.c
+++ b/library/spdm_common_lib/crypto_service.c
@@ -345,18 +345,20 @@ boolean spdm_generate_cert_chain_hash(IN spdm_context_t *spdm_context,
 
   @param  spdm_context                  A pointer to the SPDM context.
   @param  digest                       The digest data buffer.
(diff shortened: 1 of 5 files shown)


jyao1:
Looks good.

steven-bellock:
Looks good to me.

PR #82, opened 2021-08-11
Enhancement of some unit test of version, capability

yzhu52:


jyao1:
I recommend to add new unit test, instead of change the old one.

yzhu52:
Got you.

steven-bellock:
Closing this PR. Per @jyao1 we should have a new commit with the new unit test.

PR #83, opened 2021-08-12
Add transcript hash feature

### LINKED ISSUE #22: record transcript hash, instead of transcript itself

liyi77:
FIX #22


liyi77:
1. All message buffer(size 0x100 or 0x1200 ) replace with hash buffer (0x80).
2. Added the compilation option whether to use hash mode, such as: 
cmake -G"NMake Makefiles" -DARCH=ia32 -DTOOLCHAIN=VS2019 -DTARGET=Debug -DCRYPTO=openssl **-DUSETH=1** ..

jyao1:
The fix is wrong. We need use hash_init(), hash_update(), and hash_finish() to extend the digest in TRANSCRIPT_HASH branch.

jyao1:
Also, I dont think there is need to add MACRO in CMAKE, just add MACRO in spdm_config.h is good enough.

PR #89, opened 2021-08-17
Fix the situation that the first returned cert is NOT root cert

### LINKED ISSUE #70: Need handle the situation that the first cert in returned cert chain is NOT root cert
Current spdm_verify_peer_cert_chain_buffer() assumes the first cert is root cert.

liyi77:
Fix #70
spdm_verify_certificate_chain_buffer：
	No longer assume that the first certificate is the root certificate.
spdm_verify_peer_cert_chain_buffer:
	Add handler of the situation that the first returned cert is NOT root cert.
Add unit-test case.


jyao1:
I also recommend to change peer_cert_chain_provision to peer_root_cert_provision.

liyi77:
> I also recommend to change peer_cert_chain_provision to peer_root_cert_provision.

Got it

PR #90, opened 2021-08-17
Adding new unit test cases for the finish message.

vinmaciel:
This commit adds new unit test cases for the FINISH/FINISH_RSP messages.

On the requester side:
- no capabilities for key exchange
- incomplete state (up to negotiate_algorithms)
- wrong response code
- session not initiated
- with mutual authentication (success case)
- wrong MAC (all-zero and arbitrary)
- wrong MAC size (double and half sizes)

On the responder side:
- with mutual authentication (success case)
- no capabilities for key exchange
- session not initiated
- incorrect MAC (all-zero and arbitrary)
- wrong MAC size (double and half sizes)
- wrong requester signature (all-zero and arbitrary)

They were tested for x64 on Ubuntu (GCC), and for ia32 and x64 on Windows (VS2019), for debug and release.
We were unable to compile libspdm for ia32 in a x64 Ubuntu machine as we were used to with openspdm. We are still working on how to make these tests in the future. Any suggestions?

### DIFF
--- a/unit_test/test_spdm_requester/finish.c
+++ b/unit_test/test_spdm_requester/finish.c
@@ -93,6 +93,51 @@ return_status spdm_requester_finish_test_send_message(IN void *spdm_context,
 		copy_mem(m_local_buffer, &ptr[1], request_size - 1);
 		m_local_buffer_size += (request_size - 1);
 		return RETURN_SUCCESS;
+	case 0xC:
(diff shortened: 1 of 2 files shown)


jyao1:
looks good to me.

steven-bellock:
Looks good to me.

PR #96, opened 2021-08-22
Enable unit testing in workflow.

steven-bellock:
test_spdm_requester and test_spdm_responder now return the test status as an exit code. cmake.yml runs the two tests as part of the build process.


jyao1:
looks good

PR #97, opened 2021-08-23
spdm_x509_certificate_check return value bug

### LINKED ISSUE #87: spdm_x509_certificate_check return value
When check issuer_name and subject_name in spdm_x509_certificate_check:
https://github.com/DMTF/libspdm/blob/8daea2f4f61673ca39ef8a10100917c483e1125c/library/spdm_crypt_lib/crypt.c#L1664-L1673
Function will go to end if status==true and len(issuer_name)==0, I think this means that issuername does not exist and should return wrong status. So maybe should set status false then go to end.

Something similar happened in x509_get_extended_key_usage case:
https://github.com/DMTF/libspdm/blob/8daea2f4f61673ca39ef8a10100917c483e1125c/library/spdm_crypt_lib/crypt.c#L1710-L1715

liyi77:
Fix #87

### DIFF
--- a/include/hal/library/cryptlib.h
+++ b/include/hal/library/cryptlib.h
@@ -2202,24 +2202,24 @@ return_status x509_get_serial_number(IN const uint8 *cert, IN uintn cert_size,
   Retrieve the issuer bytes from one X.509 certificate.
 
   If cert is NULL, then return FALSE.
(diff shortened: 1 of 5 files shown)


jyao1:
looks good to me.

PR #105, opened 2021-08-27
Add version negotiation feature

### LINKED ISSUE #48: The Requester shall use the selected version in all future communication of other requests.
After GET_VERSION, the requester need decide which version is final selected and always use it.

After GET_CAPABILITY, the responder will know this is the version selected by the requester and always stick it it.

Current implementation just check all supported versions, which is too relax.

liyi77:
Fix #48


jyao1:
looks good to me

steven-bellock:
Looks good to me.

PR #106, opened 2021-08-27
Fix incorrect return value check on spdm_create_measurement_signature.

### LINKED ISSUE #80: Incorrectly checked return value on spdm_create_measurement_signature
In measurements.c, we call the function spdm_create_measurement_signature, and check its return value with RETURN_STATUS(x) macro.

https://github.com/DMTF/libspdm/blob/8daea2f4f61673ca39ef8a10100917c483e1125c/library/spdm_responder_lib/measurements.c#L517-L520

However the function returns boolean, and the return status macro casts the boolean to an int, to check for negative number. This can never be the case. Therefore, the return value of this is effectively ignored.

The function should be changed to either returning a return_status, or the error checking should be changed.

taprinz:
Fixes #80 

spdm_create_measurement_signature returns boolean, while the existing code checks for negative value as failure return status. This check will never be true, so we can quietly fail spdm_create_measurement_signature and continue on.

Fixing return value check to ensure we catch failure.


jyao1:
@taprinz "merge commit" is not allowed. Please always use "rebase and merge".

Please remove below:
Merge pull request #1 from taprinz/taprinz_check_retval …
Merge branch 'DMTF:main' into main
Merge branch 'DMTF:main' into main

jyao1:
@taprinz please also remove unnecessary revert
Revert "fixing incorrect return value check" …

I see only one commit is really needed. Please rebase and update the patch series to only include this valid one.

taprinz:
Closing. Will create new pull request with valid commit history.

PR #112, opened 2021-08-31
Fix88 zeroized handshake key

### LINKED ISSUE #88: session handshake key should be zeroized after handshake phase.
The SpdmSession key should be zeroized, once it is not used.

For example, the DHE shared secret should be zeroized, once the handshake keys are generated.

The handshake keys should be zeroized, once the application keys are generated.

liyi77:
Fix #88


jyao1:
please use handshake instead of hand_shake.

liyi77:
> please use handshake instead of hand_shake.

@jyao1 fixed it.

jyao1:
looks good to me

steven-bellock:
Looks good to me.

PR #113, opened 2021-08-31
Array compared to null bug

### LINKED ISSUE #51: Array compared to null
There are multiple instances of

`if (spdm_context->local_context.local_cert_chain_provision == NULL)`

that produce the warning from Coverity

> CID 344411 (#1 of 1): Array compared against 0 (NO_EFFECT)
array_null: Comparing an array to null is not useful: spdm_context->local_context.local_cert_chain_provision == NULL, since the test will always evaluate as true.

liyi77:
local_cert_chain_provision is an array used to store cert chain pointers,
should compare the member of array instead of array-self.
If slot_id is valid but local_cert_chain_provision[slot_id] is NULL,
should report internal error with SPDM_ERROR_CODE_UNSPECIFIED.

Fix #51

### DIFF
--- a/library/spdm_requester_lib/encap_certificate.c
+++ b/library/spdm_requester_lib/encap_certificate.c
@@ -57,19 +57,20 @@ return_status spdm_get_encap_response_certificate(IN void *context,
 		return RETURN_SUCCESS;
 	}
 
-	if (spdm_context->local_context.local_cert_chain_provision == NULL) {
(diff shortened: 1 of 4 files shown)


jyao1:
looks good to me

steven-bellock:
Looks good to me, but it also inspired https://github.com/DMTF/libspdm/issues/115

PR #122, opened 2021-09-16
Fix for uninitialized variables error on warning

rkongintel:
When compiling with more restrictive options, such as force error on warning, the compiler will think these two variables (session_info and return_status) may be used before initialized.  Fix this issue by always initializing the variables.


steven-bellock:
Looks good to me.

As an aside, this warning gets triggered when maybe-uninitialized is enabled. Perhaps we should add it to the Makefile?

Issue #125, opened 2021-09-19
Authentication fail because message_a reordered.


jyao1:
root cause: 
In get_version.c, spdm_append_message_a() is after spdm_negotiate_connection_version().

However, spdm_negotiate_connection_version() will change the spdm_response.version_number_entry, which causes the message_a in requester is different with the one in responder.

PR #126, opened 2021-09-19
Fix version sort issue in version negotiation in get_version request.

jyao1:
REF: https://github.com/DMTF/libspdm/issues/125


steven-bellock:
Looks good to me.

PR #136, opened 2021-09-22
Fix error handling for CHALLENGE_AUTH signature generation

### LINKED ISSUE #99: Error handling for CHALLENGE_AUTH signature generation
This is a follow-up issue of #24 .

1. Before returning Error response, we should also reset message_C buffer.
https://github.com/DMTF/libspdm/blob/8daea2f4f61673ca39ef8a10100917c483e1125c/library/spdm_responder_lib/challenge_auth.c#L202-L208

2. For the Responder, the error handling of appending response should also remove the previously appended request message.

liyi77:
Fix #99
Should reset message_c before return error.


jyao1:
Looks good

steven-bellock:
Looks good to me.

Issue #137, opened 2021-09-22
need return nonce in GetMeasurement and Challenge to the caller


jyao1:
The caller may need evaluate the nonce to see if there is any duplication.

steven-bellock:
Is this to detect a faulty random number generator?

jyao1:
Right. The detection is out of scope of libspdm. So we need provide a way to let caller detect, if the caller wants to.

steven-bellock:
Right, it is out of scope of libspdm but I think the cryptography module should handle it, not the caller. For example FIPS 140 says

> Continuous random number generator test. If a cryptographic module employs Approved or nonApproved RNGs in an >Approved mode of operation, the module shall perform the following continuous
> random number generator test on each RNG that tests for failure to a constant value.
> 1. If each call to a RNG produces blocks of n bits (where n > 15), the first n-bit block generated
> after power-up, initialization, or reset shall not be used, but shall be saved for comparison with
> the next n-bit block to be generated. Each subsequent generation of an n-bit block shall be
> compared with the previously generated block. The test shall fail if any two compared n-bit
> blocks are equal.
> 2. If each call to a RNG produces fewer than 16 bits, the first n bits generated after power-up,
> initialization, or reset (for some n > 15) shall not be used, but shall be saved for comparison with
> the next n generated bits. Each subsequent generation of n bits shall be compared with the
> previously generated n bits. The test fails if any two compared n-bit sequences are equal.

jyao1:
This is more like attestation. 
The cryptography module need guarantee. And the remote verifier may check:
1) If the nonce from device is random.
2) If the nonce from local environment is random.

steven-bellock:
Two comments.

- From the verifier's point of view the responder's nonce is immaterial to the overall security of Measurements and Challenge. The responder's nonce is meant to protect the responder and, with the cryptography algorithms that SPDM supports, probably contributes very little to that protection.
- This is part of a larger topic on the interplay between verifier and requester and where the verifier and requester reside within the overall system, and how libspdm can support the different configurations. I can raise that as a topic at the next meeting.

steven-bellock:
@raghuncstate also has opinions on this. :)

raghuncstate:
https://github.com/DMTF/Security-TF/issues/1438 discussed the issue. Steven's first point is correct. The responder nonce makes very little difference from a security point of view.

[...]

PR #139, opened 2021-09-22
need return nonce in GetMeasurement and Challenge to the caller

jyao1:
The caller may need evaluate the nonce to see if there is any duplication.

REF: https://github.com/DMTF/libspdm/issues/137


steven-bellock (changes requested):
Per the meeting today, please modify spdm_challenge_ex and spdm_get_measurement_ex to take the nonce value as an input. Also need to include key exchange.

steven-bellock (approved):

Issue #138, opened 2021-09-22
need return trust anchor in GetCertificate to the caller


jyao1:
The caller may need to know which trust anchor is used to pass the certificate verification.

PR #140, opened 2021-09-22
need return trust anchor in GetCertificate to the caller

jyao1:
The caller may need to know which trust anchor is used to pass the certificate verification.

REF: https://github.com/DMTF/libspdm/issues/138


steven-bellock:
Looks good to me.

Issue #92, opened 2021-08-18
certificate chain in transcript hash should include the CERT_CHAIN header returned in GET_CERTIFICATE


jyao1:
Current implement only includes from root to leaf certificate. Instead, spdm_generate_cert_chain_hash() should be used.

steven-bellock:
Issue on the specification side: https://github.com/DMTF/Security-TF/issues/1443

liyi77:
> Current implement only includes from root to leaf certificate. Instead, spdm_generate_cert_chain_hash() should be used.

When it is time to calculate the transcript, it is difficult to get the slot_id in current code, so it is inconvenient to use spdm_generate_cert_chain_hash() .
I recommend that use spdm_get_local_cert_chain_buffer() instead of spdm_get_local_cert_chain_data() to fix this issue.

jyao1:
Currently, hash in GET_DIGEST == hash of cert chai in GET_CERTIFICATE == hash in CHALLENGE_AUTH.
For remote, we can cache the hash of cert chain from peer directly, instead of recalculate.
For local, yes, or you can cache the local hash of cert chain too.

jyao1:
> > Current implement only includes from root to leaf certificate. Instead, spdm_generate_cert_chain_hash() should be used.
> 
> When it is time to calculate the transcript, it is difficult to get the slot_id in current code, so it is inconvenient to use spdm_generate_cert_chain_hash() .
> I recommend that use spdm_get_local_cert_chain_buffer() instead of spdm_get_local_cert_chain_data() to fix this issue.

Yes, you are right. I will use spdm_get_local_cert_chain_buffer()

PR #146, opened 2021-09-23
Include CERT_CHAIN header in the transcript hash.

jyao1:
REF: https://github.com/DMTF/libspdm/issues/92


steven-bellock (approved):

steven-bellock:
Looks good to me.

Issue #124, opened 2021-09-18
unit test fail with mbedtls


jyao1:
[  ERROR   ] --- 0x800000000000001a != 0x0
[   LINE   ] --- C:\home\DMTF\libspdm\unit_test\test_spdm_requester\get_certificate.c:2046: error: Failure!

Finally failed at mbedtls: x509_crt.c:

        /* Look for a parent in trusted CAs or up the chain */
        ret = x509_crt_find_parent( child, trust_ca, &parent,
                                       &parent_is_trusted, &signature_is_good,
                                       ver_chain->len - 1, self_cnt, rs_ctx );

        /* No parent? We're done here */
        if( parent == NULL )
        {
            *flags |= MBEDTLS_X509_BADCERT_NOT_TRUSTED;
            return( 0 );
        }

rkongintel:
I've noticed a bunch of failures when building with the following option as well: x64, VS201,Release, openssl

steven-bellock:
Also need to include Mbed TLS to the CI.

rkongintel:
Here are the errors I see with x64,VS2019, Debug, openssl:

C:\code\dmtf_libspdm\build\bin>git log -1
commit df3db1ba161f4164d5aca21dd394a940411b156b (HEAD -> main, origin/main, origin/HEAD)
Author: yi1 li <yi1.li@intel.com>
Date:   Thu Sep 2 14:31:41 2021 +0800

    Add test case about first returned cert is not root.

    Signed-off-by: yi1 li <yi1.li@intel.com>

C:\code\dmtf_libspdm\build\bin>git status
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean

C:\code\dmtf_libspdm\build\bin>test_spdm_responder.exe > NULL
[  PASSED  ] 6 test(s).
[  PASSED  ] 22 test(s).
[  PASSED  ] 19 test(s).
[  PASSED  ] 9 test(s).
[  ERROR   ] --- 0x4 != 0x408
[   LINE   ] --- C:\code\dmtf_libspdm\unit_test\test_spdm_responder\certificate.c:78: error: Failure!
[  ERROR   ] --- 0x4 != 0x8
[   LINE   ] --- C:\code\dmtf_libspdm\unit_test\test_spdm_responder\certificate.c:391: error: Failure!
[  ERROR   ] --- 0x5 != 0x1
[...]

PR #149, opened 2021-09-23
Skip the first cert when verifying integrity

liyi77:
REF: #124
Skip the first certificate of a certificate chain without root certificate when verifying integrity.


liyi77:
For openssl, one cert can verify itself even that it is not a self-signed cert, but this is an error case for mbedtls.
I think we should skip the first cert for the certificate chain without root certificate.

jyao1:
Looks good to me

steven-bellock:
Looks good to me.

PR #152, opened 2021-09-23
Add conditional compile switches for CERT and CHAL capabilities.

rkongintel:
libspdm/issues/2.

### DIFF
--- a/include/industry_standard/spdm.h
+++ b/include/industry_standard/spdm.h
@@ -330,8 +330,6 @@ typedef struct {
 #define SPDM_REGISTRY_ID_CXL 7
 #define SPDM_REGISTRY_ID_JEDEC 8
 
-
-#if SPDM_ENABLE_CAPABILITY_CERT_CAP
 ///
 /// SPDM GET_DIGESTS request
 ///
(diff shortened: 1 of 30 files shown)


jyao1:
I don't recommend to use SPDM_ENABLE_*_CAP in spdm.h.
The spdm.h just follows SPDM spec, and SPDM spec does not have such concept.

All rest seems good to me.

rkongintel:
@jyao1 Are you saying I should leave the request/response structures in spdm.h always visibly defined to the compiler (not #if'd out)? I was doing that to ensure we don't miss any code that we should be disabling. Any idea how else I could make sure all code that should be disabled, actually is disabled?

jyao1:
> @jyao1 Are you saying I should leave the request/response structures in spdm.h always visibly defined to the compiler (not #if'd out)? I was doing that to ensure we don't miss any code that we should be disabling. Any idea how else I could make sure all code that should be disabled, actually is disabled?

Right. That is my idea. The original purpose of spdm_lib_config.h is to configure the spdm_lib, but not the spec.

I am worried, because that enforces people to include spdm_lib_config.h to properly configure the specification. E.g. spdm-dump, or UEFI consumer.  That is extra complexity.

I can understand that your requirement seems to have a way to configure the specification. As such, no definition will be appeared. I think it is valid request. Capability is one example, I feel SPDM version is another example in the future.

If so, I am thinking to have a separate .h to perform that, such as spdm_spec_config.h. Worth discussing.

jyao1:
Is there any size estimation after we use these flags?

rkongintel:
I did not collect any measurements. Figured I would after we did our changes for all the capabilities.

PR #154, opened 2021-09-23
Fix for set but unused variable "count".

rkongintel:
Causes compile error on linux build.


steven-bellock:
Looks good to me.

jyao1:
looks good to me

Issue #86, opened 2021-08-16
Report error when Windows Build with -DTARGET=Release -DCRYPTO=mbedtls


yzhu52:
cmake -G"NMake Makefiles" -DARCH=x64 -DTOOLCHAIN=VS2019 -DTARGET=Release -DCRYPTO=mbedtls .. 
nmake copy_sample_key
nmake

Error:
[ 99%] Linking C executable ..\..\..\bin\test_size_of_spdm_requester.exe
LINK: command "C:\PROGRA~2\MICROS~2\2019\PROFES~1\VC\Tools\MSVC\1428~1.29~\bin\Hostx64\x64\link.exe /nologo @CMakeFiles\test_size_of_spdm_requester.dir\objects1.rsp /out:..\..\..\bin\test_size_of_spdm_requester.exe /implib:..\..\..\bin\test_size_of_spdm_requester.lib /pdb:D:\SPDM\0816\libspdm\build\bin\test_size_of_spdm_requester.pdb /version:0.0 /DLL /ENTRY:ModuleEntryPoint /NOLOGO /SUBSYSTEM:EFI_BOOT_SERVICE_DRIVER /NODEFAULTLIB /IGNORE:4086 /OPT:REF /LTCG /subsystem:console ..\..\..\lib\memlib.lib ..\..\..\lib\debuglib.lib ..\..\..\lib\spdm_requester_lib.lib ..\..\..\lib\spdm_common_lib.lib ..\..\..\lib\cryptstublib_dummy.lib ..\..\..\lib\mbedtlslib.lib ..\..\..\lib\rnglib.lib ..\..\..\lib\cryptlib_mbedtls.lib ..\..\..\lib\malloclib_null.lib ..\..\..\lib\spdm_crypt_lib.lib ..\..\..\lib\spdm_secured_message_lib.lib ..\..\..\lib\spdm_transport_mctp_lib.lib ..\..\..\lib\spdm_device_secret_lib_null.lib ..\..\..\lib\intrinsiclib.lib Kernel32.lib MSVCRTD.lib vcruntimed.lib ucrtd.lib Gdi32.lib User32.lib Winmm.lib Advapi32.lib ws2_32.lib /MANIFEST /MANIFESTFILE:..\..\..\bin\test_size_of_spdm_requester.exe.manifest" failed (exit code 1257) with the following output:
Generating code
[...]

PR #156, opened 2021-09-24
Disable /GL in the intrinsiclib of RELEASE version 

liyi77:
REF: #86


liyi77:
It seems that all config for VS2019 can be compiled normally, but I still confused that why ia32 can compile before but x64 cannot.

jyao1:
OK

steven-bellock:
Looks good to me.

PR #158, opened 2021-09-24
record transcript hash instead of data.

jyao1:
REF: https://github.com/DMTF/libspdm/issues/22

The context size is reduced from 0x12c10 (76K) to 0x7078 (29K).

The scope is message_b/c/mut_b/mut_c, message_m(measurement),
and message_k/f(key_exchange/finish).

We cannot do for message_a, because the algo is not negotiated yet.
We still need cache partial message_k, before finished_key is ready.

Test:
1) Pass unit test with RECORD_TRANSCRIPT_DATA on/off
2) Pass emu test with responder (data) + requester (hash)
3) Pass emu test with responder (hash) + requester (data)

### DIFF
--- a/include/library/spdm_lib_config.h
+++ b/include/library/spdm_lib_config.h
@@ -27,7 +27,7 @@
 #define MAX_SPDM_CONNECTION_STATE_CALLBACK_NUM 4
 
 // If cache transcript data or transcript hash
-//#define RECORD_TRANSCRIPT_DATA 1
+#define LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT 0
 
 //
 // Crypto Configuation
--- a/include/library/spdm_lib_config.h
+++ b/include/library/spdm_lib_config.h
@@ -25,6 +25,9 @@
 #define MAX_SPDM_SESSION_STATE_CALLBACK_NUM 4
 #define MAX_SPDM_CONNECTION_STATE_CALLBACK_NUM 4
 
+// If cache transcript data or transcript hash
+//#define RECORD_TRANSCRIPT_DATA 1
+
 //
 // Crypto Configuation
 // In each category, at least one should be selected.
--- a/include/library/spdm_lib_config.h
+++ b/include/library/spdm_lib_config.h
@@ -19,7 +19,8 @@
 #define MAX_SPDM_CERT_CHAIN_BLOCK_LEN 1024
 
 #define MAX_SPDM_MESSAGE_BUFFER_SIZE 0x1200
-#define MAX_SPDM_MESSAGE_SMALL_BUFFER_SIZE 0x100
+#define MAX_SPDM_MESSAGE_SMALL_BUFFER_SIZE 0x100  // to hold message_a before negotiate
+#define MAX_SPDM_MESSAGE_MEDIUM_BUFFER_SIZE 0x300 // to hold message_k before finished_key is ready
 
 #define MAX_SPDM_REQUEST_RETRY_TIMES 3
 #define MAX_SPDM_SESSION_STATE_CALLBACK_NUM 4
--- a/library/spdm_common_lib/context_data.c
+++ b/library/spdm_common_lib/context_data.c
@@ -1122,12 +1122,19 @@ return_status spdm_append_message_k(IN void *context, IN void *session_info,
 		spdm_hash_update (spdm_context->connection_info.algorithm.base_hash_algo,
 			spdm_session_info->session_transcript.digest_context_th, message, message_size);
 		if (!finished_key_ready) {
+			//
+			// append message only if finished_key is NOT ready.
+			//
 			append_managed_buffer(
 				&spdm_session_info->session_transcript.temp_message_k, message, message_size);
 		}
 
 		//
-		// append message only if finished_key is NOT ready.
+		// Above action is to calculate HASH for message_k.
+		// However, we cannot use similar way to calculate HMAC. (chicken-egg problem)
+		// HMAC need finished_key, and finished_key calculation need message_k.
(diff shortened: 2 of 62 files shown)


include/library/spdm_lib_config.h
```diff
@@ -19,12 +19,16 @@
...
-#define MAX_SPDM_MESSAGE_SMALL_BUFFER_SIZE 0x100
+#define MAX_SPDM_MESSAGE_SMALL_BUFFER_SIZE 0x100  // to hold message_a before negotiate
+#define MAX_SPDM_MESSAGE_MEDIUM_BUFFER_SIZE 0x300 // to hold message_k before finished_key is ready
 
 #define MAX_SPDM_REQUEST_RETRY_TIMES 3
 #define MAX_SPDM_SESSION_STATE_CALLBACK_NUM 4
 #define MAX_SPDM_CONNECTION_STATE_CALLBACK_NUM 4
 
+// If cache transcript data or transcript hash
+//#define RECORD_TRANSCRIPT_DATA 1
```
taprinz, line 30: nitpicky, but should we handle similarly to OPENSPDM_*_SUPPORT defines below?
Where support is enabled/disabled with 1 or 0, rather than "#ifdef"?

library/spdm_common_lib/context_data_session.c
```diff
@@ -60,10 +60,15 @@ void spdm_session_info_init(IN spdm_context_t *spdm_context,
...
 		spdm_context->local_context.psk_hint,
 		spdm_context->local_context.psk_hint_size);
+#if RECORD_TRANSCRIPT_DATA
 	session_info->session_transcript.message_k.max_buffer_size =
 		MAX_SPDM_MESSAGE_BUFFER_SIZE;
 	session_info->session_transcript.message_f.max_buffer_size =
 		MAX_SPDM_MESSAGE_BUFFER_SIZE;
+#else
+	session_info->session_transcript.temp_message_k.max_buffer_size =
+		MAX_SPDM_MESSAGE_MEDIUM_BUFFER_SIZE;
```
taprinz, line 70: would it be worth doing

```
session_info->session_transcript.temp_message_k.max_buffer_size =
         sizeof(session_info->session_transcript.temp_message_k.max_buffer_size);
```

The current is less verbose, but since the sizes are already defined in the header file by the same macro, setting max size via sizeof prevents any potential mismatch from occurring in the future.

library/spdm_common_lib/context_data.c
```diff
@@ -835,50 +1020,293 @@ return_status spdm_append_message_m(IN void *context, IN void *message,
...
+			spdm_session_info->session_transcript.digest_context_th, message, message_size);
+		if (!finished_key_ready) {
+			append_managed_buffer(
+				&spdm_session_info->session_transcript.temp_message_k, message, message_size);
+		}
+
+		//
+		// append message only if finished_key is NOT ready.
+		//
+		if (!finished_key_ready) {
```
taprinz, line 1132: nit: Move this up to block starting line 1124, since comment applies there as well.

library/spdm_common_lib/context_data.c
```diff
@@ -835,50 +1020,293 @@ return_status spdm_append_message_m(IN void *context, IN void *message,
...
+		}
+		spdm_hash_update (spdm_context->connection_info.algorithm.base_hash_algo,
+			spdm_session_info->session_transcript.digest_context_th, message, message_size);
+		if (!finished_key_ready) {
+			append_managed_buffer(
+				&spdm_session_info->session_transcript.temp_message_k, message, message_size);
+		}
+
+		//
+		// append message only if finished_key is NOT ready.
```
taprinz, line 1130: I'm a little confused as to why we both manage temporary transcripts, as well as hashes here.

Is the idea that, we keep transcripts for all session establishments scenarios ONLY until the finished key is created?
And once we have finished key, we can then create HMAC's. So that all further updates are added to HMAC context, rather than transcripts themselves?

include/library/spdm_lib_config.h
jyao1: that is OK.

I think we need clean up the *OPENSPDM* to *LIBSPDM*.

library/spdm_common_lib/context_data_session.c
jyao1: Good idea

library/spdm_common_lib/context_data.c
jyao1: OK.

library/spdm_common_lib/context_data.c
jyao1: that is right. before finished_key is ready, I am not able to HMAC.

jyao1: that is right. before finished_key is ready, I am not able to HMAC.

include/library/spdm_lib_config.h
taprinz: agreed, just didn't want to clutter this CL with it.
as for 1&0 vs #define, mostly want to be consistent, rather than strong preference for either.

jyao1: Fixed

library/spdm_common_lib/context_data_session.c
jyao1: Fixed

library/spdm_common_lib/context_data.c
jyao1: Fixed

library/spdm_common_lib/context_data.c
jyao1: fixed

jyao1:
Any additional feedback? If no, I will merge it soon. We have other P-R depending upon it.
We can file new issue, if there is feedback later.

steven-bellock:
Ok. Looks good to me for now.

PR #159, opened 2021-09-26
Fix test_size binary to help size analysis.

jyao1:
1) Always enable MAP.
2) Use cryptlib_dummy to remove impact from crypto lib.
3) Disable optimization in main test_size entrypoint.

Now the size info is:
39,424 test_size_of_spdm_requester.exe
48,128 test_size_of_spdm_responder.exe


steven-bellock:
Looks good to me.

Issue #117, opened 2021-09-09
SPDM responder has inconsistent checking against key schedule: PSK_EXCHANGE and KEY_EXCHANGE request validation


jcahill0:
The SPDM responder's PSK_EXCHANGE request validation code does check against the negotiated key schedule while the equivalent SPDM responder KEY_EXCHANGE request validation code does not.  libspdm should be consistent.

PSK validation:

		if (spdm_context->connection_info.algorithm.key_schedule !=
		    SPDM_ALGORITHMS_KEY_SCHEDULE_HMAC_HASH) {
			spdm_generate_error_response(
				spdm_context,
				SPDM_ERROR_CODE_UNSUPPORTED_REQUEST,
				SPDM_PSK_EXCHANGE, response_size, response);
			return RETURN_SUCCESS;
		}

In theory this is already covered in the SPDM Responder's NEGOTIATE_ALGORITHMS+ALGORITHMS validation, see below, but it would be reasonable for a security focused library to be more defensive with checks at each request handling.

		if (spdm_is_capabilities_flag_supported(
			    spdm_context, FALSE,
			    SPDM_GET_CAPABILITIES_REQUEST_FLAGS_KEY_EX_CAP,
			    SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_KEY_EX_CAP) ||
		    spdm_is_capabilities_flag_supported(
			    spdm_context, FALSE,
			    SPDM_GET_CAPABILITIES_REQUEST_FLAGS_PSK_CAP,
			    SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_PSK_CAP)) {
			if (spdm_context->connection_info.algorithm
				    .key_schedule !=
			    SPDM_ALGORITHMS_KEY_SCHEDULE_HMAC_HASH) {
				return RETURN_SECURITY_VIOLATION;
			}
		}

I won't be able to work on this code change, but wanted to submit the issue for awareness.

steven-bellock:
Approved. They should be consistent.

jyao1:
Now, I feel it is burden to add checks everywhere. Can we remove the check to make them consistent?

jyao1:
I also notice that there is a key difference between PSK and KEY_EXCHANGE. The PSK does not require  NEGOTIATE_ALGORITHMS. But KEY_EXCHANGE requires NEGOTIATE_ALGORITHMS.

[...]

PR #160, opened 2021-09-26
Fix117: check about key schedule should be consistent 

liyi77:
REF: #117
Double check if algorithm has been provisioned, because ALGORITHM might be skipped.


jyao1:
I also notice that there is a key difference between PSK and KEY_EXCHANGE. The PSK does not require NEGOTIATE_ALGORITHMS. But KEY_EXCHANGE requires NEGOTIATE_ALGORITHMS.

See "SPDM commands GET_CAPABILITIES and NEGOTIATE_ALGORITHMS are not required during session key establishment with the PSK option"

I change my mind now. I think we just need add comment to address the reason of inconsistency.
Adding code == Adding size.

Issue #100, opened 2021-08-25
Issues of handling Measurement Summary Hash


rw8896:
There seems to be some issues for the Responder to deal with Measurement Summary Hash.

1) Invalid "Requested measurement hash type" (values other than 0x0, 0x1, 0xff) now returns Unspecified error instead of InvalidRequest or UnsupportedRequest.
```
Requested measurement summary hash Type:
0x0 . No measurement summary hash.
0x1 . TCB measurement hash.
0xFF . All measurements hash.
All other values reserved.
```
2) Only CHALLENGE handler will verify Measurement Capability. KEY_EXCHANGE and PSK_EXCHANGE should also apply it?
https://github.com/DMTF/libspdm/commit/37798f00071935cef81d729ccd0953fabc224a21

3) return value of spdm_hash_all was not checked in spdm_generate_measurement_summary_hash (and some other places in crypto_service.c)

jyao1:
agree

steven-bellock:
Approved.
1. Should (probably) use InvalidRequest. Also need to flesh out error codes to make them less ambiguous. @raghuncstate will file an issue with the security task force to clarify.
2. Agree. @raghuncstate will file an issue with security task force to include KEY_EXCHANGE and PSK_EXCHANGE in the MEAS_CAP section of CAPABILITIES response.
3. Is covered by https://github.com/DMTF/libspdm/issues/75

raghuncstate:
Filed https://github.com/DMTF/Security-TF/issues/1473 and https://github.com/DMTF/Security-TF/issues/1472.

liyi77:
@rw8896 
1. Invalid "Requested measurement hash type" (values other than 0x0, 0x1, 0xff) now returns Unspecified error instead of InvalidRequest or UnsupportedRequest.

Does it means the return value check of spdm_generate_measurement_summary_hash()?
https://github.com/DMTF/libspdm/blob/50c2b1a389f45c3060aa040148cdd1384214a074/library/spdm_responder_lib/challenge_auth.c#L166-L172

The return value will be related to internal hash calculation fail or spdm_measurement_collection(), these cases seems need to return Unspecified.
How about adding some pre-check such if (spdm_request->header.param2 != 0 && spdm_request->header.param2 != 1 ...)
instead of modifying the existing check？

rw8896:
> @rw8896
> 
[...]

PR #161, opened 2021-09-26
Fix100: Issues of handling Measurement Summary Hash

liyi77:
REF: #100


jyao1:
It looks good.

steven-bellock:
Looks good to me.

Issue #165, opened 2021-10-01
No key synchronization when receiving ERROR messages when UPDATE_KEY with UpdateAllKeys is issued


vinmaciel:
In section 10.21.1 "Session key update synchronization" par. 495 (v1.1.1), it is stated that, when issuing a KEY_UPDATE request with UpdateAllKeys, the KEY_UPDATE_ACK response should already be encrypted with the updated receiver key.
However, when receiving an ERROR message (e.g., a Busy response), the responder would not have updated its key and, therefore, would send the ERROR message encrypted with the old key.
Currently, it appears that libspdm does not retry the decryption with the old key if the decryption with the updated one failed.

We note that the spec is flexible with how to treat the keys, as it says (in par. 491) that the correct usage of keys is responsibility of the transport layer.
And (in par. 496) that "If the sender has not received KEY_UPDATE_ACK, the sender may retry or end the session", which is not obligatory, and could result in abandoning the key update.

Since the libspdm already falls back to the old keys in case of failure, we believe that libspdm should support the two following cases:
- when decrypting the message with the updated key, accept if it is a KEY_UPDATE_ACK.
- in the case the previous decryption fails, it should decrypt with the old key and should only accept if it is an ERROR (and treat it when pertinent).

jyao1:
Agree. Error handling should be added, at least for Busy.

jyao1:
Need verify below test case to ensure it passes.

[ FAILED ] test_spdm_requester_key_update_case29
[ FAILED ] test_spdm_requester_key_update_case30
[ FAILED ] test_spdm_requester_key_update_case31
[ FAILED ] test_spdm_requester_key_update_case33

PR #166, opened 2021-10-01
Adding new test cases for the key_update message (requester side)

vinmaciel:
We add unit test cases on the requester side for the key_update message. Including the standard tests from previous messages.
These tests verify the output of the requester side of the key update, and whether the major secret was updated.
We also include the handling of retries and received ERROR responses, as in previous messages.

It includes 4 commits:

1) when receiving an ERROR response, we call the "spdm_handle_error_response_main()" function to handle it, both in the UpdateKey/UpdateAllKeys and VerifyNewKey operations.
The major secret is rolled back in most "failed" cases for UpdateKey/UpdateAllKeys, but not for VerifyNewKey.
Additionally, it can successfully retry when receiving a Busy or ResponseNotReady ERROR response.

2) new test cases for the UpdateKey operation:
- send request failed
- success case
- invalid state
- standard errors (InvalidParameters, Busy, RequestResync, ResponseNotReady)
- errors with retry (Busy, ResponseNotReady)
- unexpected errors
- buffer reset
- wrong setup (capabilities, session)
- wrong response parameters (response code, operation, token)

3) new test cases for the VerifyNewKey operation:
- standard errors (InvalidParameters, Busy, RequestResync, ResponseNotReady)
- errors with retry (Busy, ResponseNotReady)
- unexpected errors
- wrong response parameters (response code, operation, token)

4) new test cases for the UpdateAllKeys operation:
- success case
- standard errors (InvalidParameters, Busy, RequestResync, ResponseNotReady)
- errors with retry (Busy, ResponseNotReady)
- unexpected errors

We note here that there are some test cases in commit 4 that are not successful, beginning at case 28, because the library still lacks management of decryption keys in the requester when receiving ERROR messages.
Cases 28, 32, and 34 are "false success" cases, which have the same output behavior, but they are not using the correct keys.
We include a detailed explanation in issue #165.
Commits 1, 2 and 3 can be tested separately and added to the repo independently of commit 4.

### DIFF
--- a/library/spdm_requester_lib/key_update.c
+++ b/library/spdm_requester_lib/key_update.c
@@ -6,6 +6,15 @@
 
 #include "spdm_requester_lib_internal.h"
 
+#pragma pack(1)
+
+typedef struct {
+	spdm_message_header_t header;
+	uint8 dummy_data[sizeof(spdm_error_data_response_not_ready_t)];
+} spdm_key_update_response_mine_t;
+
+#pragma pack()
+
 /**
   This function sends KEY_UPDATE
   to update keys for an SPDM Session.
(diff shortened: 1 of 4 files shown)


jyao1:
Looks good. I think we can integrate this at first. Then fix the issue later.

steven-bellock:
Looks good to me.

Issue #134, opened 2021-09-21
Need ensure the crypto API implementable for crypto algorithm for 1.2WIP


jyao1:

jyao1:
This issue only ensure the API defined cryptlib.h can be used, such as SM2 or EDDSA.

The detail implementation can be tracked in
https://github.com/DMTF/libspdm/issues/173,
https://github.com/DMTF/libspdm/issues/141,
https://github.com/DMTF/libspdm/issues/142,
https://github.com/DMTF/libspdm/issues/143,
https://github.com/DMTF/libspdm/issues/144.

PR #168, opened 2021-10-03
Transcript hash prepare1.2

jyao1:
This is to fix issue https://github.com/DMTF/libspdm/issues/134 - prepare for 1.2 script.

### DIFF
--- a/library/spdm_common_lib/context_data.c
+++ b/library/spdm_common_lib/context_data.c
@@ -78,7 +78,7 @@ return_status spdm_set_data(IN void *context, IN spdm_data_type_t data_type,
 			return RETURN_INVALID_PARAMETER;
 		}
 		if (parameter->location == SPDM_DATA_LOCATION_CONNECTION) {
-			if ((uint8)(data_size /sizeof(spdm_version_number_t)) != 1) {
+			if (data_size != sizeof(spdm_version_number_t)) {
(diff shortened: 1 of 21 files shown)


library/spdm_common_lib/context_data.c
```diff
@@ -104,16 +104,14 @@ return_status spdm_set_data(IN void *context, IN spdm_data_type_t data_type,
 			return RETURN_INVALID_PARAMETER;
 		}
 		if (parameter->location == SPDM_DATA_LOCATION_CONNECTION) {
-			spdm_context->connection_info.secured_message_version
-				.spdm_version_count = (uint8)(
-				data_size / sizeof(spdm_version_number_t));
-			copy_mem(spdm_context->connection_info
-					 .secured_message_version.spdm_version,
+			if ((uint8)(data_size /sizeof(spdm_version_number_t)) != 1) {
```
steven-bellock, line 107: Remove the cast to uint8, since if data_size is large the upper bits would be truncated.

steven-bellock (changes requested):

library/spdm_common_lib/context_data.c
jyao1: done

steven-bellock:
Looks good to me.

Issue #148, opened 2021-09-23
use hook to verify remote certificate


jyao1:
REF: https://github.com/DMTF/libspdm/blob/main/library/spdm_common_lib/spdm_common_lib_internal.h

Currently, we rely on  peer_root_cert_provision to verify remote certificate.

Another approach is to allow caller provide hook function to return remove cert verification. It is a more flexible way to handle cert.

jyao1:
Other SSL libs use similar way: 
openssl: SSL_CTX_set_verify().
mbedtls: mbedtls_ssl_conf_verify().
wolfssl: wolfSSL_CTX_set_verify().

PR #172, opened 2021-10-04
Add hook to verify the peer certiticate.

jyao1:
REF: https://github.com/DMTF/libspdm/issues/148

### DIFF
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -448,6 +448,53 @@ void spdm_register_transport_layer_func(
 	IN spdm_transport_decode_message_func transport_decode_message);
 
 /**
+  Verify a SPDM cert chain in a slot.
+
+  This function shall verify:
+    1) The integrity of the certificate chain. (Root Cert Hash->Root Cert->Cert Chain)
+    2) The trust anchor. (Root Cert Hash/Root cert matches the trust anchor)
+
+  The function shall check the negotiated hash algorithm to check root cert hash.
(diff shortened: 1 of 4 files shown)


include/library/spdm_common_lib.h
```diff
@@ -447,6 +447,49 @@ void spdm_register_transport_layer_func(
...
+
+  The function shall check the negotiated hash algorithm to check root cert hash.
+  The function shall check the negotiated (req) asym algorithm to determine if it is right cert chain.
+
+  The function returns error if either of above is not satisfied.
+
+  @param  spdm_context                  A pointer to the SPDM context.
+  @param  slot_id                       The number of slot for the certificate chain.
+  @param  cert_chain_size               Indicate the size in bytes of the certificate chain.
+  @param  cert_chain                    A pointer to the buffer storing the certificate chain
```
steven-bellock, line 465: How should this behave if the Responder does not return the root certificate? Would the caller have to append the root certificate to the beginning of the certificate chain?

jyao1: This cert_chain contains everything returned from responder, including root_cert hash.

The verifier can determine which root_cert should be used to verify the chain, by comparing root_cert hash.

libspdm does not the trust anchor. Only verifier knows the trust anchor.

steven-bellock: Ok. Understood.

steven-bellock:
Looks good to me.

PR #177, opened 2021-10-05
Move spdm_reset_message_buffer_via_request_code() to internal.

### LINKED ISSUE #171: move spdm_reset_message_buffer_via_request_code() to internal func
spdm_reset_message_buffer_via_request_code() should be called by internal function.
No need to expose it in spdm_common_lib.h

jyao1:
Fix #171.


steven-bellock:
Looks good to me.

Issue #91, opened 2021-08-18
For session based spdm measurement transcript hash shall be calculated per session.


jyao1:
We should not mix the measurement transcript hash across multiple sessions. The signature should be per session.

PR #188, opened 2021-10-09
Fix91 Measurement transcript hash shall be calculated per session

liyi77:
Measurement transcript hash shall be calculated per session
    
REF: #91
    
Should not mix the measurement transcript hash across
multiple sessions. The signature should be per session.


jyao1:
looks good to me

Issue #189, opened 2021-10-12
most certificates have expired


liyi77:
valid from 2020/10/10 to 2021/10/10,
should update

joliqi:
valid expire again

PR #190, opened 2021-10-12
Update sample key

liyi77:
REF: #189


jyao1:
I think we can fix https://github.com/DMTF/spdm-emu/issues/7 as well.

steven-bellock:
Looks good to me.

PR #191, opened 2021-10-12
rename OPENSPDM_ to LIBSPDM_ in config.h

jyao1:


steven-bellock:
Looks good to me.

PR #193, opened 2021-10-12
Refine sm2 APIs.

jyao1:
The implementation is not changed yet. We will change it later in 1.2 wave.


steven-bellock:
Looks good to me.

Issue #187, opened 2021-10-09
Report test error about hmac verify


liyi77:
Test at the newest patch : 8f1d211e3a30a5d172e6f6e74e4914aa819c5f47 
build command:
cmake -G"NMake Makefiles" -DARCH=x64 -DTOOLCHAIN=VS2019 -DTARGET=Release -DCRYPTO=mbedtls ..
cmake -G"NMake Makefiles" -DARCH=x32 -DTOOLCHAIN=VS2019 -DTARGET=Debug -DCRYPTO=mbedtls ..

[  ERROR   ] --- 0x4 != 0x24
[   LINE   ] --- E:\workspace\wp_spdm\libspdm\unit_test\test_spdm_responder\respond_if_ready.c:696: error: Failure!
[  ERROR   ] --- 0xfffe != 0xffff
[   LINE   ] --- E:\workspace\wp_spdm\libspdm\unit_test\test_spdm_responder\respond_if_ready.c:785: error: Failure!
[  ERROR   ] --- 0x7f != 0x67
[   LINE   ] --- E:\workspace\wp_spdm\libspdm\unit_test\test_spdm_responder\respond_if_ready.c:886: error: Failure!
[  PASSED  ] 11 test(s).
[  FAILED  ] spdm_responder_respond_if_ready_tests: 3 test(s), listed below:
[  FAILED  ] test_spdm_responder_respond_if_ready_case6
[  FAILED  ] test_spdm_responder_respond_if_ready_case7
[  FAILED  ] test_spdm_responder_respond_if_ready_case8

 3 FAILED TEST(S)
[  PASSED  ] 7 test(s).
[  ERROR   ] --- 0x4 != 0x24
[   LINE   ] --- E:\workspace\wp_spdm\libspdm\unit_test\test_spdm_responder\finish.c:154: error: Failure!
[  ERROR   ] --- 0x4 != 0x24
[   LINE   ] --- E:\workspace\wp_spdm\libspdm\unit_test\test_spdm_responder\finish.c:844: error: Failure!
[  ERROR   ] --- 0x4 != 0x24
[   LINE   ] --- E:\workspace\wp_spdm\libspdm\unit_test\test_spdm_responder\finish.c:995: error: Failure!
[  PASSED  ] 13 test(s).
[  FAILED  ] spdm_responder_finish_tests: 3 test(s), listed below:
[  FAILED  ] test_spdm_responder_finish_case1
[  FAILED  ] test_spdm_responder_finish_case7
[...]

PR #196, opened 2021-10-13
Enable duplicate function to clone the complete hmac context

liyi77:
REF: #187
Temporary solution to the problem of context clone.
There are not any standard function in mbedtls to clone a complete hmac context,
so use hard code to fix this error.


jyao1:
Please remove MD2/MD4/MD5/RIPEMD160/SHA1/SHA224.

We only need SHA256/384/512.

Please add ASSERT() for default.

jyao1:
Looks good to me.

PR #201, opened 2021-10-16
Sm2 key exchange

jyao1:
Remove ECDH-SM2.
SPDM need SM2-Key-Exchange, not ECDH-SM2.

Change SM2 shared key size to 16 in unit test.
Align with SPDM spec.


jyao1:
user mistake, close it

Issue #198, opened 2021-10-14
Incorrect length of the NEGOTIATE_ALGORITHMS request message


Zhiqiang520:
In clause '10.4 NEGOTIATE_ALGORITHMS request and ALGORITHMS response messages', paragraph 187 'NEGOTIATE_ALGORITHMS request message format', the Offset **4** Field **Length** means Length of the entire request message, in bytes.

But the following line 73 'spdm_request.**length** = sizeof(spdm_request); ' missed the size of Field **ExtAsym**, **ExtHash** and **ExtAlgCount** in **ReqAlgStruct**.

https://github.com/DMTF/libspdm/blob/f92be5023c1a3ff036b9e43cd8de0cd06b5eb711/library/spdm_requester_lib/negotiate_algorithms.c#L71-L80

jyao1:
ext_algo is not supported.
If we open it here, then we have bigger problem in other places.

Suggest keep it as is.

Issue #197, opened 2021-10-14
Incorrect length of the ALGORITHMS response message


Zhiqiang520:
In clause '10.4 NEGOTIATE_ALGORITHMS request and ALGORITHMS response messages', paragraph 195 'Successful ALGORITHMS response message format', the Offset **4** Field **Length** means Length of the ALGORITHMS response message, in bytes.

But the following line 310 'spdm_response->**length** = (uint16)*response_size;' missed the size of Field **ExtAsymSel**, **ExtHashSel** and **ExtAlgCount** in **RespAlgStruct**.

https://github.com/DMTF/libspdm/blob/f92be5023c1a3ff036b9e43cd8de0cd06b5eb711/library/spdm_responder_lib/algorithms.c#L292-L310

Zhiqiang520:
Anyway, the current **Length** in code does not match with the SPDM specification(ver1.1.1) **strictly**, so I suggest to add some comments in the above Line 73, to avoid confusion and make it more clearer, such as 'currently ext_algo is not supported.'

PR #203, opened 2021-10-16
Add comment ext algo unsupported.

jyao1:
FIX: https://github.com/DMTF/libspdm/issues/197
FIX: https://github.com/DMTF/libspdm/issues/198


steven-bellock:
Looks good to me.

PR #207, opened 2021-10-17
Check returned heartbeat period.

### LINKED ISSUE #129: KEY_EXCHANGE - Requester accepts any HeartbeatPeriod value
SPDM version 1.1 declares that Param1 of KEY_EXCHANGE_RSP is the Heartbeat Period and that "The value of this field shall be zero if Heartbeat is not supported. Otherwise, the value shall be in units of seconds".  Nonetheless, a Requester that does not support Heartbeat messages accepts a value different from 0 and a requester that supports Heartbeat accepts a period of 0.

The requester should verify if the HEARTBEAT flag was set during the GET_CAPABILITIES/CAPABILITIES and act accordingly.

jyao1:
Fix #129

### DIFF
--- a/library/spdm_requester_lib/key_exchange.c
+++ b/library/spdm_requester_lib/key_exchange.c
@@ -205,6 +205,18 @@ return_status try_spdm_send_receive_key_exchange(
 		return RETURN_DEVICE_ERROR;
 	}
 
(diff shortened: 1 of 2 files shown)


steven-bellock:
Looks good to me.

PR #211, opened 2021-10-19
Adding new test cases for the key_update message (responder side)

vinmaciel:
We add unit test cases on the responder side for the key_update message. We also include the responder-side code for handling retries.
These tests verify the correctness of the generated response for different key_update requests, and whether the major secret was updated. It includes the following test scenarios:
- success case (all operations and retries)
- abnormal responder state (Busy, RequestResync, NotReady)
- invalid state
- buffer reset
- wrong setup (capabilities, session)
- wrong request parameters (size, operation, retry tokens)
- invalid request sequence (VerifyNewKey and retries)


jyao1:
Thank you @vinmaciel. It looks good to me.

steven-bellock (approved):

Issue #212, opened 2021-10-19
Need revert 534bafd7ff86d4b1b6c60f77d7a7e79013236065 change in context_data.c.


jyao1:
HI @rkongintel, I need revert the 534bafd7ff86d4b1b6c60f77d7a7e79013236065 change in context_data.c.

```
		#if SPDM_ENABLE_CAPABILITY_MEAS_CAP
		ASSERT((*(uint32 *)data) & SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_MEAS_CAP);
		#else // SPDM_ENABLE_CAPABILITY_MEAS_CAP
		ASSERT(((*(uint32 *)data) & SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_MEAS_CAP) == 0);
		ASSERT(((*(uint32 *)data) & SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_MEAS_FRESH_CAP) == 0);
		#endif // SPDM_ENABLE_CAPABILITY_MEAS_CAP
```

This code causes the spdm-emu fail. It is a big regression. It is not acceptable.

jyao1:
We will discuss how to add check back in https://github.com/DMTF/libspdm/issues/205.

Issue #205, opened 2021-10-17
Should not assert unsupported response capabbility in spdm_set_data()


liyi77:
https://github.com/DMTF/libspdm/blob/f0f39596dd4ea0689c6196f2d023bc98ba9663ba/library/spdm_common_lib/context_data.c#L128-L152

1. spdm_set_data() is a spdm common lib function, it will falsely assert when the requester is correctly initialized.
2. It is confused that assert when have not used a libspdm supported capabbility,
    assert when user want to use a unsupported cap is good enough.

steven-bellock:
Yeah the logic is backwards.

liyi77:
> 1. spdm_set_data() is a spdm common lib function, it will falsely assert when the requester is correctly initialized.
> 2. It is confused that assert when have not used a libspdm supported capabbility,
>    assert when user want to use a unsupported cap is good enough.

Requester and responder use same bit to indicate supported capabilities, such:

#define SPDM_GET_CAPABILITIES_REQUEST_FLAGS_HANDSHAKE_IN_THE_CLEAR_CAP BIT15
#define SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_HANDSHAKE_IN_THE_CLEAR_CAP BIT15
it is my bad that treat it as different thing just according name, topic 1 is a wrong description :(  .

But requester should not check this CAP anyway:
```
#define SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_MEAS_CAP (BIT3 | BIT4)
```
because there is no corresponding REQUSET FLAGS.

Wenxing-hou:
I will check it.

PR #213, opened 2021-10-19
revert assert check in context data, which causes spdm-emu fail.

jyao1:
FIX: https://github.com/DMTF/libspdm/issues/212

We will discuss formal solution in https://github.com/DMTF/libspdm/issues/205


rkongintel (approved):
Looks good.

PR #217, opened 2021-10-21
Fix wrong version number check.

jyao1:


jyao1:
We need support 2 styles of transcript.
1.0, 1.1 - old style
1.2, 1.3 - new style
I assume 2.0, 2.1 will also use the new style.

PR #245, opened 2021-11-01
Add test_spdm_responder_algorithms_case20 for responder algorithms un…

Zhiqiang520:
…it test.

When both of requester and responder support multiple algorithms, then defaults to choose the strongest available algorithm.


jyao1:
I know we are discussing Algorithm selection at https://github.com/DMTF/libspdm/discussions/121
Seems OK for me. Do you agree @steven-bellock ?

PR #251, opened 2021-11-02
Optimize Get Measurements code path 

### LINKED ISSUE #179: Get measurements efficiency
The Get Measurements code path seems inefficient. It looks like it the code path ends up in the spdm_measurement_collection routine. This routine collects all measurements on every single call, even though only single measurement may be requested at a higher level.  Some of the measurements, depending on their implementation, may involve some long/complex calculations, and may not be wise to repeat, or the responder may not want to cache the results. 

If my interpretation of the code path is correct, can we optimize this to only collect the measurements that were requested at a higher level? Perhaps by passing the "index" requested into the spdm_measurement_collection routine?

rkongintel:
Fix #179

This change optimizes the get measurements code path related to [https://github.com/DMTF/libspdm/issues/179](https://github.com/DMTF/libspdm/issues/179)

- 4.5K is saved off the stack by not using an extra local buffer for copying measurements when **spdm_get_response_measurements** calls **spdm_measurement_collection**. The response buffer is passed instead and **spdm_measurement_collection** will fill in the response buffer directly.
- The size of the **spdm_get_response_measurements** code is reduced from ~1.3K to ~500bytes (Release build x64). The size of **spdm_measurement_collection**, which is just an example, increases slightly by ~180 bytes.
- **spdm_measurement_collection** now receives and "**measurements_index**" so it does not need to calculate all measurements when only one is really needed.

Tested this running x64 release and x64 debug unit tests for crypt\common\requester\responder. Also tested spdm_emu x64 debug.

Please review this carefully as it is quite a change.

I believe there is also potential for optimization down the challenge_auth, key_exchange, psk_exchange code paths as they use spdm_generate_measurement_summary_hash which also uses spdm_measurement_collection.

### DIFF
--- a/include/library/spdm_device_secret_lib.h
+++ b/include/library/spdm_device_secret_lib.h
@@ -27,19 +27,22 @@
                                        It must align with measurement_specification (SPDM_MEASUREMENT_BLOCK_HEADER_SPECIFICATION_*)
   @param  measurement_hash_algo          Indicates the measurement hash algorithm.
                                        It must align with measurement_hash_algo (SPDM_ALGORITHMS_MEASUREMENT_HASH_ALGO_*)
-  @param  device_measurement_count       The count of the device measurement block.
-  @param  device_measurement            A pointer to a destination buffer to store the concatenation of all device measurement blocks.
-  @param  device_measurement_size        On input, indicates the size in bytes of the destination buffer.
+  @param  measurement_index      The index of the measurement to collect.
+  @param  measurements_count       The count of the device measurement block.
+  @param  measurements            A pointer to a destination buffer to store the concatenation of all device measurement blocks.
+  @param  measurements_size        On input, indicates the size in bytes of the destination buffer.
                                        On output, indicates the size in bytes of all device measurement blocks in the buffer.
 
-  @retval TRUE  the device measurement collection success and measurement is returned.
-  @retval FALSE the device measurement collection fail.
+  @retval RETURN_SUCCESS             Successfully returned measurement_count and optionally measurements, measurements_size.
+  @retval RETURN_BUFFER_TOO_SMALL    "measurements" buffer too small for measurements.
+  @retval RETURN_INVALID_PARAMETER   Invalid parameter passed to function.
+  @retval RETURN_***                 Any other RETURN_ error from base.h
 **/
-typedef boolean (*spdm_measurement_collection_func)(
+typedef return_status (*spdm_measurement_collection_func)(
 	IN spdm_version_number_t spdm_version,
 	IN uint8 measurement_specification, IN uint32 measurement_hash_algo,
-	OUT uint8 *device_measurement_count, OUT void *device_measurement,
-	IN OUT uintn *device_measurement_size);
+	IN uint8 measurement_index, OUT uint8 *measurement_count,
+	OUT void *measurement, IN OUT uintn *measurement_size);
 
 /**
   Sign an SPDM message data.
@@ -112,7 +115,7 @@ typedef boolean (*spdm_psk_hkdf_expand_func)(
   Collect the device measurement.
 
   libspdm will call this function to retrieve the measurements for a device.
-  The "device_measurement_index" parameter indicates the measurement requested.
+  The "measurement_index" parameter indicates the measurement requested.
 
   @param  measurement_specification     Indicates the measurement specification.
   Must be a SPDM_MEASUREMENT_BLOCK_HEADER_SPECIFICATION_* value in spdm.h.
@@ -158,12 +161,17 @@ typedef boolean (*spdm_psk_hkdf_expand_func)(
   On input, indicates the size in bytes of the destination buffer.
   On output, indicates the total size in bytes of all device measurement
   blocks in the buffer. This field should only be modified if
-  "device_measurement_index" is non-zero.
+  "measurement_index" is non-zero.
 
-  @retval TRUE  the measurement collection succeeded and measurement is returned.
-  @retval FALSE the measurement collection failed.
+  @retval RETURN_SUCCESS             Successfully returned measurement_count,
+                                     measurements, measurements_size.
+  @retval RETURN_BUFFER_TOO_SMALL    measurements buffer too small for measurements.
+  @retval RETURN_INVALID_PARAMETER   Invalid parameter passed to function.
+  @retval RETURN_NOT_FOUND           Unsupported measurement index.
+  @retval RETURN_***                 Any other RETURN_ error from base.h
+                                     indicating the type of failure
 **/
-boolean spdm_measurement_collection(
+return_status spdm_measurement_collection(
 				    IN spdm_version_number_t spdm_version,
 				    IN uint8  measurement_specification,
 				    IN uint32 measurement_hash_algo,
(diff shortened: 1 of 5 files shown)


jyao1:
Good idea to save stack size.

Feedback:
1) 
```
		ASSERT(*measurements_size >= total_size_needed);
		if (total_size_needed >= *measurements_size) {
			return FALSE;
		}
```
Is this typo ? 
Should be "if (total_size_needed > *measurements_size)"

rkongintel:
Yes. Will fix.

rkongintel:
@jyao1 Regarding #3. The reason I return true is because I needed a way to distinguish between an error where **spdm_get_response_measurements** should return **SPDM_ERROR_CODE_UNSPECIFIED** and **SPDM_ERROR_CODE_INVALID_REQUEST**.  The **spdm_measurement_collection** only returns boolean, and I did not want to make it use **return_status**, as it would be the only function in os_stub that would do this except cryptlib. Therefore, in the case where the **measurement_index** is not valid for the device, I made **spdm_measurement_collection** return TRUE, but the **measurement_count** return 0. I will add a comment in the code review to highlight this.  Do you have an alternative suggestion for how else we could fix this?

library/spdm_responder_lib/measurements.c
```diff
@@ -246,266 +247,181 @@ return_status spdm_get_response_measurements(IN void *context,
...
 		spdm_context->connection_info.version,
 		spdm_context->connection_info.algorithm.measurement_spec,
 		spdm_context->connection_info.algorithm.measurement_hash_algo,
-		&device_measurement_count, device_measurement,
-		&device_measurement_size);
+		measurements_index,
+		&measurements_count,
+		measurements,
+		&measurements_size);
+
```
rkongintel, line 303: Since spdm_measurement_collection returns boolean, there is no way to distinguish an UNSPECIFIED error and INVALID_REQUEST error (for measurements_index not being supported by the device). For the later case, I made spdm_measurement_collection return TRUE, but measurement_count == 0.

library/spdm_responder_lib/measurements.c
```diff
@@ -246,266 +247,181 @@ return_status spdm_get_response_measurements(IN void *context,
...
-						slot_id_param;
-				}
-			} else {
-				spdm_create_measurement_opaque(
-					spdm_context, spdm_response,
-					spdm_response_size);
-			}
-		} else {
-			//Block not found
+		if (measurements_count != 1) { //Block not found
```
rkongintel, line 376: This is the case where spdm_measurement_collection returns TRUE but the measurement_index was not valid for the device.  We return SPDM_ERROR_CODE_INVALID_REQUEST instead of SPDM_ERROR_CODE_UNSPECIFIED.

include/library/spdm_device_secret_lib.h
```diff
@@ -107,25 +107,66 @@ typedef boolean (*spdm_psk_hkdf_expand_func)(
 /**
   Collect the device measurement.
 
+  libspdm will call this function to retrieve the measurements for a device.
+  The "device_measurement_index" parameter indicates the measurement requested.
```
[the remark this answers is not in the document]
rkongintel: Thanks.

library/spdm_responder_lib/measurements.c
rkongintel: Agree. Should probably assert measurements count == 1 after this section too.

jyao1:
Basically, you want to use "count = 1" + "return TRUE" to indicate invalidate parameter, and "return FALSE" to indicate internal error.

I also saw code below:
```
	if (measurement_specification !=
	    SPDM_MEASUREMENT_BLOCK_HEADER_SPECIFICATION_DMTF) {
		return FALSE;
```
If we follow same rule, should we return TRUE here as well ?

I still prefer to use a ERROR Code to indicate error. More readable.
Maybe we can change to use return_status, instead of boolean as return code.
```
@retval INVALID_PARAMETER  - SPDM_ERROR_CODE_INVALID_REQUEST
@retval DEVICE_ERROR - SPDM_ERROR_CODE_UNSPECIFIED
@return other - SPDM_ERROR_CODE_UNSPECIFIED
```

rkongintel:
My gut feeling is to have spdm_measurement_collection return a status code instead of boolean. I just searched the rest of os_stub and noticed no other function was using return_status except for cryptlib functions, so I wanted to check with you before exposing this return_status type further, if you had other reasons to keep the return type boolean. But yes, I'd prefer to have a return type that indicates more than true/false as a client needs to tell more information about the failure to the spdm library.

jyao1:
Yes, we should use status_code instead of boolean.

include/library/spdm_device_secret_lib.h
```diff
@@ -23,19 +23,22 @@
...
+  @param  measurements_size        On input, indicates the size in bytes of the destination buffer.
                                        On output, indicates the size in bytes of all device measurement blocks in the buffer.
 
-  @retval TRUE  the device measurement collection success and measurement is returned.
-  @retval FALSE the device measurement collection fail.
+  @retval RETURN_SUCCESS             Successfully returned measurement_count and optionally measurements, measurements_size.
+  @retval RETURN_BUFFER_TOO_SMALL    "measurements" buffer too small for measurements.
+  @retval RETURN_INVALID_PARAMETER   Invalid parameter passed to function.
+  @retval RETURN_***                 Any other RETURN_ error from base.h
 **/
```
rkongintel, line 36: I updated this function pointer typedef to match the spdm_measurement_collection function, but it seems that this function pointer typedef, and a few others like it, are never used. Perhaps we should remove it in another pull request, since it seemed we were interested in removing function pointers for security.

library/spdm_responder_lib/measurements.c
```diff
@@ -301,12 +299,29 @@ return_status spdm_get_response_measurements(IN void *context,
 		measurements,
 		&measurements_size);
 
-	if (!ret) {
-		spdm_generate_error_response(spdm_context,
-					     SPDM_ERROR_CODE_UNSPECIFIED,
-					     0, response_size, response);
-		return RETURN_SUCCESS;
+
```
rkongintel, line 302: I'm a bit unclear on when we should be returning errors and returning success combined with calling spdm_generate_error_response. It seem that sometimes we return errors and set an error response and other times we return success and set an error response. We should clear that up.

jyao1: We should make it consistent in this patch set.

If we want to change, I prefer we change it together with all other response function.

rkongintel: Ok. Then on error, I will go with existing convention and make the spdm_get_response_measurements return RETURN_SUCCESS and error will be generated via spdm_generate_error_response.  There was one piece of code that I did not modify that return RETURN_UNSUPPORTED and calls spdm_generate_error_response, which is what confused me (see below). All other errors return RETURN_SUCCESS when calling spdm_generate_error_reponse. Should this one below be changed to RETURN_SUCCESS?

		session_state = spdm_secured_message_get_session_state(
			session_info->secured_message_context);
		if (session_state != SPDM_SESSION_STATE_ESTABLISHED) {
			spdm_generate_error_response(
				spdm_context,
				SPDM_ERROR_CODE_UNEXPECTED_REQUEST, 0,
				response_size, response);
			return RETURN_UNSUPPORTED;
		}

jyao1:
Looks good to me.

steven-bellock (approved):

Issue #247, opened 2021-11-01
null cryptlib


rkongintel:
Should we provide a null version of cryptlib under os_stub? This would be a copy of cryptlib_mbedtls\cryptlib_openssl, but have just empty top level functions that only assert and return failure. This would perhaps save a step for clients that wish to use their own internal crypto libraries?

jyao1:
Good idea.
It can be a template for cryptolib_pkcs11 in the future (https://github.com/DMTF/libspdm/issues/1)

steven-bellock:
This is also a solution for a Responder that does not support cryptography.

PR #262, opened 2021-11-03
move cryptlib_dummy to os_stub

jyao1:
Fix https://github.com/DMTF/libspdm/issues/247


steven-bellock (changes requested):
Please rename to cryptlib_null to be consistent with the other _null directories.

steven-bellock (approved):

Issue #32, opened 2021-07-07
need upgrade to latest openssl such as 1.1.1l


jyao1:
See https://www.openssl.org/news/vulnerabilities-1.1.1.html
https://www.openssl.org/source/

jyao1:
we can update to latest version before production.

PR #269, opened 2021-11-05
Update the openssllib to 1.1.1l

Wenxing-hou:
REF: #32 


jyao1:
looks good

steven-bellock (approved):

Issue #268, opened 2021-11-05
openssl build will fail the unit test.


jyao1:
[  PASSED  ] 15 test(s).
[  PASSED  ] 29 test(s).
[  PASSED  ] 31 test(s).
[  PASSED  ] 17 test(s).
[  ERROR   ] --- 0x0 != 0x800000000000001a
[   LINE   ] --- C:\home\DMTF\libspdm\unit_test\test_spdm_requester\get_certificate.c:2334: error: Failure!
[  PASSED  ] 18 test(s).
[  FAILED  ] spdm_requester_get_certificate_tests: 1 test(s), listed below:
[  FAILED  ] test_spdm_requester_get_certificate_case19

PR #270, opened 2021-11-05
build with -DCRYPTO=openssl, then test_spdm_requester_get_certificate…

Zhiqiang520:
Fix https://github.com/DMTF/libspdm/issues/268

…_case19 unit test will fail. So comment it temporarily.


jyao1:
This is only a workaround. The formal patch should be to fix issue in openssl wrapper lib - to add time check.

jyao1:
Please 
1) Don't comment code. Remove the code, please.
2) Don't use merge, please rebase your patch to latest main and force push it.

jyao1:
Looks good

steven-bellock (approved):

PR #273, opened 2021-11-09
Add fuzz test case

Xiaohanjlll:


jyao1:
defer after code freeze.

Issue #277, opened 2021-11-09
test_spdm_requester crashes with OpenSSL


steven-bellock:
> [ RUN      ] test_spdm_requester_get_certificate_case1
/home/runner/work/_temp/a6d3e837-cbab-4f86-83fe-737952fe3251.sh: line 2:  7070 Aborted                 (core dumped) ./test_spdm_requester
Error: Process completed with exit code 134.

jyao1:
Have you run "nmake copy_sample_key" ?

steven-bellock:
Yes.

> -- Build files have been written to: /home/runner/work/libspdm/libspdm/build
Built target copy_sample_key
[  0%] Building C object os_stub/cryptlib_openssl/CMakeFiles/cryptlib_openssl.dir/cipher/aead_aes_gcm.c.o
[  0%] Building C object os_stub/openssllib/CMakeFiles/openssllib.dir/openssl/crypto/aes/aes_cbc.c.o

steven-bellock:
Probably related to https://github.com/DMTF/libspdm/pull/270

jyao1:
#270 is fix.

I do not see any problem on my side. Weird.

jyao1:
BTW: Do you do "git submodule update --recursive" ? We just upgrade openssl version.

steven-bellock:
Yes.
Did you check "cmake -DARCH=x64 -DTOOLCHAIN=GCC -DTARGET=Release -DCRYPTO=openssl .." ?

jyao1:
You are right. I only test Debug. but not release.

Yes, I can reproduce the error with Release build now.

jyao1:
Just curious, why CI did not fail before? I do merge only after CI passes.
But anyway, I will investigate it.

Wenxing-hou:
I find the first bad patch to cause the error :_enable time checks for openssl._  [https://github.com/DMTF/libspdm/commit/c8becfca3163f98937a43e0215c9a256f7f9acd8](url)
Before the patch, build with  cmake -DARCH=x64 -DTOOLCHAIN=GCC -DTARGET=<Release|Debug> -DCRYPTO=openssl .. pass ,and pass in the unit_test.

steven-bellock:
That's https://github.com/DMTF/libspdm/pull/270

liyi77:
I found that time() will cause a segfault in there:
[...]

PR #278, opened 2021-11-09
Rename files

steven-bellock:
Signed-off-by: Steven Bellock sbellock@nvidia.com


jyao1:
@steven-bellock. I would like to prove it. But CI fails. Is there any concern?

steven-bellock:
For this pull request it should be fine to merge, since it builds correctly. CI fails due to https://github.com/DMTF/libspdm/issues/277 but that failure is during testing.

PR #288, opened 2021-11-13
Rename functions

steven-bellock:
Rename client-visible functions in spdm_common_lib.h, spdm_requester_lib.h, and spdm_responder_lib.h from spdm_ to libspdm_
@jyao1 this will break spdm-emu so perhaps wait before merging the pull request.

Signed-off by: Steven Bellock, sbellock@nvidia.com


jyao1:
please let me know when you plan to merge.

PR #289, opened 2021-11-15
Fix the bug for test_spdm_requester crashes with OpenSSL in Linux

Wenxing-hou:

Fix: #277 

Update the time_t type to ensure that:
time_t is 4 bytes for 32bit machine and 8 bytes for 64bit machine.

### DIFF
--- a/os_stub/openssllib/include/crt_support.h
+++ b/os_stub/openssllib/include/crt_support.h
@@ -103,7 +103,7 @@
 typedef uintn size_t;
 typedef uintn u_int;
 typedef intn ssize_t;
-typedef int32 time_t;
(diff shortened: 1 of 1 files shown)


jyao1:
please add more comment:
time_t is 4 bytes for 32bit machine and 8 bytes for 64bit machine.

steven-bellock:
> time_t is 4 bytes for 32bit machine and 8 bytes for 64bit machine.
time_t is migrating to 64 bits on all architectures due to https://en.wikipedia.org/wiki/Year_2038_problem
For example see https://musl.libc.org/time64.html

jyao1:
approved. We can see if there is need to change to int64 in the future.

Issue #214, opened 2021-10-19
spdm_send_receive_data() error handling is unable to distingish between a DecryptError from responder and device error


jcahill0:
The spdm_send_receive_data() error handling uses limited return statuses that cause an application to be unable to distinguish between a DecryptError ERROR response from SPDM Responder and a generic error of the device.  The two are important to distinguish because a DecryptError signifies a sufficient error occurred in the session to warrant terminating the session and possibly starting a new session.  A generic device error signifies the application may not want to even talk to the device/SPDM responder, or start just start again at attestation.

Starting at spdm_receive_spdm_response() call, this calls spdm_receive_response() directly without any changing of the error status.
```
    return spdm_receive_response(spdm_context, session_id, FALSE,
                     response_size, response);
```

Next, the spdm_receive_response() call always checks first if the passed in session and received response's session match.  Then it throws RETURN_DEVICE_ERROR regardless of the actual response from the device/SPDM responder:
```
    message_session_id = NULL;
    is_message_app_message = FALSE;
    status = spdm_context->transport_decode_message(
        spdm_context, &message_session_id, &is_message_app_message,
        FALSE, message_size, message, response_size, response);

    if (session_id != NULL) {
        if (message_session_id == NULL) {
            DEBUG((DEBUG_INFO,
                   "spdm_receive_spdm_response[%x] GetSessionId - NULL\n",
                   (session_id != NULL) ? *session_id : 0x0));
            return RETURN_DEVICE_ERROR;
        }
```

One way to obtain the desired result, is to nuance this inner if-statement with checks against the actual message.
```
    message_session_id = NULL;
    is_message_app_message = FALSE;
    status = spdm_context->transport_decode_message(
        spdm_context, &message_session_id, &is_message_app_message,
        FALSE, message_size, message, response_size, response);

    if (session_id != NULL) {
        if (message_session_id == NULL) {
            spdm_message_header_t *spdm_response;
            spdm_response = response;
            DEBUG((DEBUG_INFO,
                   "spdm_receive_spdm_response[%x] GetSessionId - NULL\n",
                   (session_id != NULL) ? *session_id : 0x0));
            if (    spdm_response.header.request_response_code == SPDM_ERROR
                &&  spdm_response.header.param1 == SPDM_ERROR_CODE_DECRYPT_ERROR ) {
               return RETURN_SECURITY_VIOLATION;
            }
            else
            {
                return RETURN_DEVICE_ERROR;
            }
        }
```

[...]

PR #296, opened 2021-11-16
return DecryptError ERROR response rather than generic error

qizhangz:
Fix : #214 

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_send_receive.c
+++ b/library/spdm_requester_lib/libspdm_req_send_receive.c
@@ -110,14 +110,14 @@ return_status libspdm_receive_response(IN void *context, IN uint32_t *session_id
             DEBUG((DEBUG_INFO,
                    "spdm_receive_spdm_response[%x] GetSessionId - NULL\n",
                    (session_id != NULL) ? *session_id : 0x0));
-            return RETURN_DEVICE_ERROR;
+            goto error;
         }
         if (*message_session_id != *session_id) {
             DEBUG((DEBUG_INFO,
                    "spdm_receive_spdm_response[%x] GetSessionId - %x\n",
                    (session_id != NULL) ? *session_id : 0x0,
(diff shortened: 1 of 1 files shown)


jcahill0 (changes requested):
We can get to line 152 even if there is a decryption error on the SPDM requestor side (i.e. decrypting a Secure response from SPDM responder).  Need to check the MsgType before checking the the SPDM request_response_code.  Otherwise we could be interpreting that byte incorrectly because it could be a Secure MsgType response message.

We cannot directly use is_message_app_message because it assumes SPDM message (line 102), and there is no guarantee that the registered transport decode message routine (e.g. spdm_transport_mctp_decode_message) will properly change it to TRUE when a Secure mesage is received (e.g. another thread takes the session down via GET_VERSION which triggers line 275 in libspdm_mctp_common.c).

qizhangz:
Hi, 

I'm a newcomer to SPDM and have a few questions regarding you comments. Thank you!

> We can get to line 152 even if there is a decryption error on the SPDM requestor side (i.e. decrypting a Secure response from SPDM responder). Need to check the MsgType before checking the the SPDM request_response_code. Otherwise we could be interpreting that byte incorrectly because it could be a Secure MsgType response message.

[qizhangz] chech what MesgType in which structure?
> 
> We cannot directly use is_message_app_message because it assumes SPDM message (line 102), and there is no guarantee that the registered transport decode message routine (e.g. spdm_transport_mctp_decode_message) will properly change it to TRUE when a Secure mesage is received (e.g. another thread takes the session down via GET_VERSION which triggers line 275 in libspdm_mctp_common.c).

[qizhangz] If the message is not SPDM message, like an app message or secure message, how check its request_response_code and error code?

jyao1:
Looks good. Merge at first.
@jcahill0, if you have feedback later, please submit a new issue.

PR #306, opened 2021-11-19
Add slot_mask parameter to try_challenge() callers and appropriate un…

jcahill0:
Added slot_mask parameter to try_challenge() callers, added unit test for multiple slots set in the mask, and extended existing unit tests to check on RETURN_SUCCESS; new MBedTLS requires C99 and breaks older compilers that don't use C99 by default, so added "-std=c99" to CMakefile.


jyao1:
recommend to separate c99 change to another P-R.

jyao1:
Please also resolve conflict because of new naming.

jyao1:
@jcahill0 , would you please rebase your patch and submit again according to review feedback ?

jyao1:
@jcahill0 , any chance to rebase this patch in this week?

jyao1:
it is handled by #429

PR #307, opened 2021-11-23
Add appropriate assert for capabbility in spdm_set_data()

Wenxing-hou:
Fix: #205 

### DIFF
--- a/library/spdm_common_lib/libspdm_com_context_data.c
+++ b/library/spdm_common_lib/libspdm_com_context_data.c
@@ -129,6 +129,27 @@ return_status libspdm_set_data(IN void *context, IN spdm_data_type_t data_type,
 		if (data_size != sizeof(uint32)) {
 			return RETURN_INVALID_PARAMETER;
 		}
+
+		#if !SPDM_ENABLE_CAPABILITY_CERT_CAP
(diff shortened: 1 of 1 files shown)


liyi77:
Hi Wenxin, have you test it with spdm_emu?
I think there will be problems when testing it with emu.

Wenxing-hou:
Hi Liyi. I have tested it with spdm_emu. And the test pass. Thanks.

liyi77:
Have you try to disable SPDM_ENABLE_CAPABILITY_MEAS_CAP and test it?
like: #define SPDM_ENABLE_CAPABILITY_MEAS_CAP 0
I found some error at this case, 
BUT it maybe not trigger by this patch.

Wenxing-hou:
Yes, I have tested the disable SPDM_ENABLE_CAPABILITY_XXX_CAP case yesterday morning. Some error is found. But it is not this patch's reason. It is because that spdm-emu call wrong function when disable SPDM_ENABLE_CAPABILITY_XXX_CAP. I will generate a issue in spdm-emu. Thanks.

jyao1:
looks good

steven-bellock (approved):

Issue #246, opened 2021-11-01
The priority table of BaseAsymAlgo and ReqBaseAsymAlg should be same


Zhiqiang520:
Both of BaseAsymAlgo and ReqBaseAsymAlg are asymmetric key signature algorithms, I think their priority tables for asymmetric key signature algorithms selection should be same. But currently they are different(Line44 and Line62).

https://github.com/DMTF/libspdm/blob/68169b89611ee04ec7a8de3a12ae80e97b881dfa/library/spdm_responder_lib/algorithms.c#L44-L78

jyao1:
Most likely, the requester is host. The responder is device. I think it is OK to use different policy.  

Also we are discussing priority at https://github.com/DMTF/libspdm/discussions/121.

Zhiqiang520:
OK, got it. Agree.

jyao1:
11/8/21 - prefer to be same by default. but allow it be overridden to be different.

Wenxing-hou:
I will check it.

PR #308, opened 2021-11-23
Change the default priority table of ReqBaseAsymAlg

Wenxing-hou:
Fix: #246 
The priority table of BaseAsymAlgo and ReqBaseAsymAlg should be same.

ECC is better than RSA. Larger keys are better than smaller keys.


jyao1:
looks good

Wenxing-hou:
I have resolved the conflict.

steven-bellock (approved):

Issue #272, opened 2021-11-08
need a way to pass non-zero timeout value to spdm IO.


jyao1:
Current libspdm always uses 0, which means infinite wait for response.

jcahill0:
Few ideas:

1. Change the API for every command to add a timeout parameter
  a. Brute force and puts a lot of work on the client of the lib
2. Add a per-SPDM context parameter to set the SPDM RTT
  b. libspdm can then more elegantly pass down RTT, and if a crypt-calc is involved, add CT Exponent as appropriate
3. Nuance the timeout parameter to the transport callbacks to do the work of using the right timeout.  This is appropriate in one way as it is much like how RTT and timeouts are defined and layered in the specs, but also inappropriate because it still puts a ton of work on the client that is less elegant than 2 above.  Transport would need to handle
   a. No timeout
   b. RTT timeout
   c. RTT + CT Exponent timeout
   d. Custom timeout
4. A hybrid of 1 and 2 above.  Certain calls should 1 above with RTT and RTT+CT Exponent, e.g. GET_VERSION and CHALLENGE.  Others should use 2 above for a Secure Msg Request that goes out and waits for a Secure Msg Response.

jcahill0:
One other level of decision is whether libspdm will bake in each transport's RTT or the client will do this.  I'm for the client doing it as new transports can come up without SPDM being directly involved: SPDM over MCTP over <X> binding, where <X> is a new binding to MCTP and has a unique MT1 and MT2.

jyao1:
@jcahill0, please let me know if you want to lead the discussion? I would like to assign the owner to you, if you are OK with that.

jyao1:
11/15/2021 - For SPDM message, let user set in the context once.
For SPDM vendor-defined / secure application message, let user input every time.

jyao1:
Also need consider how to add unit test for the timer.

qizhangz:
I will check it

qizhangz:
I made PR to add timeout for SPDM message. Could you please review if it is what you want? Thank you!

https://github.com/DMTF/libspdm/pull/309

steven-bellock:
Will defer to STABLE 0.2.

PR #309, opened 2021-11-23
Add timeout for SPDM message

qizhangz:
Fix: #272

### DIFF
--- a/include/hal/library/sleeplib.h
+++ b/include/hal/library/sleeplib.h
@@ -1,0 +1,22 @@
+/**
+    Copyright Notice:
+    Copyright 2021 DMTF. All rights reserved.
+    License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
+**/
+
+/** @file
+  Provides random number generator services.
+**/
+
+#ifndef __SLEEP_LIB_H__
+#define __SLEEP_LIB_H__
+
+/**
+  Suspends the execution of the current thread until the time-out interval elapses.
+
+  @param[in] milliseconds     The time interval for which execution is to be suspended, in milliseconds.
+
+**/
+void libspdm_sleep(uint64_t milliseconds);
+
+#endif // __SLEEP_LIB_H__
--- a/library/spdm_requester_lib/libspdm_req_challenge.c
+++ b/library/spdm_requester_lib/libspdm_req_challenge.c
@@ -321,6 +321,8 @@ return_status libspdm_challenge(IN void *context, IN uint8_t slot_id,
 	return_status status;
 
 	spdm_context = context;
+	spdm_context->req_timeout = spdm_context->connection_info.capability.rtt;
+	spdm_context->resp_timeout = 2 << spdm_context->connection_info.capability.ct_exponent;
 	retry = spdm_context->retry_times;
 	do {
 		status = try_spdm_challenge(spdm_context, slot_id,
@@ -367,6 +369,8 @@ return_status libspdm_challenge_ex(IN void *context, IN uint8_t slot_id,
 	return_status status;
 
 	spdm_context = context;
+	spdm_context->req_timeout = spdm_context->connection_info.capability.rtt;
+	spdm_context->resp_timeout = 2 << spdm_context->connection_info.capability.ct_exponent;
 	retry = spdm_context->retry_times;
 	do {
 		status = try_spdm_challenge(spdm_context, slot_id,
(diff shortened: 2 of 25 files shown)


library/spdm_requester_lib/libspdm_req_challenge.c
```diff
@@ -5,6 +5,7 @@
 **/
 
 #include "internal/libspdm_requester_lib.h"
+#include <math.h>
```
steven-bellock, line 8: Code in /spdm_requester_lib or /spdm_responder_lib can only use C freestanding headers. And, since this operation is a power-of-two you can use bit-shifts.

library/spdm_requester_lib/libspdm_req_handle_error_response.c
```diff
@@ -145,10 +152,25 @@ return_status spdm_handle_response_not_ready(IN spdm_context_t *spdm_context,
...
-					       response_size, response,
-					       expected_response_code,
-					       expected_response_size);
+	wt = (uint64)pow(2, extend_error_data->rd_exponent)/1000;
+	wt_max = wt * extend_error_data->rd_tm - spdm_context->connection_info.capability.rtt;
+	retry = (wt == 0)? 1 : wt_max / wt;
+	retry = (retry < 1) ? 1 : retry;
+
+	do {
+#ifdef _MSC_VER
```
steven-bellock, line 161: Implementations of system functions belong in os_stub/. This code would call something like libspdm_sleep that would then call into the appropriate system function (Windows, Linux, user-defined, etc). We need to make sure we support different environments, such as RTOSes.

steven-bellock (changes requested):

library/spdm_requester_lib/libspdm_req_encap_request.c
```diff
@@ -5,6 +5,7 @@
 **/
 
 #include "internal/libspdm_requester_lib.h"
+#include <math.h>
```
jyao1, line 8: I don't think we allow math.h

library/spdm_requester_lib/libspdm_req_encap_request.c
```diff
@@ -394,8 +395,15 @@ return_status spdm_encapsulated_request(IN spdm_context_t *spdm_context,
...
-return_status libspdm_send_receive_encap_request(IN void *spdm_context,
+return_status libspdm_send_receive_encap_request(IN void *context,
 					      IN uint32 *session_id)
 {
+	spdm_context_t *spdm_context;
+
+	spdm_context = context;
+	spdm_context->req_timeout = spdm_context->connection_info.capability.rtt;
+	spdm_context->resp_timeout =
+		(uint64)pow(2, spdm_context->connection_info.capability.ct_exponent);
```
jyao1, line 406: I dont think we allow pow()

jyao1:
Hi @steven-bellock , what do you think about sleep_lib?

steven-bellock:
> Hi @steven-bellock , what do you think about sleep_lib?

Ah sorry. I did not see that it had been updated. I'll have a look today.

library/spdm_requester_lib/libspdm_req_challenge.c
```diff
@@ -321,6 +321,8 @@ return_status libspdm_challenge(IN void *context, IN uint8_t slot_id,
 	return_status status;
 
 	spdm_context = context;
+	spdm_context->req_timeout = spdm_context->connection_info.capability.rtt;
```
steven-bellock, line 324: This req_timeout isn't in the specification, but it would be the maximum amount of time the libspdm integrator can tolerate before a request is accepted by the Responder. It's not RTT.

include/hal/library/sleeplib.h
```diff
@@ -0,0 +1,22 @@
+/**
+    Copyright Notice:
+    Copyright 2021 DMTF. All rights reserved.
+    License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
+**/
+
+/** @file
+  Provides random number generator services.
```
steven-bellock, line 8: Delete.

steven-bellock (changes requested):
@qizhangz @jyao1 @jcahill0 I think we should produce a rough design before coding can begin. In particular how the asynchronous wait case would be handled and how all this would be tested.

qizhangz:
> @qizhangz @jyao1 @jcahill0 I think we should produce a rough design before coding can begin. In particular how the asynchronous wait case would be handled and how all this would be tested.

I will reconsider the implementation as all comments. So I close this PR and will submit a new PR for the new implementation.
Thank you!

Issue #290, opened 2021-11-15
'dhe_context' should be checked whether it is a NULL Pointer before subsequent using.


Zhiqiang520:
`dhe_context` returned in following line 133, it may be a NULL Pointer.
It should be checked whether it is a NULL Pointer before subsequent using.
But this pointer `dhe_context` was directly used in the following line 137.

https://github.com/DMTF/libspdm/blob/063a6e7c26432d5adfa6a273a232fbdfa519532d/library/spdm_requester_lib/libspdm_req_key_exchange.c#L133-L137

jyao1:
11/15/2021 - approved.

Wenxing-hou:
I will check it.

PR #310, opened 2021-11-24
Add dhe_context check for NULL pointer return

Wenxing-hou:
Fix: #290 

'dhe_context' should be checked whether it is a NULL Pointer before subsequent using


jyao1:
looks good

steven-bellock (approved):

PR #316, opened 2021-11-25
Add fuzzing automation test script

Zhiqiang520:
Add fuzzing automation test script

### DIFF
--- a/unit_test/fuzzing/fuzzing.sh
+++ b/unit_test/fuzzing/fuzzing.sh
@@ -1,0 +1,187 @@
+#!/bin/bash
+
+# Before run this script, please Download and install AFL. Unzip and follow docs\QuickStartGuide.txt. Build it with make.
+# If command 'screen' not found, please install with: sudo apt install screen.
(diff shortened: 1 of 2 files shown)


unit_test/fuzzing/fuzzing.sh
```diff
@@ -0,0 +1,157 @@
...
+if [[ $2 = "ON" ]]; then
+    cd $fuzzing_out
+    mkdir coverage_log
+    cd coverage_log
+    lcov --capture --directory $libspdm_path --output-file coverage.info
+    genhtml coverage.info --output-directory .
+fi
+
+cd $script_path
+python3 ./summary.py $fuzzing_out
```
jyao1, line 157: Do we must use python? Can we just use normal shell script?
If have to, then we need document python requirement, e.g python3 version.

Zhiqiang520: Python is not necessary, and we can just use the normal shell script. I have updated.

jyao1: Thank you!

jyao1:
looks good

Issue #305, opened 2021-11-19
test_spdm_requester_get_version fuzzing test ASSERT and hang


Zhiqiang520:
Verified on 3ff9685
1. For x64 builds, use a x64 Native Tools Command Prompt for Visual Studio... command prompt.
cd libspdm
mkdir build_x64_Debug_mbedtls
cd build_x64_Debug_mbedtls
cmake -G"NMake Makefiles" -DARCH=x64 -DTOOLCHAIN=VS2019 -DTARGET=Debug -DCRYPTO=mbedtls ..
nmake copy_sample_key
nmake
cd bin
test_spdm_requester_get_version.exe **testinput_requester_hang**
spdm_send_spdm_request[0] (0x4):
0000: 10 84 00 00
**ASSERT**: C:\libspdm\unit_test\spdm_transport_test_lib\test.c(157): ((transport_message_size - sizeof(test_message_header_t)) & (alignment - 1)) == 0
^C
C:\libspdm\build_x64_Debug_mbedtls\bin>

2. The content of binary file "**testinput_requester_hang**":
![image](https://user-images.githubusercontent.com/90754803/142607610-bb58dcab-e8f6-46e9-af1c-4864dda66422.png)

3. The following line 125 alignment = 4, and the following line 157 ASSERT the length of "Successful VERSION response message" should be multiples of 4.
https://github.com/DMTF/libspdm/blob/3ff968553368dd5517ca8143cc620f48c66e65f3/unit_test/spdm_transport_test_lib/test.c#L125-L157

3. In clause '10.2 GET_VERSION request and VERSION response messages', paragraph 166 'Successful VERSION response message format', the Offset 6 Field "VersionNumberEntry1:\<n\>". the length of "Successful VERSION response message" maybe not multiples of 4.

qizhangz:
I will check this issue.

qizhangz:
It seems not an issue.  The content of of binary file "testinput_requester_hang" need to be 4 byte aligned according the encode function test_decode_message (). So the content should be as the following with two byte NULL appended.

0000: 10 04 00 00 00 04 11 11 11 11 11 11 11 11 00 00

jyao1:
11/22/21 - approved. Need fix the fuzzing test case harness.

Xiaohanjlll:
I will check this issue.

Issue #281, opened 2021-11-10
Get_capabilities fuzzing test folder name typo


Zhiqiang520:
https://github.com/DMTF/libspdm/blob/main/unit_test/fuzzing/test_spdm_responder_get_capabilities/get_capabilities.c

I think the above folder name `test_spdm_responder_get_capabilities` should be  `test_spdm_requester_get_capabilities`, because `get_capabilities` is requester side message.

jyao1:
11/15/2021 - approved

Xiaohanjlll:
I will check this issue.

PR #323, opened 2021-11-30
Add and update fuzzy tests

Xiaohanjlll:
Fix: #305 
Fix: #281 


unit_test/fuzzing/test_spdm_requester_encap_key_update/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
+cmake_minimum_required(VERSION 2.6)
+INCLUDE_DIRECTORIES(${LIBSPDM_DIR}/unit_test/fuzzing/test_spdm_requester_encap_key_update
+                    ${LIBSPDM_DIR}/include
+                    ${LIBSPDM_DIR}/include
```
steven-bellock, line 4: Delete redundant line.

steven-bellock (changes requested):
You can delete all instances of ${LIBSPDM_DIR}/include/hal  That is no longer required as an include path.

unit_test/fuzzing/test_spdm_requester_encap_key_update/CMakeLists.txt
Xiaohanjlll: Thanks.I have deleted redundant line.

steven-bellock (approved):

Issue #328, opened 2021-12-01
Build error with CLANG


Wenxing-hou:
When build libspdm with CLANG toolchain, the error is found:
```
/usr/bin/ld: ../../lib/libmemlib.a: error adding symbols: archive has no index; run ranlib to add one
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [unit_test/test_spdm_requester/CMakeFiles/test_spdm_requester.dir/build.make:550: bin/test_spdm_requester] Error 1
make[1]: *** [CMakeFiles/Makefile2:1120: unit_test/test_spdm_requester/CMakeFiles/test_spdm_requester.dir/all] Error 2
make: *** [Makefile:84: all] Error 2
```

PR #330, opened 2021-12-02
Fix build error with clang

Wenxing-hou:
Fix: #328 
Fix build error with CLANG


jyao1:
looks good to me.

steven-bellock (approved):

Issue #331, opened 2021-12-02
openssl build failed with AFL in Linux


Zhiqiang520:
openssl build failed with AFL in Linux.
Verified on https://github.com/DMTF/libspdm/commit/09a5f5c62ebc41f9a00e573f92615247c798d3a5
Build command:
```
cd libspdm
mkdir build
cd build
cmake -DARCH=x64 -DTOOLCHAIN=AFL -DTARGET=Release -DCRYPTO=openssl -DGCOV=ON ..
make copy_sample_key
make
```

Build error:
[ 37%] Building C object os_stub/openssllib/CMakeFiles/openssllib.dir/openssl/crypto/o_str.c.o
afl-cc 2.52b by <lcamtuf@google.com>
/home/test/libspdm/os_stub/openssllib/openssl/crypto/o_str.c: In function ‘openssl_strerror_r’:
/home/test/libspdm/os_stub/openssllib/openssl/crypto/o_str.c:253:13: error: implicit declaration of function ‘strerror_r’; did you mean ‘strerror’? [-Werror=implicit-function-declaration]
  253 |     return !strerror_r(errnum, buf, buflen);
      |             ^~~~~~~~~~
      |             strerror
cc1: all warnings being treated as errors
make[2]: *** [os_stub/openssllib/CMakeFiles/openssllib.dir/build.make:3911: os_stub/openssllib/CMakeFiles/openssllib.dir/openssl/crypto/o_str.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1451: os_stub/openssllib/CMakeFiles/openssllib.dir/all] Error 2
make: *** [Makefile:84: all] Error 2

steven-bellock:
Ah sorry about this. You need to add -std=c99 to the CmakeFile

Zhiqiang520:
Yes, if following https://github.com/DMTF/libspdm/commit/9d375c321e94df0eb02c155595a4886f09bd08e7 and add `-std=c99` to CMAKE_C_FLAGS  for AFL TOOLCHAIN in the CmakeFile, then build will pass.

PR #332, opened 2021-12-02
fix the openssl build fail issue with AFL in Linux

Zhiqiang520:
Fix the openssl build fail issue with AFL in Linux.
Fix: #331


jyao1:
Suggested by @steven-bellock 

Looks good to me.

Issue #341, opened 2021-12-03
build error cannot find -lrnglib_std


qizhangz:
fuzzing test is not updated with rnglib instead of rnglib_std

/usr/bin/ld: cannot find -lrnglib_std
collect2: error: ld returned 1 exit status
make[2]: *** [unit_test/fuzzing/test_spdm_requester_get_capabilities/CMakeFiles/test_spdm_requester_get_capabilities.dir/build.make:145: bin/test_spdm_requester_get_capabilities] Error 1
make[1]: *** [CMakeFiles/Makefile2:2126: unit_test/fuzzing/test_spdm_requester_get_capabilities/CMakeFiles/test_spdm_requester_get_capabilities.dir/all] Error 2
make: *** [Makefile:104: all] Error 2

qizhangz:
I will check this

qizhangz:
PR raised:
https://github.com/DMTF/libspdm/pull/342

PR #342, opened 2021-12-03
update fuzzing with rnglib instead of rnglib_std

qizhangz:
Fix: #341 


jyao1:
looks good

steven-bellock (approved):

Issue #338, opened 2021-12-02
test_spdm_responder_algorithms Fuzzing test ASSERT and hang 


Zhiqiang520:
Verified on 3e0e1b9
```
cd libspdm
mkdir build
cd build
cmake -DARCH=x64 -DTOOLCHAIN=AFL -DTARGET=Release -DCRYPTO=mbedtls -DGCOV=ON ..
make copy_sample_key
make
cd bin
~/libspdm/build/bin$ ./test_spdm_responder_algorithms ../../unit_test/fuzzing/seeds/test_spdm_responder_algorithms/algorithms.raw
ASSERT: /home/test/libspdm/library/spdm_responder_lib/libspdm_rsp_algorithms.c(191): !(((spdm_context->local_context.capability.flags & (0x00000008 | 0x00000010)) == 0) ^ (spdm_context->local_context.algorithm.measurement_spec == 0))
^C
```

qizhangz:
I will check this

Xiaohanjlll:
The fuzzy test case does not match the new code, version 0c9e989cd19ce94469b637814b69b7e9c95c3f2a

PR #347, opened 2021-12-03
Fix responder_algorithms Fuzzing test ASSERT and hang

Xiaohanjlll:
Fix: #338 


jyao1:
looks good

Issue #98, opened 2021-08-23
Unneeded and unwanted #includes


steven-bellock:
https://github.com/DMTF/libspdm/blob/8daea2f4f61673ca39ef8a10100917c483e1125c/library/spdm_requester_lib/get_capabilities.c#L8

jyao1:
agree

steven-bellock:
Also 

https://github.com/DMTF/libspdm/blob/77c33789864cdd55d2738ff64ecf5ce80db95eb2/os_stub/cryptlib_mbedtls/sys_call/mem_allocation.c#L14

PR #348, opened 2021-12-03
Remove / rename unwanted #includes of stdio.h

steven-bellock:
Fixes https://github.com/DMTF/libspdm/issues/98


steven-bellock:
Ah I guess each commit needs a "Signed-off-by: Steven Bellock sbellock@nvidia.com" rather than just a single one for the pull request?

jyao1:
Right, add signed-off for every commit, please.

steven-bellock:
Will make a new pull request. Third time is the charm.

PR #355, opened 2021-12-06
Update cmake.yml

Wenxing-hou:
Update cmake.yml to automatic build check every PR for main branch.


steven-bellock (approved):
We can use this for now to help with pull requests but I would like to add workflow_dispatch back in later.

Issue #356, opened 2021-12-06
Should not always assume Random method will return success


qizhangz:
random_bytes() is a wrapper of openssl or mbedtls's random method, and it is a bool function, it might fail. But in spdm_get_random_number(), it is defined as a void function, and ignore the return value from random_bytes(). We should not always assume the random function return success. If attackers exhaust the rand entropy , the random function fill the random buffer with all zero or some error message, but we treat it as a random value and used in our crypto code, that will make security problems. 

```
./library/spdm_crypt_lib/libspdm_crypt_crypt.c line 2502: 
void spdm_get_random_number(IN uintn size, OUT uint8 *rand)
{
        random_bytes(rand, size);
        return;
}
```

qizhangz:
I will check this.

PR #357, opened 2021-12-06
Add check for return value of spdm_get_random_number().

qizhangz:
Fix: #356 

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_measurements.c
+++ b/library/spdm_responder_lib/libspdm_rsp_measurements.c
@@ -42,7 +42,9 @@ boolean spdm_create_measurement_signature(IN spdm_context_t *spdm_context,
     ptr = (void *)((uint8_t *)response_message + response_message_size -
                measurment_sig_size);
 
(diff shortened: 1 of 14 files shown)


library/spdm_responder_lib/libspdm_rsp_measurements.c
```diff
@@ -83,7 +85,9 @@ void spdm_create_measurement_opaque(IN spdm_context_t *spdm_context,
 	ptr = (void *)((uint8_t *)response_message + response_message_size -
 		       measurment_no_sig_size);
 
-	spdm_get_random_number(SPDM_NONCE_SIZE, ptr);
+	if(!spdm_get_random_number(SPDM_NONCE_SIZE, ptr)) {
```
steven-bellock, line 88: It looks like spdm_create_measurement_opaque needs a return value, but we can do that in a separate pull request.

steven-bellock (approved):

jyao1:
ref: https://github.com/DMTF/libspdm/issues/360

PR #365, opened 2021-12-07
Add CI for format check and mbedtls build check

Wenxing-hou:
Add format check for later patch:
If the later patch include Tab, the check will not pass.
And the check will skip on openssl/mbedtls/cmocka/README.

Add build with mbedtls :
Add mbedtls build to CI check.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -6,11 +6,31 @@ on:
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [ main ]
+    #Changes to the following files do not trigger deployment
+    paths-ignore:
+     - README.md
+     - 'os_stub/openssllib/openssl/**'
+     - 'os_stub/mbedtlslib/mbedtls/**'
+     - 'unit_test/cmockalib/cmocka/**'
 
 env:
   BUILD_TYPE: Release
 
 jobs:
+  format_check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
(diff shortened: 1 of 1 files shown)


.github/workflows/cmake.yml
```diff
@@ -6,12 +6,32 @@ on:
...
-  build:
+  format_check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
+      - name: Check code format
```
steven-bellock, line 28: Please add a comment that this is checking for tabs. This will work for now, but later on we'll probably want something more sophisticated, and we may have files where tabs are required (Python, YAML, etc).

.github/workflows/cmake.yml
```diff
@@ -36,3 +56,29 @@ jobs:
...
+
+      - name: Test Requester
+        run: |
+          cd build/bin
+          ./test_spdm_requester
+
+      - name: Test Responder
+        run: |
+          cd build/bin
+          ./test_spdm_responder
```
steven-bellock, line 84: Add newline at EOF to appease Git / Github.

steven-bellock (changes requested):

.github/workflows/cmake.yml
Wenxing-hou: Thanks. I have added the new line.

.github/workflows/cmake.yml
Wenxing-hou: Thanks. I have added a comment. And I have made the Tab check better, which the check only run on *.c/*.h file. And the check will skip on openssl/mbedtls/cmocka.

steven-bellock (approved):

Issue #351, opened 2021-12-04
Stack space in spdm_responder_dispatch_message


richkong88:
Seems like we use significant stack space in the responder dispatch code path. In just reviewing the code (not measuring programmatically), I estimate we can use 22K+ of stack space. This is due to multiple large buffers of size MAX_SPDM_MESSAGE_BUFFER_SIZE (0x1200).

This is the flow I see and the stack space used at each level of the stack:

spdm_responder_dispatch_message 9K
-----ctxt->receive_message
---------spdm_process_message
-------------spdm_process_request
-----------------ctxt->transport_decode_message (spdm_transport_mctp_decode_message) 9K
-------------spdm_build_response 4.5K
-----------------ctxt->transport_encode_message (spdm_transport_mctp_encode_message) 9K
-----ctxt->send_message

Total stack usage from spdm_responder_dispatch_message -> spdm_build_response->transport_encode_message
is (9K + 4.5K + 9K). It would be good if someone else can confirm this flow is using that much.

This is just for tracking now.

jyao1:
I think you are probably right.
We need do some optimization work to make it smaller. The basic idea is to preserve the header.

qizhangz:
I will check this

steven-bellock:
Right now will defer to STABLE 0.2. It might make it into 0.1.

qizhangz:
With PR #409, the stack is reduced by 20K in the responder dispatch code path. Context size is increased by 12K.
Can this be accepted?

jyao1:
@rkongintel , would you please review it?

rkongintel:
@jyao1 Yes. I will review.

jyao1:
continue working on it after #409.

jyao1:
Some idea for the buffer management.

We need provide option to allow external (libspdm integrator) to provides the buffer to libspdm - to achieve zero copy.
1) libspdm integrator provides `send_buffer` to libspdm - libspdm can manage the pointer inside of `send_buffer`.
2) libspdm integrator provides `receive_buffer` to libspdm - libspdm can manage the pointer inside of `receive_buffer`.
3) libspdm integrator provides `scratch_buffer` to libspdm - libspdm can manage the pointer inside of `scratch_buffer`. Sizeof `scratch_buffer` can be calculated based upon transport capability.

The usage of scratch buffer to hold a copy of messages, such as decrypted send_buffer, decrypted receive_buffer, last_message.

ekohandel:
[...]

PR #371, opened 2021-12-07
optimize stack in the responder dispatch code path

qizhangz:
Fix: #351


include/internal/libspdm_common_lib.h
```diff
@@ -81,6 +81,12 @@ typedef struct {
     boolean basic_mut_auth_requested;
     uint8_t mut_auth_requested;
     uint8_t heartbeat_period;
+
+    void *request_buffer;
+    void *response_buffer;
+
+    void *secured_message;
+    void *app_message;
```
jyao1, line 89: This is internal data structure. We cannot use malloc() in emu.

Issue #75, opened 2021-08-10
Many crypto (and crypto reliant) functions do not have their return values properly checked.


taprinz:
Many crypto functions, or functions that rely on crypto, have their return values entirely ignored. Discovered this when playing around with GCC's __attribute__((warn_unused_result).

Especially as we leave the crypto & other implementation's up to the user to provide, we should be sure that we check the return values, as a user's error may be quietly ignored. 

Crypto functions:
- random_seed
- random_bytes
- spdm_hash_all
- spdm_hmac_all_with_response_finished_key
- spdm_hmac_all_with_request_finished_key

Crypto-reliant functions:
- spdm_generate_cert_chain_hash
- spdm_secured_message_dhe_generate_key
- spdm_create_update_session_data_key
- spdm_activate_update_session_data_key
- spdm_generate_finished_key
- spdm_generate_aead_key_and_iv

There are other functions that do not have their return values properly checked - I limited this issue to crypto functions in order to try and limit the scope. I may have missed others as well, as manually adding __attribute__((warn_unused_result)) isn't a perfect method.

jyao1:
Agree.

jyao1:
random function issue is also reported at https://github.com/DMTF/libspdm/issues/65

taprinz:
I have started work on this again. My plan was to check every function with a return value to ensure it does not have any case where the return value isn't checked. Currently working via GCC's \_\_attribute\_\_((warn_unused_result)) - unfortunately, has to be done manually.

My thinking for doing so for all functions is as such: if a function is capable of failing, regardless of how unlikely it is, we NEED to be checking its return value. If we think checking the return of a given function is unnecessary, that function should return void and be incapable of failing by design.

taprinz:
To follow up on the above, I can limit the scope to just crypto & adjacent functions. But I would really argue for the need to handle all return values.

jyao1:
Agree - "if a function is capable of failing, regardless of how unlikely it is, we NEED to be checking its return value."

taprinz:
I'm going by, header by header, adding the GCC attribute to all functions. Currently, only include/library/spdm_crypt_lib.h is remaining as last file to process.

Then, need to address:
- spdm_hmac_*with*_finished_key
- append_managed_buffer

The reason for the above three is that I need to be careful with the currently missing retvals, as they might need to handle structure deallocation (like spdm_hash_free) on failure. Want to make sure I don't miss anything.

[...]

PR #375, opened 2021-12-08
First batch of fixes for missing return value checks

taprinz:
First round of fixes for #75 .

Going to do these in batches since there's a lot of ground to cover.
This fixes (most) functions that had their return values not checked in:
- include/internal/libspdm_common_lib.h
- include/internal/libspdm_requester_lib.h
- include/internal/libspdm_responder_lib.h
- include/internal/libspdm_secured_message_lib.h
- include/library/spdm_secured_message_lib.h

There's a few omissions from the above two that I'll handle in a separate follow up:
- spdm_hmac_\*_with_\*_finished_key
- append_managed_buffer

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_encap_challenge_auth.c
+++ b/library/spdm_requester_lib/libspdm_req_encap_challenge_auth.c
@@ -112,10 +112,9 @@ return_status spdm_get_encap_response_challenge_auth(
     ptr = (void *)(spdm_response + 1);
     result = spdm_generate_cert_chain_hash(spdm_context, slot_id, ptr);
     if (!result) {
-        libspdm_generate_encap_error_response(
+        return libspdm_generate_encap_error_response(
             spdm_context, SPDM_ERROR_CODE_UNSPECIFIED, 0,
             response_size, response);
-        return RETURN_SUCCESS;
     }
     ptr += hash_size;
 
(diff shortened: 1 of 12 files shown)


taprinz:
Build failure is due to Requester KEY_UPDATE tests failing now that certain functions can report failure.
Test cases 28, 29, 30, 31, 32, 34.

I'm looking to patch these up - but is anyone familiar with these tests specifically in general? I want to make sure I'm correctly identifying that we're failing earlier or with a different error code than expected. And not that these failures are functions that are unexpectedly broken.

jyao1:
> Build failure is due to Requester KEY_UPDATE tests failing now that certain functions can report failure.
> Test cases 28, 29, 30, 31, 32, 34.

Do you mean the test failure before your patch? or the failure is caused by your patch?

taprinz:
I can defer review of this portion of fixes until I am able to investigate those test failures a bit more.

Re: your second comment, moving discussion to #75, so we can discuss the scope of expected changes.

taprinz:
Investigated test failures. Was unintentional behavior on my end - I was clobbering a status in one scenario, that was expected to be already false. The clobbered status was then successful, causing the error to be ignored (ironically).

Anyways, I have pushed the fix.

jyao1 (approved):

PR #384, opened 2021-12-09
Rename OPAQUE_CONTEXT_DATA to APP_CONTEXT_DATA.

### LINKED ISSUE #274: the name SPDM_DATA_OPAQUE_CONTEXT_DATA is confusing
Because it includes "OPAQUE", people may think it is used to set the "Opaque data" for measurement/challenge/keyExchange etc.

I think we need replace "opaque" with some other name.

jyao1:
Fix #274.

The reason is that "opaque" is used in SPDM spec with special meaning.
Here we just need a pointer for application, not related to opaque in SPDM spec.

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -345,9 +345,9 @@ typedef struct {
     uint8_t retry_times;
 
     //
-    // Opaque context data for use by application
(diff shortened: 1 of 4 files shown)


steven-bellock:
Eventually we will want to have examples of how this might be used.

Issue #367, opened 2021-12-07
fuzzing in Linux with LLVM LibFuzzer build error


Zhiqiang520:
fuzzing in Linux with LLVM LibFuzzer build error.
Verified on 1edab81
Ubuntu 20.04.2 LTS
Download [LLVM 13.0.0](https://github.com/llvm/llvm-project/releases/download/llvmorg-13.0.0/clang+llvm-13.0.0-x86_64-linux-gnu-ubuntu-20.04.tar.xz) and extract it.
` tar -xvf clang+llvm-13.0.0-x86_64-linux-gnu-ubuntu-20.04.tar.xz`
```
cd libspdm
mkdir build_llvm
cd build_llvm
export CLANG_PATH=~/clang+llvm-13.0.0-x86_64-linux-gnu-ubuntu-20.04/bin
export ASAN_SYMBOLIZER_PATH=$CLANG_PATH/llvm-symbolizer
export PATH=$PATH:~/clang+llvm-13.0.0-x86_64-linux-gnu-ubuntu-20.04/bin
cmake -DARCH=x64 -DTOOLCHAIN=LIBFUZZER -DTARGET=Release -DCRYPTO=mbedtls ..
make copy_sample_key
make
[ 45%] Linking C executable ../../bin/test_spdm_requester
/usr/bin/ld: CMakeFiles/test_spdm_requester.dir/test_spdm_requester.c.o: in function `main':
/home/test/libspdm/unit_test/test_spdm_requester/test_spdm_requester.c:42: multiple definition of `main'; /home/test/clang+llvm-13.0.0-x86_64-linux-gnu-ubuntu-20.04/lib/clang/13.0.0/lib/linux/libclang_rt.fuzzer-x86_64.a(fuzzer.o):/home/brian/src/llvm_releases/llvm-project/llvm/utils/release/final/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:19: first defined here
/usr/bin/ld: /home/test/clang+llvm-13.0.0-x86_64-linux-gnu-ubuntu-20.04/lib/clang/13.0.0/lib/linux/libclang_rt.fuzzer-x86_64.a(fuzzer.o): in function `main':
/home/brian/src/llvm_releases/llvm-project/llvm/utils/release/final/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:20: undefined reference to `LLVMFuzzerTestOneInput'
[...]

PR #392, opened 2021-12-10
Fix libfuzzer build error in Linux

Wenxing-hou:
Fix: #367 

Fix build error for libfuzzer in Linux
Because the main function has been defined in libfuzzer and the libfuzzer doesn't allow multiple definition of main function, the switch is added to fix the error.

Update test.md for Linux libfuzzer build.


steven-bellock:
My commit is to fix a small typo. Didn't want to send it back to @Wenxing-hou just for that.

steven-bellock (approved):

jyao1:
thank you very much @steven-bellock

PR #404, opened 2021-12-13
[spdm_requester_lib] fixed potential buffer overflow

shmarovfedor:
In function `spdm_negotiate_connection_version` there is a potential buffer overflow (line 115) where the indexing variable (`res_index`) is incremented before the array access.

This bug was found by ESBMC as a part of our work of applying formal verification to libspdm.


jyao1:
CI fails. Please add your signed-off.

steven-bellock (approved):

PR #409, opened 2021-12-14
optimize stack in the responder dispatch code path.

qizhangz:
Fix: #351 

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -353,6 +353,15 @@ typedef struct {
     /* Register for the last KEY_UPDATE token and operation (responder only)*/
     
     uint8_t last_update_request[4];
+
+    //
+    // reserved for request and response in the main
+    // dispatch function in SPDM responder
+    // this buffer is the transport message recived from spdm_context->receive_message()
+    // or sent to spdm_context->send_message(). This message may be SPDM transport message
+    // or secured SPDM transport message.
+    //
+    uint8_t request_response[LIBSPDM_MAX_MESSAGE_BUFFER_SIZE];
 } spdm_context_t;
 
 /**
--- a/include/library/spdm_responder_lib.h
+++ b/include/library/spdm_responder_lib.h
@@ -103,6 +103,9 @@ return_status libspdm_build_response(IN void *spdm_context, IN uint32_t *session
   The alternative is: an SPDM responder may receive the request message directly
   and call this function to process it, then send the response message.
 
+  Note: request and response may point to the same buffer. The implementation^M
+  must handle that case.
+
   @param  spdm_context                  A pointer to the SPDM context.
   @param  session_id                    Indicates if it is a secured message protected via SPDM session.
                                        If *session_id is NULL, it is a normal message.
@@ -245,3 +248,4 @@ return_status libspdm_register_connection_state_callback_func(
 void libspdm_init_key_update_encap_state(IN void *spdm_context);
 
 #endif
+
--- a/library/spdm_responder_lib/libspdm_rsp_communication.c
+++ b/library/spdm_responder_lib/libspdm_rsp_communication.c
@@ -16,6 +16,9 @@
   The alternative is: an SPDM responder may receive the request message directly
   and call this function to process it, then send the response message.
 
+  Note: request and response may point to the same buffer. The implementation
+  must handle that case.
+
   @param  spdm_context                  A pointer to the SPDM context.
   @param  session_id                    Indicates if it is a secured message protected via SPDM session.
                                        If *session_id is NULL, it is a normal message.
@@ -41,6 +44,8 @@ return_status libspdm_process_message(IN void *context, IN OUT uint32_t **sessio
     return_status status;
     spdm_context_t *spdm_context;
     boolean is_app_message;
+    uint32_t tmp_session_id;
+    uint32_t *session_id_ptr;
 
     spdm_context = context;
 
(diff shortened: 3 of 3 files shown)


include/internal/libspdm_common_lib.h
```diff
@@ -285,7 +309,7 @@ typedef struct {
     // Cached plain text command
     // If the command is cipher text, decrypt then cache it.
     //
-    uint8_t last_spdm_request[MAX_SPDM_MESSAGE_BUFFER_SIZE];
+    void *last_spdm_request;
```
jyao1, line 312: Please comment where it points to.

include/internal/libspdm_common_lib.h
```diff
@@ -265,6 +265,30 @@ typedef struct {
...
+    // be SPDM transport message or secured SPDM transport message.
+    //
+    uint8_t request[MAX_SPDM_MESSAGE_BUFFER_SIZE];
+    //
+    // reserved bufeer for response in the main dispatch function
+    // in SPDM responder. This buffer is the transport message
+    // sent to spdm_context->send_message(). This message may
+    // be SPDM transport message or secured SPDM transport message.
+    //
+    uint8_t response[MAX_SPDM_MESSAGE_BUFFER_SIZE];
```
jyao1, line 281: Please describe why we cannot merge those two buffer.

include/internal/libspdm_common_lib.h
```diff
@@ -265,6 +265,30 @@ typedef struct {
...
+    //
+    // reserved buffer for secured_message in decode/encode
+    // function of transport layer message.
+    // This buffer is encrypted message in encode or decode function.
+    //
+    uint8_t secured_message[MAX_SPDM_MESSAGE_BUFFER_SIZE];
+
+    // This message is decrypted message in encode or decode function.
+    uint8_t app_message[MAX_SPDM_MESSAGE_BUFFER_SIZE];
```
jyao1, line 290: Please follow the other comment to describe this data structure.

jyao1:
Question: It seems only MCTP transport is updated. Do we also need change PCI_DOE ?

library/spdm_transport_mctp_lib/libspdm_mctp_mctp.c
```diff
@@ -59,14 +62,28 @@ uint32_t spdm_mctp_get_max_random_number_count(void)
...
                   OUT void *transport_message)
 {
     uintn aligned_message_size;
     uintn alignment;
     mctp_message_header_t *mctp_message_header;
+    void *tmp_buffer;
+    uintn tmp_buffer_size;
+
+    // To requester message, use APP_message as tmp buffer to save message.
+    // because message and transport_message use the same buffer.
```
jyao1, line 79: This is tricky, and documented in API header.
Or should the caller avoid this?

library/spdm_transport_mctp_lib/libspdm_mctp_mctp.c
```diff
@@ -59,14 +62,28 @@ uint32_t spdm_mctp_get_max_random_number_count(void)
...
                   IN OUT uintn *transport_message_size,
                   OUT void *transport_message)
 {
     uintn aligned_message_size;
     uintn alignment;
     mctp_message_header_t *mctp_message_header;
+    void *tmp_buffer;
+    uintn tmp_buffer_size;
+
+    // To requester message, use APP_message as tmp buffer to save message.
```
jyao1, line 78: "to requester message"
Do you mean a responder message?

jyao1 (approved):

include/internal/libspdm_common_lib.h
```diff
@@ -265,6 +265,37 @@ typedef struct {
 #define spdm_context_struct_VERSION 0x1
 
 typedef struct {
+    //
+    // reserved buffer for request in the main dispatch function
+    // in SPDM responder. This buffer is the transport message
+    // recived from spdm_context->receive_message(). This message may
+    // be SPDM transport message or secured SPDM transport message.
+    // This buffer may be shared by last_spdm_request.
```
jyao1, line 273: I think sharing with last_spdm_request may be problematic.

In SPDM 1.2, request message is transport layer message, while last_spdm_request is SPDM message, which might be larger than transport layer message if CHUNK is supported.

Can we copy it to app_message ?

library/spdm_transport_mctp_lib/libspdm_mctp_mctp.c
```diff
@@ -149,8 +179,17 @@ return_status mctp_decode_message(OUT uint32_t **session_id,
         if (session_id != NULL) {
             *session_id = NULL;
         }
+        // To requester message, pass the SPDM message to last_spdm_request.
+        if (is_requester) {
+            libspdm_set_last_spdm_request(spdm_context, (uint8_t *)transport_message + sizeof(mctp_message_header_t),
```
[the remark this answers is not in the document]
rkongintel: I agree this set of changes is overly complicated. 
Maybe should rethink about an alternate method.
Perhaps as a first pass, it is enough to move request/response buffers to context, and keep extra working temp buffers (secured/app buffers) on the stack.

rkongintel:
I think this may be too complicated and error prone. 

Perhaps we should just start with moving just the request/response to context and leave other code the same?

jyao1:
@rkongintel, I agree. I have same feeling.

For SPDM 1.2, the TransportMessageSize != MaxSPDMMessageSize. We need rework the flow anyway.
Maybe we can defer at that time.

qizhangz:
So I simplified the change with with moving just the request/response to context. The stack is reduced by 8K size and context size is increased by 4K size.

qizhangz:
@rkongintel Could you please review this PR again? Thank you!

jyao1:
it seems OK as first step.
Let's keep #351 open and we need continue working on this.

steven-bellock (approved):

PR #414, opened 2021-12-15
Add libspdm_ suffix to publicly visible symbols.

steven-bellock:

spdm-emu needs to be updated as well.

### DIFF
--- a/include/industry_standard/spdm.h
+++ b/include/industry_standard/spdm.h
@@ -14,6 +14,13 @@
 
 #pragma pack(1)
 
+#define SPDM_MAX_VERSION_COUNT 5
+#define SPDM_MAX_SLOT_COUNT 8
+#define SPDM_MAX_OPAQUE_DATA_SIZE 1024
+
+#define SPDM_NONCE_SIZE 32
+#define SPDM_RANDOM_DATA_SIZE 32
+
 ///
 /// SPDM response code (1.0)
 ///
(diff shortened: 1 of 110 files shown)


jyao1:
@steven-bellock, CI failed.
I guess it is caused by new patch checked in after you made the change.

include/industry_standard/spdm.h
```diff
@@ -14,6 +14,13 @@
 
 #pragma pack(1)
 
+#define SPDM_MAX_VERSION_COUNT 5
```
jyao1, line 17: I think spdm does not define MAX count.
But if we count SPDM 1.2, we need redefine it to 3.

We can handle that later, in another patch.

steven-bellock:
Yeah I'm updating things.

include/industry_standard/spdm.h
steven-bellock: According to https://github.com/DMTF/Security-TF/issues/1690 it can be as large as 256 if the Responder is allowed to duplicate entries. :(

jyao1:
Would you please not use "merge commit", but use "rebase and merge"?
 f255006 is not desired.

steven-bellock:
> Would you please not use "merge commit", but use "rebase and merge"? [f255006](https://github.com/DMTF/libspdm/commit/f25500661e80feedb667e7193993478993bff1e2) is not desired.

Done.

PR #415, opened 2021-12-15
Additional return value checks fixed

taprinz:
Addresses #75 

The purpose of this PR was to fix any missing return value checks for functions that may not have declarations in header files. This is so that they were not overlooked when doing the large swaths of checking via adding __attribute__((warn_unused_result)) to every function declaration. This didn't catch many/any functions that were only defined internally, but did provide a few warnings for other functions declared in headers we haven't gotten do yet.

Also updates the spdm_create_update_session_data_key and spdm_activate_update_session_data_key to be single operation. With error checks, these could fail halfway and have halfway completed result, which is a pain to deal with. Changing to single operation at a time, to try and prevent.

### DIFF
--- a/library/spdm_secured_message_lib/libspdm_secmes_session.c
+++ b/library/spdm_secured_message_lib/libspdm_secmes_session.c
@@ -306,29 +306,38 @@ spdm_generate_session_handshake_key(IN void *spdm_secured_message_context,
                hash_size);
     DEBUG((DEBUG_INFO, "\n"));
 
-    spdm_generate_finished_key(
+    status = spdm_generate_finished_key(
         secured_message_context,
         secured_message_context->handshake_secret
             .request_handshake_secret,
         secured_message_context->handshake_secret.request_finished_key);
+    if (RETURN_ERROR(status)) {
+        return status;
+    }
 
-    spdm_generate_finished_key(
+    status = spdm_generate_finished_key(
         secured_message_context,
         secured_message_context->handshake_secret
             .response_handshake_secret,
         secured_message_context->handshake_secret.response_finished_key);
+    if (RETURN_ERROR(status)) {
+        return status;
+    }
 
-    spdm_generate_aead_key_and_iv(secured_message_context,
+    status = spdm_generate_aead_key_and_iv(secured_message_context,
                       secured_message_context->handshake_secret
                           .request_handshake_secret,
                       secured_message_context->handshake_secret
                           .request_handshake_encryption_key,
                       secured_message_context->handshake_secret
                           .request_handshake_salt);
+    if (RETURN_ERROR(status)) {
+        return status;
+    }
     secured_message_context->handshake_secret
         .request_handshake_sequence_number = 0;
 
-    spdm_generate_aead_key_and_iv(
+    status = spdm_generate_aead_key_and_iv(
         secured_message_context,
         secured_message_context->handshake_secret
             .response_handshake_secret,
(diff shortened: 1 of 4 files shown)


library/spdm_secured_message_lib/libspdm_secmes_session.c
```diff
@@ -306,42 +306,54 @@ spdm_generate_session_handshake_key(IN void *spdm_secured_message_context,
                hash_size);
     DEBUG((DEBUG_INFO, "\n"));
 
-    spdm_generate_finished_key(
+    status = spdm_generate_finished_key(
```
taprinz, line 309: Issue for a separate PR, but would we ever want to consider scrubbing of session information on failure to calculate secrets, as in here? that way, we don't have half calculated or incorrect secrets in the session context structure?

jyao1 (approved):

library/spdm_secured_message_lib/libspdm_secmes_session.c
```diff
@@ -595,21 +616,27 @@ spdm_create_update_session_data_key(IN void *spdm_secured_message_context,
...
                 .request_data_secret,
             secured_message_context->application_secret
                 .request_data_encryption_key,
             secured_message_context->application_secret
                 .request_data_salt);
+        if (RETURN_ERROR(status)) {
+            return status;
+        }
         secured_message_context->application_secret
             .request_data_sequence_number = 0;
```
taprinz, line 641: @jyao1 or others - there's the possibility here that we want to update both keys, and updating requester keys succeeds but updating responder keys fails. It would leave us in a tricky situation to determine whether we want to say the requester backup is valid.

It looks like we would only ever call this function to update the requester or update the responder - we're only doing one at a time right now. But if we want to avoid having a state issue, we should probably make this an 

`   else if ((action & SPDM_KEY_UPDATE_ACTION_RESPONDER) != 0) {`

so that we don't have to worry about a halfway complete operation.

library/spdm_secured_message_lib/libspdm_secmes_session.c
jyao1: I think we had better to clear the session info.

library/spdm_secured_message_lib/libspdm_secmes_session.c
jyao1: halfway completion is headache.
Since we only call this function to update one party, how about we make it mandatory?
We can assert only SPDM_KEY_UPDATE_ACTION_RESPONDER or SPDM_KEY_UPDATE_ACTION_REQUESTER is called. Then we dont need worry this case.

I try to *avoid* the complicated situation, instead of *resolve* it.

library/spdm_secured_message_lib/libspdm_secmes_session.c
taprinz: Yeah it is. We can leave it as responsibility of caller to clear, in case there's ever desire to retry versus clear session.

library/spdm_secured_message_lib/libspdm_secmes_session.c
taprinz: Agreed, which is why I figured we change the function slightly such that it is exclusive to one or other actions, but not both. I can do in this PR.

taprinz: Updated to resolve. We probably deserve a separate item/investigation for how we handle halfway completed results elsewhere as well.

PR #416, opened 2021-12-16
Fixing missing return values for spdm_hash_all

taprinz:
Addresses #75

Fixes missing return values for spdm_hash_all. For certain usages of spdm_hash_all, which were only meant for debug code, gated the usage behind DEBUG_CODE macro to ensure the functions wouldn't be called in Release builds.

### DIFF
--- a/library/spdm_common_lib/libspdm_com_crypto_service.c
+++ b/library/spdm_common_lib/libspdm_com_crypto_service.c
@@ -205,14 +205,18 @@ boolean spdm_calculate_m1m2(IN void *context, IN boolean is_mut,
             return FALSE;
         }
 
-        // debug only
-        spdm_hash_all(
-            spdm_context->connection_info.algorithm.base_hash_algo,
-            get_managed_buffer(&m1m2),
-            get_managed_buffer_size(&m1m2), hash_data);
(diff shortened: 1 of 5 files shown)


library/spdm_common_lib/libspdm_com_crypto_service.c
```diff
@@ -205,14 +205,18 @@ boolean spdm_calculate_m1m2(IN void *context, IN boolean is_mut,
...
+        DEBUG_CODE(
+            if (!spdm_hash_all(
+                spdm_context->connection_info.algorithm.base_hash_algo,
+                get_managed_buffer(&m1m2),
+                get_managed_buffer_size(&m1m2), hash_data)) {
+                return FALSE;
+            }
+            DEBUG((DEBUG_INFO, "m1m2 Mut hash - "));
+            internal_dump_data(hash_data, hash_size);
+            DEBUG((DEBUG_INFO, "\n"));
```
taprinz, line 218: I guess these don't need to be inside DEBUG_CODE. But it looks a little cleaner all in one, especially since these function calls rely on the result of the former.

jyao1 (approved):

PR #417, opened 2021-12-16
Address spdm_hmac_* and spdm_hash_* return values

taprinz:
Addresses #75.

Addresses most remaining functions - these functions use allocated contexts heavily, so have to be more careful when checking for and handling error, so that we do not unintentionally disrupt state.

Note: I did NOT address the libspdm_append_message_k and libspdm_append_message_f buffers yet. These functions are not written in a way to handle errors gracefully. If we fail at any point during these functions, the message hashes and buffers can enter an unrecoverable state. The changes required to gracefully handle this error would need close care when reviewing, and so I will defer it to a later PR. This probably deserves its own issue.

### DIFF
--- a/os_stub/spdm_device_secret_lib_sample/cert.c
+++ b/os_stub/spdm_device_secret_lib_sample/cert.c
@@ -92,8 +92,13 @@ boolean read_responder_root_public_certificate(IN uint32_t base_hash_algo,
     cert_chain->length = (uint16_t)cert_chain_size;
     cert_chain->reserved = 0;
 
-    spdm_hash_all(base_hash_algo, file_data, file_size,
+    res = spdm_hash_all(base_hash_algo, file_data, file_size,
               (uint8_t *)(cert_chain + 1));
+    if (!res) {
+        free(file_data);
+        free(cert_chain);
+        return res;
+    }
     copy_mem((uint8_t *)cert_chain + sizeof(spdm_cert_chain_t) + digest_size,
          file_data, file_size);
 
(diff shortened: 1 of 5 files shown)


library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -1000,15 +1048,33 @@ return_status libspdm_append_message_mut_b(IN void *context, IN void *message,
...
+                return RETURN_DEVICE_ERROR;
+            }
+            result = spdm_hash_init (spdm_context->connection_info.algorithm.base_hash_algo,
+                spdm_context->transcript.digest_context_mut_m1m2);
+            if (!result) {
+                spdm_hash_free (spdm_context->connection_info.algorithm.base_hash_algo,
+                    spdm_context->transcript.digest_context_mut_m1m2);
+                spdm_context->transcript.digest_context_mut_m1m2 = NULL;
+                return RETURN_DEVICE_ERROR;
+            }
```
taprinz, line 1067: For this and similar functions, the idea is to not initialize the transcript at all, unless the full transcript initialization is successful. Otherwise, we can end up in some unknown state.

os_stub/spdm_device_secret_lib_sample/cert.c
```diff
@@ -461,8 +481,13 @@ boolean read_responder_root_public_certificate_by_size(
...
     cert_chain->reserved = 0;
 
-    spdm_hash_all(base_hash_algo, file_data, file_size,
+    res = spdm_hash_all(base_hash_algo, file_data, file_size,
               (uint8_t *)(cert_chain + 1));
+    if (!res) {
+        free(file_data);
+        free(cert_chain);
+        return res;
+    }
```
taprinz, line 490: Some of these changes will go away once spdm_hash_all fixes are merged.
Was just difficult to develop both without ensuring there won't be merge conflicts later

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -1032,15 +1098,31 @@ return_status libspdm_append_message_mut_c(IN void *context, IN void *message,
     return append_managed_buffer(&spdm_context->transcript.message_mut_c,
                      message, message_size);
 #else
+    boolean result;
```
taprinz, line 1101: note: put this in brackets like others

jyao1 (approved):

taprinz:
it appears I clobbered this in a misguided attempt to merge latest. Will reattempt and push new PR.

PR #418, opened 2021-12-16
Fixing return values for most spdm_hash* and spdm_hmac* functions

taprinz:
Addresses #75.

Addresses most remaining functions - these functions use allocated contexts heavily, so have to be more careful when checking for and handling error, so that we do not unintentionally disrupt state.

Note: I did NOT address the libspdm_append_message_k and libspdm_append_message_f buffers yet. These functions are not written in a way to handle errors gracefully. If we fail at any point during these functions, the message hashes and buffers can enter an unrecoverable state. The changes required to gracefully handle this error would need close care when reviewing, and so I will defer it to a later PR. This probably deserves its own issue.

### DIFF
--- a/library/spdm_common_lib/CMakeLists.txt
+++ b/library/spdm_common_lib/CMakeLists.txt
@@ -5,7 +5,7 @@ INCLUDE_DIRECTORIES(${LIBSPDM_DIR}/include
 )
 
 SET(src_spdm_common_lib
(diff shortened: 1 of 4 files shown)


taprinz:
This is redo of https://github.com/DMTF/libspdm/pull/417, which I managed to accidentally delete while trying to merge latest.

jyao1 (approved):

PR #420, opened 2021-12-16
Add libspdm_ suffix to spdm_lib_config.h

steven-bellock:


steven-bellock:
@jyao1 I deleted `DEFAULT_SECURE_MCTP_PADDING_SIZE` since that doesn't appear to be used anywhere.

Issue #249, opened 2021-11-01
Convert // comments to /* */ comments


steven-bellock:

jyao1:
Hey
I know other project does not like /* */ and convert to //.

Any special reason for this issue?

steven-bellock:
// was introduced with C99. C89 only supports /* */, and it's something that the integrator can't override.

jyao1:
I see your point. Agree.

Wenxing-hou:
I will check it.

PR #428, opened 2021-12-19
Convert // comments to /* */ comments

Wenxing-hou:
Fix: #249 


jyao1:
Looks good

PR #432, opened 2021-12-20
Move SPDM_MAX_VERSION_COUNT to config.h

### LINKED ISSUE #431: move "#define SPDM_MAX_VERSION_COUNT 5" from spdm.h back to libspdm_config.h
Since SPDM_MAX_VERSION_COUNT is NOT defined in SPDM spec, I think we should move it back to config.h

jyao1:
Fix #431

### DIFF
--- a/library/spdm_common_lib/libspdm_com_context_data.c
+++ b/library/spdm_common_lib/libspdm_com_context_data.c
@@ -73,15 +73,10 @@ return_status libspdm_set_data(IN void *context, IN libspdm_data_type_t data_typ
 
     switch (data_type) {
     case LIBSPDM_DATA_SPDM_VERSION:
-        if (data_size >
-            sizeof(spdm_version_number_t) * SPDM_MAX_VERSION_COUNT) {
-            return RETURN_INVALID_PARAMETER;
-        }
+        ASSERT (data_size <= sizeof(spdm_version_number_t) * SPDM_MAX_VERSION_COUNT);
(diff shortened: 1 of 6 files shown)


library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -73,15 +73,10 @@ return_status libspdm_set_data(IN void *context, IN libspdm_data_type_t data_typ
...
-            sizeof(spdm_version_number_t) * SPDM_MAX_VERSION_COUNT) {
-            return RETURN_INVALID_PARAMETER;
-        }
+        ASSERT (data_size <= sizeof(spdm_version_number_t) * SPDM_MAX_VERSION_COUNT);
         if (parameter->location == LIBSPDM_DATA_LOCATION_CONNECTION) {
-            if (data_size != sizeof(spdm_version_number_t)) {
-                /* Only have one connected version */
-                return RETURN_INVALID_PARAMETER;
-            }
+            /* Only have one connected version */
```
steven-bellock, line 78: How does integrator specify that they support both 1.0 and 1.1?

jyao1: this is version for connection. We can only choose 1.0 or 1.1.
Not local provision.

steven-bellock: Ah yeah that makes sense.

steven-bellock (approved):

Issue #401, opened 2021-12-13
most of the fuzzing test cases have no scenario description


Zhiqiang520:
most of the fuzzing test cases have no scenario description. It is difficult for future fail analyze and test case maintain.
1. for example:
https://github.com/DMTF/libspdm/blob/5f108919e509eccdc09410e50f85dd804f097eb6/unit_test/fuzzing/test_responder/test_spdm_responder_digests/digests.c#L163-L167
2. the same issue is in [capabilities.c](https://github.com/DMTF/libspdm/blob/main/unit_test/fuzzing/test_responder/test_spdm_responder_capabilities/capabilities.c), [algorithms.c](https://github.com/DMTF/libspdm/blob/main/unit_test/fuzzing/test_responder/test_spdm_responder_algorithms/algorithms.c), [certificate.c](https://github.com/DMTF/libspdm/blob/main/unit_test/fuzzing/test_responder/test_spdm_responder_certificate/certificate.c) and so on.

Xiaohanjlll:
I will check this issue.

PR #436, opened 2021-12-20
Add scenario description for fuzzy test case

Xiaohanjlll:
Add scenario description for fuzzy test case.
Optimize and organize test cases.
Fix: #401 


jyao1:
Please separate description from optimization.

Issue #151, opened 2021-09-23
Support multiple root cert provision


jyao1:
Current SPDM_DATA_PEER_PUBLIC_ROOT_CERT and peer_root_cert_provision only supports 1 root cert.

REF: https://github.com/DMTF/libspdm/blob/main/include/library/spdm_common_lib.h
REF: https://github.com/DMTF/libspdm/blob/main/library/spdm_common_lib/spdm_common_lib_internal.h

However, we need support use case to have multiple root cert enrolled.

Wenxing-hou:
I will check it.

PR #447, opened 2021-12-21
Support multiple root cert provision

Wenxing-hou:
Fix: #151 
Support multiple root cert provision.

1. Change spdm_context->local_context.peer_root_cert_provision and spdm_context->local_context.peer_root_cert_provision_size to array.
The multiple root cert store in the array. The MAX_ROOT_CERT_SUPPORT is 10 by default.

2. Set data for local root cert:

- Iterate through the all root_cert_provision array in order.
- Find the first NULL root_cert_provision[i] to set root cert.
- When the all root_cert_provision is full, return RETURN_OUT_OF_RESOURCES.

3. verify local root cert and peer cert chain:

- Iterate through the all root_cert_provision array in order.
- Compare every root_cert_hash, until find the match root cert.
- otherwise, return FALSE.

4. Update the unit_test. Change root_cert_provision to root_cert_provision array.

5. Add new unit tests for spdm_verify_peer_cert_chain_buffer and libspdm_set_data.

6. Update user_guide.md to guide setting multi root certs


include/industry_standard/spdm.h
```diff
@@ -19,6 +19,7 @@
 #define SPDM_MAX_SLOT_COUNT 8
 #define SPDM_MAX_OPAQUE_DATA_SIZE 1024
 
+#define MAX_ROOT_CERT_SUPPORT 10
```
jyao1, line 22: please move this to config.h, and add LIBSPDM_ prefix.

jyao1:
please add unit test.

include/industry_standard/spdm.h
Wenxing-hou: Yes, I have changed it. Thanks.

jyao1:
Looks good

steven-bellock (approved):

PR #451, opened 2021-12-22
Spdm1.2

jyao1:
This PR is to migrate the patch from [SPDM1.2WIP](https://github.com/DMTF/libspdm/tree/SPDM1.2WIP) to main, after SPDM1.2 spec is published.
Once it is migrated, we can delete SPDM1.2WIP branch.

Fix: #173
Fix: #144
Fix: #143
Fix: #142
Fix: #141

This is first patch set to support SPDM 1.2.
1) It adds support for SPDM 1.2 transcript hash.
2) It adds multiple SPDM 1.2 crypto algorithms.
3) It supports SPDM 1.2 version negotiation.

The patch set is validated with spdm_emu and spdm_dump.
1) SPDM 1.1 configuration still works in unit test and emu.
2) SPDM 1.2 configuration with new transcript hash works in emu.
3) SPDM 1.2 crypto algorithm works in unit test.
4) SPDM 1.2 crypto algorithm in emu. (EDDSA_25519, EDDSA_448, SM2_P256/SM3)

Feature not included yet: (will file individual issue)
1) SPDM 1.2 new capabilities
2) SPDM 1.2 new commands.


jyao1:
@steven-bellock, would you please review this patch set ?
I have rebased the WIP several times. I hope to check in SPDM1.2 stuff soon, so that we can continue working on that.

steven-bellock (approved):

Issue #450, opened 2021-12-22
unit test: test_spdm_common error 


Wenxing-hou:
When build pass, run test_spdm_common. The error is found.

```
[ RUN      ] test_spdm_common_context_data_case4
[  ERROR   ] --- 0x0 != 0x80000005
[   LINE   ] --- C:\mywork\code\libspdm\unit_test\test_spdm_common\context_data.c:180: error: Failure!
[  FAILED  ] test_spdm_common_context_data_case4
```

jyao1:
Do you run with X64 arch?
It seems the test assume X64 arch.

steven-bellock:
@raghuncstate introduced these tests. Now that the field is called LIBSPDM_DATA_APP_CONTEXT_DATA instead of OPAQUE_DATA we need to rename the word "opaque".

steven-bellock:
But yes `data_return_size = 4;` should be something like `data_return_size = sizeof(void *) - 1;`

jyao1:
agree

Wenxing-hou:
I will check it.

PR #456, opened 2021-12-23
Change the hard code to fix test_spdm_common error  and make the test case id correct

Wenxing-hou:
Fix: #450 

Change the hard code to fix test_spdm_common unit test error.
Because the size of viod* is different on different machine, test_spdm_common unit test error is happen in ia32. Change the hard code to fix it.

Make the test case id correct.


jyao1:
Looks good to me.

steven-bellock (approved):

PR #457, opened 2021-12-23
Enhance C89 compatibility.

steven-bellock:
The last element in an enum cannot have a trailing comma; that was introduced in C99. The bit-field type must be either unsigned int or signed int.

### DIFF
--- a/include/industry_standard/spdm.h
+++ b/include/industry_standard/spdm.h
@@ -110,10 +110,10 @@ typedef struct {
 /* SPDM VERSION structure*/
 
 typedef struct {
-    uint16_t alpha : 4;
-    uint16_t update_version_number : 4;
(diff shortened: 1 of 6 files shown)


jyao1:
1. For bit definition, I am worried that if the size of structure is changed. Would you please confirm if the size is changed or not?
If the size is changed, then the data structure cannot be used. We had better *remove* bit definition at all.

2. why it removes basic_mut_auth_requested ? I don't think that is for C89.
Can we file a new issue and discuss how to handle deprecated field in SPDM 1.2 ?

steven-bellock:
Yes, I don't think we should be using bit-fields at all. Let me break this pull request up into separate pull requests.

jyao1:
Cool. Thanks. Let's remove bit field, then.

PR #458, opened 2021-12-24
Enhance C89 compatibility.

steven-bellock:
The last element in an enum cannot have a trailing comma; that was introduced in C99.


jyao1:
looks good

Issue #402, opened 2021-12-13
The requester should check the SPDMVersion filed of response message received from responder


Zhiqiang520:
The requester should check the SPDMVersion filed of response message received from responder.
1. for example:
Line88 and Line97 only check `spdm_response.header.request_response_code`, I think `spdm_response.header.spdm_version` also should be checked, whether it match `spdm_request.header.spdm_version`.
https://github.com/DMTF/libspdm/blob/5f108919e509eccdc09410e50f85dd804f097eb6/library/spdm_requester_lib/libspdm_req_get_digests.c#L80-L99
2. the same issue is in [libspdm_req_get_certificate.c](https://github.com/DMTF/libspdm/blob/main/library/spdm_requester_lib/libspdm_req_get_certificate.c), [libspdm_req_get_measurements.c](https://github.com/DMTF/libspdm/blob/main/library/spdm_requester_lib/libspdm_req_get_measurements.c), [libspdm_req_key_exchange.c](https://github.com/DMTF/libspdm/blob/main/library/spdm_requester_lib/libspdm_req_key_exchange.c) and so on.

Wenxing-hou:
I will check it.

PR #468, opened 2021-12-29
Add check the SPDMVersion filed of response message

Wenxing-hou:
Fix: #402 


jyao1:
looks good

steven-bellock (approved):

Issue #147, opened 2021-09-23
cache peer_used_cert_chain_buffer_hash instead of peer_used_cert_chain_buffer to save context size


jyao1:
REF: https://github.com/DMTF/libspdm/blob/main/library/spdm_common_lib/spdm_common_lib_internal.h

We can use uint8 peer_used_cert_chain_buffer_hash[MAX_HASH_SIZE] to replace uint8 peer_used_cert_chain_buffer[MAX_SPDM_CERT_CHAIN_SIZE];

jyao1:
NOTE: We need the leaf cert public key to validate the digital signature.

jyao1:
We can also consider how to optimize large_managed_buffer_t certificate_chain_buffer in spdm_encap_context_t.

qizhangz:
I will check it.

PR #473, opened 2021-12-30
cache peer_used_cert_chain_buffer_hash instead of

qizhangz:
peer_used_cert_chain_buffer.

about 4K is saved for context size.

Fix: #147 

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -98,12 +98,18 @@ typedef struct {
     spdm_device_capability_t capability;
     spdm_device_algorithm_t algorithm;
     spdm_version_number_t secured_message_version;
-    
-    /* Peer CertificateChain*/
-    
+
+    /* Peer CertificateChain */
+#if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT
     uint8_t peer_used_cert_chain_buffer[LIBSPDM_MAX_CERT_CHAIN_SIZE];
     uintn peer_used_cert_chain_buffer_size;
-    
+#else
+    uint8_t peer_used_cert_chain_buffer_hash[LIBSPDM_MAX_HASH_SIZE];
+    uint32_t   peer_used_cert_chain_buffer_hash_size;
+    /* leaf cert public key of the peer */
+    void *peer_used_leaf_cert_public_key;
+#endif
+
     /* Local Used CertificateChain (for responder, or requester in mut auth)*/
     
     uint8_t *local_used_cert_chain_buffer;
--- a/library/spdm_common_lib/libspdm_com_context_data.c
+++ b/library/spdm_common_lib/libspdm_com_context_data.c
@@ -52,6 +52,9 @@ return_status libspdm_set_data(IN void *context, IN libspdm_data_type_t data_typ
     uint8_t slot_id;
     uint8_t mut_auth_requested;
     uint8_t root_cert_index;
+#if !LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT
+    boolean status;
+#endif
 
     if (!context || !data || data_type >= LIBSPDM_DATA_MAX) {
         return RETURN_INVALID_PARAMETER;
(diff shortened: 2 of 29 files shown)


include/library/spdm_common_lib.h
```diff
@@ -893,6 +893,26 @@ boolean libspdm_get_peer_cert_chain_data(IN void *spdm_context,
...
+  @param  spdm_context                  A pointer to the SPDM context.
+  @param  algo                          SPDM base_asym_algo or req_base_asym_alg.
+  @param  cert_chain_data               Certitiface chain data without spdm_cert_chain_t header.
+  @param  cert_chain_data_size          size in bytes of the certitiface chain data.
+  @param  public_key                    Pointer to new-generated asymmetric context which contain the retrieved public key component.
+
+  @retval  TRUE   public key was retrieved successfully.
+  @retval  FALSE  Fail to retrieve public key from X509 certificate.
+**/
+boolean libspdm_get_public_key_from_cert_chain(IN void *context,
```
jyao1, line 910: 1) Please move this function to spdm_crypt_lib.
2) Please change name to be libspdm_get_leaf_cert_public_key_from_cert_chain().
3) Please describe the expectation of the cert_chain - validated or not validated.

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -316,11 +316,21 @@ return_status libspdm_set_data(IN void *context, IN libspdm_data_type_t data_typ
...
-                 .peer_used_cert_chain_buffer,
-             data, data_size);
+        libspdm_hash_all(
+            spdm_context->connection_info.algorithm.base_hash_algo,
+            data, data_size,
+            spdm_context->connection_info.peer_used_cert_chain_buffer_hash);
+
+        spdm_context->connection_info.peer_used_cert_chain_buffer_hash_size =
+            libspdm_get_hash_size(spdm_context->connection_info.algorithm.base_hash_algo);
+        libspdm_get_public_key_from_cert_chain(spdm_context,
```
jyao1, line 326: Need error handling here.

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -316,11 +316,21 @@ return_status libspdm_set_data(IN void *context, IN libspdm_data_type_t data_typ
...
+            spdm_context->connection_info.algorithm.base_asym_algo,
+            data, data_size,
+            &spdm_context->connection_info.public_key_with_base_asym_algo);
+        libspdm_get_public_key_from_cert_chain(spdm_context,
```
jyao1, line 330: Need error handling here.

Why we need both base_asym_algo and req_base_asym_alg ? Should we only use one of them?
E.g. requester only uses base_asym_algo, and responder only uses req_base_asym_alg.

include/internal/libspdm_common_lib.h
```diff
@@ -95,12 +95,14 @@ typedef struct {
...
-    uint8_t peer_used_cert_chain_buffer[LIBSPDM_MAX_CERT_CHAIN_SIZE];
-    uintn peer_used_cert_chain_buffer_size;
-    
+
+    /* Peer CertificateChain hash*/
+    uint8_t peer_used_cert_chain_buffer_hash[LIBSPDM_MAX_HASH_SIZE];
+    uint32_t   peer_used_cert_chain_buffer_hash_size;
+    /* public key from leaf cert */
+    void *public_key_with_base_asym_algo;
+    void *public_key_with_req_base_asym_alg;
```
jyao1, line 104: This definition is confusing. Please only use below:
/* leaf cert public key of the peer */
void *peer_used_leaf_cert_public_key;

Please describe why we need distinguish requester (base_asym_algo) from responder (req_base_asym_alg).

jyao1:
1. Have you tested spdm-emu and spdm-dump to ensure they are still good?

2. I feel that we had better keep the old full CertificateChain data if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT is TRUE.

steven-bellock (approved):

qizhangz:
> 1. Have you tested spdm-emu and spdm-dump to ensure they are still good?
>    spdm-dump uses: LIBSPDM_DATA_PEER_USED_CERT_CHAIN_BUFFER, but I did not see the update ?
Yes. I tested spdm-emu, spdm-dump, unit test and fuzzing test.
> 2. I feel that we had better keep the old full CertificateChain data if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT is TRUE.
done.

PR #474, opened 2021-12-30
Enhance C89 compatibility - remove Bit flags

jyao1:
This PR removes bit flags definition in spdm.h, according to discussion in #457.

1) spdm_version_number_t - done
2) spdm_negotiate_algorithms_struct_table_alg_count_t - done
3) spdm_challenge_auth_response_attribute_t - done
4) spdm_get_measurements_request_slot_id_parameter_t - done
5) SPDM_MEASUREMENTS_BLOCK_MEASUREMENT_TYPE - done

PR for spdm_emu - https://github.com/DMTF/spdm-emu/pull/32
PR for spdm_dump - https://github.com/DMTF/spdm-dump/pull/12

### DIFF
--- a/include/industry_standard/spdm.h
+++ b/include/industry_standard/spdm.h
@@ -478,11 +478,6 @@ typedef struct {
     uint8_t SlotIDParam; /* BIT[0:3]=slot_id, BIT[4:7]=reserved*/
 } spdm_get_measurements_request_t;
 
-typedef struct {
-    uint8_t slot_id : 4;
-    uint8_t reserved : 4;
-} spdm_get_measurements_request_slot_id_parameter_t;
-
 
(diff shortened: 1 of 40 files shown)


jyao1:
@steven-bellock, this PR is to resolve the missing piece of https://github.com/DMTF/libspdm/pull/457 - remove bit field.
Please review if it is OK.

I will add more SPDM 1.2 definition after this spdm.h cleanup is merged.

steven-bellock (approved):
To reiterate why why we're doing this (for folks reading in the future): in C89 the type of a bit-field must be either `signed int` or `unsigned int`. This works fine in GCC but Visual Studio has a bug where, for some reason, it thinks that the entire bitfield is an int. So it's just easier to get rid of bit-fields altogether. Plus C bit-fields are notoriously underspecified and not very portable.

PR #481, opened 2022-01-04
Add check version for mutual function and add missing check in spdm_requester_lib

Wenxing-hou:
Fix: #402 

Add check version for mutual function in spdm_responder_lib.

Add missing check the SPDM Version filed of response message in spdm_requester_lib.


jyao1:
looks good

steven-bellock (approved):

Issue #237, opened 2021-10-29
Need to better handle version negotiation


liyi77:
@rw8896
I was wondering the error handling of "After GET_CAPABILITY, the responder will know this is the version selected by the requester and always stick it it."

After the version is negotiated (either from cache or set by GET_CAPABILITIES), If requester sends a command with a version other than the negotiated one, current responder seems to ignore the error and continue to process it with the previously negotiated version.

Shouldn't the responder validate the version against the negotiated one and return error if it doesn't match?

rw8896:
@liyi77 Thanks for copying my question here.

Current code is a little bit confusing about **supported versions** vs **negotiated version**:
- `spdm_version_number_t version;` in `spdm_connection_info_t` is the negotiated version and would be more convenient to use the same type as `uint8 spdm_version;` in `spdm_message_header_t` 
- `spdm_is_version_supported()` actually means `spdm_is_version_negotiated()`

Here are the proposed changes:

1) For any received command (except GET_VERSION), the responder should check if the command uses the negotiated version
- GET_VERSION is always 1.0. No need to check it against negotiated version.
- Negotiated version is determined ((either from successful GET_CAPABILITIES or cached negotiation) if connection state >= SPDM_CONNECTION_STATE_AFTER_CAPABILITIES  

A pseudo code like this might work in spdm_build_response so we don't need to add the check in each handler function
```
spdm_message_header_t *spdm_request = (spdm_message_header_t *)spdm_context->last_spdm_request;
if ((spdm_request->request_response_code != SPDM_GET_VERSION) && (connection state >= SPDM_CONNECTION_STATE_AFTER_CAPABILITIES)) {
    if (!spdm_is_version_negotiated(spdm_request->spdm_version))
        return Error; /* 0x41? */
}
```

2) requester and responder should fill header.spdm_version using selected/negotiated version except GET_VERSION to replace the hardcoded spdm_version in current code, e.g. 
`spdm_response->header.spdm_version = SPDM_MESSAGE_VERSION_11;`
- not very sure about Error response with Error code 0x41 though

liyi77:
It doesn't seem important since only versions 1.1 and 1.0 are currently supported.
If libspdm supports more versions in the future, it will be a bit confusing.

jyao1:
This code is not scalable, if we need support 1.2, 1.3, etc.

I think we need write code like:
```
spdm_request.header.spdm_version = get_negotiated_version(spdm_context);
```

[...]

PR #483, opened 2022-01-04
Add check version for of requester message

Wenxing-hou:
Fix: #237
For any received command (except GET_VERSION),
the responder should check if the command uses the negotiated version.

### DIFF
--- a/unit_test/test_spdm_responder/capabilities.c
+++ b/unit_test/test_spdm_responder/capabilities.c
@@ -650,9 +650,6 @@ void test_spdm_responder_capabilities_case7(void **state)
     spdm_context->connection_info.connection_state =
         LIBSPDM_CONNECTION_STATE_AFTER_VERSION;
 
-
-    spdm_context->connection_info.version = SPDM_MESSAGE_VERSION_11 << SPDM_VERSION_NUMBER_SHIFT_BIT;
-
     response_size = sizeof(response);
     status = spdm_get_response_capabilities(
         spdm_context, m_spdm_get_capabilities_request3_size,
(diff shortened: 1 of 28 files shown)


library/spdm_responder_lib/libspdm_rsp_capabilities.c
```diff
@@ -157,6 +157,11 @@ return_status spdm_get_response_capabilities(IN void *context,
     spdm_context = context;
     spdm_request = request;
 
+    if (spdm_request->header.spdm_version != spdm_get_connection_version(spdm_context)) {
```
jyao1, line 160: How does the code work?

The responder need use this spdm_version to set connection version.

unit_test/test_spdm_responder/capabilities.c
```diff
@@ -436,6 +436,7 @@ void test_spdm_responder_capabilities_case1(void **state)
     spdm_test_context = *state;
     spdm_context = spdm_test_context->spdm_context;
     spdm_test_context->case_id = 0x1;
+    spdm_context->connection_info.version = SPDM_MESSAGE_VERSION_10 << SPDM_VERSION_NUMBER_SHIFT_BIT;
```
jyao1, line 439: This is wrong.

Responder need use GET_CAPABILITIES to determine the version.

jyao1:
1. Please read my comment in issue carefully.
Special cases:
spdm_get_response_version: 1.0 should be used.
spdm_get_response_capabilities(): the spdm_version should be set to spdm_context->connection_info.version.

2. Have you run spdm-emu test?
I think it should fail with this patch, because the patch checks version in spdm_get_response_capabilities(), which is wrong.

unit_test/test_spdm_responder/capabilities.c
Wenxing-hou: Yes. I have changed the code.  The libspdm_rsp_capabilities.c use follow code to set version. So we can't use the spdm_context->connection_info.version before setting.
```
        if (local_ver == version) {
            spdm_context->connection_info.version = version << SPDM_VERSION_NUMBER_SHIFT_BIT;
            return TRUE;
        }
```

jyao1:
looks good now.

Issue #478, opened 2022-01-02
try_spdm_get_certificate needs to ensure forward progress


steven-bellock:
try_spdm_get_certificate uses a do-while loop to fetch portions of a certificate chain from the Responder:

```
do
{
.
.
} while (spdm_response.remainder_length != 0);
```

However it needs to ensure the response's `remainder_length` field eventually goes to 0, else it can be held in an infinite loop by the Responder.

jyao1:
Any recommendation on what criteria can be used to break the loop ?

steven-bellock:
```
if (remainder_length[n] < remainder_length[n -1])
/* Keep going. */
else
/* Return error. */
```

jyao1:
Makes sense.
And I think we can add more check.
```
total_length = portion_length[0] + remainder_length[0];
if (offset[n] + portion_length[n] + remainder_length[n] == total_length) {
  /* Keep going. */
} else {
  /* Return error. */
}
```

Wenxing-hou:
I will check it.

PR #484, opened 2022-01-04
Add check for try_spdm_get_certificate to make sure get cert correctly

Wenxing-hou:
Fix: #478 

Add total cert chain buffer length check and add remainder length check.

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_get_certificate.c
+++ b/library/spdm_requester_lib/libspdm_req_get_certificate.c
@@ -56,6 +56,7 @@ return_status try_spdm_get_certificate(IN void *context, IN uint8_t slot_id,
     uintn spdm_response_size;
     large_managed_buffer_t certificate_chain_buffer;
     spdm_context_t *spdm_context;
+    uint16_t total_responder_cert_chain_buffer_length;
 
     spdm_context = context;
     if (!spdm_is_capabilities_flag_supported(
@@ -93,7 +94,11 @@ return_status try_spdm_get_certificate(IN void *context, IN uint8_t slot_id,
         spdm_request.header.param2 = 0;
         spdm_request.offset = (uint16_t)get_managed_buffer_size(
             &certificate_chain_buffer);
-        spdm_request.length = length;
+        if (spdm_request.offset == 0) {
+            spdm_request.length = length;
+        } else {
+            spdm_request.length = MIN(length, spdm_response.remainder_length);
+        }
         DEBUG((DEBUG_INFO, "request (offset 0x%x, size 0x%x):\n",
                spdm_request.offset, spdm_request.length));
 
@@ -144,8 +149,7 @@ return_status try_spdm_get_certificate(IN void *context, IN uint8_t slot_id,
             status = RETURN_DEVICE_ERROR;
             goto done;
         }
-        if (spdm_response.portion_length >
-            LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN) {
+        if ((spdm_response.portion_length > spdm_request.length) || (spdm_response.portion_length == 0)) {
             status = RETURN_DEVICE_ERROR;
             goto done;
         }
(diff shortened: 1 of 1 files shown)


library/spdm_requester_lib/libspdm_req_get_certificate.c
```diff
@@ -158,11 +160,28 @@ return_status try_spdm_get_certificate(IN void *context, IN uint8_t slot_id,
...
         }
+        if (spdm_request.offset == 0) {
+            total_responder_cert_chain_buffer_length = spdm_response.portion_length + spdm_response.remainder_length;
+        }
+        else if (spdm_request.offset + spdm_response.portion_length + spdm_response.remainder_length != total_responder_cert_chain_buffer_length) {
+            status = RETURN_DEVICE_ERROR;
+            goto done;
+        }
+
+        if (spdm_request.offset == 0) {
```
jyao1, line 171: Do we still need this logic?

Line 163~169 already covered below check.

library/spdm_requester_lib/libspdm_req_get_certificate.c
```diff
@@ -158,11 +160,28 @@ return_status try_spdm_get_certificate(IN void *context, IN uint8_t slot_id,
             status = RETURN_DEVICE_ERROR;
             goto done;
         }
+        if (spdm_request.offset == 0) {
+            total_responder_cert_chain_buffer_length = spdm_response.portion_length + spdm_response.remainder_length;
```
steven-bellock, line 164:
```
        if (spdm_response.portion_length >
            LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN) {
            status = RETURN_DEVICE_ERROR;
            goto done;
        }
```
can be moved inside here and use `total_responder_cert_chain_buffer_length`.

steven-bellock (changes requested):

library/spdm_requester_lib/libspdm_req_get_certificate.c
Wenxing-hou: Yes, the logic  `else if (spdm_request.offset + spdm_response.portion_length + spdm_response.remainder_length != total_responder_cert_chain_buffer_length)` is strong and already covers the check: `if (remainder_length[n] < remainder_length[n - 1])` .

And I have changed the code. Thanks.

library/spdm_requester_lib/libspdm_req_get_certificate.c
Wenxing-hou: As we can see, the logic` else if (spdm_request.offset + spdm_response.portion_length + spdm_response.remainder_length != total_responder_cert_chain_buffer_length)` is strong.

So I have deleted the check:
```
        if (spdm_response.portion_length >
            LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN) {
            status = RETURN_DEVICE_ERROR;
            goto done;
        }
```
And is my way right? Look forward to receiving your reply. Thanks.

jyao1: I dont think it is right.
"spdm_response.portion_length <= LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN" cannot be enforced by total length check.

In addition, we need add another check
"spdm_response.portion_length > 0" to ensure we always move forward.

Wenxing-hou: Yes. I have added the check.

library/spdm_requester_lib/libspdm_req_get_certificate.c
```diff
@@ -158,11 +162,20 @@ return_status try_spdm_get_certificate(IN void *context, IN uint8_t slot_id,
...
+        }
+        else if ((spdm_request.offset + spdm_response.portion_length + spdm_response.remainder_length != total_responder_cert_chain_buffer_length)
+                 || (spdm_response.portion_length <= 0)) {
```
jyao1, line 169: Portion_length is uint16_t, it cannot < 0.
Please change to (spdm_response.portion_length == 0).

Wenxing-hou: Yes. Even if the responder sends negative values, the values requester read is positive values. And I have changed the code location to prevent stop for first GET_CERTIFICATE request.

jyao1:
Looks good to me.

steven-bellock (approved):
Looks good. As a follow-up we should have a unit test that simulates a misbehaving CERTIFICATE response to check that the Requester can detect it.

Issue #485, opened 2022-01-05
Need update all cert date


Wenxing-hou:
Some certs have out of date. It makes unit test wrong. We need update the cert date.

```
C:\mywork\code\libspdm\build\bin>test_spdm_requester.exe >NUL
[  ERROR   ] --- 0x80000007 != 0x0
[   LINE   ] --- C:\mywork\code\libspdm\unit_test\test_spdm_requester\get_certificate.c:1985: error: Failure!
[  PASSED  ] 18 test(s).
[  FAILED  ] spdm_requester_get_certificate_tests: 1 test(s), listed below:
[  FAILED  ] test_spdm_requester_get_certificate_case13
```

jyao1:
Please 1. update all cert to 10 years.
2. add a script to regenerate all certs.

PR #486, opened 2022-01-05
Update all certs date to 10 years and update the script to regenerate all certs

Wenxing-hou:
Fix: #485


jyao1:
looks good

jyao1:
Merge it soon, since it blocked CI.

Issue #480, opened 2022-01-02
libspdm_com_opaque_data.c should not hardcode secured message binding version.


jyao1:
In current code, SPDM_MESSAGE_VERSION_11 is hardcoded in libspdm_com_opaque_data.c as the  secured message binding version.

The secured message version should be from secured_message_version in spdm_local_context_t, and negotiated value should be saved to spdm_connection_info_t.

Wenxing-hou:
I will check it.

PR #507, opened 2022-01-10
Fix the hardcode for secured message binding version

Wenxing-hou:
Fix: #480 

Sort for secured message binding version.
Find the first match version in loop.

Fix the hardcode for secured message binding version.

Make sort function consistent and remove sort function to common lib.

### DIFF
--- a/library/spdm_common_lib/libspdm_com_opaque_data.c
+++ b/library/spdm_common_lib/libspdm_com_opaque_data.c
@@ -181,8 +181,8 @@ spdm_process_opaque_data_supported_version_data(IN spdm_context_t *spdm_context,
     secured_message_opaque_element_supported_version_t
         *opaque_element_support_version;
     spdm_version_number_t *versions_list;
-    uint8_t secured_message_version_index;
-    uint8_t versions_list_index;
+    spdm_version_number_t common_version;
+    boolean result;
 
     if (spdm_context->local_context.secured_message_version
             .spdm_version_count == 0) {
@@ -230,24 +230,18 @@ spdm_process_opaque_data_supported_version_data(IN spdm_context_t *spdm_context,
     }
     versions_list = (void *)(opaque_element_support_version + 1);
 
-    spdm_version_number_sort(versions_list, spdm_context->local_context.secured_message_version.spdm_version_count);
-    spdm_version_number_sort(spdm_context->local_context.secured_message_version.spdm_version, spdm_context->local_context.secured_message_version.spdm_version_count);
-
-    for (secured_message_version_index = 0;
-        secured_message_version_index < spdm_context->local_context.secured_message_version.spdm_version_count;
-        secured_message_version_index ++) {
-        for (versions_list_index = 0;
-            versions_list_index < spdm_context->local_context.secured_message_version.spdm_version_count;
-            versions_list_index ++) {
-            if (spdm_get_version_from_version_number(versions_list[versions_list_index]) ==
-                spdm_get_version_from_version_number(spdm_context->local_context.secured_message_version.spdm_version[secured_message_version_index])) {
-                spdm_context->connection_info.secured_message_version = versions_list[versions_list_index];
-                return RETURN_SUCCESS;
-            }
-        }
+    result = spdm_negotiate_connection_version(&common_version, spdm_context->local_context.secured_message_version.spdm_version,
+                                    spdm_context->local_context.secured_message_version.spdm_version_count,
+                                    versions_list,
+                                    opaque_element_support_version->version_count);
+    if (result == FALSE) {
+        return RETURN_UNSUPPORTED;
     }
+    copy_mem(&(spdm_context->connection_info.secured_message_version),
+    &(common_version),
+    sizeof(spdm_version_number_t));
 
-    return RETURN_UNSUPPORTED;
+    return RETURN_SUCCESS;
 }
 
 /**
--- a/library/spdm_common_lib/libspdm_com_opaque_data.c
+++ b/library/spdm_common_lib/libspdm_com_opaque_data.c
@@ -7,6 +7,38 @@
 #include "internal/libspdm_common_lib.h"
 
 /**
+  Sort SPDMversion in descending order.
+
+  @param  spdm_context                A pointer to the SPDM context.
+  @param  ver_set                    A pointer to the version set.
+  @param  ver_num                    Version number.
+*/
+void spdm_secured_version_number_sort(IN OUT spdm_version_number_t *ver_set, IN uintn ver_num)
+{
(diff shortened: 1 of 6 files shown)


jyao1:
We cannot assume "local_context.secured_message_version.spdm_version_count is 1". It is configurable from libspdm consumer.

Wenxing-hou:
Yes. I have updated the code and description.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -6,6 +6,38 @@
...
 #include "internal/libspdm_common_lib.h"
 
+/**
+  Sort SPDMversion in descending order.
+
+  @param  spdm_context                A pointer to the SPDM context.
+  @param  ver_set                    A pointer to the version set.
+  @param  ver_num                    Version number.
+*/
+void spdm_secured_version_number_sort(IN OUT spdm_version_number_t *ver_set, IN uintn ver_num)
```
jyao1, line 16: Can we use same spdm_version_number_sort() ?

Wenxing-hou: Yes. I will make these two sort function same and remove the sort function to common lib.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -224,11 +264,24 @@ spdm_process_opaque_data_supported_version_data(IN spdm_context_t *spdm_context,
         return RETURN_UNSUPPORTED;
     }
     versions_list = (void *)(opaque_element_support_version + 1);
-    if ((versions_list[0] >> SPDM_VERSION_NUMBER_SHIFT_BIT) != SPDM_MESSAGE_VERSION_11) {
-        return RETURN_UNSUPPORTED;
+
+    spdm_secured_version_number_sort(versions_list, spdm_context->local_context.secured_message_version.spdm_version_count);
+    spdm_secured_version_number_sort(spdm_context->local_context.secured_message_version.spdm_version, spdm_context->local_context.secured_message_version.spdm_version_count);
+
+    for (secured_message_version_index = 0;
```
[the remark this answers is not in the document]
Wenxing-hou: Yes. I have updated the code.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -144,10 +146,15 @@ spdm_build_opaque_data_supported_version_data(IN spdm_context_t *spdm_context,
     opaque_element_support_version->version_count = 1;
 
     versions_list = (void *)(opaque_element_support_version + 1);
-    versions_list[0] = SPDM_MESSAGE_VERSION_11 << SPDM_VERSION_NUMBER_SHIFT_BIT;
+    for (secured_message_version_index = 0;
```
jyao1, line 149: why not use
```
copy_mem(
    versions_list,
    spdm_context->local_context.secured_message_version.spdm_version,
    spdm_context->local_context.secured_message_version.spdm_version_count * sizeof(spdm_version_number_t)
    );
```

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -220,15 +229,28 @@ spdm_process_opaque_data_supported_version_data(IN spdm_context_t *spdm_context,
          SECURED_MESSAGE_OPAQUE_ELEMENT_SMDATA_DATA_VERSION) ||
         (opaque_element_support_version->sm_data_id !=
          SECURED_MESSAGE_OPAQUE_ELEMENT_SMDATA_ID_SUPPORTED_VERSION) ||
-        (opaque_element_support_version->version_count != 1)) {
+        (opaque_element_support_version->version_count != spdm_context->local_context.secured_message_version.spdm_version_count)) {
```
jyao1, line 232: Why we have such restriction ?

library/spdm_common_lib/libspdm_com_opaque_data.c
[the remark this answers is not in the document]
jyao1: I do not see the update. How do you align the code?

The original code is problematic, because it compares the full version.
But only major/minor is required, patch/alpha should be ignored. See https://github.com/DMTF/libspdm/blob/main/library/spdm_requester_lib/libspdm_req_get_version.c#L109, spdm_get_version_from_version_number() function.

library/spdm_common_lib/libspdm_com_opaque_data.c
jyao1: ok

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -220,15 +225,29 @@ spdm_process_opaque_data_supported_version_data(IN spdm_context_t *spdm_context,
          SECURED_MESSAGE_OPAQUE_ELEMENT_SMDATA_DATA_VERSION) ||
         (opaque_element_support_version->sm_data_id !=
          SECURED_MESSAGE_OPAQUE_ELEMENT_SMDATA_ID_SUPPORTED_VERSION) ||
-        (opaque_element_support_version->version_count != 1)) {
+        (opaque_element_support_version->version_count < 0)) {
```
jyao1, line 228: How could unsigned value < 0 ???

I think we need use "opaque_element_support_version->version_count == 0"

jyao1:
looks good.

jyao1:
@steven-bellock , would you please review this?

steven-bellock:
> @steven-bellock , would you please review this?

Yes, I'll look at it today.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -220,15 +225,25 @@ spdm_process_opaque_data_supported_version_data(IN spdm_context_t *spdm_context,
...
     versions_list = (void *)(opaque_element_support_version + 1);
-    if ((versions_list[0] >> SPDM_VERSION_NUMBER_SHIFT_BIT) != SPDM_MESSAGE_VERSION_11) {
+
+    result = spdm_negotiate_connection_version(&common_version, spdm_context->local_context.secured_message_version.spdm_version,
+                                    spdm_context->local_context.secured_message_version.spdm_version_count,
+                                    versions_list,
+                                    opaque_element_support_version->version_count);
+    if (result == FALSE) {
         return RETURN_UNSUPPORTED;
+    } else {
```
steven-bellock, line 239: Can remove the `else` statement since the `return RETURN_UNSUPPORTED;` at line 246 is not needed.

steven-bellock (changes requested):

steven-bellock (approved):

Issue #509, opened 2022-01-11
Correct the error code for messages with mismatched version


rw8896:
Current code used SPDM_ERROR_CODE_INVALID_REQUEST but 1.2 spec mentioned:

If the Requester uses a
different version than the selected version in other Requests, the Responder should return an ERROR message with
ErrorCode = VersionMismatch or the Responder can silently discard the Request.

jyao1:
agree

Wenxing-hou:
Yes. I will check it.

PR #511, opened 2022-01-12
Update the error code for messages with mismatched version

Wenxing-hou:
Fix: #509


jyao1:
looks good

steven-bellock (approved):

Issue #510, opened 2022-01-11
Need to add strong check for get cert in libspdm_rsp_encap_get_certificate.c 


Wenxing-hou:
It needs to ensure the response's remainder_length field eventually goes to 0, else it can be held in an infinite loop by the Responder.

jyao1:
Similar to https://github.com/DMTF/libspdm/issues/478

PR #518, opened 2022-01-16
Add encap get cert check to avoid infinite loop

Wenxing-hou:
Fix: #510 


library/spdm_responder_lib/libspdm_rsp_encap_get_certificate.c
```diff
@@ -8,6 +8,7 @@
 
 #if LIBSPDM_ENABLE_CAPABILITY_CERT_CAP
 
+uint16_t total_responder_cert_chain_buffer_length;
```
jyao1, line 11: We cannot use global variable.

Wenxing-hou: Yes. We should observe the README. I have updated the code.

steven-bellock (approved):

Issue #493, opened 2022-01-06
Requester slot_id checks should be assertions


steven-bellock:
https://github.com/DMTF/libspdm/blob/2a7afd9ed271d9ea3e6e719d59d0325bc2ff15ca/library/spdm_requester_lib/libspdm_req_challenge.c#L86-L92
https://github.com/DMTF/libspdm/blob/2a7afd9ed271d9ea3e6e719d59d0325bc2ff15ca/library/spdm_requester_lib/libspdm_req_get_certificate.c#L81-L83
https://github.com/DMTF/libspdm/blob/2a7afd9ed271d9ea3e6e719d59d0325bc2ff15ca/library/spdm_requester_lib/libspdm_req_key_exchange.c#L101-L107

jyao1:
1/10/2022 - probably need review all check to see if it makes sense to use ASSERT.

PR #528, opened 2022-01-20
Change runtime checks to asserts

steven-bellock:
Fixes https://github.com/DMTF/libspdm/issues/493


jyao1:
looks good

Issue #529, opened 2022-01-21
spdm_process_opaque_data_version_selection_data() logic error


jyao1:
https://github.com/DMTF/libspdm/blob/main/library/spdm_common_lib/libspdm_com_opaque_data.c

spdm_process_opaque_data_version_selection_data():
```
    for (secured_message_version_index = 0;
        secured_message_version_index < spdm_context->local_context.secured_message_version.spdm_version_count;
        secured_message_version_index ++) {

        if (spdm_get_version_from_version_number(opaque_element_version_section->selected_version) ==
            spdm_get_version_from_version_number(spdm_context->local_context.secured_message_version.spdm_version[secured_message_version_index])) {
            break;
        }
        return RETURN_UNSUPPORTED; <== ??? Why it returns unsupported here?
    }
    return RETURN_SUCCESS;
```

Wenxing-hou:
I will check it.

jyao1:
Also this error can be caught by "C4702: unreachable code".

I think we need remove /wd4702.

PR #530, opened 2022-01-21
Fix the spdm_process_opaque_data_version_selection_data() logic error

Wenxing-hou:
Fix: #529 


jyao1:
Please remove /wd4702 compiler option, which is useful to detect such issue earlier.

jyao1:
Since #531 will handle /wd4702, this patch seems good.

steven-bellock (approved):

PR #531, opened 2022-01-22
Remove /wd4702 compiler option

Wenxing-hou:
Remove wd4702 compiler option, which is useful to detect such issue earlier.


jyao1:
looks good

steven-bellock (approved):

Issue #467, opened 2021-12-28
try_spdm_get_version frees m1m2 hash after clearing the base hash algorithm information


jcahill0:
When the SPDM requestor gets through attestation (all the way to CHALLENGE), then starts all over at GET_VERSION, an assert is hit when trying to grab the hash free function.  It is hit because the function try_spdm_get_version incorrectly clears the base hash algorithm information via libspdm_reset_context.  Below is the reset context that clears algorithms substructure:
```
void libspdm_reset_context(IN void *context)
{
    spdm_context_t *spdm_context;
    uintn index;

    spdm_context = context;
    /*Clear all info about last connection*/
    zero_mem(&spdm_context->connection_info.capability, sizeof(spdm_device_capability_t));
    zero_mem(&spdm_context->connection_info.algorithm, sizeof(spdm_device_algorithm_t));
```
After sending the request, try_spdm_get_version tries to free the hash within libspdm_reset_message_b, but the algorithm.base_hash_algo is 0 from earlier.
```
void libspdm_reset_message_b(IN void *context)
{
    spdm_context_t *spdm_context;

    spdm_context = context;
#if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT
    reset_managed_buffer(&spdm_context->transcript.message_b);
#else
    if (spdm_context->transcript.digest_context_m1m2 != NULL) {
        libspdm_hash_free (spdm_context->connection_info.algorithm.base_hash_algo,
            spdm_context->transcript.digest_context_m1m2);
        spdm_context->transcript.digest_context_m1m2 = NULL;
    }
#endif
}
```
Therefore the assert in libspdm_hash_free_func is hit because it cannot find the right freeing function since base_hash_algo is 0.
```
libspdm_hash_free_func get_spdm_hash_free_func(IN uint32_t base_hash_algo)
{
    switch (base_hash_algo) {
...
    ASSERT(FALSE);
    return NULL;
}
```

It may be worthwhile to add a few test cases to make sure transitions from different states back to GET_VERSION are correct.

jyao1:
I think we need move libspdm_reset_message_*() before libspdm_reset_context().

Wenxing-hou:
I will check it.

PR #533, opened 2022-01-24
Move libspdm_reset_message_*() before libspdm_reset_context()

Wenxing-hou:
Fix: #467 

The req_get_vesrion should free hash first when the m1m2 didn't equal 0.
Then req_get_vesrion reset context.


jyao1:
looks good

steven-bellock (approved):
There's still the larger issue of tests that encompass multiple request / responses, that would be able to catch this particular issue.

Issue #525, opened 2022-01-20
spdm_general_opaque_data_table_header_t definition error


jyao1:
https://github.com/DMTF/libspdm/blob/main/include/industry_standard/spdm.h#L364

```
typedef struct {
    uint8_t total_elements;
    uint16_t reserved;
    /*opaque_element_table_t  opaque_list[];*/
} spdm_general_opaque_data_table_header_t;
```

The reserved field should be 3 bytes, not 2 bytes.

See: SPDM 1.2: Table 92 — General opaque data format

Wenxing-hou:
I will check it.

PR #535, opened 2022-01-25
Define spdm_general_opaque_data_table_header_t correctly

Wenxing-hou:
Fix: #525 


jyao1:
looks good.

Wenxing-hou:
I have passed  the spdm-emu and spdm-dump test.

Issue #489, opened 2022-01-05
Responder need send SPDM_ERROR_CODE_DECRYPT_ERROR or drop error message based upon policy.


jyao1:
Some other standard spec may require to drop the error message in a session on integrity check failure.

We need a policy to control if a decryption failure will generate SPDM_ERROR_CODE_DECRYPT_ERROR  response, or drop the request silently.

jyao1:
Related issue #488

Wenxing-hou:
I will check it.

PR #537, opened 2022-01-26
Add a plocicy to control send DECRYPT_ERROR or drop request silently

Wenxing-hou:
Fix: #489

Add error_return_policy to control send DECRYPT_ERROR or drop request.
If the handle_error_return_policy BIT0 set, drop the request silently.

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -376,6 +376,13 @@ typedef struct {
      * or secured SPDM transport message.
      **/
     uint8_t request_response[LIBSPDM_MAX_MESSAGE_BUFFER_SIZE];
+
+    /**
+     * The BIT0 control to generate SPDM_ERROR_CODE_DECRYPT_ERROR response or drop the request silently.
+     * If the BIT0 is not set, generate SPDM_ERROR_CODE_DECRYPT_ERROR response.
+     * If the BIT0 set, drop the request silently.
+     **/
+    uint8_t handle_error_return_policy;
 } spdm_context_t;
 
 /**
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -115,6 +115,13 @@ typedef enum {
 
     LIBSPDM_DATA_APP_CONTEXT_DATA,
 
+    /**
+     * The LIBSPDM_DATA_HANDLE_ERROR_RETURN_POLICY BIT0 control to generate SPDM_ERROR_CODE_DECRYPT_ERROR response or drop the request silently.
+     * If the BIT0 is not set, generate SPDM_ERROR_CODE_DECRYPT_ERROR response.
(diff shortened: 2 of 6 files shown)


jyao1:
This is incorrect.

Why you use bit 7 as indicator ?

 if(((session_info->session_policy) & BIT7) != 0) { /*if the policy bit7 set, return decrypt error*/

Wenxing-hou:
I have changed the code. And the code passed on unit test and spdm-emu text

include/internal/libspdm_common_lib.h
```diff
@@ -373,6 +373,10 @@ typedef struct {
     // or secured SPDM transport message.
     //
     uint8_t request_response[LIBSPDM_MAX_MESSAGE_BUFFER_SIZE];
+
+    /*The BIT0 control to generate SPDM_ERROR_CODE_DECRYPT_ERROR response or drop the request silently.*/
+    /*If the BIT0 is not set, generate SPDM_ERROR_CODE_DECRYPT_ERROR response.*/
```
jyao1, line 378: Please state what if BIT0 is set, what happens.

include/library/spdm_common_lib.h
```diff
@@ -114,6 +114,10 @@ typedef enum {
 
     LIBSPDM_DATA_APP_CONTEXT_DATA,
 
+    /*The BIT0 control to generate SPDM_ERROR_CODE_DECRYPT_ERROR response or drop the request silently.*/
+    /*If the BIT0 is not set, generate SPDM_ERROR_CODE_DECRYPT_ERROR response.*/
+    LIBSPDM_DATA_HANDLE_ERROR_RETURN_POLICY,
```
jyao1, line 119: Please state what if BIT0 is set, what happens.

Wenxing-hou: OK. I have added the comments.

jyao1:
It seems this patch has completeness problem. Below error is not fixed.

```
    if (spdm_context->last_spdm_error.error_code != 0) {
        
        /* Error in libspdm_process_request(), and we need send error message directly.*/
        
        my_response_size = sizeof(my_response);
        zero_mem(my_response, sizeof(my_response));
        switch (spdm_context->last_spdm_error.error_code) {
        case SPDM_ERROR_CODE_DECRYPT_ERROR:
            /* session ID is valid. Use it to encrypt the error message.*/
            status = libspdm_generate_error_response(
                spdm_context, SPDM_ERROR_CODE_DECRYPT_ERROR, 0,
                &my_response_size, my_response);
            break;
```

jyao1 (approved):

steven-bellock (approved):

jyao1:
I just thought more cases to close session, such as GetMeasurement in session.
We can approve this at first, and there will need to be a follow-up pull request to address issues.

PR #539, opened 2022-01-26
Add rnglib_null to support size calculation.

jyao1:
It is commented by default.
It will be used to calculate size with real cryptolib used such as mbedtls.


steven-bellock (changes requested):

jyao1:
Updated according to the feedback.

steven-bellock (approved):

Issue #454, opened 2021-12-22
C89 Compatibility Issues


steven-bellock:
While attempting to enable C89 compilation I ran into the following fundamental issues.
- C89 does not support 64-bit integers (`long long int`). 64-bit integers are mainly used in the secure message lib to keep track of sequence numbers and IVs. The `timeout` parameter is also 64-bit. Luckily GCC supports `long long int` in C89.
- C89 does not support variadic macros. This is mainly a debug library issue, specifically the macro `_DEBUG_PRINT(PrintLevel, ...)`. The benefit of having it as a macro is that it can be optimized away via a #define. The alternative would be to ditch the macro and perform a function call directly to `debug_print(...)` and rely on link-time optimizations to get rid of the function call if not needed.

So the question is how important is it to work around these limitations?

jyao1:
I try to be realistic. The importance depends upon if there is any usage of C89 only compiler for the libspdm project.

My suggest is to check if there is anything broken.
If nothing is broken, then we dont fix.
If something is broken, let's review what is broken is how to fix/work-around it.

steven-bellock:
The use of `static_assert` is also an issue but we can work around that with `#error` which is a c89 construct.

steven-bellock:
I got close to full support, but the 64-bit integers are a sticking point. An integrator will have to do a little work for it to compile under a C89 compiler.

PR #551, opened 2022-01-28
Add _Static_assert support for C89 and C99

ekohandel:
_Static_assert was only introduced in C11, this implementation
which I have taken from glibc:
https://patchwork.ozlabs.org/project/glibc/patch/54D40347.3020508@cs.ucla.edu/
allows usage of _Static_assert on earlier compliers.

This change partially addresses the C89 compatibility brought up in #454

### DIFF
--- a/include/hal/base.h
+++ b/include/hal/base.h
@@ -645,6 +645,10 @@ typedef uintn *BASE_LIST;
 #define STATIC_ASSERT(expression, message)
 #elif _MSC_EXTENSIONS
 #define STATIC_ASSERT static_assert
+#elif !defined(_Static_assert)
+#define STATIC_ASSERT(expression, message) \
+    extern int (*__Static_assert_function (void)) \
+        [!!sizeof (struct { int __error_if_negative: (expression) ? 2 : -1; })]
 #else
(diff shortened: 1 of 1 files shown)


jyao1:
I will reject this patch, because of license reason.

glibc is LGPL license.

libspdm is BSD license.

So we cannot include any code with LGPL license.

ekohandel:
> I will reject this patch, because of license reason.
> 
> glibc is LGPL license.
> 
> libspdm is BSD license.
> 
> So we cannot include any code with LGPL license.
> 
> if we want to add this feature, we can only 1) refer C11, (or) 2) write code by yourself.

I wasn't sure if this was acceptable hence I was very clear on the source of the above code. I will explore alternatives.

steven-bellock:
I don't think we need `static_assert` at all. For the compile-time checks we're interested in, mainly checking that #defines are legal, we can use `#error` similar to what MbedTLS does [here](https://github.com/ARMmbed/mbedtls/blob/6249603e7c23aa15327ae740d1b8f66c32d3204d/include/mbedtls/check_config.h).

PR #552, opened 2022-01-28
Protect redefinition of common defines

ekohandel:
MAX, MIN, NULL, TRUE, and FALSE are very common defines, redefining them
causes warnings if they are already defined.

This change only redefines these common macros if they are not already
defined.

### DIFF
--- a/include/hal/base.h
+++ b/include/hal/base.h
@@ -217,17 +217,23 @@
 
 /* Boolean true value.*/
 
+#ifndef TRUE
 #define TRUE ((boolean)(1 == 1))
+#endif
 
 
 /* Boolean false value.*/
 
+#ifndef FALSE
 #define FALSE ((boolean)(0 == 1))
+#endif
 
 
 /* NULL pointer (void *)*/
 
+#ifndef NULL
 #define NULL ((void *)0)
+#endif
 
 
 /* Maximum values for common data Types*/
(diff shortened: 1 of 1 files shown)


jyao1:
looks good

steven-bellock (approved):
This is fine in the short-term, especially if it is causing issues in your build system. Long-term I think it should be
- `NULL` comes from `stddef.h`.
- `true`, `false`, and the `bool` datatype come from `stdbool.h` or are user-defined if the compiler only supports c89.
- `MIN` and `MAX` should be `LIBSPDM_MIN` and `LIBSPDM_MAX` if they are user-visible, else their #defines should be moved to an internal header.

ekohandel:
> This is fine in the short-term, especially if it is causing issues in your build system. Long-term I think it should be
> 
> * `NULL` comes from `stddef.h`.
> * `true`, `false`, and the `bool` datatype come from `stdbool.h` or are user-defined if the compiler only supports c89.
> * `MIN` and `MAX` should be `LIBSPDM_MIN` and `LIBSPDM_MAX` if they are user-visible, else their #defines should be moved to an internal header.

@steven-bellock, yeah this is a short term fix to address warnings in our build environment. I agree with your statements, would be great to move to using standard library provided defines.

Possibly a larger discussion but it would be nice to move to standard types for `uintn`, `intn`, `boolean`, and `char8` as well.

steven-bellock:
`uintn` and `intn` are on their way out, `boolean` will become standard `bool`, and `char8` is just `char`. I'm not sure why `char8` exists. The `boolean` and `char8` conversion can probably begin right away; I'll make issues for them. The `uintn` conversion needs more investigation, since it's used in pointer arithmetic.

Issue #490, opened 2022-01-05
Responder need support a callback function to drop the response.


jyao1:
Other standard may support drop a request in SPDM vendor defined message on error. (not return response.) Our code need support this case.

Current code seems always assume the response is returned without checking my_response_size == 0.

```
            status = ((libspdm_get_response_func)
                      spdm_context->get_response_func)(
                spdm_context, session_id, is_app_message,
                spdm_context->last_spdm_request_size,
                spdm_context->last_spdm_request,
                &my_response_size, my_response);


    status = spdm_context->transport_encode_message(
        spdm_context, session_id, is_app_message, FALSE,
        my_response_size, my_response, response_size, response);
    if (RETURN_ERROR(status)) {
        DEBUG((DEBUG_INFO, "transport_encode_message : %p\n", status));
        return status;
    }
```

jyao1:
Related issue #489

Wenxing-hou:
I will check it.

PR #574, opened 2022-01-30
Add a check when callback function drop the response

Wenxing-hou:
Fix: #490 

When the callback function drop the response, the return response_size is 0.
And the return status is RETURN_UNSUPPORTED.


library/spdm_responder_lib/libspdm_rsp_receive_send.c
```diff
@@ -418,6 +418,14 @@ return_status libspdm_build_response(IN void *context, IN uint32_t *session_id,
             status = RETURN_NOT_FOUND;
         }
     }
+    /*if return the status: Responder drop the response*/
+    /* just ignore this message */
+    /* return UNSUPPORTED and clear response_size to continue the dispatch without send response.*/
```
jyao1, line 423: Please add above comment to "libspdm_get_response_func" API definition in "spdm_responder_lib.h"

Wenxing-hou: Yes. I have added the comment.

jyao1 (approved):

PR #577, opened 2022-01-31
Initial draft of status codes and handling

steven-bellock:
@jyao1 @taprinz @richkong88 

Some things to consider / contemplate.
- I got rid of the error origin because in a lot of cases the origin is either unknown or not easily attributable to another entity.
- Having these unique error codes has the potential to increase code size beyond an acceptable level for some integrators. It's possible we might offer an optimization where libpsdm only distinguishes between error (status == 1) and success (status == 0). This would mean libspdm would be prohibited from taking different actions based on the status, ie it couldn't do `if status == LIBSPDM_STATUS_INVALID_STATE_PEER) do_something;`
- The list of error codes is obviously not complete.


steven-bellock:
@ekohandel Do you know why this failed the [style compliance check](https://github.com/DMTF/libspdm/runs/5014133499?check_suite_focus=true)?

ekohandel:
> @ekohandel Do you know why this failed the [style compliance check](https://github.com/DMTF/libspdm/runs/5014133499?check_suite_focus=true)?

It is complaining about larger then 4 space indentation on the line after `#define`:
```c
#define LIBSPDM_STATUS_SUCCESS \
	LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_SUCCESS, LIBSPDM_SOURCE_SUCCESS, 0x0000)
```
It wants it to be:
```c
#define LIBSPDM_STATUS_SUCCESS \
    LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_SUCCESS, LIBSPDM_SOURCE_SUCCESS, 0x0000)
```

But this points out that I need to improve logging and output the changes uncrustify is expecting.

jyao1:
I think the initial draft is good. Let's move forward.

I guess we might revisit the draft, after we see more real cases in the code.

ekohandel:
> > @ekohandel Do you know why this failed the [style compliance check](https://github.com/DMTF/libspdm/runs/5014133499?check_suite_focus=true)?
> 
> It is complaining about larger then 4 space indentation on the line after `#define`:
> 
> ```c
> #define LIBSPDM_STATUS_SUCCESS \
> 	LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_SUCCESS, LIBSPDM_SOURCE_SUCCESS, 0x0000)
> ```
> 
> It wants it to be:
> 
> ```c
> #define LIBSPDM_STATUS_SUCCESS \
>     LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_SUCCESS, LIBSPDM_SOURCE_SUCCESS, 0x0000)
> ```
> 
> But this points out that I need to improve logging and output the changes uncrustify is expecting.

@steven-bellock, I added #580 to output the diff expected by uncrustify when it fails.

include/library/spdm_return_status.h
```diff
@@ -0,0 +1,90 @@
+#ifndef SPDM_RETURN_STATUS_H
+#define SPDM_RETURN_STATUS_H
+
+typedef uint32_t libspdm_return_t;
+
+/* Returns 0 if status is LIBSPDM_STATUS_SUCCESS else it returns 1. */
```
taprinz, line 6: Would be nice to have a comment/diagram at the top that shows the bits of the return status, which bits correspond to what fields, and what is reserved/unused

taprinz:
> Some things to consider / contemplate.
> * Having these unique error codes has the potential to increase code size beyond an acceptable level for some integrators. It's possible we might offer an optimization where libpsdm only distinguishes between error (status == 1) and success (status == 0). This would mean libspdm would be prohibited from taking different actions based on the status, ie it couldn't do `if status == LIBSPDM_STATUS_INVALID_STATE_PEER) do_something;`

It would be good to have an idea of how much code size might increase (although obviously hard to guess impact without having performed all the replacements). I think the ability to at least have some ability to take actions based on the status would be beneficial - for example, for certain errors we could maybe determine whether a retry is possible, or if we need to error out hard.

The makeup & definitions look good to me otherwise, though.

richkong88:
I know this isn't in the current spec, but I found it useful for some internal work, to return a slightly different error value depending on which if statement triggered an error.

For example, in the following code, there are 3 different error cases, but they all return the same thing. 

```    if (spdm_context->connection_info.connection_state !=
        LIBSPDM_CONNECTION_STATE_AFTER_CAPABILITIES) {
        return libspdm_generate_error_response(spdm_context,
                                               SPDM_ERROR_CODE_UNEXPECTED_REQUEST,
                                               0, response_size, response);
    }

    if (request_size < sizeof(spdm_negotiate_algorithms_request_t)) {
        return libspdm_generate_error_response(spdm_context,
                                               SPDM_ERROR_CODE_INVALID_REQUEST, 0,
                                               response_size, response);
    }
    if (request_size <
        sizeof(spdm_negotiate_algorithms_request_t) +
        sizeof(uint32_t) * spdm_request->ext_asym_count +
        sizeof(uint32_t) * spdm_request->ext_hash_count +
        sizeof(spdm_negotiate_algorithms_common_struct_table_t) *
        spdm_request->header.param1) {
        return libspdm_generate_error_response(spdm_context,
                                               SPDM_ERROR_CODE_INVALID_REQUEST, 0,
                                               response_size, response);
    }```

 I thought there was an "origin" discussed before within the error structure, which we could differentiate situations similar to this. I'm wondering if we could bring that back into the error fields returned at least internally to libspdm, as it could facilitate debug, especially in environments where its tough to get a debugger. I know I saved a lot of time by knowing the exact if statement that failed by having extra error data.

steven-bellock:
> > Some things to consider / contemplate.
> > 
> > * Having these unique error codes has the potential to increase code size beyond an acceptable level for some integrators. It's possible we might offer an optimization where libpsdm only distinguishes between error (status == 1) and success (status == 0). This would mean libspdm would be prohibited from taking different actions based on the status, ie it couldn't do `if status == LIBSPDM_STATUS_INVALID_STATE_PEER) do_something;`
> 
> It would be good to have an idea of how much code size might increase (although obviously hard to guess impact without having performed all the replacements). I think the ability to at least have some ability to take actions based on the status would be beneficial - for example, for certain errors we could maybe determine whether a retry is possible, or if we need to error out hard.
> 
> The makeup & definitions look good to me otherwise, though.

@taprinz We can leave it is-as and then if it becomes a problem revisit the issue. I don't think libspdm will have many cases where it evaluates the status; it just passes it up to the integrator. If we need to implement simple status codes it it isn't very difficult.

> I thought there was an "origin" discussed before within the error structure, which we could differentiate situations similar to this. I'm wondering if we could bring that back into the error fields returned at least internally to libspdm, as it could facilitate debug, especially in environments where its tough to get a debugger. I know I saved a lot of time by knowing the exact if statement that failed by having extra error data.

@richkong88 Origin wouldn't help in the case of, for example, isolating which parameter of a message is invalid. Seems like you'd want to use `DEBUG()` whenever an error is encountered that has the additional info. In the case of validating a message the return status would be `LIBSPDM_STATUS_INVALID_MESS_FIELD` and then the `DEBUG()` would state exactly which field is bad.

include/library/spdm_return_status.h
steven-bellock: It is done.

richkong88:
> @richkong88 Origin wouldn't help in the case of, for example, isolating which parameter of a message is invalid. Seems like you'd want to use `DEBUG()` whenever an error is encountered that has the additional info. In the case of validating a message the return status would be `LIBSPDM_STATUS_INVALID_MESS_FIELD` and then the `DEBUG()` would state exactly which field is bad.

Yeah, having DEBUG print is ideal, but may be difficult in some environments, or if you are the requester and don't have access to the responder source. I feel like we the "error data" passed back in the requester and it is usually 0. If we could make use of it and send back some non-zero code instead, it could be helpful to speed up integration. In any case, this is against spec, so I'll bring it up with the spec owners and drive it there.

steven-bellock:
> > @richkong88 Origin wouldn't help in the case of, for example, isolating which parameter of a message is invalid. Seems like you'd want to use `DEBUG()` whenever an error is encountered that has the additional info. In the case of validating a message the return status would be `LIBSPDM_STATUS_INVALID_MESS_FIELD` and then the `DEBUG()` would state exactly which field is bad.
> 
> Yeah, having DEBUG print is ideal, but may be difficult in some environments, or if you are the requester and don't have access to the responder source. I feel like we the "error data" passed back in the requester and it is usually 0. If we could make use of it and send back some non-zero code instead, it could be helpful to speed up integration. In any case, this is against spec, so I'll bring it up with the spec owners and drive it there.

@xiaoyuruan beat you to it. https://github.com/DMTF/Security-TF/issues/1784

jyao1:
Question on severity [bit 31]:
1) In original proposal, BIT-31 is ERROR. Here is is severity. Are they same? or do they have different meaning?
2) What if BIT-31 is 0? if non-severity error == warning ?

jyao1:
> Yeah, having DEBUG print is ideal, but may be difficult in some environments, or if you are the requester and don't have access to the responder source. I feel like we the "error data" passed back in the requester and it is usually 0. If we could make use of it and send back some non-zero code instead, it could be helpful to speed up integration. In any case, this is against spec, so I'll bring it up with the spec owners and drive it there.

I don't believe spec will be updated shortly. I recommend we separate them to two different issues.
First, we need a solution in Q1, under current restriction - no error data. There is no need to add origin.
Second, if spec is updated later, we can figure out how to incorporate. We may add origin in reserved field, which is a compatible change.

richkong88:
> > > @richkong88 Origin wouldn't help in the case of, for example, isolating which parameter of a message is invalid. Seems like you'd want to use `DEBUG()` whenever an error is encountered that has the additional info. In the case of validating a message the return status would be `LIBSPDM_STATUS_INVALID_MESS_FIELD` and then the `DEBUG()` would state exactly which field is bad.
> > 
> > 
> > Yeah, having DEBUG print is ideal, but may be difficult in some environments, or if you are the requester and don't have access to the responder source. I feel like we the "error data" passed back in the requester and it is usually 0. If we could make use of it and send back some non-zero code instead, it could be helpful to speed up integration. In any case, this is against spec, so I'll bring it up with the spec owners and drive it there.
> 
> @xiaoyuruan beat you to it. https://github.com/DMTF/Security-TF/issues/1784

Xiaoyu is on my team. I brought it up to him to bring into the spec :)

steven-bellock:
@jyao1 I've updated the severity so that it is four bits now.

@richkong88 It is truly a small world we inhabit.

jyao1 (approved):

PR #578, opened 2022-02-01
Add CI build matrix strategy

ekohandel:
This change adds a matrix strategy in the build step to start adding
combinations of configurations the library supports.

Not all combinations are built right now, nor do they all compile so we
will need to fill these out as we iron that out.


jyao1 (approved):

steven-bellock (approved):
This is fine for now but that `exclude:` list might grow large as more parameters are added.

ekohandel:
> This is fine for now but that `exclude:` list might grow large as more parameters are added.

Totally agree! I wasn't sure this would be the best approach, as we adjust the configurations the strategy here can change. We can also just craft full combination of CMAKE configuration parameters as part of the strategy.

PR #579, opened 2022-02-01
Proposal for copy_mem_s function that adds the destination size as parameter.

richkong88:
@jyao1 @steven-bellock @taprinz 
This is a proposal for a copy_mem_s function that behaves similar to memcpy_s. This is just a review for the function change itself.  Once this new function is approved, I will add another PR to change all the calling locations to the new function and remove the older copy_mem function.

### DIFF
--- a/include/hal/library/memlib.h
+++ b/include/hal/library/memlib.h
@@ -40,7 +40,7 @@
  *
  * @return   0 on success. non-zero on error.
  *
-**/
+ **/
 int copy_mem_s(OUT void* dst_buf, IN size_t dst_len, IN const void* src_buf, IN size_t len);
 
 int copy_mem(OUT void* dst_buf, IN const void* src_buf, IN uintn len);
--- a/include/hal/library/memlib.h
+++ b/include/hal/library/memlib.h
@@ -21,15 +21,16 @@
  * This function copies "len" bytes from "src_buf" to "dst_buf".
  *
  * Asserts and returns a non-zero value if any of the following are true:
- *   1) ("src_buf" or "dst_buf" are NULL) and "len" is greater 0.
+ *   1) "src_buf" or "dst_buf" are NULL.
  *   2) "src_buf" and "dst_buf" overlap.
  *   3) "len" is greater than "dst_len".
  *   4) "len" or "dst_len" is greater than (MAX_ADDRESS - "dst_buf" + 1).
  *   5) "len" or "dst_len" is greater than (MAX_ADDRESS - "src_buf" + 1).
  *   6) "len" or "dst_len" is greater than (SIZE_MAX >> 1).
  *
- * In case of error, the "dst_buf" is left unmodifed. This behavior is different
- * than memcopy_s, which zeros the "dst_buf" if "dst_buf" is valid.
+ * In case of error, the "dst_buf" is left unmodifed.
+ * This behavior is different than C11 memcopy_s, which zeros the "dst_buf"
+ * if there is a runtime violation.
  *
  * @param    dst_buf   Destination buffer to copy to.
  * @param    dst_len   Maximum length in bytes of the destination buffer.
--- a/include/hal/library/memlib.h
+++ b/include/hal/library/memlib.h
@@ -24,9 +24,12 @@
  *   1) ("src_buf" or "dst_buf" are NULL) and "len" is greater 0.
  *   2) "src_buf" and "dst_buf" overlap.
  *   3) "len" is greater than "dst_len".
- *   4) "len" is greater than (MAX_ADDRESS - "dst_buf" + 1).
- *   5) "len" is greater than (MAX_ADDRESS - "src_buf" + 1).
- *   6) "len" is greater than (SIZE_MAX >> 1).
+ *   4) "len" or "dst_len" is greater than (MAX_ADDRESS - "dst_buf" + 1).
+ *   5) "len" or "dst_len" is greater than (MAX_ADDRESS - "src_buf" + 1).
+ *   6) "len" or "dst_len" is greater than (SIZE_MAX >> 1).
+ *
+ * In case of error, the "dst_buf" is left unmodifed. This behavior is different
+ * than memcopy_s, which zeros the "dst_buf" if "dst_buf" is valid.
  *
  * @param    dst_buf   Destination buffer to copy to.
  * @param    dst_len   Maximum length in bytes of the destination buffer.
@@ -36,7 +39,7 @@
  * @return   0 on success. non-zero on error.
  *
 **/
-int copy_mem_s(OUT void* dst_buf, IN uintn dst_len, IN const void* src_buf, IN uintn len);
+int copy_mem_s(OUT void* dst_buf, IN size_t dst_len, IN const void* src_buf, IN size_t len);
 
 int copy_mem(OUT void* dst_buf, IN const void* src_buf, IN uintn len);
 
--- a/include/hal/library/memlib.h
+++ b/include/hal/library/memlib.h
@@ -15,8 +15,6 @@
 #ifndef __BASE_MEMORY_LIB__
 #define __BASE_MEMORY_LIB__
 
-#include <stddef.h>
-
 /**
  * Copies bytes from a source buffer to a destination buffer.
  *
@@ -41,7 +39,7 @@
  * @return   0 on success. non-zero on error.
  *
  **/
-int copy_mem_s(OUT void* dst_buf, IN size_t dst_len, IN const void* src_buf, IN size_t len);
+int copy_mem_s(OUT void* dst_buf, IN uintn dst_len, IN const void* src_buf, IN uintn len);
 
 int copy_mem(OUT void* dst_buf, IN const void* src_buf, IN uintn len);
 
--- a/include/hal/library/memlib.h
+++ b/include/hal/library/memlib.h
@@ -16,24 +16,29 @@
 #define __BASE_MEMORY_LIB__
 
 /**
- * Copies a source buffer to a destination buffer, and returns the destination buffer.
+ * Copies bytes from a source buffer to a destination buffer.
  *
- * This function copies length bytes from source_buffer to destination_buffer, and returns
- * destination_buffer.  The implementation must be reentrant, and it must handle the case
- * where source_buffer overlaps destination_buffer.
+ * This function copies "len" bytes from "src_buf" to "dst_buf".
  *
- * If length is greater than (MAX_ADDRESS - destination_buffer + 1), then ASSERT().
- * If length is greater than (MAX_ADDRESS - source_buffer + 1), then ASSERT().
+ * Asserts and returns a non-zero value if any of the following are true:
+ *   1) ("src_buf" or "dst_buf" are NULL) and "len" is greater 0.
+ *   2) "src_buf" and "dst_buf" overlap.
+ *   3) "len" is greater than "dst_len".
+ *   4) "len" is greater than (MAX_ADDRESS - "dst_buf" + 1).
+ *   5) "len" is greater than (MAX_ADDRESS - "src_buf" + 1).
+ *   6) "len" is greater than (SIZE_MAX >> 1).
  *
- * @param  destination_buffer   The pointer to the destination buffer of the memory copy.
- * @param  source_buffer        The pointer to the source buffer of the memory copy.
- * @param  length              The number of bytes to copy from source_buffer to destination_buffer.
+ * @param    dst_buf   Destination buffer to copy to.
+ * @param    dst_len   Maximum length in bytes of the destination buffer.
+ * @param    src_buf   Source buffer to copy from.
+ * @param    len       The number of bytes to copy.
  *
- * @return destination_buffer.
+ * @return   0 on success. non-zero on error.
  *
- **/
-void *copy_mem(OUT void *destination_buffer, IN const void *source_buffer,
-               IN uintn length);
+**/
+int copy_mem_s(OUT void* dst_buf, IN uintn dst_len, IN const void* src_buf, IN uintn len);
+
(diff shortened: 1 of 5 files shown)


os_stub/memlib/copy_mem.c
```diff
@@ -9,35 +9,65 @@
...
- * If length is greater than (MAX_ADDRESS - destination_buffer + 1), then ASSERT().
- * If length is greater than (MAX_ADDRESS - source_buffer + 1), then ASSERT().
+ * Asserts and returns a non-zero value if any of the following are true:
+ *   1) ("src_buf" or "dst_buf" are NULL) and "len" is greater 0.
+ *   2) "src_buf" and "dst_buf" overlap.
+ *   3) "len" is greater than "dst_len".
+ *   4) "len" is greater than (MAX_ADDRESS - "dst_buf" + 1).
+ *   5) "len" is greater than (MAX_ADDRESS - "src_buf" + 1).
+ *   6) "len" is greater than (SIZE_MAX >> 1).
  *
```
jyao1, line 26: I think we need add 4~6) shall also be applicable to "dst_len".

os_stub/memlib/copy_mem.c
```diff
@@ -9,35 +9,65 @@
...
- * This function copies length bytes from source_buffer to destination_buffer, and returns
- * destination_buffer.  The implementation must be reentrant, and it must handle the case
- * where source_buffer overlaps destination_buffer.
+ * This function copies "len" bytes from "src_buf" to "dst_buf".
  *
- * If length is greater than (MAX_ADDRESS - destination_buffer + 1), then ASSERT().
- * If length is greater than (MAX_ADDRESS - source_buffer + 1), then ASSERT().
+ * Asserts and returns a non-zero value if any of the following are true:
+ *   1) ("src_buf" or "dst_buf" are NULL) and "len" is greater 0.
```
jyao1, line 20: for 1), I recommend to remove "len is greater 0.". the src_buf and dst_buf shall not be NULL in all cases.

os_stub/memlib/copy_mem.c
```diff
@@ -9,35 +9,65 @@
  **/
 
 #include "base.h"
+#include "library/debuglib.h"
 
 /**
- * Copies a source buffer to a destination buffer, and returns the destination buffer.
+ * Copies bytes from a source buffer to a destination buffer.
  *
```
jyao1, line 16: I recommend we add description on the difference between this API and C11 memcpy_s().
For example, C11 requires to fill dst_buf with 0, in case of runtime violation. But it is not in this API.

include/hal/library/memlib.h
```diff
@@ -16,24 +16,29 @@
...
- * @return destination_buffer.
+ * @return   0 on success. non-zero on error.
  *
- **/
-void *copy_mem(OUT void *destination_buffer, IN const void *source_buffer,
-               IN uintn length);
+**/
+int copy_mem_s(OUT void* dst_buf, IN uintn dst_len, IN const void* src_buf, IN uintn len);
+
+int copy_mem(OUT void* dst_buf, IN const void* src_buf, IN uintn len);
```
steven-bellock, line 41: Is libspdm always going to use `copy_mem` ? Are there any cases where it would use copy_mem_s directly?

richkong88: I left copy_mem in temporarily just so I wouldn't have to change all the code across libspdm at this time. I intend to get rid of the 3 parameter version completely, leaving the 4 parameter version as the only option. I'm open to whether the 4 parameter version is named copy_mem or copy_mem_s.

os_stub/memlib/copy_mem.c
richkong88: Agree.

os_stub/memlib/copy_mem.c
richkong88: One of the test cases (test_spdm_responder_challenge_auth_case1) fails if I make this change. I'll investigate.

os_stub/memlib/copy_mem.c
richkong88: I think 3 implies 4-6 for "dst_len", but I can call it out explicitly.

richkong88: Ok.

richkong88: I think 4-6 are applicable for "dst_len", but the additional checks (4-6) against "dst_len" are unnecessary due to 3. Can someone else confirm?

os_stub/memlib/copy_mem.c
richkong88: A few places call copy_mem with null source and size 0. I'll push the code that corrects it for those places, caught by current unit tests.

include/hal/library/memlib.h
steven-bellock: If the old `copy_mem` is being retired then they new one should keep the `copy_mem` mantle.

include/hal/library/memlib.h
```diff
@@ -16,24 +16,29 @@
...
+ * @param    len       The number of bytes to copy.
  *
- * @return destination_buffer.
+ * @return   0 on success. non-zero on error.
  *
- **/
-void *copy_mem(OUT void *destination_buffer, IN const void *source_buffer,
-               IN uintn length);
+**/
+int copy_mem_s(OUT void* dst_buf, IN uintn dst_len, IN const void* src_buf, IN uintn len);
```
steven-bellock, line 39: Once it's in place the return type can be `libspdm_return_t` and return somewhat detailed information. Also the `uintn`s will be converted to `uint32_t`s.

os_stub/memlib/copy_mem.c
```diff
@@ -9,35 +9,68 @@
...
-void *copy_mem(OUT void *destination_buffer, IN const void *source_buffer,
-               IN uintn length)
+ * @return   0 on success. non-zero on error.
+ *
+**/
+int copy_mem_s(OUT void* dst_buf, IN uintn dst_len, IN const void* src_buf, IN uintn len)
 {
-    volatile uint8_t *pointer_dst;
-    const volatile uint8_t *pointer_src;
+    volatile uint8_t* dst;
```
steven-bellock, line 40: This is me being a C language lawyer but when incrementing/decrementing a pointer by a certain number of bytes the type should be `char`. It just so happens that `uint8_t` is a `unsigned char`.

include/hal/library/memlib.h
richkong88: Sounds good.

include/hal/library/memlib.h
richkong88: Very well.

os_stub/memlib/copy_mem.c
richkong88: Can you elaborate more on this one? I tend to favor unsigned types unless there is a need to be signed.

os_stub/memlib/copy_mem.c
jyao1: 3) is not enough.

For example: len == 10, dst_len == 50, (size_max >> 1) == 20.

It can pass all checks above. But it is invalid, because dst_len is too large. There must be something wrong.

os_stub/memlib/copy_mem.c
jyao1: I cannot understand the comment. @steven-bellock , could you please share the source of restriction? C standard? compiler?

include/hal/library/memlib.h
jyao1: can we use size_t to replace uintn, instead of uint32_t ?

os_stub/memlib/copy_mem.c
richkong88: Ok. Thanks.

os_stub/memlib/copy_mem.c
steven-bellock: Signedness doesn't play a role here, it's the definition of a byte and its association with `char`. Although musl uses `unsigned char` for some reason. https://git.musl-libc.org/cgit/musl/tree/src/string/memcpy.c#n7

steven-bellock: I guess because of `(uintptr_t)s` it needs to be of a specific signedness.

include/hal/library/memlib.h
steven-bellock: In this specific case `size_t` would be consistent with `memcpy`. However `memcpy_s` uses `rsize_t` which is only in c11.

jyao1: Not sure if we really need rsize_t, to me size_TD is good enough.

os_stub/memlib/copy_mem.c
jyao1: What is conclusion here? I am lost.

steven-bellock: Heh forget I brought it up. In the grand scheme of things `uint8_t` is fine since it will always be `unsigned char`.

richkong88:
@ekohandel Can you clarify the error on the compliance check? Thanks.

ekohandel:
> @ekohandel Can you clarify the error on the compliance check? Thanks.

@richkong88, the [log](https://github.com/DMTF/libspdm/runs/5056538295?check_suite_focus=true) outputs the failure.

richkong88:
> > @ekohandel Can you clarify the error on the compliance check? Thanks.
> 
> @richkong88, the [log](https://github.com/DMTF/libspdm/runs/5056538295?check_suite_focus=true) outputs the failure.

Yes, I checked the log, but it's not clear to me.

ekohandel:
> > > @ekohandel Can you clarify the error on the compliance check? Thanks.
> > 
> > 
> > @richkong88, the [log](https://github.com/DMTF/libspdm/runs/5056538295?check_suite_focus=true) outputs the failure.
> 
> Yes, I checked the log, but it's not clear to me.

The specific compliance check is doing style checks `(Compliance Checks / Style Compliance)` so you are missing some spaces to align things as the style checker expects.

richkong88:
> 
> The specific compliance check is doing style checks `(Compliance Checks / Style Compliance)` so you are missing some spaces to align things as the style checker expects.

Ok. It looks like it expects parameters carried over onto the next line to be lined up with the first parameter from the previous line, whereas I just indent by 4 spaces on the next line. Was also missing a space at the end of the function header.  The "-" lines in the log say what it is currently, and the "+" lines say what is expected.

ekohandel:
> > The specific compliance check is doing style checks `(Compliance Checks / Style Compliance)` so you are missing some spaces to align things as the style checker expects.
> 
> Ok. It looks like it expects parameters carried over onto the next line to be lined up with the first parameter from the previous line, whereas I just indent by 4 spaces on the next line. Was also missing a space at the end of the function header. The "-" lines in the log say what it is currently, and the "+" lines say what is expected.

Yeah, the easiest way to generated a report was to apply the changes and output a git diff. So that is what you are seeing.

richkong88:
I left out the changes for size_t, because they were causing problems with linux builds, even though I was including stddef.h. Can leave that for a future endeavor.

jyao1:
> I left out the changes for size_t, because they were causing problems with linux builds, even though I was including stddef.h. Can leave that for a future endeavor.

Agree

os_stub/memlib/copy_mem.c
jyao1: Thank you. I see you updated the code. Would you please update comment as well?

richkong88: I think I updated the copy function comments over a couple different change sets. If you look at the final changes, they should be in there?

include/hal/library/memlib.h
```diff
@@ -16,24 +16,32 @@
...
  *
- * This function copies length bytes from source_buffer to destination_buffer, and returns
- * destination_buffer.  The implementation must be reentrant, and it must handle the case
- * where source_buffer overlaps destination_buffer.
+ * This function copies "len" bytes from "src_buf" to "dst_buf".
  *
- * If length is greater than (MAX_ADDRESS - destination_buffer + 1), then ASSERT().
- * If length is greater than (MAX_ADDRESS - source_buffer + 1), then ASSERT().
+ * Asserts and returns a non-zero value if any of the following are true:
+ *   1) ("src_buf" or "dst_buf" are NULL) and "len" is greater 0.
```
jyao1, line 24: "len is greater 0." should be removed?

os_stub/memlib/copy_mem.c
jyao1: "len is greater 0" is still there in the final, which is not expected. Would you please double check?

include/hal/library/memlib.h
```diff
@@ -16,24 +16,32 @@
...
+ *   3) "len" is greater than "dst_len".
+ *   4) "len" or "dst_len" is greater than (MAX_ADDRESS - "dst_buf" + 1).
+ *   5) "len" or "dst_len" is greater than (MAX_ADDRESS - "src_buf" + 1).
+ *   6) "len" or "dst_len" is greater than (SIZE_MAX >> 1).
  *
- * @param  destination_buffer   The pointer to the destination buffer of the memory copy.
- * @param  source_buffer        The pointer to the source buffer of the memory copy.
- * @param  length              The number of bytes to copy from source_buffer to destination_buffer.
+ * In case of error, the "dst_buf" is left unmodifed. This behavior is different
+ * than memcopy_s, which zeros the "dst_buf" if "dst_buf" is valid.
```
jyao1, line 32: I would say:
This behavior is different with C11 memcpy_s, which zeros the "dst_buf" if there is runtime violation.

include/hal/library/memlib.h
richkong88: Ah, yes. Sorry, I thought you were referring to updating "dst_len"

include/hal/library/memlib.h
richkong88: ok.

jyao1 (approved):

PR #591, opened 2022-02-03
fix build issue with new test cases.

jyao1:


jyao1:
Merge it quickly, to fix build error.

PR #597, opened 2022-02-07
Add full response version info in get_version API.

jyao1:
It will be used in conformance test.


steven-bellock:
Let's talk about this one at Monday's meeting, since this gets into modifying the integrator-facing API to enable testing.

steven-bellock (approved):

PR #601, opened 2022-02-07
Zero copy_mem_s dst_buf on some errors. Fix copy_mem return value bug..

richkong88:

### DIFF
--- a/include/hal/library/memlib.h
+++ b/include/hal/library/memlib.h
@@ -18,29 +18,41 @@
 /**
  * Copies bytes from a source buffer to a destination buffer.
  *
- * This function copies "len" bytes from "src_buf" to "dst_buf".
+ * This function copies "src_len" bytes from "src_buf" to "dst_buf".
  *
  * Asserts and returns a non-zero value if any of the following are true:
  *   1) "src_buf" or "dst_buf" are NULL.
- *   2) "src_buf" and "dst_buf" overlap.
- *   3) "len" is greater than "dst_len".
- *   4) "len" or "dst_len" is greater than (MAX_ADDRESS - "dst_buf" + 1).
- *   5) "len" or "dst_len" is greater than (MAX_ADDRESS - "src_buf" + 1).
- *   6) "len" or "dst_len" is greater than (SIZE_MAX >> 1).
+ *   2) "src_len" or "dst_len" is greater than (SIZE_MAX >> 1).
+ *   3) "src_len" is greater than "dst_len".
+ *   4) "src_buf" and "dst_buf" overlap.
  *
- * In case of error, "dst_len" bytes of "dst_buf" is zeroed, if "dst_buf"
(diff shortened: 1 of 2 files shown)


os_stub/memlib/copy_mem.c
```diff
@@ -41,32 +42,40 @@ int copy_mem_s(OUT void* dst_buf, IN uintn dst_len, IN const void* src_buf, IN u
...
+        dst_len > MAX_ADDRESS - (uintn)dst + 1 ||
+        dst_len > MAX_ADDRESS - (uintn)src + 1 ||
+        dst_len > (SIZE_MAX >> 1)) {
         ASSERT(0);
         return -1;
     }
 
+    /* Gaurd against invalid source. Zero "dst" in this case. */
+    if (src == NULL) {
+        set_mem(dst_buf, dst_len, 0);
```
jyao1, line 60: you can use zero_mem() API.

include/hal/library/memlib.h
```diff
@@ -28,9 +28,9 @@
  *   5) "len" or "dst_len" is greater than (MAX_ADDRESS - "src_buf" + 1).
  *   6) "len" or "dst_len" is greater than (SIZE_MAX >> 1).
  *
- * In case of error, the "dst_buf" is left unmodifed.
- * This behavior is different than C11 memcopy_s, which zeros the "dst_buf"
- * if there is a runtime violation.
+ * In case of error, "dst_len" bytes of "dst_buf" is zeroed, if "dst_buf"
+ * points to a non-NULL value and "dst_len" does not exceed
+ * ((MAX_ADDRESS - "dst_buf" + 1) or (MAX_ADDRESS - "src_buf" + 1) or (SIZE_MAX >> 1)).
  *
```
jyao1, line 34: I suggest you can mention, it is aligned to C11 standard

Please double check MSFT memcpy_s() - https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/memcpy-s-wmemcpy-s, to see they are aligned.
Also also mention that in the comment, in case someone wants to use C11 version or MSFT version memcpy_s().

jyao1: If memcopy_s is different in windows and linux world, how we can implement copy_mem_s() via memcopy_s() ?

richkong88: I was going to follow the cppreference [description](https://en.cppreference.com/w/c/string/byte/memcpy) of memcpy_s and document the differences with MS and safeclib. In particular, cppreference does not allow for NULL dest or source, and only says non-zero is returned on error, rather than a specific error code. 

If clients want to call safeclib or the microsoft version, they need to be aware of the differences. I will push a change you can review.

jyao1:
@richkong88 , any more update for this one?

richkong88:
No more update by me. I was waiting for yours @jyao1 and @steven-bellock approval to get checked in. Then I'll proceed with changing the functions calls across unit_test and library.

PR #617, opened 2022-02-09
Allow build command line configurations

ekohandel:
It is desirable to be able to configure the library at build time via
the command line. This change introduces the ability to set the CFLAGS
environment variable which cmake uses to initialize any options it
passes to the build targets.

cmake ignores the CFLAGS environment variable if CMAKE_C_FLAGS is set.
In order to allow us to use CFLAGS, ADD_COMPILE_OPTIONS is used instead
of explicitly setting CMAKE_C_FLAGS.


jyao1 (approved):

steven-bellock (approved):

qizhangz:
The change in CMakeLists.txt need to be ported to spdm-emu.

Issue #618, opened 2022-02-09
session_id is used wrongly for watchdog.


qizhangz:
when session_id is NULL, should use latest_session_id for watchdog.

    } else {
        switch (spdm_response->request_response_code) {
        case SPDM_FINISH_RSP:
            if (spdm_is_capabilities_flag_supported(
                    spdm_context, false,
                    SPDM_GET_CAPABILITIES_REQUEST_FLAGS_HANDSHAKE_IN_THE_CLEAR_CAP,
                    SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_HANDSHAKE_IN_THE_CLEAR_CAP)) {
                spdm_set_session_state(
                    spdm_context,
                    `spdm_context->latest_session_id`,
                    LIBSPDM_SESSION_STATE_ESTABLISHED);
                session_state_established = true;
            }
            break;
        default:
            /* No session state update needed */
            break;
        }
    }

qizhangz:
spdm-emu will fail due to this bug.

PR #619, opened 2022-02-09
correct session id for watchdog.

qizhangz:
Fix: #618 


jyao1 (approved):

jyao1:
merge it soon, because spdm emu will fail without it.

Issue #625, opened 2022-02-10
AddressSanitizer detected the direct memory leak with the initial corpus and aborted the execution.


Zhiqiang520:
1. Verified on 2474af3
Ubuntu 20.04.2 LTS

```
cd libspdm
mkdir build_libfuzz_mbedtls_2474af3
cd build_libfuzz_mbedtls_2474af3
cmake -DARCH=x64 -DTOOLCHAIN=LIBFUZZER -DTARGET=Debug -DCRYPTO=mbedtls -DGCOV=ON ..
make copy_sample_key
make
cd bin
cp -r ~/libspdm/unit_test/fuzzing/seeds ./
./test_spdm_responder_heartbeat_ack ./seeds/test_spdm_responder_heartbeat_ack/heartbeat.raw
INFO: Seed: 2568356757
INFO: Loaded 1 modules   (12952 inline 8-bit counters): 12952 [0x8c7eba, 0x8cb152),
INFO: Loaded 1 PC tables (12952 PCs): 12952 [0x8cb158,0x8fdad8),
./test_spdm_responder_heartbeat_ack: Running 1 inputs 1 time(s) each.
Running: ./seeds/test_spdm_responder_heartbeat_ack/heartbeat.raw

=================================================================
==7335==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 1344 byte(s) in 1 object(s) allocated from:
    #0 0x5312cd in malloc (/home/test/libspdm/build_libfuzz_mbedtls_2474af3/bin/test_spdm_responder_heartbeat_ack+0x5312cd)
    #1 0x5eca36 in read_responder_public_certificate_chain /home/test/libspdm/os_stub/spdm_device_secret_lib_sample/cert.c:298:26
    #2 0x561359 in test_spdm_responder_heartbeat /home/test/libspdm/unit_test/fuzzing/test_responder/test_spdm_responder_heartbeat_ack/heartbeat_ack.c:54:5
    #3 0x561b2c in run_test_harness /home/test/libspdm/unit_test/fuzzing/test_responder/test_spdm_responder_heartbeat_ack/heartbeat_ack.c:103:5
    #4 0x56378e in LLVMFuzzerTestOneInput /home/test/libspdm/unit_test/fuzzing/spdm_unit_fuzzing_common/toolchain_harness.c:144:5
[...]

PR #630, opened 2022-02-11
fix the direct memory leak with the initial corpus for test_spdm_resp…

Zhiqiang520:
…onder_encap_key_update and so on

fix: #625


jyao1:
Would you please isolate bug fix in the code, and bug fix in the test case?

Zhiqiang520:
> Would you please isolate bug fix in the code, and bug fix in the test case?

I have updated the PR, now it only bug fix in the test case. I have created a standalone PR for bug fix in the code.

Issue #503, opened 2022-01-08
SPDM 1.2 need handle data_transfer_size and max_spdm_msg_size capability


jyao1:
Missing 1.2 feature, after https://github.com/DMTF/libspdm/pull/477

For chunk, ref https://github.com/DMTF/libspdm/issues/504

jyao1:
1/10/2022 - not sure how to use two fields without chunking. Maybe add them together with chunking.

qizhangz:
I will check this

jyao1:
This issue can only handle data_transfer_size.

max_spdm_msg_size  can be done in #504

PR #635, opened 2022-02-14
handle data_transfer_size

qizhangz:
Fix: #503 

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_send_receive.c
+++ b/library/spdm_requester_lib/libspdm_req_send_receive.c
@@ -190,6 +190,11 @@ return_status spdm_send_spdm_request(IN spdm_context_t *spdm_context,
     spdm_session_info_t *session_info;
     libspdm_session_state_t session_state;
 
+    if ((spdm_context->connection_info.capability.data_transfer_size != 0) &&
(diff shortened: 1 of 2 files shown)


library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -190,6 +190,11 @@ return_status spdm_send_spdm_request(IN spdm_context_t *spdm_context,
     spdm_session_info_t *session_info;
     libspdm_session_state_t session_state;
 
+    if ((spdm_context->connection_info.capability.data_transfer_size != 0) &&
+        (request_size > spdm_context->connection_info.capability.data_transfer_size)) {
+        return RETURN_DEVICE_ERROR;
```
jyao1, line 195: I think we need return better error code, such as RETURN_BAD_BUFFER_SIZE

steven-bellock: For now I would say just leave it as `RETURN_DEVICE_ERROR`. It will get changed during the return status sweep.

jyao1: May I know what is problem of using RETURN_BAD_BUFFER_SIZE ?

steven-bellock (approved):

PR #638, opened 2022-02-15
Use copy_mem_s in unit_test/fuzzing

richkong88:
Move all of unit_test/fuzzing to use copy_mem_s

### DIFF
--- a/unit_test/fuzzing/spdm_unit_fuzzing_common/toolchain_harness.c
+++ b/unit_test/fuzzing/spdm_unit_fuzzing_common/toolchain_harness.c
@@ -139,7 +139,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
     if (size > max_buffer_size) {
         size = max_buffer_size;
     }
-    copy_mem(test_buffer, data, size);
+    copy_mem_s(test_buffer, max_buffer_size, data, size);
     /* 2. Run test*/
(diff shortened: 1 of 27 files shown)


steven-bellock:
Are we eventually going to check the return value?

richkong88:
> Are we eventually going to check the return value?

I generally don't on these low level functions. They do assert in debug. I look at the release mode behavior as more of a last ditch effort to prevent a buffer overflow for a case that should generally never occur, but if it does could be a bad security bug, so prevent the overflow. 

Using the function is safer than writing my own for loop to do the memory copy, and a for loop in its place would not have a return value to check anyway.  I'm interested to others thoughts on the topic though. 

For ULTs I'm not sure its really necessary with asserts already there. Just be sure you run ULTs in debug mode in addition to release.

jyao1:
> I generally don't on these low level functions. They do assert in debug. I look at the release mode behavior as more of a last ditch effort to prevent a buffer overflow for a case that should generally never occur, but if it does could be a bad security bug, so prevent the overflow.
> 
> Using the function is safer than writing my own for loop to do the memory copy, and a for loop in its place would not have a return value to check anyway. I'm interested to others thoughts on the topic though.
> 
> For ULTs I'm not sure its really necessary with asserts already there. Just be sure you run ULTs in debug mode in addition to release.

agree

jyao1:

PR #639, opened 2022-02-15
Use copy_mem_s in unit_test/ various tests.

richkong88:


jyao1:

PR #640, opened 2022-02-15
Use copy_mem_s in unit_test/test_spdm_responder/*

richkong88:
Change code under unit_test/test_spdm_responder to use copy_mem_s.


jyao1:

PR #648, opened 2022-02-17
Use MbedTLS's build system

ekohandel:
In order to better decouple libspdm from MbedTLS's internals, we should
not be reaching into MbedTLS's source directories.

This change uses build scripts and configurations provided by MbedTLS to
create the mbedcrypto, mbedx509, and mbedtls artifacts for use by libspdm.

This change also positions us better for integrating out of tree builds
of mbedtls as mbedcrypto, mbedx509, and mbedtls are the normal
artifacts from a standalone mbedtls build. It also positions us better
for upgrading to later versions of MbedTLS as it reduces coupling to
internals of MbedTLS.


jyao1 (approved):

jyao1:
Good isolation!

Issue #570, opened 2022-01-29
Consider to remove IN, OUT, OPTIONAL and replace them by const.


jyao1:
Current code is using IN, OUT, OPTIONAL macro to provide description on parameter.
This cannot be enforced by compiler. A better way is to use const.
IN == const.
OUT == no const.
OPTIONAL - discard.

richkong88:
Code is also using INOUT. We should be using const anyway, but I do like the IN, INOUT, OUT. Perhaps it should be specified in the comments?

jyao1:
Agree. We don't need update comment.
This issue only need remove IN/OUT/OPTIONAL modifier in the code.

qizhangz:
I will check this

PR #651, opened 2022-02-17
remove IN, OUT, OPTIONAL macro.

qizhangz:
Fix: #570 


jyao1:
We dont need add "const" for basic type such as "uint32_t". We just need "const" for pointer, such as "uint32_t *".

jyao1:
If there is issue, we can resolve in other patch.

PR #662, opened 2022-02-18
Add Linux script to execute code beautifier.

steven-bellock:
Windows batch file to come in separate pull request.


steven-bellock:
> @steven-bellock, should we just have the CI workflow call this script then? We probably don't want to duplicate the logic.

Yeah maybe have the default behavior be to modify the source files when run, and we could add a `--check` option to just check the files.

script/format_nix.sh
```diff
@@ -0,0 +1,38 @@
...
+    exit 1
+fi
+
+# Change directory to top of repository.
+cd `dirname $0`
+cd ../
+
+# Exclude non-libspdm submodules.
+EXCLUDE_PATH1="./unit_test/test_size/intrinsiclib/ia32/*"
+EXCLUDE_PATH2="./os_stub/mbedtlslib/mbedtls/*"
```
ekohandel, line 20: I think we can use `git ls-files | sed "/unit_test\/test_size\/intrinsiclib\/ia32/d" | sed -n '/.\.[ch]$/p' | uncrustify -q -c ./.uncrustify.cfg --replace --no-backup -F -`. This way we don't have to explicitly mention the submodules and is forward compatible with more submodules being added.

steven-bellock: That's modifying files in `mbedtls/` and `openssl/`. How about I keep this as-is and then we can refactor it later.

ekohandel: Really? That is surprising! There is a specific `--recurse-submodules` option to recurse submodules and I didn't see it modify them on my testing :(.

ekohandel (approved):

Issue #576, opened 2022-01-31
Need add LIBSPDM_ or libspdm_ for any external symbol or MACRO


jyao1:
For example, https://github.com/DMTF/libspdm/issues/222

steven-bellock:
In addition, symbols and MACROS that are currently publicly viewable but don't need to be should be moved to an internal header so that the integrator doesn't have to see them and we don't have to document them.

qizhangz:
I will check this

jyao1:
Please also resolve missing camel case issue. such as:
```
SpdmGetEncapResponseFuncViaRequestCode
SpdmProcessEncapsulatedRequest
mSpdmGetResponseStruct
SpdmRequesterSendMessage
SpdmRequesterReceiveMessage
SpdmResponderSendMessage
SpdmResponderReceiveMessage
mSpdmRequesterGetVersionTestContext
```

Also remove unused `SPDM_TEST_CONTEXT_FROM_SPDM_PROTOCOL` and `SpdmGetResponseFinish`.

Issue #222, opened 2021-10-22
DEBUG macros


rkongintel:
I noticed the use of DEBUG macro which compiles to _DEBUG then _DEBUG_PRINT. 

debuglib.h
#define _DEBUG_PRINT(PrintLevel, ...)                                          \
               do {                                                                   \
                              debug_print(PrintLevel, ##__VA_ARGS__);                        \
               } while (FALSE)
#define _DEBUG(expression) _DEBUG_PRINT expression

#if !defined(MDEPKG_NDEBUG)
#define DEBUG(expression)                                                      \
               do {                                                                   \
                              _DEBUG(expression);                                            \
               } while (FALSE)
#else
#define DEBUG(expression)
#endif

I think we should use a different name than DEBUG and _DEBUG for debug prints as these seem common enough names that clients may already define this for their platforms. For example, standard Windows builds in VC++ define _DEBUG and NDEBUG to distinguish debug/release builds. 

Since _DEBUG only defines to _DEBUG_PRINT, perhaps we can reduce the 3 macros down to 2 and get rid of _DEBUG. The DEBUG and _DEBUG_PRINT will be good enough and change the name so it is DEBUG_PRINT and DEBUG_PRINT_EX to reduce chance of potential build environment conflict?

Example:

#define DEBUG_PRINT_EX(PrintLevel, ...)                                          \
               do {                                                                   \
                              debug_print(PrintLevel, ##__VA_ARGS__);                        \
               } while (FALSE)

#if !defined(MDEPKG_NDEBUG)
#define DEBUG_PRINT(expression)                                                      \
               do {                                                                   \
                              DEBUG_PRINT_EX(expression);                                            \
               } while (FALSE)
#else
#define DEBUG_PRINT(expression)
#endif

steven-bellock:
Voted to close at 6-Dec-2021 meeting.

PR #689, opened 2022-03-01
add libspdm_ for any external symbol or MACRO.

qizhangz:
for debuglib.

Fix: #576 
Fix: #222


jyao1:
See https://github.com/DMTF/libspdm/issues/222 as well

@richkong88 , would you please take a look as well?

richkong88:
Looks ok to me.

PR #707, opened 2022-03-07
Transition GET_VERSION to new error codes

steven-bellock:
Add spdm_return_status.h to library\spdm_common_lib.h.
Create a new macro to check and return on error.

### DIFF
--- a/include/library/spdm_return_status.h
+++ b/include/library/spdm_return_status.h
@@ -43,6 +43,14 @@ typedef uint32_t libspdm_return_t;
 #define LIBSPDM_STATUS_SUCCESS \
     LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_SUCCESS, LIBSPDM_SOURCE_SUCCESS, 0x0000)
 
+#define LIBSPDM_RET_ON_ERR(status) \
+    do { \
+        if (LIBSPDM_STATUS_IS_ERROR(status)) { \
+            return (status); \
+        } \
+    } \
+    while (0)
+
 /* Core errors. */
 
 /* Unable to complete operation due to unsupported capabilities by the caller. */
@@ -68,6 +76,14 @@ typedef uint32_t libspdm_return_t;
 /* The received message contains one or more invalid message fields. */
 #define LIBSPDM_STATUS_INVALID_MESS_FIELD \
     LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_ERROR, LIBSPDM_SOURCE_CORE, 0x0005)
+
+/* The received message's size is invalid. */
+#define LIBSPDM_STATUS_INVALID_MESS_SIZE \
+    LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_ERROR, LIBSPDM_SOURCE_CORE, 0x0006)
+
+/* Unable to derive a common set of versions, algorithms, etc. */
+#define LIBSPDM_STATUS_NEGOTIATION_FAIL \
+    LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_ERROR, LIBSPDM_SOURCE_CORE, 0x0007)
 
 /* Cryptography errors. */
 
--- a/include/library/spdm_return_status.h
+++ b/include/library/spdm_return_status.h
@@ -86,6 +86,14 @@ typedef uintn libspdm_return_t;
 #define LIBSPDM_STATUS_NEGOTIATION_FAIL \
     LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_ERROR, LIBSPDM_SOURCE_CORE, 0x0007)
 
+/* Received a Busy error message. */
+#define LIBSPDM_STATUS_BUSY_PEER \
+    LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_ERROR, LIBSPDM_SOURCE_CORE, 0x0008)
+
+/* Received an unexpected error message. */
+#define LIBSPDM_STATUS_ERROR_PEER \
+    LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_ERROR, LIBSPDM_SOURCE_CORE, 0x0009)
+
 /* Cryptography errors. */
 
 /* Verification of the provided signature failed. */
@@ -109,7 +117,6 @@ typedef uintn libspdm_return_t;
 /* Unable to receive message from peer. */
 #define LIBSPDM_STATUS_RECEIVE_FAIL \
     LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_ERROR, LIBSPSM_SOURCE_TRANSPORT, 0x0001)
-
 
 /* Random number generation errors. */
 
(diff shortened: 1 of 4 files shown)


steven-bellock:
@richkong88 @taprinz I've defined `LIBSPDM_STATUS_INVALID_MESS_FIELD` and `LIBSPDM_STATUS_INVALID_MESS_SIZE` and you can see how they're used in the pull request. Trying to strike a balance between having tons of different error codes and simplicity, especially since we need to list all the `@retval`s for documentation.

library/spdm_requester_lib/libspdm_req_get_version.c
```diff
@@ -126,14 +115,14 @@ return_status libspdm_try_get_version(libspdm_context_t *spdm_context,
...
-                         sizeof(spdm_version_number_t));
+        return LIBSPDM_STATUS_NEGOTIATION_FAIL;
     }
 
+    libspdm_copy_mem(&(spdm_context->connection_info.version),
+                     sizeof(spdm_context->connection_info.version),
+                     &(common_version),
+                     sizeof(spdm_version_number_t));
+
     if (version_number_entry_count != NULL && version_number_entry != NULL) {
```
steven-bellock, line 126: @jyao1 I am slightly confused by this code. This is only to be used for testing purposes? Are you assuming that the compiler will remove this code if it's not being used for testing?

library/spdm_requester_lib/libspdm_req_get_version.c
```diff
@@ -52,51 +51,43 @@ return_status libspdm_try_get_version(libspdm_context_t *spdm_context,
     libspdm_reset_message_buffer_via_request_code(spdm_context, NULL,
                                                   spdm_request.header.request_response_code);
 
-    status = libspdm_send_spdm_request(spdm_context, NULL,
-                                       sizeof(spdm_request), &spdm_request);
-    if (RETURN_ERROR(status)) {
-        return status;
-    }
+    LIBSPDM_RET_ON_ERR(libspdm_send_spdm_request(spdm_context, NULL,
```
jyao1, line 54: I don't have concern to define LIBSPDM_RET_ON_ERR().

But I do have concern to use it in this way, because it makes the step by step debug hard.

I prefer we take 2 steps:
```
    status = libspdm_send_spdm_request(spdm_context, NULL, sizeof(spdm_request), &spdm_request);
    LIBSPDM_RET_ON_ERR(status);
```

As such, when I trace the flow I can know what error status is returned.

include/library/spdm_return_status.h
```diff
@@ -69,6 +77,14 @@ typedef uint32_t libspdm_return_t;
 #define LIBSPDM_STATUS_INVALID_MESS_FIELD \
     LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_ERROR, LIBSPDM_SOURCE_CORE, 0x0005)
 
+/* The received message's size is invalid. */
+#define LIBSPDM_STATUS_INVALID_MESS_SIZE \
```
jyao1, line 81: Now, I feel that we may consider to use `MSG` instead of `MESS`. :)

library/spdm_requester_lib/libspdm_req_get_version.c
jyao1: I will expose the libspdm_get_version() API in requesterlib.

As such, a consumer may want use libspdm_get_version(), libspdm_get_capabilities(), libspdm_negotiate_algorithms() directly instead of libspdm_init_connection()

library/spdm_requester_lib/libspdm_req_get_version.c
richkong88: I concur that this should be split across two statements. It'll make debugging easier.

jyao1:
looks good to me now.
If you can make it pass CI, we can merge it.

unit_test/test_spdm_requester/get_version.c
```diff
@@ -480,13 +480,13 @@ void libspdm_test_requester_get_version_case5(void **state)
     spdm_test_context->case_id = 0x5;
 
     status = libspdm_get_version(spdm_context, NULL, NULL);
-    assert_int_equal(status, RETURN_NO_RESPONSE);
+    assert_int_equal(status, LIBSPDM_STATUS_INVALID_MSG_FIELD);
 }
 
 /**
  * Test 6: on the first try, receiving a Busy ERROR message, and on retry, receiving
```
steven-bellock, line 487: @jyao1 I don't understand how this test works. It says that GET_VERSION is called twice, but it looks like it's only called once.

jyao1: That is because case 5 will return SPDM_ERROR_CODE_BUSY.

Then the GET_VERSION will be sent again.

steven-bellock: What about [line 141](https://github.com/DMTF/libspdm/blob/1d2f6f30dd585cb39dccbf299db25b10dcb01710/unit_test/test_spdm_requester/get_version.c#L141), that has both the `SPDM_ERROR_CODE_BUSY` and the successful `VERSION` response?

jyao1: Right. Typo, case 6.
It use "static uintn sub_index1 = 0;" to track how many times the interface is called.

If it is first time, the test responder return BUSY. Then the case expect the requester issues GET_VERSION again.
Second time, the test responder return SUCCESS.

steven-bellock: Right but in `libspdm_test_requester_get_version_case6` `libspdm_get_version` is only called once, not twice. And `libspdm_test_requester_get_version_case6` is only called once according to `cmocka_unit_test(libspdm_test_requester_get_version_case6)` (line 726).

jyao1: Yes, libspdm_get_version() internal implementation will call GET_VERSION twice.

steven-bellock: Ah I understand now. I had thought that the Responder could not reply to a `GET_VERSION` request with a `Busy` error, but it's only `ResponseNotReady` that it can't reply with.

steven-bellock: As an aside, I filed https://github.com/DMTF/libspdm/issues/717. Thanks @jyao1

unit_test/test_spdm_requester/get_version.c
```diff
@@ -542,7 +538,7 @@ void libspdm_test_requester_get_version_case8(void **state)
     spdm_test_context->case_id = 0x8;
 
     status = libspdm_get_version(spdm_context, NULL, NULL);
-    assert_int_equal(status, RETURN_DEVICE_ERROR);
+    assert_int_equal(status, LIBSPDM_STATUS_INVALID_MSG_FIELD);
 }
 
 /**
```
steven-bellock, line 544: @jyao1 We can get rid of Test 9, correct? It doesn't do what it says it does and is effectively the same as Test 8.

steven-bellock, line 544: @jyao1 We can get rid of Test 9, correct? It doesn't do what it says it does and is effectively the same as Test 8.

jyao1: Agree. I think there is misunderstanding for NOT_READY for GET_VERSION in the beginning.

We should remove it now.

steven-bellock: I've removed Test 9.

steven-bellock:
I will update the documentation for `libspdm_init_connection` once I finish capabilities and algorithms.

PR #721, opened 2022-03-14
Initial commit for new error codes for GET_CAPABILITIES

steven-bellock:
Still need to update unit tests 30 and 31.


steven-bellock:
@jyao1 @richkong88 This adds new error codes `LIBSPDM_STATUS_RESYNCH_PEER` and `LIBSPDM_STATUS_BUFFER_FULL`. I talked with @raghuncstate on what the Requester should do if it issues a `GET_VERSION` request and the Responder returns a `RequestResynch` error. We think that would be weird but not illegal. It is up to the integrator to decide whether they want to re-issue the `GET_VERSION`.

Issue #725, opened 2022-03-14
test_spdm_requester_key_exchange hang with the AFL generated specific mutation seed


Zhiqiang520:
1.verify on https://github.com/DMTF/libspdm/commit/57ac200e7c0c8f633f1259a71d2f20c1c5f7b014
Ubuntu 20.04.2 LTS
```
cd libspdm
mkdir build_fuzz_mbedtls_57ac200
cd build_fuzz_mbedtls_57ac200
export AFL_PATH=~/afl-2.52b/
export PATH=$PATH:$AFL_PATH
cmake -DARCH=x64 -DTOOLCHAIN=AFL -DTARGET=Release -DCRYPTO=mbedtls -DGCOV=ON ..
make copy_sample_key
make
cd bin
./test_spdm_requester_key_exchange ./test_spdm_requester_key_exchange_hang
ClientRandomData (0x20) - e3 76 1c df f9 91 bb 2f b1 23 0d 2d 75 d2 ce f7 43 4a 5e 7f 77 83 bf e5 00 e1 71 b5 56 c1 ff bb
ClientKey (0x40):
0000: 2e fe ab f9 ef ce ee cf 2a ad a7 5c 21 08 99 f7 61 50 62 68 fe 4e 53 d0 3c 2e 75 62 65 cc d7 5e
0020: 92 ba 81 a1 6d 43 b6 07 b6 21 10 ad 18 09 8d 88 08 6a 31 91 27 38 cb 8a f7 d1 be 49 24 91 8e 39
libspdm_send_spdm_request[0] (0x7e):
0000: 11 e4 00 00 ff ff 00 00 e3 76 1c df f9 91 bb 2f b1 23 0d 2d 75 d2 ce f7 43 4a 5e 7f 77 83 bf e5
0020: 00 e1 71 b5 56 c1 ff bb 2e fe ab f9 ef ce ee cf 2a ad a7 5c 21 08 99 f7 61 50 62 68 fe 4e 53 d0
0040: 3c 2e 75 62 65 cc d7 5e 92 ba 81 a1 6d 43 b6 07 b6 21 10 ad 18 09 8d 88 08 6a 31 91 27 38 cb 8a
0060: f7 d1 be 49 24 91 8e 39 14 00 46 54 4d 44 01 01 00 00 00 00 05 00 01 01 01 00 10 00 00 00
m_libspdm_local_buffer_size (0xf8):
[...]

Issue #722, opened 2022-03-14
test_spdm_requester_challenge hang with the AFL generated specific mutation seed


Zhiqiang520:
1. verify on 57ac200
Ubuntu 20.04.2 LTS
```
cd libspdm
mkdir build_fuzz_mbedtls_57ac200
cd build_fuzz_mbedtls_57ac200
export AFL_PATH=~/afl-2.52b/
export PATH=$PATH:$AFL_PATH
cmake -DARCH=x64 -DTOOLCHAIN=AFL -DTARGET=Release -DCRYPTO=mbedtls -DGCOV=ON ..
make copy_sample_key
make
cd bin
./test_spdm_requester_challenge ./test_spdm_requester_challenge_hang
ClientNonce - 08 50 a0 d7 b1 af c3 da aa 23 12 48 d5 a5 4b fe e5 32 e3 68 b8 56 0d f8 2c 19 c6 6f b7 98 d1 32
libspdm_send_spdm_request[0] (0x24):
0000: 11 83 00 00 08 50 a0 d7 b1 af c3 da aa 23 12 48 d5 a5 4b fe e5 32 e3 68 b8 56 0d f8 2c 19 c6 6f
0020: b7 98 d1 32
0000: 11 83 00 00 08 50 a0 d7 b1 af c3 da aa 23 12 48 d5 a5 4b fe e5 32 e3 68 b8 56 0d f8 2c 19 c6 6f
0020: b7 98 d1 32 d1 03 00 01 15 10 9b 97 d0 6c 86 62 56 db 7d 1b 3c be 2f 45 bf 7e e3 5f 3b e6 7f 90
0040: 54 b6 0f 22 2b 7c 20 d8 bb d5 7b 0d 82 57 c0 89 4c cc 5a 5a d4 a6 94 7c 12 48 70 75 4a 05 96 8a
0060: b7 88 07 d9 f0 cd 29 e1 00 00
0000: 11 83 00 00 08 50 a0 d7 b1 af c3 da aa 23 12 48 d5 a5 4b fe e5 32 e3 68 b8 56 0d f8 2c 19 c6 6f
0020: b7 98 d1 32 d1 03 00 01 15 10 9b 97 d0 6c 86 62 56 db 7d 1b 3c be 2f 45 bf 7e e3 5f 3b e6 7f 90
[...]

PR #728, opened 2022-03-15
The caller need guarantee the version is correct, both of MajorVersio…

Zhiqiang520:
…n and MinorVersion should be less than 10.
fix: #722
fix: #725


jyao1:
Would you please go through all the test cases to ensure the same patch is applied for all?

PR #730, opened 2022-03-15
Add Xtensa architecture support

ekohandel:
This changes introduces support for Cadence's Xtensa architecture


steven-bellock:
The `processor_bind.h` files are being phased out. @ekohandel can you workaround this issue locally? Especially since it doesn't seem like you need to update CMakeFiles.txt.

ekohandel:
> The `processor_bind.h` files are being phased out. @ekohandel can you workaround this issue locally? Especially since it doesn't seem like you need to update CMakeFiles.txt.

Absolutely, I wasn't sure if we were moving in the improved direction of removing those files. I will keep this local for now until they are removed.

jyao1:
Once we fix https://github.com/DMTF/libspdm/issues/867, then this is not needed.

So far, I think we need add this.

include/hal/xtensa/processor_bind.h
```diff
@@ -0,0 +1,61 @@
...
+ *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
+ **/
+
+#ifndef __PROCESSOR_BIND_H__
+#define __PROCESSOR_BIND_H__
+
+
+/* Define the processor type so other code can make processor based choices*/
+
+#define MDE_CPU_XTENSA
```
jyao1, line 13: Align with the latest style, it should be `LIBSPDM_CPU_XTENSA`

include/hal/xtensa/processor_bind.h
```diff
@@ -0,0 +1,61 @@
...
+
+
+#ifndef LIBSPDM_STDINT_ALT
+#include <stdint.h>
+#else
+#include LIBSPDM_STDINT_ALT
+#endif
+
+typedef char char8;
```
jyao1, line 22: remove this.

include/hal/xtensa/processor_bind.h
```diff
@@ -0,0 +1,61 @@
...
+
+
+/* Maximum legal xtensa address*/
+
+#define MAX_ADDRESS 0xFFFFFFFF
+
+
+/* Maximum legal xtensa intn values.*/
+
+#define MAX_INTN ((intn)0x7FFFFFFF)
```
jyao1, line 59: remove this

include/hal/xtensa/processor_bind.h
```diff
@@ -0,0 +1,61 @@
...
+#endif
+
+typedef char char8;
+
+#ifndef LIBSPDM_STDBOOL_ALT
+#include <stdbool.h>
+#else
+#include LIBSPDM_STDBOOL_ALT
+#endif
+
```
jyao1, line 29: We need add below to align with other arch.
```
#ifndef LIBSPDM_STDDEF_ALT
#include <stddef.h>
#else
#include LIBSPDM_STDDEF_ALT
#endif
```

Issue #799, opened 2022-03-28
Need update readme for new device IO APIs.


jyao1:

jyao1:
Also need document the threat model for the scratch buffer.
If it can be used to hold decrypted message. If yes, then who should clear it.

PR #808, opened 2022-03-28
Update readme for io buffer redesign.

jyao1:
Fix: https://github.com/DMTF/libspdm/issues/799

### DIFF
--- a/doc/user_guide.md
+++ b/doc/user_guide.md
@@ -30,20 +30,43 @@ Please refer to spdm_client_init() in [spdm_requester.c](https://github.com/DMTF
 
 1. Initialize SPDM context
 
-   1.1, allocate buffer for the spdm_context and initialize it.
+   1.1, allocate buffer for the spdm_context, initialize it, and setup scratch_buffer.
+   The spdm_context may include the decrypted secured message or session key.
+   The scratch buffer may include the decrypted secured message.
+   The spdm_context and scratch buffer shall be zeroed before freed or reused.
 
    ```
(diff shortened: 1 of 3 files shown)


doc/user_guide.md
```diff
@@ -30,20 +30,41 @@ Please refer to spdm_client_init() in [spdm_requester.c](https://github.com/DMTF
...
+   1.1, allocate buffer for the spdm_context, initialize it, and setup scratch_buffer.
+   The spdm_context may include the decrypted secured message or session key.
+   The scratch buffer may include the decrypted secured message.
+   The spdm_context and scratch buffer shall be zeroed before freed or reused.
 
    ```
    spdm_context = (void *)malloc (libspdm_get_context_size());
    libspdm_init_context (spdm_context);
+
+   scratch_buffer_size = libspdm_get_sizeof_required_scratch_buffer(m_spdm_context);
```
steven-bellock, line 42: Integrator should also be able to get the size statically to be used in a linker script.

steven-bellock (changes requested):

doc/user_guide.md
jyao1: Are you proposing to remove get_scratch_buffer_size, but define a MACRO in spdm_common_lib.h ?

steven-bellock: Both ways should be available. It can be called `LIBSPDM_SCRATCH_BUFFER_SIZE` or something similar.

jyao1: Doc updated

steven-bellock (approved):

PR #812, opened 2022-03-29
correct path of spdm_lib_config.h in spdm_crypt_lib.h.

qizhangz:
Fix: #811 


jyao1:
I think we need fix https://github.com/DMTF/libspdm/blob/main/include/library/spdm_device_secret_lib.h#L11 as well.

Issue #744, opened 2022-03-17
GET_DIGESTS API / Implementation changes


steven-bellock:
The current API / implementation of `libspdm_get_digest` is inadequate for many use cases. In particular it assumes that when the Responder has multiple certificate chain slots that the certificate chains are identical and produce the same digests, as well as returning an error if the Responder's digest does not match the Requester's digest. 

Will update this issue with more information once I'm done with the error code work.

jyao1:
Current libspdm_verify_peer_digests() logic is:
```
If the peer cert chain is provision, then
--> For each digest in digest list, do
------> compare the digest with cert chain hash
------> If match, then return pass
--> Endfor
--> Return fail
Else
--> Pass
Endif
```

So, if you want to skip the check and return all digest, just not provision a peer cert chain.
BTW: You may still provision a root cert, it has no impact.

steven-bellock:
Thanks, I misread the code. In that case the only issue in the code is that https://github.com/DMTF/libspdm/blob/714eceee428fb85d1b48705e0bee9f7b267a08a7/library/spdm_requester_lib/libspdm_req_get_digests.c#L187-L190 should be moved before the verification step.

steven-bellock:
For the API there is some information that can technically be recovered from `total_digest_buffer` which libspdm could provide. In particular
- whether the `DIGESTS` response is identical to the previous `DIGESTS` response.
- whether the slot id of the provisioned certificate changed since the last `DIGESTS` response.

Right now the integrator must cache `total_digest_buffer` to find this information. If we wanted libpsdm to provide this information then `total_digest_buffer` would be stored in the context.

jyao1:
Yes. I think we can add total_digest_buffer to context size.

PR #821, opened 2022-03-30
Cache total digest buffer to context.

jyao1:
Fix: https://github.com/DMTF/libspdm/issues/744


library/spdm_requester_lib/libspdm_req_get_digests.c
```diff
@@ -184,6 +184,12 @@ libspdm_return_t libspdm_try_get_digest(void *context, uint8_t *slot_mask,
                          spdm_response->digest, digest_size * digest_count);
     }
 
+    spdm_context->connection_info.peer_digest_slot_mask = spdm_response->header.param2;
```
steven-bellock, line 187: The verification of the digests needs to happen after `total_digest_buffer` has been copied.

steven-bellock (changes requested):

library/spdm_requester_lib/libspdm_req_get_digests.c
steven-bellock: It's fine if it returns an error but wouldn't the integrator want to know the new digest values if verification fails?

jyao1: The problem is: if error is returned, how the caller knows the value is filled or not.

jyao1: But in this case, why you need provision the whole cert chain to trigger the verification process?

steven-bellock: That's what's in the code now
```
if ((cert_chain_buffer != NULL) && (cert_chain_buffer_size != 0)) {
/* Check if digest of cert_chain_buffer matches a digest in the DIGEST response. */
```

jyao1: > That's what's in the code now

I am sorry, what is feedback ?

steven-bellock (approved):

PR #839, opened 2022-04-06
Fix memory leaks found in the new CI

Xiaohanjlll:


library/spdm_requester_lib/libspdm_req_get_digests.c
```diff
@@ -191,9 +191,12 @@ libspdm_return_t libspdm_try_get_digest(void *context, uint8_t *slot_mask,
         spdm_response->digest, digest_size * digest_count);
 
     spdm_context->connection_info.connection_state = LIBSPDM_CONNECTION_STATE_AFTER_DIGESTS;
-    status = LIBSPDM_STATUS_SUCCESS;
+
+    libspdm_release_receiver_buffer (spdm_context);
+    return LIBSPDM_STATUS_SUCCESS;
```
jyao1, line 196: you dont need duplicate the code.
You can use 
```
if (status != LIBSPDM_STATUS_SUCCESS) {
    libspdm_reset_message_b(spdm_context);
}
```

library/spdm_requester_lib/libspdm_req_get_measurements.c
```diff
@@ -516,9 +516,11 @@ libspdm_return_t libspdm_try_get_measurement(void *context, const uint32_t *sess
                          measurement_record_data_length);
     }
 
-    status = LIBSPDM_STATUS_SUCCESS;
+    libspdm_release_receiver_buffer (spdm_context);
+    return LIBSPDM_STATUS_SUCCESS;
 
 receive_done:
+    libspdm_reset_message_m(spdm_context, session_info);
```
jyao1, line 523: I dont think you can always reset message_m.

According to spec:
```
422 An error response with ErrorCode=ResponseNotReady shall not re-initialize L1/L2 - Requester and
Responder shall continue to construct L1/L2 with GET_MEASUREMENTS and MEASUREMENTS . An error
response with any error code other than ResponseNotReady shall re-initialize L1/L2 to null.
```

PR #853, opened 2022-04-18
spdm_common_lib: libspdm_com_opaque_data assumes only one version count

### LINKED ISSUE #852: spdm_common_lib: libspdm_com_opaque_data assumes only one version count
libspdm_com_opaque_data.c assumes only one version count.

For instance:
https://github.com/DMTF/libspdm/blob/66c74309c183f0d35c791f6cc34ca5c1e0099218/library/spdm_common_lib/libspdm_com_opaque_data.c#L57-L62

longlongyang:
fix #852 by checking version count.

Signed-off-by: Longlong Yang <longlong.yang@intel.com>

### DIFF
--- a/library/spdm_common_lib/libspdm_com_opaque_data.c
+++ b/library/spdm_common_lib/libspdm_com_opaque_data.c
@@ -38,11 +38,11 @@ size_t libspdm_get_opaque_data_version_selection_data_size(
 }
 
 /**
- * Return the size in bytes of opaque data supproted version.
+ * Return the size in bytes of opaque data supported version.
  *
  * This function should be called in KEY_EXCHANGE/PSK_EXCHANGE request generation.
  *
- * @return the size in bytes of opaque data supproted version.
+ * @return the size in bytes of opaque data supported version.
  **/
 size_t libspdm_get_opaque_data_supported_version_data_size(
     libspdm_context_t *spdm_context)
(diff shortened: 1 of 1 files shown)


include/internal/libspdm_common_lib.h
```diff
@@ -962,10 +962,12 @@ bool libspdm_verify_psk_finish_req_hmac(libspdm_context_t *spdm_context,
  *
  * This function should be called in KEY_EXCHANGE/PSK_EXCHANGE request generation.
  *
+ * @param  version_count                 Secure version count.
+ * 
  * @return the size in bytes of opaque data supproted version.
  **/
 size_t libspdm_get_opaque_data_supported_version_data_size(
-    libspdm_context_t *spdm_context);
+    libspdm_context_t *spdm_context, uint8_t version_count);
```
jyao1, line 970: Why you need version_count as input?
Why not just use spdm_context->local_context.secured_message_version.spdm_version_count ?

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -231,13 +225,31 @@ libspdm_process_opaque_data_supported_version_data(libspdm_context_t *spdm_conte
...
         return LIBSPDM_STATUS_INVALID_MSG_FIELD;
     }
+
+    /* update the version_count from the received buffer */
+    version_count = opaque_element_support_version->version_count;
+
+    if (data_in_size !=
+        libspdm_get_opaque_data_supported_version_data_size(spdm_context, version_count)) {
+        return LIBSPDM_STATUS_INVALID_MSG_SIZE;
+    }
```
longlongyang, line 235: @jyao1 
Because, I think this function will also be called to calculate the supported version data size from the received buffer(The requester). And spdm_context->local_context.secured_message_version.spdm_version_count is the version_count from the responder side, It can't cover both cases, Am I right?

jyao1: It does cover both cases.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -231,13 +258,29 @@ libspdm_process_opaque_data_supported_version_data(libspdm_context_t *spdm_conte
...
+    if ((opaque_element_table_header->id != SPDM_REGISTRY_ID_DMTF) ||
+        (opaque_element_table_header->vendor_len != 0) ||
+        (opaque_element_table_header->opaque_element_data_len !=
+         sizeof(secured_message_opaque_element_supported_version_t) +
+         sizeof(spdm_version_number_t) * version_count)) {
+        return LIBSPDM_STATUS_INVALID_MSG_FIELD;
+    }
+
+    if (data_in_size !=
+        libspdm_get_untrusted_opaque_data_supported_version_data_size(spdm_context, version_count)) {
```
jyao1, line 273: should it be "<" ?

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -185,25 +215,28 @@ libspdm_process_opaque_data_supported_version_data(libspdm_context_t *spdm_conte
...
         return LIBSPDM_STATUS_SUCCESS;
     }
 
-    if (data_in_size !=
-        libspdm_get_opaque_data_supported_version_data_size(spdm_context)) {
-        return LIBSPDM_STATUS_INVALID_MSG_SIZE;
-    }
     if (libspdm_get_connection_version (spdm_context) >= SPDM_MESSAGE_VERSION_12) {
+        if (data_in_size < sizeof(spdm_general_opaque_data_table_header_t) + sizeof(secured_message_opaque_element_table_header_t) + sizeof(secured_message_opaque_element_supported_version_t)) {
+            return LIBSPDM_STATUS_INVALID_MSG_FIELD;
```
jyao1, line 228: I think you may use libspdm_get_untrusted_opaque_data_supported_version_data_size(spdm_context, 1) here, because 1 is minimal.

library/spdm_common_lib/libspdm_com_opaque_data.c
longlongyang: libspdm_get_untrusted_opaque_data_supported_version_data_size works to return 
1. general_opaque_data_table_header_t +
2. secured_message_opaque_element_table_header_t +
3. secured_message_opaque_element_supported_version_t +
4. spdm_version_number_t * version count =
all the opaque data field, right? 
If right, then data_in_size should equals to the return of libspdm_get_untrusted_opaque_data_supported_version_data_size

library/spdm_common_lib/libspdm_com_opaque_data.c
longlongyang: Yeah, Make sense, good point!

library/spdm_common_lib/libspdm_com_opaque_data.c
jyao1: Fine, let's keep "!=" in this patch, because I found more places need change.

Later, we should consider the situation that opaque data include more entries.

jyao1 (approved):

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -58,17 +58,49 @@ size_t libspdm_get_opaque_data_supported_version_data_size(
...
-               sizeof(spdm_version_number_t);
+               sizeof(spdm_version_number_t) *
+               spdm_context->local_context.secured_message_version.spdm_version_count;
     }
     /* Add Padding*/
     return (size + 3) & ~3;
 }
 
+/**
+ * Return the size in bytes of opaque data supproted version.
```
steven-bellock, line 75: "supproted" -> "supported"

longlongyang: There are 4 "supproted" typo in total in this file. And they are all fixed in the latest commit, please check. Thank you.

PR #859, opened 2022-04-21
Enable const qualifier discard warning

### LINKED ISSUE #572: Need enable compiler warning for discards qualifiers
Related fix:
#566 
#565 
#563 
#561 

We need enable such warning treat error by default, and catch them in CI in the future.

ekohandel:
Resolves #572

This change enables generation of warnings for const qualifier discards.
The enforcement is excluded from the 3rd party libraries as well as the
openssl and mbedtl wrapper libraries cryptlib_openssl cryptlib_mbedtls.

The new device IO improvements caused further propagation of removal of
const than expected as message decoding and encoding are happening via
pointer manipulation when possible instead of memory copies.


steven-bellock (approved):

jyao1:
@ekohandel , May I know how do you resolve https://github.com/DMTF/libspdm/issues/572 ?

I do not see the change to enable such warning treat error by default.

ekohandel:
> @ekohandel , May I know how do you resolve #572 ?
> 
> I do not see the change to enable such warning treat error by default.

@jyao1, it is a bit hard to see due to the long lines in the root `CMakeLists.txt` file but `-Wcast-qual` has been added to the compile options and all warnings are treated as errors resulting in compile failure.

jyao1:
> @jyao1, it is a bit hard to see due to the long lines in the root `CMakeLists.txt` file but `-Wcast-qual` has been added to the compile options and all warnings are treated as errors resulting in compile failure.

I found it. It does not show until I click to expand it.
Look good to me.

PR #869, opened 2022-04-28
Buffer use after free

LeeTroy:
Fixing #868

The session_id is a pointer point to receiver buffer, and it will be
dereferenced during construst response message. However, the buffer
is released after process the request message. So it may cause
use-after-free error.

Signed-off-by: Troy Lee <troy_lee@aspeedtech.com>


jyao1:
Thank you very much. The root cause is right.

The fix need some improvement, because we allow send_buffer and receive_buffer are same.

As such, we must release_sender_buffer **before** acquire_receive_buffer.

jyao1:
Original flow:
```
libspdm_release_receiver_buffer()
// use session_id
libspdm_acquire_sender_buffer()
```
The fix should be:
```
// use session_id
libspdm_release_receiver_buffer()
libspdm_acquire_sender_buffer()
```
Would you please update in above way?

Issue #558, opened 2022-01-28
Is it mandatory that the requester must support DMTF measurement?


rw8896:
In spdm_get_response_algorithms(), the responder checks that the requester must also support DMTF Measurement if the responder supports MEAS_CAP.

https://github.com/DMTF/libspdm/blob/529224184c4f5ee7fb33d191d0985690944418f8/library/spdm_responder_lib/libspdm_rsp_algorithms.c#L494-L502

But in spec, it seems not mandatory that the requester must support DMTF measurement.
In Table 15 — NEGOTIATE_ALGORITHMS request message format, it didn't require the requester to set at least one bit in MeasurementSpecification. It only says the requester can set more than one bit.
Bit mask. The MeasurementSpecification field in Table 44 — Measurement block
format defines the values in this field. The Requester can set more than one bit to
indicate multiple measurement specification support.

However, the spec also asks that the responder "shall" select one of the measurement specifications supported by the Requester. 

It becomes ambiguous if there is no measurement specification the responder could select, especially if the requester didn't even support or need measurement.

Spec-wise, could people here help clarify my understanding above?
Implementation-wise, I was wondering if the responder should be that restricted.
It should be okay for the responder to allow MeasurementSpecificationSel to be zero in ALGORITHMS response and return errors if later the requester sends commands requiring the measurements.
I seemed to see this issue in openspdm issues but I couldn't find it now.
Appreciate if someone could help clarify this.

steven-bellock:
> I seemed to see this issue in openspdm issues but I couldn't find it now.

It's in https://github.com/DMTF/libspdm/issues/329.

richard-chien:
May I know the status of this issue? In libspdm_rsp_algorithms.c, the responder assumes the requester always supports one and only one measurement specification, which is DMTF measurement. This would cause problems (see line 517) when the requester doesn't support any measurement specification or supports multiple measurement specifications in the future. Should this be taken into account?

jyao1:
I agree it is a problem.
@richard-chien, do you have a real use case that we need resolve immediately?

richard-chien:
> I agree it is a problem. @richard-chien, do you have a real use case that we need resolve immediately?
HPE iLO currently doesn't support measurement capabilities so it set bit 0 of MeasurementSpecification to 0. However, the device (responder) implements the same logic as mentioned above and therefore hits the error at line 517. This has resulted in interoperability issues.

steven-bellock:
[...]

PR #876, opened 2022-05-11
handle case that requester does not support measurement.

jyao1:
Fix: https://github.com/DMTF/libspdm/issues/558

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_challenge_auth.c
+++ b/library/spdm_responder_lib/libspdm_rsp_challenge_auth.c
@@ -79,11 +79,15 @@ libspdm_return_t libspdm_get_response_challenge_auth(void *context,
                                                SPDM_ERROR_CODE_INVALID_REQUEST, 0,
                                                response_size, response);
     }
-    if (!libspdm_is_capabilities_flag_supported(spdm_context, false, 0,
-                                                SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_MEAS_CAP) &&
-        spdm_request->header.param2 > 0) {
-        return libspdm_generate_error_response (spdm_context, SPDM_ERROR_CODE_INVALID_REQUEST,
-                                                SPDM_CHALLENGE, response_size, response);
+    if (spdm_request->header.param2 > 0) {
+        if (!libspdm_is_capabilities_flag_supported(
+                spdm_context, false, 0,
+                SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_MEAS_CAP) ||
+            (spdm_context->connection_info.algorithm.measurement_spec == 0) ||
+            (spdm_context->connection_info.algorithm.measurement_hash_algo == 0) ) {
+            return libspdm_generate_error_response (spdm_context, SPDM_ERROR_CODE_INVALID_REQUEST,
+                                                    SPDM_CHALLENGE, response_size, response);
+        }
     }
 
     slot_id = spdm_request->header.param1;
--- a/library/spdm_responder_lib/libspdm_rsp_key_exchange.c
+++ b/library/spdm_responder_lib/libspdm_rsp_key_exchange.c
@@ -82,13 +82,16 @@ libspdm_return_t libspdm_get_response_key_exchange(void *context,
                                                0, response_size, response);
     }
 
-    if (!libspdm_is_capabilities_flag_supported(
-            spdm_context, false,
-            0, SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_MEAS_CAP) &&
-        spdm_request->header.param1 > 0) {
-        return libspdm_generate_error_response(
-            spdm_context, SPDM_ERROR_CODE_INVALID_REQUEST,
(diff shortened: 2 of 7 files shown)


steven-bellock:
That returns `SPDM_ERROR_CODE_UNSPECIFIED` to the Requester instead of `SPDM_ERROR_CODE_INVALID_REQUEST`.

steven-bellock:
If `MeasurementSpecificationSel` is zero Responder also needs to return an error if Requester sends `GET_MEASUREMENTS`.

jyao1:
> If `MeasurementSpecificationSel` is zero Responder also needs to return an error if Requester sends `GET_MEASUREMENTS`; also `ALGORITHMS.MeasurementHashAlgo` needs to be zero.

Yes, that is already done in libspdm_measurement_collection. That is why I removed ASSERT.

jyao1:
Update:

1) Add check If MeasurementSpecificationSel can be zero then this brings in more checks on the Responder side. If MeasurementSpecificationSel is zero then

CHALLENGE.Param2 must be zero.
KEY_EXCHANGE.Param1 must be zero.
PSK_EXCHANGE.Param1 must be zero.

2) If MeasurementSpecificationSel is zero Responder also needs to return an error if Requester sends GET_MEASUREMENTS

3) ALGORITHMS.MeasurementHashAlgo needs to be zero.

library/spdm_responder_lib/libspdm_rsp_challenge_auth.c
```diff
@@ -79,11 +79,15 @@ libspdm_return_t libspdm_get_response_challenge_auth(void *context,
...
-                                                SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_MEAS_CAP) &&
-        spdm_request->header.param2 > 0) {
-        return libspdm_generate_error_response (spdm_context, SPDM_ERROR_CODE_INVALID_REQUEST,
-                                                SPDM_CHALLENGE, response_size, response);
+    if (spdm_request->header.param2 > 0) {
+        if (!libspdm_is_capabilities_flag_supported(
+                spdm_context, false, 0,
+                SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_MEAS_CAP) ||
+            (spdm_context->connection_info.algorithm.measurement_spec == 0) ||
+            (spdm_context->connection_info.algorithm.measurement_hash_algo == 0) ) {
```
steven-bellock, line 87: Aren't these equivalent? If Requester sets `measurement_spec` to zero then `measurement_hash_algo` will be zero, right?

jyao1: It should be at this time.
But I do not know what it will be in the future or in some corner case. This is more a defensive way.

library/spdm_responder_lib/libspdm_rsp_key_exchange.c
```diff
@@ -82,13 +82,16 @@ libspdm_return_t libspdm_get_response_key_exchange(void *context,
...
-        spdm_request->header.param1 > 0) {
-        return libspdm_generate_error_response(
-            spdm_context, SPDM_ERROR_CODE_INVALID_REQUEST,
-            SPDM_KEY_EXCHANGE, response_size, response);
+    if (spdm_request->header.param1 > 0) {
+        if (!libspdm_is_capabilities_flag_supported(
+                spdm_context, false,
+                0, SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_MEAS_CAP) ||
+            (spdm_context->connection_info.algorithm.measurement_spec == 0) ||
+            (spdm_context->connection_info.algorithm.measurement_hash_algo == 0) ) {
```
steven-bellock, line 90: `spdm_context->connection_info.algorithm.measurement_hash_algo` is set to zero if the Responder supports measurements but its `MeasurementHashAlgo` is raw bit stream only (0), correct? In that case it obviously shouldn't return an error.

library/spdm_responder_lib/libspdm_rsp_measurements.c
```diff
@@ -175,6 +175,12 @@ libspdm_return_t libspdm_get_response_measurements(void *context,
             spdm_context, SPDM_ERROR_CODE_UNSUPPORTED_REQUEST,
             SPDM_GET_MEASUREMENTS, response_size, response);
     }
+    if ((spdm_context->connection_info.algorithm.measurement_spec == 0) ||
+        (spdm_context->connection_info.algorithm.measurement_hash_algo == 0) ) {
```
steven-bellock, line 179: `spdm_context->connection_info.algorithm.measurement_hash_algo` is set to zero if the Responder supports measurements but its `MeasurementHashAlgo` is raw bit stream only (0), correct? In that case it obviously shouldn't return an error.

jyao1: measurement_hash_algo is set to BIT0 (1) if raw bit stream only.

It is BIT0, the value is 1, not 0.

steven-bellock: Ah yeah that's right. I keep forgetting that's a bit mask.

Issue #780, opened 2022-03-26
libspdm_test_requester_key_update_case32() is incorrect


jyao1:
In test description:
```
/**
 * Test 32: the requester is setup correctly (see Test 27), but receives an
 * ERROR message indicating the ResponseNotReady status of the responder, when
 * updating all keys.
 * Expected behavior: client returns a Status of RETURN_DEVICE_ERROR, and
 * no keys should be updated.
 **/
```

But the code checks `assert_int_equal(status, RETURN_UNSUPPORTED);`, which is incorrect.

After trace code, we found the flow is:
libspdm_key_update()
--> libspdm_send_spdm_request: KEY_UPDATE
--> libspdm_receive_spdm_response: NOT_READY
------> libspdm_handle_error_response_main ()
--------> libspdm_requester_respond_if_ready ()
----------> libspdm_send_spdm_request: SPDM_RESPOND_IF_READY
----------> libspdm_receive_spdm_response: 
---------------> libspdm_transport_test_decode_message ()
-------------------> libspdm_decode_secured_message(): Fail at 
```
        if (libspdm_const_compare_mem(record_header1 + 1, &sequence_num_in_header,
                                      sequence_num_in_header_size) != 0) {
```
This is not expected.
Need investigate why the libspdm_decode_secured_message() fails.

PR #879, opened 2022-05-15
increase sequence number for next NOT Ready message

qizhangz:
Fix: #780 


unit_test/test_spdm_requester/key_update.c
```diff
@@ -3253,6 +3253,8 @@ libspdm_return_t libspdm_requester_key_update_test_receive_message(
                          secured_message_context->aead_iv_size);
         secured_message_context->application_secret
         .response_data_sequence_number = curr_rsp_sequence_number;
+        secured_message_context->application_secret_backup
+        .response_data_sequence_number = 0;
```
jyao1, line 3257: I am not sure why we need reset to 0 in decode.

We need figure out why encode sequence is 0, in second message.

Issue #759, opened 2022-03-22
Questions about M1/M2 calculation.


rw8896:
Hi @jyao1 @steven-bellock 

I tried to look for answers from the code but I probably missed something.
Would you help clarify the implementation of these two scenarios from spec 10.9.2.1?

1) GET_VERSION ,GET_CAPABILITIES ,NEGOTIATE_ALGORITHMS Skipped
In 1.1, A2=Null
In 1.2, A2=VCA
How do the requester and responder handle the difference according to the negotiated version?

2) CHALLENGE completion => Completion of CHALLENGE sets M1/M2 to null
On the responder side, it seems to assume M1/M2 will be reset by the next valid message.
Curious why we didn't choose to explicitly set M1/M2 to null in libspdm_get_response_challenge_auth()?

steven-bellock:
1. libspdm gets around that by not officially supporting `CACHE_CAP` (yet). Note that the 1.0/1.1 behavior, where `A2=null`, is a security vulnerability, as it allows for downgrade attacks.
2. Resetting M1/M2 can go either way, either resetting at end of response or beginning of request. @jyao1 do you think this should be an issue?

jyao1:
For 1), it is known limitation. see https://github.com/DMTF/libspdm/issues/784.

For 2), I think not an issue. But we had better clean the resource immediately. I would treat it as enhancement.

steven-bellock:
#2 will be handled in this issue. It's an enhancement.

PR #881, opened 2022-05-16
set M1/M2 to null in libspdm_get_response_challenge_auth().

qizhangz:
Fix: #759 


library/spdm_responder_lib/libspdm_rsp_challenge_auth.c
```diff
@@ -248,6 +248,8 @@ libspdm_return_t libspdm_get_response_challenge_auth(void *context,
                                      LIBSPDM_CONNECTION_STATE_AUTHENTICATED);
     }
 
+    libspdm_reset_message_c(spdm_context);
```
jyao1, line 251: I think we need reset message_b as well.

jyao1, line 251: I think we need reset message_b as well.

steven-bellock (approved):

PR #882, opened 2022-05-17
First drop of chunk_get proof-of-concept.

richkong88:
Just a proof-of-concept on chunk_get implementation to get ideas. Please do NOT commit.

### DIFF
--- a/include/library/spdm_lib_config.h
+++ b/include/library/spdm_lib_config.h
@@ -171,6 +171,11 @@
 #define LIBSPDM_ENABLE_SET_CERTIFICATE_CAP 1
 #endif
 
+#ifndef LIBSPDM_ENABLE_CHUNK_CAP
+#define LIBSPDM_ENABLE_CHUNK_CAP 1
+#endif
+
+
 /*
  * MinDataTransferSize = 42
  *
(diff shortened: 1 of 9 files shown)


include/library/spdm_lib_config.h
```diff
@@ -270,12 +275,17 @@
...
+ * If chunking is supported, it should be at least LIBSPDM_SENDER_RECEIVE_BUFFER_SIZE
+ * LIBSPDM_MAX_SPDM_MSG_SIZE.
  *
  * The value is NOT configurable.
- * The value MAY be chnaged in different libspdm version.
+ * The value MAY be changed in different libspdm version.
  * It is exposed here, just in case the libspdm consumer wants to configure the setting at build time.
  */
-#define LIBSPDM_SCRATCH_BUFFER_SIZE (LIBSPDM_SENDER_RECEIVE_BUFFER_SIZE)
+#define LIBSPDM_SCRATCH_BUFFER_SIZE (LIBSPDM_SENDER_RECEIVE_BUFFER_SIZE + LIBSPDM_MAX_SPDM_MSG_SIZE)
```
jyao1, line 289: do we need include LIBSPDM_MAX_SPDM_MSG_SIZE if LIBSPDM_ENABLE_CHUNK_CAP is 0 ?

jyao1:
Overall, it looks good.

Question: How do we trigger the flow in unit test?

include/library/spdm_lib_config.h
richkong88: You are right. LIBSPDM_SCRATCH_BUFFER should depend on if LIBSPDM_ENABLE_CHUNK_CAP is enabled or not.

richkong88:
> Overall, it looks good.
> 
> Question: How do we trigger the flow in unit test?

Thanks. Good question. I'm still trying to figure this part out. I was just testing by modifying the spdm_lib_config.h parameters (eg. LIBSPDM_DATA_TRANSFER_SIZE) so it would force larger responses to go through the CHUNK_GET flow.  Is it possible to dynamically override some of these defines when compiling for ULT?

jyao1:
> Is it possible to dynamically override some of these defines when compiling for ULT?

Sure, we can override.
In my definition, dynamic == runtime override, static == compiling time override.
Would you please clarify what you mean when you have both "dynamically" and "compiling" ?

richkong88:
> > Is it possible to dynamically override some of these defines when compiling for ULT?
> 
> Sure, we can override. In my definition, dynamic == runtime override, static == compiling time override. Would you please clarify what you mean when you have both "dynamically" and "compiling" ?

I meant force the compiler to change the values for the LIBSPDM_DATA_TRANSFER_SIZE define without modifying the code. I think we can do this with the -DLIBSPDM_DATA_TRANSFER_SIZE=X, since the define is protected with #ifndef #endif in the spdm_lib_config file. I think we'd want to run the ULTs with their default values, then run them again with the config settings to force exercising the chunking flows.

jyao1:
> I think we can do this with the -DLIBSPDM_DATA_TRANSFER_SIZE=X

yes we can

richkong88:
Looking more at unit tests now. This is more complicated than just changing a define. We'll have to write a bunch of unit_test cases. I'll work on this.

richkong88:
Closing since this was just POC to get feedback.

PR #885, opened 2022-05-17
Expose APIs for validator.

jyao1:


steven-bellock:
@jyao1 Why can't the validator #include the `internal/*` headers? If we don't intend for integrators to use these functions then they shouldn't be exposed to the integrator.

jyao1:
> @jyao1 Why can't the validator #include the `internal/*` headers? If we don't intend for integrators to use these functions then they shouldn't be exposed to the integrator.

I treat a validator be an integrator.
I don't think a validator can use internal header. It is different from the unit test.

Issue #886, opened 2022-05-17
libspdm_rsp_psk_exchange not processing psk_hint field


ecominetti:
I am creating some additional tests for psk_exchange messages and notice that the responder does not process the psk_hint field (which is optional) in libspdm_rsp_psk_exchange.c

Example:

In the unit_test/test_spdm_responder/psk_exchange.c, the psk_hint field is created on lines 94-96 and it is stored on the appropriate fields of spdm_context on lines 97-99.
Then, the request message also receives the value (lines 101-102 and 110-113).
However, the psk_hint field from the request message is not processed in the libspdm_rsp_psk_exchange.c, where it uses the value already set in the local_context (the value is used inside the libspdm_generate_session_handshake_key, line 299).
Because the value in the unit_test message is copied from the one stored on the spdm_context, no error is detected.

Suggested fix:

Before libspdm_generate_session_handshake_key is called in libspdm_rsp_psk_exchange.c, check the psk_hint size. If 0, set psk_hint to null. Else, set spdm_context->local_context.psk_hint and  spdm_context->local_context.psk_hint_size to the ones received in the request message.

steven-bellock:
Shouldn't it be in `session_info->secured_message_context`? That's what gets passed to `libspdm_generate_session_handshake_key`.

qizhangz:
> Shouldn't it be in `session_info->secured_message_context`? That's what gets passed to `libspdm_generate_session_handshake_key`.

it will be passed from spdm_context->local_context to session_info->secured_message_context in the below flow.
libspdm_assign_session_id() -> libspdm_session_info_init() -> libspdm_secured_message_set_psk_hint()

steven-bellock:
Ah thanks. But the use of the Responder's `local_context` to store the Requester's fields is strange, since the Responder's `local_context` is for long-standing parameters that won't change and are independent of the Requester. The way it's being used in the PR is a temporary storage location before being stored into `session_info`. It should probably in `connection_info` or passed as parameters to `libspdm_assign_session_id`.

steven-bellock:
Ah thanks. But the use of the Responder's `local_context` to store the Requester's fields is strange, since the Responder's `local_context` is for long-standing parameters that won't change and are independent of the Requester. The way it's being used in the PR is a temporary storage location before being stored into `session_info`. It should probably in `connection_info` or passed as parameters to `libspdm_assign_session_id`.

PR #893, opened 2022-05-20
process psk_hint field in libspdm_rsp_psk_exchange().

qizhangz:
Fix: #886 

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_psk_exchange.c
+++ b/library/spdm_responder_lib/libspdm_rsp_psk_exchange.c
@@ -198,6 +198,15 @@ libspdm_return_t libspdm_get_response_psk_exchange(void *context,
     spdm_response->header.request_response_code = SPDM_PSK_EXCHANGE_RSP;
     spdm_response->header.param1 = spdm_context->local_context.heartbeat_period;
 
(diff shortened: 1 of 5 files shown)


library/spdm_responder_lib/libspdm_rsp_psk_exchange.c
```diff
@@ -198,6 +198,15 @@ libspdm_return_t libspdm_get_response_psk_exchange(void *context,
...
     spdm_response->header.param1 = spdm_context->local_context.heartbeat_period;
 
+    if (spdm_request->psk_hint_length == 0) {
+        spdm_context->local_context.psk_hint_size = 0;
+        spdm_context->local_context.psk_hint = NULL;
+    } else {
+        spdm_context->local_context.psk_hint_size = spdm_request->psk_hint_length;
+        spdm_context->local_context.psk_hint = (const uint8_t *)request +
+                                               sizeof(spdm_psk_exchange_request_t);
+    }
```
jyao1, line 208: local_context is designed to be provisioned by integrator.

Here we need a way to store info, I think we need move it to connection_info.

steven-bellock (approved):

Issue #181, opened 2021-10-06
Handling of Challenges and Measurements 


steven-bellock:
The SPDM 1.2 specification says

> Because issuing GET_MEASUREMENTS clears the M1/M2 message transcript, it is recommended that a Requester does not send this message until it has received at least one successful CHALLENGE_AUTH response message from the Responder. This ensures that the information in message pairs GET_DIGESTS/DIGESTS and GET_CERTIFICATES/CERTIFICATES has been authenticated at least once.

When executing GET_MEASUREMENTS and MEASUREMENTS outside of a secure session libspdm checks that the connection state is at least SPDM_CONNECTION_STATE_AUTHENTICATED for both the Requester and the Responder, which requires a successful CHALLENGE_AUTH response. However this assumes that CHAL_CAP == 1, which is not mandated by the SPDM specification. A responder with no cryptography capabilities should still be able to successfully return an unsigned measurement.

Going further, even if the Requester and Responder both support CHALLENGE it should not be up to the Responder to enforce this recommendation, as the recommendation is directed towards the Requester.

steven-bellock:
This was fixed by https://github.com/DMTF/libspdm/pull/898

PR #898, opened 2022-05-22
Change measurements state check.

jyao1:
Authentication is recommended by not required.


steven-bellock:
This fixes https://github.com/DMTF/libspdm/issues/181

steven-bellock (approved):

Issue #835, opened 2022-04-01
the requester need free resource after the transcript is calculated.


jyao1:
See https://github.com/DMTF/libspdm/pull/834.

It freed the resource in the caller, which is work-around. We should fix it internally if it is success.

PR #907, opened 2022-05-25
the requester need free resource after the transcript is calculated.

qizhangz:
Fix: #835 


library/spdm_common_lib/libspdm_com_crypto_service.c
```diff
@@ -751,6 +751,11 @@ bool libspdm_generate_challenge_auth_signature(libspdm_context_t *spdm_context,
     result = libspdm_calculate_m1m2_hash(spdm_context, is_requester,
                                          &m1m2_hash_size, &m1m2_hash);
 #endif
+    if (is_requester) {
+        libspdm_reset_message_mut_c(spdm_context);
+    } else {
+        libspdm_reset_message_c(spdm_context);
+    }
```
jyao1, line 758: should we also reset message_b ?

jyao1 (approved):

steven-bellock (approved):

PR #910, opened 2022-05-26
Optimize function libspdm_calculate_l1l2

lsun100:
This commit passes a libspdm_large_managed_buffer_t pointer directly
to libspdm_calculate_l1l2 to save a l1l2 message copy. It also
saves 'LIBSPDM_MAX_MESSAGE_BUFFER_SIZE' bytes of stack usage.

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -552,68 +552,6 @@ libspdm_is_capabilities_flag_supported(const libspdm_context_t *spdm_context,
                                        uint32_t requester_capabilities_flag,
                                        uint32_t responder_capabilities_flag);
 
-#if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT
-/*
- * This function calculates m1m2.
- *
- * @param  spdm_context                  A pointer to the SPDM context.
- * @param  is_mut                        Indicate if this is from mutual authentication.
- * @param  m1m2_buffer_size               size in bytes of the m1m2
- * @param  m1m2_buffer                   The buffer to store the m1m2
- *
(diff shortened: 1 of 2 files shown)


include/internal/libspdm_common_lib.h
```diff
@@ -596,7 +596,7 @@ bool libspdm_calculate_m1m2_hash(void *context, bool is_mut,
  * @retval RETURN_SUCCESS  l1l2 is calculated.
  */
 bool libspdm_calculate_l1l2(void *context, void *session_info,
-                            size_t *l1l2_buffer_size, void *l1l2_buffer);
+                            libspdm_large_managed_buffer_t *l1l2);
```
steven-bellock, line 599: It looks like the scope of this function is entirely in `library/spdm_common_lib/libspdm_com_crypto_service.c` and the header declaration can be removed and the function can be made `static`. @jyao1 Do you think other files will need to access this?

steven-bellock:
@lsun100 If @jyao1 is ok with this pattern I'd like to see `libspdm_calculate_m1m2` updated as well, so that the two functions are aligned.

jyao1:
This is internal function. I think it is OK to change.

jyao1 (approved):

jyao1:
yes, I think libspdm_calculate_m1m2() should be updated too.

jyao1:
@lsun100, do you wan to add optimization for m1m2 in this patch or other patch?

lsun100:
> libspdm_calculate_m1m2
I'll update this patch with the above comments. Thanks!

lsun100:
@jyao1/@steven-bellock, I made an update according to the comments. thanks!

PR #911, opened 2022-05-26
CHUNK_GET implementation.

richkong88:
Official version of CHUNK_GET for review.

### DIFF
--- a/include/internal/libspdm_requester_lib.h
+++ b/include/internal/libspdm_requester_lib.h
@@ -424,7 +424,6 @@ libspdm_return_t libspdm_send_spdm_request(libspdm_context_t *spdm_context,
  * @param  session_id                    Indicate if the response is a secured message.
  *                                       If session_id is NULL, it is a normal message.
  *                                       If session_id is NOT NULL, it is a secured message.
- * @param  support_large_response        Perform large response handling if necessary.
  * @param  response_size                 size in bytes of the response data buffer.
  * @param  response                      A pointer to a destination buffer to store the response.
  *                                       The caller is responsible for having
@@ -437,7 +436,6 @@ libspdm_return_t libspdm_send_spdm_request(libspdm_context_t *spdm_context,
  **/
 libspdm_return_t libspdm_receive_spdm_response(libspdm_context_t *spdm_context,
                                                const uint32_t *session_id,
-                                               bool support_large_response,
                                                size_t *response_size,
                                                void **response);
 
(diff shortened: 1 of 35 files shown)


include/internal/libspdm_requester_lib.h
```diff
@@ -399,20 +422,22 @@ libspdm_return_t libspdm_send_spdm_request(libspdm_context_t *spdm_context,
...
+ *                                       either implicit or explicit ownership of the buffer.
+ *                                       For normal message, response pointer still point to original transport_message.
+ *                                       For secured message, response pointer will point to the scratch buffer in spdm_context.
  *
  * @retval RETURN_SUCCESS               The SPDM response is received successfully.
  * @retval RETURN_DEVICE_ERROR          A device error occurs when the SPDM response is received from the device.
  **/
 libspdm_return_t libspdm_receive_spdm_response(libspdm_context_t *spdm_context,
                                                const uint32_t *session_id,
+                                               bool support_large_response,
```
jyao1, line 440: One question for this: Do you see the absolute need to add this parameter?
What if we always support `support_large_response` ?

richkong88: Well, the only disadvantage I see with removing it is you could be unintentionally supporting CHUNK_GET on responses that theoretically should not need it. So a bug could be hidden.  An alternative though is remove the parameter and in the large response handler code after the large response is retrieved, assert that the response code is one of those responses we expect as valid for large response. That is cleaner.

jyao1: you can cache the request opcode in spdm_context, by using request opcode, you can already distinguish that without introducing new field.

richkong88: Good point. The request code is actually already there in the last_spdm_request member. The CHUNK_GET does not overwrite that member.

richkong88: I will push a change that removes the parameter, and will always try CHUNK_GET on an large response. Then it will assert if any of the responses are unexpected ones that shouldn't require a large response. If it passes CI.  I'll then create a new branch and pull request so as to not have any change history to the libspdm requester files that called this `libspdm_receive_spdm_response `function with the old `support_large_response` parameter.

jyao1:
Thanks. Would you please revert the change for each libspdm_req_<command> ?
It seems they are just coding style difference.

richkong88:
> Thanks. Would you please revert the change for each libspdm_req_<command> ? It seems they are just coding style difference.
> 
> You can squash and force push a new one.

Yeah, I'll create a new branch that has no changes or added history to the libspdm_req_<command> files.

Issue #774, opened 2022-03-25
Ensure that hashes / signatures over transcripts are consistent


steven-bellock:
`LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT` is a compile-time configuration setting that, when enabled, stores the entire transcript in memory before it is hashed / signed. To reduce memory requirements `LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT` can be disabled, in which case a running calculation is made over the transcript. We need to ensure that the two methods produce the same results.

jyao1:
agree. I think we can enable CI to test:
1) requester (DATA_SUPPORT = 0) <=> responder (DATA_SUPPORT = 1)
2) requester (DATA_SUPPORT = 1) <=> responder (DATA_SUPPORT = 2)

Wenxing-hou:
The spdm-emu patch [Add CI check for enable LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT](https://github.com/DMTF/spdm-emu/pull/117) has been merged and checked the consistent successfully.

I think this issue can be closed.

PR #914, opened 2022-05-27
Avoid adding VCA multiple times into digest_context_l1l2

lsun100:
This commit moves the VCA logic inside the block of
'digest_context_l1l2 == NULL' in libspdm_append_message_m(). It
could avoid VCA being added into the digest_context_l1l2 each
time the libspdm_append_message_m() function is called, which
causes incompatability between requestor and respondor that are
compiled with different LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT.

### DIFF
--- a/library/spdm_common_lib/libspdm_com_context_data.c
+++ b/library/spdm_common_lib/libspdm_com_context_data.c
@@ -1387,24 +1387,24 @@ libspdm_return_t libspdm_append_message_m(void *context, void *session_info,
                     spdm_context->transcript.digest_context_l1l2 = NULL;
                     return LIBSPDM_STATUS_CRYPTO_ERROR;
                 }
-            }
-            if ((spdm_context->connection_info.version >> SPDM_VERSION_NUMBER_SHIFT_BIT) >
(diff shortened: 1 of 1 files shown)


lsun100:
This pull request was intended to solve a measurement checking failure found on my setup when requestor is compiled with LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT=0 while responder is compiled with LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT=1. Hope the proposed changes make sense. thanks.

jyao1 (approved):

lsun100:
> Related to issue #774
> 
> @lsun100 Did you also test Requester (LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT = 1) and Responder (LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT = 0) ?

Not really. I was using the ported Responder which can only do LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT=1 for now. But I'll give it a try.

jyao1:
OK. I will merge after get your confirmation.

PR #923, opened 2022-06-03
Store csr for libspdm_req_get_csr more clearly

Wenxing-hou:

### DIFF
--- a/include/library/spdm_requester_lib.h
+++ b/include/library/spdm_requester_lib.h
@@ -715,13 +715,17 @@ libspdm_return_t libspdm_generate_encap_extended_error_response(
  * This function sends GET_CSR
  * to get csr from the device.
  *
- * @param  context                      A pointer to the SPDM context.
- * @param  cert_chain                   The pointer for the certificate chain to set.
- *                                      The cert chain is a full SPDM certificate chain, including Length and Root Cert Hash.
- * @param  cert_chain_size              The size of the certificate chain to set.
- * @param  session_id                   Indicates if it is a secured message protected via SPDM session.
- *                                      If session_id is NULL, it is a normal message.
- *                                      If session_id is NOT NULL, it is a secured message.
+ * @param[in]  context                      A pointer to the SPDM context.
+ * @param[in]  requester_info               requester info to gen CSR
+ * @param[in]  requester_info_length        The len of requester info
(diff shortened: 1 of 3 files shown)


include/library/spdm_requester_lib.h
```diff
@@ -731,7 +734,7 @@ libspdm_return_t libspdm_get_csr(void * context,
                                  void * requester_info, uint16_t requester_info_length,
                                  void * opaque_data, uint16_t opaque_data_length,
                                  const uint32_t *session_id,
-                                 void * csr, size_t csr_max_len);
+                                 void **csr, size_t *csr_len);
```
jyao1, line 737: I think the interface should be `void *csr, size_t *csr_len`

on input, *csr_len indicates the max csr buffer size.
on output, *csr_len indicates the actual csr buffer size.

We still need the caller to allocate the buffer, instead of the callee.

jyao1, line 737: I think the interface should be `void *csr, size_t *csr_len`

on input, *csr_len indicates the max csr buffer size.
on output, *csr_len indicates the actual csr buffer size.

We still need the caller to allocate the buffer, instead of the callee.

library/spdm_requester_lib/libspdm_req_get_csr.c
```diff
@@ -145,7 +146,8 @@ libspdm_return_t libspdm_try_get_csr(void *context,
         goto receive_done;
     }
 
-    libspdm_copy_mem(csr, csr_max_len, spdm_response + 1, spdm_response->csr_length);
+    libspdm_copy_mem(csr, *csr_len, spdm_response + 1, spdm_response->csr_length);
```
jyao1, line 149: I think you need check 
```
if (*csr_len < spdm_response->csr_length) {
    *csr_len = spdm_response->csr_length;
    return BUFFER_TOO_SMALL;
}
```

Wenxing-hou: Yes. I have added the check.

Wenxing-hou:
And the [emu patch](https://github.com/DMTF/spdm-emu/pull/83) has been changed too.

jyao1 (approved):

steven-bellock (approved):

PR #926, opened 2022-06-03
Fix and add definition for PCI idekm.

jyao1:


jyao1:
It is for https://github.com/DMTF/spdm-emu/pull/82

steven-bellock (approved):

PR #930, opened 2022-06-08
libspdm_rsp_psk_exchange not processing psk_hint field.

qizhangz:
Fix: #886 

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_psk_exchange.c
+++ b/library/spdm_requester_lib/libspdm_req_psk_exchange.c
@@ -275,6 +275,9 @@ libspdm_return_t libspdm_try_send_receive_psk_exchange(
         status = LIBSPDM_STATUS_SESSION_NUMBER_EXCEED;
         goto receive_done;
     }
+    libspdm_session_info_set_psk_hint(session_info,
+                                      spdm_context->local_context.psk_hint,
+                                      spdm_context->local_context.psk_hint_size);
 
     measurement_summary_hash_size = libspdm_get_measurement_summary_hash_size(
         spdm_context, true, measurement_hash_type);
--- a/os_stub/spdm_device_secret_lib_sample/lib.c
+++ b/os_stub/spdm_device_secret_lib_sample/lib.c
@@ -1071,11 +1071,10 @@ bool libspdm_psk_handshake_secret_hkdf_expand(
     bool result;
     uint8_t handshake_secret[64];
 
-    if ((psk_hint == NULL) && (psk_hint_size == 0)) {
+    if (psk_hint_size == 0) {
         psk = LIBSPDM_TEST_PSK_DATA_STRING;
         psk_size = sizeof(LIBSPDM_TEST_PSK_DATA_STRING);
-    } else if ((psk_hint != NULL) && (psk_hint_size != 0) &&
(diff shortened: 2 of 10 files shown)


library/spdm_requester_lib/libspdm_req_psk_exchange.c
```diff
@@ -268,9 +269,13 @@ libspdm_return_t libspdm_try_send_receive_psk_exchange(
     if (heartbeat_period != NULL) {
         *heartbeat_period = spdm_response->header.param1;
     }
+    use_psk.use_psk = true;
+    use_psk.psk_hint = spdm_context->local_context.psk_hint;;
```
steven-bellock, line 273: There are two semicolons. `;;`

include/library/spdm_crypt_lib.h
```diff
@@ -25,6 +25,7 @@
 #define LIBSPDM_MAX_AEAD_KEY_SIZE 32
 #define LIBSPDM_MAX_AEAD_IV_SIZE 12
 #define LIBSPDM_MAX_AEAD_TAG_SIZE 16
+#define LIBSPDM_MAX_PSK_HINT_SIZE 128
```
steven-bellock, line 28: `LIBSPDM_MAX_PSK_HINT_SIZE` needs to be defined by the integrator.

steven-bellock (changes requested):

include/library/spdm_common_lib.h
```diff
@@ -918,7 +918,7 @@ libspdm_get_secured_message_context_via_session_info(void *spdm_session_info);
  * @return session info associated with this new session ID.
  **/
 void *libspdm_assign_session_id(void *spdm_context, uint32_t session_id,
-                                bool use_psk);
+                                void *use_psk);
```
jyao1, line 921: we cannot change common_lib.h

qizhangz: move the change to internal/libspdm_common_lib.h

include/library/spdm_crypt_lib.h
jyao1: We already have `LIBSPDM_PSK_MAX_HINT_LENGTH` in spdm_lib_config.h. Why we need `LIBSPDM_MAX_PSK_HINT_SIZE` here???

os_stub/spdm_device_secret_lib_sample/lib.c
```diff
@@ -1107,10 +1107,10 @@ bool libspdm_psk_master_secret_hkdf_expand(
     uint8_t salt1[64];
     uint8_t master_secret[64];
 
-    if ((psk_hint == NULL) && (psk_hint_size == 0)) {
+    if (psk_hint_size == 0) {
         psk = LIBSPDM_TEST_PSK_DATA_STRING;
         psk_size = sizeof(LIBSPDM_TEST_PSK_DATA_STRING);
-    } else if ((psk_hint != NULL) && (psk_hint_size != 0) &&
+    } else if ((psk_hint_size != 0) &&
```
steven-bellock, line 1113: At this point `psk_hint_size != 0`. No need to check it.

steven-bellock (changes requested):

jyao1 (approved):

steven-bellock (approved):

Issue #854, opened 2022-04-18
Need consider multiple elements in opaque data 


jyao1:
Current libspdm_process_opaque_data_supported_version_data() assumes there is 1 element for key exchange - SECURED_MESSAGE_OPAQUE_ELEMENT_SMDATA_ID_SUPPORTED_VERSION.

It is true if libspdm is the sender, but it may cause problem if libspdm is the receiver, because other implementation may send more entries.

I think libspdm should consider multiple elements in libspdm_process_opaque_data_supported_version_data().

PR #931, opened 2022-06-08
Add multiple element support in opaque data for when libspdm is receiver

Wenxing-hou:
Fix: #854 

Add multiple element support in opaque data for when libspdm is receiver

**There are some checks:**

- ensure total data size is right;
- element header id;
- version count consistent;
- check DMTF element, when the element id is PDM_REGISTRY_ID_DMTF;

**And add unit_test for libspdm receive multiple element case:**

- right multiple element case for 1.2;
- wrong multiple element case for 1.2;
- right multiple element case for 1.1;
- wrong multiple element case for 1.1;

And the fuzzing test is passed.


library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -81,23 +81,32 @@ size_t libspdm_get_opaque_data_supported_version_data_size(
  * @return the size in bytes of opaque data supported version.
  **/
 size_t libspdm_get_untrusted_opaque_data_supported_version_data_size(
-    libspdm_context_t *spdm_context, uint8_t version_count)
+    libspdm_context_t *spdm_context, uint8_t version_count, uint8_t element_num)
```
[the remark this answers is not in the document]
Wenxing-hou: Thanks. I have changed the code. I know your opinion: only the DMTF id element have version count.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -233,64 +239,98 @@ libspdm_process_opaque_data_supported_version_data(libspdm_context_t *spdm_conte
...
 
-    if ((opaque_element_table_header->id != SPDM_REGISTRY_ID_DMTF) ||
-        (opaque_element_table_header->vendor_len != 0) ||
-        (opaque_element_table_header->opaque_element_data_len !=
-         sizeof(secured_message_opaque_element_supported_version_t) +
-         sizeof(spdm_version_number_t) * version_count)) {
-        return LIBSPDM_STATUS_INVALID_MSG_FIELD;
-    }
+        /*check element header id*/
+        if ((opaque_element_table_header->id < SPDM_REGISTRY_ID_DMTF) ||
```
jyao1, line 268: SPDM_REGISTRY_ID_DMTF is 0.

How could `opaque_element_table_header->id < SPDM_REGISTRY_ID_DMTF` happen?

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -233,64 +239,98 @@ libspdm_process_opaque_data_supported_version_data(libspdm_context_t *spdm_conte
...
+            (opaque_element_table_header->id > SPDM_REGISTRY_ID_JEDEC)) {
+            return LIBSPDM_STATUS_INVALID_MSG_FIELD;
+        }
```
jyao1, line 271: I recommend to remove the whole ID check.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -220,6 +219,13 @@ libspdm_process_opaque_data_supported_version_data(libspdm_context_t *spdm_conte
     spdm_version_number_t common_version;
     uint8_t version_count;
     bool result;
+    uint8_t element_index;
+    uint8_t element_num;
+    size_t current_element_len;
+    size_t total_element_len;
+    size_t total_opaque_len;
+
+    total_element_len = 0;
```
jyao1, line 228: I think we also need update libspdm_process_opaque_data_version_selection_data().

I dont see that update for that function.

library/spdm_common_lib/libspdm_com_opaque_data.c
Wenxing-hou: I have deleted the code.

library/spdm_common_lib/libspdm_com_opaque_data.c
Wenxing-hou: I have updated the libspdm_process_opaque_data_version_selection_data.
image

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -401,64 +440,111 @@ libspdm_process_opaque_data_version_selection_data(libspdm_context_t *spdm_conte
...
-                                                    selected_version)
-            ==
-            libspdm_get_version_from_version_number(spdm_context->local_context.
-                                                    secured_message_version
-                                                    .spdm_version[secured_message_version_index]))
-        {
-            return LIBSPDM_STATUS_SUCCESS;
+    for (element_index = 0; element_index < element_num;
+         element_index++) {
+
```
jyao1, line 484: Since most checks are duplicated, would you please consider create a function to get the opaque_element via sm_data_id?

Wenxing-hou: I create function: `libspdm_process_opaque_data` , which can be called by `libspdm_process_opaque_data_version_selection_data` and `libspdm_process_opaque_data_supported_version_data`.
There is a parameter: `process_supported_version` to distinguish.

And I have moved the unit test on libspdm_common unit test.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -192,21 +191,24 @@ libspdm_build_opaque_data_supported_version_data(libspdm_context_t *spdm_context
...
  *
  * @retval RETURN_SUCCESS               The opaque data supported version is processed successfully.
  * @retval RETURN_UNSUPPORTED           The data_in is NOT opaque data supported version.
  **/
-libspdm_return_t
-libspdm_process_opaque_data_supported_version_data(libspdm_context_t *spdm_context,
-                                                   size_t data_in_size,
-                                                   const void *data_in)
+libspdm_return_t libspdm_process_opaque_data(libspdm_context_t *spdm_context, size_t data_in_size,
+                                             const void *data_in, bool process_supported_version)
```
jyao1, line 211: This is not a good idea. What if there are more opaque data need to handle? 
A BOOL cannot handle more than 2 state.

Why not take my feedback to create a common function to element based upon ID ?

library/spdm_common_lib/libspdm_com_opaque_data.c
jyao1: This is not a good idea. What if there are more opaque data need to handle?
A BOOL cannot handle more than 2 state.

Why not take my feedback to create a common function to return element based upon ID ?

Wenxing-hou: Because the many check is over before getting sm_data_id, I think the `libspdm_process_opaque_data` can be named `libspdm_process_opaque_data_negotiate_vesion`.

jyao1: I think we need split the ID specific check from common element check.

The common element check goes to common function.
The ID specific check still stays in original function.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -192,6 +191,82 @@ libspdm_build_opaque_data_supported_version_data(libspdm_context_t *spdm_context
...
+ * @param[in]  data_in                     A pointer to the multi element.
+ * @param[in]  element_num                 element number.
+ * @param[in]  element_id                  element id.
+ * @param[out] get_element_ptr             pointer to store finded element
+ *
+ * @retval true                            get element successfully
+ * @retval false                           get element failed
+ **/
+bool libspdm_get_element_from_opaque_data(size_t data_in_size, const void *data_in,
+                                          uint8_t element_num, uint8_t element_id,
```
jyao1, line 211: I dont think we need element_num here. Please remove it.

Element_num should be get from the data_in.

jyao1 (changes requested):
Please move the shared code (such as opaque_element_table_header calculation) to common function.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -192,6 +191,112 @@ libspdm_build_opaque_data_supported_version_data(libspdm_context_t *spdm_context
...
+ * @param[in]  data_in                     A pointer to the multi element opaque data.
+ * @param[in]  element_id                  element id.
+ * @param[out] get_element_ptr             pointer to store finded element
+ *
+ * @retval true                            get element successfully
+ * @retval false                           get element failed
+ **/
+bool libspdm_get_element_from_opaque_data(libspdm_context_t *spdm_context,
+                                          size_t data_in_size, const void *data_in,
+                                          uint8_t element_id, const void **get_element_ptr)
```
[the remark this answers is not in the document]
Wenxing-hou: Yes. I have moved common check to `libspdm_get_element_from_opaque_data` and added the sm_data_id to identify secured message element.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -192,6 +191,120 @@ libspdm_build_opaque_data_supported_version_data(libspdm_context_t *spdm_context
...
+    if (libspdm_get_connection_version (spdm_context) >= SPDM_MESSAGE_VERSION_12) {
+        spdm_general_opaque_data_table_header = data_in;
+        if (spdm_general_opaque_data_table_header->total_elements < 1) {
+            return LIBSPDM_STATUS_INVALID_MSG_FIELD;
+        }
+        opaque_element_table_header =
+            (const void *)(spdm_general_opaque_data_table_header + 1);
+
+        element_num = spdm_general_opaque_data_table_header->total_elements;
+        data_element_size = data_in_size - sizeof(spdm_general_opaque_data_table_header_t);
```
jyao1, line 251: Please do check for data_in_size is bigger enough before minus.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -192,6 +191,120 @@ libspdm_build_opaque_data_supported_version_data(libspdm_context_t *spdm_context
...
+             SECURED_MESSAGE_OPAQUE_VERSION) ||
+            (general_opaque_data_table_header->total_elements < 1)) {
+            return LIBSPDM_STATUS_INVALID_MSG_FIELD;
+        }
+        opaque_element_table_header =
+            (const void *)(general_opaque_data_table_header + 1);
+
+        element_num = general_opaque_data_table_header->total_elements;
+        data_element_size = data_in_size -
+                            sizeof(secured_message_general_opaque_data_table_header_t);
```
jyao1, line 266: Please do check for data_in_size is bigger enough before minus.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -192,6 +191,120 @@ libspdm_build_opaque_data_supported_version_data(libspdm_context_t *spdm_context
...
+    }
+
+    for (element_index = 0; element_index < element_num;
+         element_index++) {
+
+        /*check element header id*/
+        if (opaque_element_table_header->id > SPDM_REGISTRY_ID_JEDEC) {
```
jyao1, line 273: please check data_element_size before refer any data.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -192,6 +191,120 @@ libspdm_build_opaque_data_supported_version_data(libspdm_context_t *spdm_context
...
+            if (secured_message_element_header->sm_data_id == sm_data_id) {
+                /*get element by element id*/
+                *get_element_ptr = opaque_element_table_header;
+                result = true;
+            }
+        }
+
+        /*move to next element*/
+        current_element_len = sizeof(secured_message_opaque_element_table_header_t) +
+                              opaque_element_table_header->opaque_element_data_len;
```
jyao1, line 289: please check data_element_size with opaque_element_data_len

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -192,6 +191,120 @@ libspdm_build_opaque_data_supported_version_data(libspdm_context_t *spdm_context
...
+ * @param[in]  sm_data_id                  sm_data_id to identifiy for the Secured Message data type.
+ * @param[out] get_element_ptr             pointer to store finded element
+ *
+ * @retval true                            get element successfully
+ * @retval false                           get element failed
+ **/
+bool libspdm_get_element_from_opaque_data(libspdm_context_t *spdm_context,
+                                          size_t data_in_size, const void *data_in,
+                                          uint8_t element_id, uint8_t sm_data_id,
+                                          const void **get_element_ptr)
```
jyao1, line 213: As complete API, please return element_len

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -392,51 +492,42 @@ libspdm_process_opaque_data_version_selection_data(libspdm_context_t *spdm_conte
...
+    if ((opaque_element_table_header->vendor_len != 0) ||
         (opaque_element_table_header->opaque_element_data_len !=
          sizeof(secured_message_opaque_element_version_selection_t))) {
         return LIBSPDM_STATUS_INVALID_MSG_FIELD;
     }
-    opaque_element_version_section = (void *)(opaque_element_table_header + 1);
+    opaque_element_version_section = (const void *)(opaque_element_table_header + 1);
     if ((opaque_element_version_section->sm_data_version !=
          SECURED_MESSAGE_OPAQUE_ELEMENT_SMDATA_DATA_VERSION) ||
         (opaque_element_version_section->sm_data_id !=
```
jyao1, line 533: why we need check again?

I think we need move sm_data_version check to common function and remove check here.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -448,14 +539,12 @@ libspdm_process_opaque_data_version_selection_data(libspdm_context_t *spdm_conte
          secured_message_version_index <
          spdm_context->local_context.secured_message_version.spdm_version_count;
```
jyao1, line 540: please add spdm_version_count check with element_length.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -231,28 +344,21 @@ libspdm_process_opaque_data_supported_version_data(libspdm_context_t *spdm_conte
...
-    opaque_element_support_version =
-        (const void *)(opaque_element_table_header + 1);
+    opaque_element_table_header = (const secured_message_opaque_element_table_header_t*)
+                                  get_element_ptr;
+
+    /*check for supported vesion data*/
+    opaque_element_support_version = (const void *)(opaque_element_table_header + 1);
     if ((opaque_element_support_version->sm_data_version !=
          SECURED_MESSAGE_OPAQUE_ELEMENT_SMDATA_DATA_VERSION) ||
         (opaque_element_support_version->sm_data_id !=
```
jyao1, line 364: Same comment as previous function.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -192,6 +191,133 @@ libspdm_build_opaque_data_supported_version_data(libspdm_context_t *spdm_context
...
+        total_element_len += current_element_len;
+
+        if (data_element_size < total_element_len) {
+            return false;
+        }
+
+        if (opaque_element_table_header->id == element_id) {
+            secured_message_element_header =
+                (const void *)(opaque_element_table_header + 1);
+            if ((secured_message_element_header->sm_data_id == sm_data_id) ||
```
jyao1, line 297: Why use "||" ?

I think it should be "&&"

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -192,6 +191,133 @@ libspdm_build_opaque_data_supported_version_data(libspdm_context_t *spdm_context
...
+                 SECURED_MESSAGE_OPAQUE_ELEMENT_SMDATA_DATA_VERSION)) {
+                /*get element by element id*/
+                *get_element_ptr = opaque_element_table_header;
+                *get_element_len = current_element_len;
+                result = true;
+            }
+        }
+
+        /*move to next element*/
+        opaque_element_table_header =
```
jyao1, line 308: How do you know this new opaque_element_table_header is valid?

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -231,62 +358,45 @@ libspdm_process_opaque_data_supported_version_data(libspdm_context_t *spdm_conte
...
-    if ((opaque_element_support_version->sm_data_version !=
-         SECURED_MESSAGE_OPAQUE_ELEMENT_SMDATA_DATA_VERSION) ||
-        (opaque_element_support_version->sm_data_id !=
-         SECURED_MESSAGE_OPAQUE_ELEMENT_SMDATA_ID_SUPPORTED_VERSION) ||
-        (opaque_element_support_version->version_count == 0)) {
+    opaque_element_table_header = (const secured_message_opaque_element_table_header_t*)
+                                  get_element_ptr;
+
+    /*check for supported vesion data*/
+    opaque_element_support_version = (const void *)(opaque_element_table_header + 1);
```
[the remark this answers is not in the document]
Wenxing-hou: When libspdm get element successfully from opaque data use `libspdm_get_element_from_opaque_data`,
this means the opaque_element_table_header is valid. 
And the follow check in `libspdm_get_element_from_opaque_data` is passed.
```
        if (opaque_element_table_header->id == element_id) {
            secured_message_element_header =
                (const void *)(opaque_element_table_header + 1);
            if ((secured_message_element_header->sm_data_id == sm_data_id) &&
                (secured_message_element_header->sm_data_version ==
                 SECURED_MESSAGE_OPAQUE_ELEMENT_SMDATA_DATA_VERSION)) {
                /*get element by element id*/
                *get_element_ptr = opaque_element_table_header;
                *get_element_len = current_element_len;
                result = true;
            }
        }
```

And the  `opaque_element_support_version = (const void *)(opaque_element_table_header + 1);`, 
` opaque_element_support_version` is same as `secured_message_element_header `.
so  opaque_element_support_version is valid.

library/spdm_common_lib/libspdm_com_opaque_data.c
jyao1: The problem is that you move to the next, but you did not check the size.

The pointer may already out of range, before you deference it.

library/spdm_common_lib/libspdm_com_opaque_data.c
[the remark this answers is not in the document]
jyao1: But you did not check the size, before you start dereference the version_count.

PR #935, opened 2022-06-09
Chunk Send POC 1.

richkong88:
Draft Chunk Send POC implementation to solicit feedback. No ULTs yet. Tested via my own requester/responder talking to each other.

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_send_receive.c
+++ b/library/spdm_requester_lib/libspdm_req_send_receive.c
@@ -373,7 +373,9 @@ libspdm_return_t libspdm_handle_large_request(
              * than the DATA_TRANSFER_SIZE. In this case an ERROR_LARGE_RESPONSE
              * is returned directly in the response buffer rather than part of
              * the CHUNK_SEND_ACK. Store this error response in scratch buffer
-             * to be handled when reading response */
+             * to be handled when reading response. Also note that in this case
+             * of large response, the CHUNK_SEND_ACK portion is not sent.
+             * Only the response portion that requires the CHUNK_GET is sent */
             if (response_size < send_info->large_message_capacity) {
                 libspdm_copy_mem(
                     send_info->large_message, send_info->large_message_capacity,
@@ -407,13 +409,17 @@ libspdm_return_t libspdm_handle_large_request(
                 break;
             }
             if (spdm_response->header.param2 != send_info->chunk_handle) {
-                status = LIBSPDM_STATUS_INVALID_MSG_SIZE;
+                status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
                 break;
             }
             if (send_info->chunk_seq_no != spdm_response->chunk_seq_no) {
                 status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
                 break;
             }
+
+            /* Guard against the case where we have a bad responder that:
+             * 1) Sends all bytes but does not set LAST_CHUNK bit.
+             * 2) Dos not send all bytes, but sets LAST_CHUNK bit. */
             if ((send_info->chunk_bytes_transferred >= send_info->large_message_size
                  && !(spdm_response->header.param1 & SPDM_CHUNK_SEND_REQUEST_ATTRIBUTE_LAST_CHUNK))
                 || (send_info->chunk_bytes_transferred < send_info->large_message_size
@@ -512,9 +518,11 @@ libspdm_return_t libspdm_send_spdm_request(libspdm_context_t *spdm_context,
         && request_size > spdm_context->connection_info.capability.data_transfer_size) {
 
         #if LIBSPDM_ENABLE_CHUNK_CAP
-        /* since libspdm_send_request is not called with the original request in
-         * this flow, backup it to last_spdm_request, because the caller wants to
-         * compare it with response */
+        /* libspdm_send_request is not called with the original request in this flow.
+         * This leads to the last_spdm_request field not having the original request value.
+         * The caller assumes the request has been copied to last_spdm_request,
+         * so that it can compare last_spdm_request's fields with response fields
+         * Therefore the request must be copied to last_spdm_request here. */
 
         if (((const spdm_message_header_t*) request)->request_response_code != SPDM_RESPOND_IF_READY
             && ((const spdm_message_header_t*) request)->request_response_code != SPDM_CHUNK_GET
@@ -528,7 +536,6 @@ libspdm_return_t libspdm_send_spdm_request(libspdm_context_t *spdm_context,
         status = libspdm_handle_large_request(
             spdm_context, session_id, request_size, request);
         #else  /* LIBSPDM_ENABLE_CHUNK_CAP*/
-        /* TODO: Is this the right error handling? */
         status = LIBSPDM_STATUS_BUFFER_TOO_SMALL;
         #endif /* LIBSPDM_ENABLE_CHUNK_CAP*/
     }
@@ -568,8 +575,6 @@ libspdm_return_t libspdm_receive_spdm_response(libspdm_context_t *spdm_context,
     spdm_message_header_t *spdm_response;
     size_t response_capacity;
     libspdm_chunk_info_t *send_info;
-    size_t i;
-    uint8_t *ptr;
     #endif /* LIBSPDM_ENABLE_CHUNK_CAP */
 
     if ((session_id != NULL) &&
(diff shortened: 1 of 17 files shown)


jyao1:
Looks good to me in general.

library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -207,6 +232,229 @@ libspdm_return_t libspdm_receive_response(void *context, const uint32_t *session
...
+                break;
+            }
+
+            if (response_size < sizeof(spdm_chunk_send_ack_response_t)) {
+                status = LIBSPDM_STATUS_INVALID_MSG_SIZE;
+                break;
+            }
+            if (spdm_response->header.param1
+                & SPDM_CHUNK_SEND_ACK_RESPONSE_ATTRIBUTE_EARLY_ERROR_DETECTED) {
+                status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
```
steven-bellock, line 406: The error response will be present in `CHUNK_SEND_ACK.ResponseToLargeRequest`.

library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -207,6 +232,229 @@ libspdm_return_t libspdm_receive_response(void *context, const uint32_t *session
...
+                break;
+            }
+            if (spdm_response->header.param2 != send_info->chunk_handle) {
+                status = LIBSPDM_STATUS_INVALID_MSG_SIZE;
```
steven-bellock, line 410: This should be `LIBSPDM_STATUS_INVALID_MSG_FIELD`.

library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -207,6 +232,229 @@ libspdm_return_t libspdm_receive_response(void *context, const uint32_t *session
...
+                break;
+            }
+            if (send_info->chunk_seq_no != spdm_response->chunk_seq_no) {
+                status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
+                break;
+            }
+            if ((send_info->chunk_bytes_transferred >= send_info->large_message_size
+                 && !(spdm_response->header.param1 & SPDM_CHUNK_SEND_REQUEST_ATTRIBUTE_LAST_CHUNK))
```
steven-bellock, line 418: What's going on here? `CHUNK_SEND_ACK.Param1.LastChunk` doesn't exist.

library/spdm_responder_lib/libspdm_rsp_chunk_send.c
```diff
@@ -1,15 +1,244 @@
+/**
```
steven-bellock, line 1: Have the file name match the response name: `libspdm_rsp_chunk_send_ack.c`.

library/spdm_responder_lib/libspdm_rsp_chunk_send.c
```diff
@@ -1,15 +1,244 @@
...
+            response_size, response);
+    }
+
+    send_info = &spdm_context->chunk_context.send;
+
+    if (!send_info->chunk_in_use) {
+
+        large_message_size = *(const uint32_t*) (spdm_request + 1);
+        chunk = (((const uint8_t*) (spdm_request + 1)) + sizeof(uint32_t));
+        calc_chunk_size =
```
steven-bellock, line 108: I think @jyao1 said that `request_size` is only an approximation due to transport data paths that don't have byte granularity. Ie, the data path is 8-bytes.

library/spdm_responder_lib/libspdm_rsp_chunk_send.c
```diff
@@ -1,15 +1,244 @@
...
+            ((uint32_t)request_size - (uint32_t)(chunk - (const uint8_t*) spdm_request));
+
+        if (spdm_request->chunk_seq_no != 0
+            || spdm_request->chunk_size != calc_chunk_size
+            || (uint32_t)request_size != spdm_context->local_context.capability.data_transfer_size
+            || large_message_size >= LIBSPDM_MAX_SPDM_MSG_SIZE
+            ) {
+            status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
```
steven-bellock, line 116: `LastChunk` shall not be set on the first `CHUNK_SEND` message.

library/spdm_responder_lib/libspdm_rsp_receive_send.c
```diff
@@ -564,6 +557,9 @@ libspdm_return_t libspdm_build_response(void *context, const uint32_t *session_i
             LIBSPDM_DEBUG((LIBSPDM_DEBUG_ERROR,
                            "Warning: Could not save chunk. Scratch buffer too small.\n"));
 
+            /* Should we really be returning LARGE_RESPONSE is we aren't supporting chunking?
```
steven-bellock, line 560: `LargeResponse` is the correct error code when `DataTransferSize < Response <= MaxSPDMmsgSize`. `ResponseTooLarge` is the correct error code when `Response > MaxSPDMmsgSize`. I'll file an issue with the specification to clarify this.

steven-bellock: https://github.com/DMTF/Security-TF/issues/2060

steven-bellock (changes requested):

library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -253,8 +506,38 @@ libspdm_return_t libspdm_send_spdm_request(libspdm_context_t *spdm_context,
...
+                spdm_context->last_spdm_request, sizeof(spdm_context->last_spdm_request),
+                request, request_size);
+            spdm_context->last_spdm_request_size = request_size;
+        }
+
+        status = libspdm_handle_large_request(
+            spdm_context, session_id, request_size, request);
+        #else  /* LIBSPDM_ENABLE_CHUNK_CAP*/
+        /* TODO: Is this the right error handling? */
+        status = LIBSPDM_STATUS_BUFFER_TOO_SMALL;
```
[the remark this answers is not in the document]
jyao1: I don't think we need LIBSPDM_STATUS_BUFFER_TOO_SMALL_PEER.

If the requester no such capability, the caller should ensure request_size is smaller.

This check is done at line 482. Why we need handle here again?

library/spdm_requester_lib/libspdm_req_send_receive.c
richkong88: Not clear what the ask is here?

library/spdm_requester_lib/libspdm_req_send_receive.c
richkong88: I'm detecting the error case where we have a bad responder that is
1) Not setting the LastChunk when All bytes have been sent
Or 
2) Setting the LastChunk bit when Not All bytes have been sent

steven-bellock:
I filed https://github.com/DMTF/Security-TF/issues/2061 against the specification. After re-reading `EarlyErrorDetected` I think the specification is currently correct and clear. If the Responder detects a non-LargeResponse error after the last `CHUNK_SEND` has been sent then the Responder replies with `CHUNK_SEND_ACK.EarlyErrorDetected` set to 0 and the error is present in `CHUNK_SEND_ACK.ResponseToLargeRequest`.

library/spdm_requester_lib/libspdm_req_send_receive.c
steven-bellock: This means that the Responder has detected an error in the request and has returned the error to the Requester. libspdm needs to parse the error response and return that to the integrator: `LIBSPDM_STATUS_RESYNCH_PEER`, `LIBSPDM_STATUS_BUSY_PEER`, etc.

library/spdm_requester_lib/libspdm_req_send_receive.c
steven-bellock: The Requester sets `CHUNK_SEND.Param1.LastChunk`. The check should be done by the Responder.

library/spdm_responder_lib/libspdm_rsp_receive_send.c
steven-bellock: Although I think if a endpoint does not support chunking then it should set `DataTransferSize == MaxSPDMmsgSize` in which case it would be `ResponseTooLarge`.

library/spdm_requester_lib/libspdm_req_send_receive.c
richkong88: Is the error response of type `spdm_error_response_t` which has an ERROR_CODE and ERROR_DATA? So this would be different type than `LIBSPDM_STATUS_RESYNC_PEER`, `LIBSPDM_STATUS_BUSY_PEER`, etc of type `libspdm_return_t`. I'm not clear on what mapping of values should be used between `spdm_error_response_t` and `libspdm_return_t`.

library/spdm_requester_lib/libspdm_req_send_receive.c
richkong88: Ah! Right.

library/spdm_responder_lib/libspdm_rsp_chunk_send.c
richkong88: Yes. Fixed.

library/spdm_requester_lib/libspdm_req_send_receive.c
richkong88: I will save the `spdm_error_response_t` to the internal scratch space and make the this code break and return success. That will lead to the error response being read later on in the flow by `libspdm_receive_spdm_response`.

library/spdm_requester_lib/libspdm_req_send_receive.c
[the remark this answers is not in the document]
richkong88: The check at 482 also is ANDed with the capability flag. The check here is against commands that aren't VERSION and CAPABILITIES, which do not support chunking.

jyao1:
@richkong88 , since you added ULT in latest code, you may update the description - "No ULTs yet."

jyao1:
https://github.com/DMTF/libspdm/pull/958 is merged.

PR #937, opened 2022-06-10
fix the private key memory leak

Wenxing-hou:


os_stub/spdm_device_secret_lib_sample/lib.c
```diff
@@ -310,7 +310,8 @@ bool libspdm_gen_csr(uint32_t base_hash_algo, uint32_t base_asym_algo, bool *nee
     result = libspdm_gen_x509_csr(hash_nid, asym_nid, csr_len, csr_pointer, csr_buffer_size,
                                   requester_info, requester_info_length,
                                   context, subject_name);
-
+    libspdm_asym_free(base_asym_algo, context);
+    free(prikey);
```
jyao1, line 314: you need free at line 302 as well.

Wenxing-hou: Yes. I add the code: `free(prikey);`.

Issue #933, opened 2022-06-09
Dynamic memory stored in 'cached_req_info' allocated through function 'libspdm_read_cached_requester_info' is lost


Zhiqiang520:
1. Dynamic memory stored in 'cached_req_info' allocated through function 'libspdm_read_cached_requester_info' at line 272 is lost at line 282.

https://github.com/DMTF/libspdm/blob/064203def5f6b2c2d3e8fa12879caf731e186a4f/os_stub/spdm_device_secret_lib_sample/lib.c#L272-L282

The TRACEBACK is as following:
```
libspdm\os_stub\spdm_device_secret_lib_sample\lib.c:272: Dynamic memory stored in 'cached_req_info' is allocated by calling function 'libspdm_read_cached_requester_info'.
libspdm\os_stub\spdm_device_secret_lib_sample\lib.c:152: '*req_info' is allocated by function 'libspdm_read_input_file'.
libspdm\unit_test\spdm_unit_test_common\support.c:65: '*file_data' is allocated by function 'malloc'.
```
2. Dynamic memory stored in 'cached_req_info' allocated through function 'libspdm_read_cached_requester_info' at line 272 also may be lost at line 288, 296, 302, 314.
https://github.com/DMTF/libspdm/blob/064203def5f6b2c2d3e8fa12879caf731e186a4f/os_stub/spdm_device_secret_lib_sample/lib.c#L287-L288
https://github.com/DMTF/libspdm/blob/064203def5f6b2c2d3e8fa12879caf731e186a4f/os_stub/spdm_device_secret_lib_sample/lib.c#L295-L296
https://github.com/DMTF/libspdm/blob/064203def5f6b2c2d3e8fa12879caf731e186a4f/os_stub/spdm_device_secret_lib_sample/lib.c#L301-L302
https://github.com/DMTF/libspdm/blob/064203def5f6b2c2d3e8fa12879caf731e186a4f/os_stub/spdm_device_secret_lib_sample/lib.c#L314

Issue #932, opened 2022-06-09
Dynamic memory stored in 'csr_p' allocated through function 'libspdm_gen_csr' is lost


Zhiqiang520:
Dynamic memory stored in 'csr_p' allocated through function 'libspdm_gen_csr' at line 127 is lost at line 159.

https://github.com/DMTF/libspdm/blob/064203def5f6b2c2d3e8fa12879caf731e186a4f/library/spdm_responder_lib/libspdm_rsp_csr.c#L127-L130

https://github.com/DMTF/libspdm/blob/064203def5f6b2c2d3e8fa12879caf731e186a4f/library/spdm_responder_lib/libspdm_rsp_csr.c#L156-L160

The TRACEBACK is as following:
```
libspdm\library\spdm_responder_lib\libspdm_rsp_csr.c:127: Dynamic memory stored in 'csr_p' is allocated by calling function 'libspdm_gen_csr'
libspdm\os_stub\spdm_device_secret_lib_sample\lib.c:279: '*csr_pointer' is allocated by function 'libspdm_read_cached_csr'.
libspdm\os_stub\spdm_device_secret_lib_sample\lib.c:234: '*csr_pointer' is allocated by function 'libspdm_read_input_file'.
libspdm\unit_test\spdm_unit_test_common\support.c:65: '*file_data' is allocated by function 'malloc'.
```

PR #941, opened 2022-06-12
Fix get csr memory leak

Wenxing-hou:
Fix: #932 
Fix: #933 

### DIFF
--- a/include/library/spdm_device_secret_lib.h
+++ b/include/library/spdm_device_secret_lib.h
@@ -234,17 +234,18 @@ bool libspdm_write_certificate_to_nvm(uint8_t slot_id, const void * cert_chain,
  * @param[in]      base_asym_algo        asym public key to set
  * @param[in]      need_reset            device need reset for gen csr
  *
- * @param[out]     csr_len               CSR len for DER format
- * @param[in]      csr_pointer           For input, csr_pointer is address to store CSR.
- * @param[out]     csr_pointer           For input, csr_pointer is address for stored CSR.
- * @param[in]      csr_buffer_size       The size of store CSR buffer.
- *
  * @param[in]      requester_info        requester info to gen CSR
  * @param[in]      requester_info_length The len of requester info
+ *
+ * @param[in]      csr_len               For input，csr_len is the size of store CSR buffer.
+ *                                       For output，csr_len is CSR len for DER format
+ * @param[in]      csr_pointer           For input, csr_pointer is buffer address to store CSR.
+ *                                       For output, csr_pointer is address for stored CSR.
+ *                                       The csr_pointer address will be changed.
  *
  * @retval  true   Success.
  * @retval  false  Failed to gen CSR.
  **/
 bool libspdm_gen_csr(uint32_t base_hash_algo, uint32_t base_asym_algo, bool *need_reset,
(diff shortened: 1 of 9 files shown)


jyao1:
I think below function should be changed:
```
bool libspdm_gen_csr(uint32_t base_hash_algo, uint32_t base_asym_algo, bool *need_reset,
                     size_t *csr_len, uint8_t **csr_pointer, size_t csr_buffer_size,
                     uint8_t *requester_info, size_t requester_info_length);
```

to
```
bool libspdm_gen_csr(uint32_t base_hash_algo, uint32_t base_asym_algo, bool *need_reset,
                     uint8_t *requester_info, size_t requester_info_length,
                     uint8_t *csr, size_t *csr_len);
```

for input csr_len is csr buffer max size.
for output csr_len is csr buffer actual size.

os_stub/spdm_device_secret_lib_sample/lib.c
```diff
@@ -266,21 +266,33 @@ bool libspdm_gen_csr(uint32_t base_hash_algo, uint32_t base_asym_algo, bool *nee
...
-        /*get the cahed requester info and csr*/
+        /*get the cached requester info and csr*/
         if ((result) &&
             (cached_req_info_length == requester_info_length) &&
             (libspdm_const_compare_mem(cached_req_info, requester_info, requester_info_length)) &&
-            (libspdm_read_cached_csr(base_asym_algo, csr_pointer, csr_len))) {
+            (libspdm_read_cached_csr(base_asym_algo, &cached_csr, csr_len))) {
+
+            /*get and save cached csr*/
+            libspdm_copy_mem(*csr_pointer, csr_buffer_size, cached_csr, *csr_len);
```
jyao1, line 283: You cannot do the copy directly.

You can compare the size and return BUFFER_TOO_SMALL, if the destination is not big enough.

Wenxing-hou: Yes. I have added the check and deleted useless parameter :csr_buffer_size.

include/library/spdm_device_secret_lib.h
```diff
@@ -234,17 +234,17 @@ bool libspdm_write_certificate_to_nvm(uint8_t slot_id, const void * cert_chain,
...
+ *                                       For input, csr_pointer is address for stored CSR.
+ *
  * @retval  true   Success.
  * @retval  false  Failed to gen CSR.
  **/
 bool libspdm_gen_csr(uint32_t base_hash_algo, uint32_t base_asym_algo, bool *need_reset,
-                     size_t *csr_len, uint8_t **csr_pointer, size_t csr_buffer_size,
-                     uint8_t *requester_info, size_t requester_info_length);
+                     uint8_t *requester_info, size_t requester_info_length,
+                     size_t *csr_len, uint8_t **csr_pointer);
```
jyao1, line 250: Why you need **csr_pointer? Why not use *csr, as I suggested?

jyao1: OK, then please describe that clearly in the comment

Wenxing-hou: Yes. I have added the comment. Thanks.

steven-bellock (approved):

Issue #887, opened 2022-05-18
Responder shall not respond to the VCA request message w/ ErrorCode=ResponseNotReady


hiption:
in SPDM 1.2, there are some statements about ErrorCode=ResponseNotReady as follows;
"A Responder shall not respond to the GET_VERSION request message with ErrorCode=ResponseNotReady ."
"A Responder shall not respond to GET_CAPABILITIES request message with ErrorCode=ResponseNotReady ."
"A Responder shall not respond to NEGOTIATE_ALGORITHMS request message with ErrorCode=ResponseNotReady ."

But, there are some test cases as follows;
https://github.com/DMTF/libspdm/blob/b42d264dd770d6f14c2a2278c9942806e2c554fd/unit_test/test_spdm_responder/version.c#L199-L235

https://github.com/DMTF/libspdm/blob/b42d264dd770d6f14c2a2278c9942806e2c554fd/unit_test/test_spdm_responder/capabilities.c#L579-L618

https://github.com/DMTF/libspdm/blob/b42d264dd770d6f14c2a2278c9942806e2c554fd/unit_test/test_spdm_responder/algorithms.c#L1061-L1102

It seems to conflict between Spec. and implementation.

PR #954, opened 2022-06-17
remove ResponseNotReady unit test in vca

Wenxing-hou:
Fix: #887 


jyao1:
please dont change case index. Just delete the function and make it reserved.

Wenxing-hou:
I have change the code. Thanks.

PR #1015, opened 2022-07-06
Remove illegal check of reserved field

### LINKED ISSUE #1014: Requester should not check conditionally reserved fields
The SPDM specification says
>Unless otherwise specified, field values marked as Reserved shall be written as zero ( 0 ), ignored when read, not
modified and not interpreted as an error if not zero, and used in transcript hash calculations as is.

and `CHALLENGE_AUTH.Param2` says
>This field is reserved if the public key of the Responder was provisioned to the Requester in a trusted environment.

However if the public key has been provisioned libspdm checks that `Param2` is 0, in violation of the specification : https://github.com/DMTF/libspdm/blob/30adb0aa91384eb478ecc3a9d3e39b0b694dc737/library/spdm_requester_lib/libspdm_req_challenge.c#L181-L188

steven-bellock:
Fixes #1014


steven-bellock:
Apparently we did not have unit tests for this check.

PR #1024, opened 2022-07-08
Convert functions to internal linkage

### LINKED ISSUE #389: Convert functions from external linkage to internal linkage
libspdm_requester_lib, libspdm_responder_lib, and libspdm_common_lib expose functions that are local to a compilation unit. Need to add `static`to them. We can do the usual
```
#if LIBSPDM_TEST
#define STATIC static
#else
#define STATIC
#endif
```
so that they can exposed for unit testing if needed.

steven-bellock:
Fixes #389

This also moves some statically allocated variables to the stack.


library/spdm_requester_lib/libspdm_req_encap_request.c
```diff
@@ -53,11 +39,24 @@ void libspdm_register_get_encap_response_func(void *context,
  *
  * @return GET_ENCAP_RESPONSE function according to the request code.
  **/
-libspdm_get_encap_response_func
+static libspdm_get_encap_response_func
 libspdm_get_encap_response_func_via_request_code(uint8_t request_response_code)
 {
     size_t index;
 
+    libspdm_get_encap_response_struct_t m_libspdm_get_encap_response_struct[] = {
```
jyao1, line 47: please remove "m_" prefix, which means module global variable.

PR #1029, opened 2022-07-11
General cleanup

steven-bellock:
1. Move #includes or #defines from include/library to include/internal.
2. Move more statically allocated variables to stack.


library/spdm_crypt_lib/libspdm_crypt_crypt.c
```diff
@@ -4255,15 +4243,23 @@ bool libspdm_verify_leaf_cert_eku_spdm_OID(const uint8_t *cert, size_t cert_size
     size_t index;
     size_t len;
 
+    /* SPDM defined OID */
+    #define OID_SPDM_EXTENSION {0x2B, 0x06, 0x01, 0x04, 0x01, 0x83, 0x1C, 0x82, 0x12, 0x06}
```
steven-bellock, line 4247: @jyao1 Should some of these OIDs live in `industry_standard/spdm.h` since they're defined in the specification?

jyao1: ok. Agree. The spec defined data can be in spdm.h

steven-bellock:
@jyao1 https://github.com/DMTF/spdm-emu/pull/100 has the cleanup on the `spdm-emu` side. That can go in first since it's just adding an #include.

jyao1:
Have you also checked spdm-dump and SPDM-Responder-Validator?

steven-bellock:
> Have you also checked spdm-dump and SPDM-Responder-Validator?

Nope. Will look at those next.

steven-bellock:
spdm-dump is fine. SPDM-Responder-Validator gets built when spdm-emu gets built, right?

Issue #1030, opened 2022-07-11
OpaqueData in MEASUREMENTS


hiption:
There are redundant code for inserting OpaqueData.

https://github.com/DMTF/libspdm/blob/af332d221e58f99b3e9b6a32e8950e255ae6d887/library/spdm_responder_lib/libspdm_rsp_measurements.c#L54-L60

https://github.com/DMTF/libspdm/blob/af332d221e58f99b3e9b6a32e8950e255ae6d887/library/spdm_responder_lib/libspdm_rsp_measurements.c#L102-L108

steven-bellock:
This has been in the code base for a very long time.

PR #1036, opened 2022-07-13
Make the common code reuse in libspdm_rsp_measurements

Wenxing-hou:
Fix: #1030 

Because the `libspdm_create_measurement_opaque` and `libspdm_create_measurement_signature` have common code and they are internal function, I change the function parameters to make the common code reuse.

And add a unit test for code test.

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_measurements.c
+++ b/library/spdm_responder_lib/libspdm_rsp_measurements.c
@@ -9,6 +9,41 @@
 #if LIBSPDM_ENABLE_CAPABILITY_MEAS_CAP
 
 /**
+ * This function creates the opaque data to response message.
+ * @param  spdm_context                  A pointer to the SPDM context.
+ * @param  response_message              The measurement response message with empty signature to be filled.
+ * @param  response_message_size         Total size in bytes of the response message including signature.
+ * @param  fill_response_ptr             A pointer to fill response.
+ *
+ * @retval true                          measurement opaque data and nonce is created.
+ * @retval false                         measurement opaque data and nonce is not created.
+ **/
+static bool libspdm_create_measurement_opaque(libspdm_context_t *spdm_context,
+                                              void *response_message,
+                                              size_t response_message_size,
+                                              uint8_t *fill_response_ptr)
+{
+    if(!libspdm_get_random_number(SPDM_NONCE_SIZE, fill_response_ptr)) {
+        return false;
+    }
+    fill_response_ptr += SPDM_NONCE_SIZE;
+
+    *(uint16_t *)fill_response_ptr =
+        (uint16_t)spdm_context->local_context.opaque_measurement_rsp_size;
+    fill_response_ptr += sizeof(uint16_t);
+
+    if (spdm_context->local_context.opaque_measurement_rsp != NULL) {
+        libspdm_copy_mem(fill_response_ptr,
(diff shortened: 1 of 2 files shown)


library/spdm_responder_lib/libspdm_rsp_measurements.c
```diff
@@ -8,6 +8,38 @@
 
 #if LIBSPDM_ENABLE_CAPABILITY_MEAS_CAP
 
+/**
+ * This function creates the opaque data to response message.
+ * @param  spdm_context                  A pointer to the SPDM context.
+ * @param  response_message              The measurement response message with empty signature to be filled.
+ * @param  response_message_size          Total size in bytes of the response message including signature.
+ * @param  fill_response_ptr            A pointer to fill response.
+ **/
```
steven-bellock, line 17: Add @retvals

library/spdm_responder_lib/libspdm_rsp_measurements.c
```diff
@@ -39,76 +71,25 @@ static bool libspdm_create_measurement_signature(libspdm_context_t *spdm_context
         spdm_context->local_context.opaque_measurement_rsp_size +
         signature_size;
     LIBSPDM_ASSERT(response_message_size > measurment_sig_size);
-    ptr = (void *)((uint8_t *)response_message + response_message_size -
-                   measurment_sig_size);
+    fill_response_ptr = (void *)((uint8_t *)response_message + response_message_size -
```
steven-bellock, line 74: Is this `(void *)` cast needed?

steven-bellock (changes requested):

library/spdm_responder_lib/libspdm_rsp_measurements.c
Wenxing-hou: I have added the @retval. Thanks.

library/spdm_responder_lib/libspdm_rsp_measurements.c
Wenxing-hou: The `(void *)` is not needed now. I have deleted it.

steven-bellock (approved):

library/spdm_responder_lib/libspdm_rsp_measurements.c
```diff
@@ -423,8 +408,10 @@ libspdm_return_t libspdm_get_response_measurements(void *context,
             }
         }
     } else {
+        fill_response_ptr = (void *)((uint8_t *)spdm_response +
+                                     spdm_response_size - measurements_no_sig_size);
         if(!libspdm_create_measurement_opaque(spdm_context, spdm_response,
-                                              spdm_response_size)) {
+                                              spdm_response_size, fill_response_ptr)) {
```
hiption, line 414: inserting nonce and opaque fields in MEASURENTS is mandatory. 
So, it is better to move to out of  if statement and remove "libspdm_create_measurement_opaque()" from "libspdm_create_measurement_signature()."

Wenxing-hou: Yes, inserting nonce and opaque fields in MEASURENTS is mandatory.  But code should ensure create opaque and nonce successfully. So add the check in if statement.

About why add `libspdm_create_measurement_opaque()` in `libspdm_create_measurement_signature()`,it can reduce the redundant code.  Thanks.

PR #1040, opened 2022-07-14
Cleanup device secret lib

steven-bellock:


steven-bellock:
`spdm-emu` and `spdm-dump` appear to be ok with this change.

jyao1:
conflict?

Issue #1013, opened 2022-07-06
CHALLENGE_AUTH's slot mask should return all populated slots


steven-bellock:
In the SPDM specification `CHALLENGE_AUTH.Param2` says
>Slot mask. The bit in position K of this byte shall be set to 1b if and only if slot number K contains a certificate chain for the protocol version in the SPDMVersion field.

Currently libspdm only sets the bit that corresponds to the slot number in the request (`CHALLENGE.Param1`) rather than all the populated slots. It should return all populated slots the same as in the `DIGESTS` response https://github.com/DMTF/libspdm/blob/30adb0aa91384eb478ecc3a9d3e39b0b694dc737/library/spdm_responder_lib/libspdm_rsp_digests.c#L132-L140

PR #1042, opened 2022-07-14
challenge_auth param2 slot mask should return all populated slots

Wenxing-hou:
Fix: #1013 

Because the populated slot store cert_chain may be not continuous, the get slot number logic is changed.

And the PR has passed unit test and emu test.

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -1132,4 +1132,14 @@ void libspdm_get_receiver_buffer (
     void **receiver_buffer,
     size_t *receiver_buffer_size);
 
+/**
+ * Get the certificate slot mask
+ *
+ * @param[in]   context              A pointer to the SPDM context.
+ *
+ * @retval slot_mask                 get slot mask
+ **/
+uint8_t libspdm_get_cert_slot_mask (
+    libspdm_context_t *spdm_context);
+
 #endif /* SPDM_COMMON_LIB_INTERNAL_H */
--- a/library/spdm_common_lib/libspdm_com_crypto_service.c
+++ b/library/spdm_common_lib/libspdm_com_crypto_service.c
@@ -493,6 +493,29 @@ bool libspdm_generate_cert_chain_hash(libspdm_context_t *spdm_context,
 }
 
 /**
+ * Get the certificate slot mask
+ *
+ * @param[in]   context              A pointer to the SPDM context.
+ *
+ * @retval slot_mask                 get slot mask
+ **/
+uint8_t libspdm_get_cert_slot_mask(libspdm_context_t *spdm_context)
+{
+    size_t index;
+    uint8_t slot_mask;
+
+    slot_mask = 0;
+    for (index = 0; index < SPDM_MAX_SLOT_COUNT; index++) {
+        if (spdm_context->local_context
+            .local_cert_chain_provision[index] != NULL) {
+            slot_mask |= (1 << index);
+        }
+    }
+
+    return slot_mask;
+}
+
+/**
  * This function verifies the digest.
  *
  * @param  spdm_context                  A pointer to the SPDM context.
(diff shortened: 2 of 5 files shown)


steven-bellock (changes requested):
Since this logic is mostly the same throughout requests and responses it should probably live in a common function in `include/internal/libspdm_common_lib.h`.

Wenxing-hou:
I have added a function: `libspdm_get_cert_slot_mask`. Thanks.
This function can also be used for the issue :[digest responder check slot will fail when the slot store cert_chain uncontinuously ](https://github.com/DMTF/libspdm/issues/1043).

steven-bellock:
With #1043 we'll probably want to get rid of `local_context.slot_count` and have a function, `libspdm_get_cert_slot_count`, that scans for populated slots similar to `libspdm_get_cert_slot_mask`. In that case `libspdm_get_cert_slot_mask` would look like
```
uint8_t libspdm_get_cert_slot_mask(libspdm_context_t *spdm_context);
```
Alternatively, we could have one function that provides both the slot mask and the slot count since the for-loop can produce both values
```
void libspdm_get_cert_slot_mask_and_count (libspdm_context_t *spdm_context, uint8_t *slot_mask, uint8_t *slot_count);
```

Wenxing-hou:
Thanks. I have changed the function to `libspdm_get_cert_slot_mask_and_count`.

steven-bellock (changes requested):

library/spdm_requester_lib/libspdm_req_encap_challenge_auth.c
```diff
@@ -101,12 +103,21 @@ libspdm_return_t libspdm_get_encap_response_challenge_auth(
...
+
     if (slot_id == 0xFF) {
         spdm_response->header.param2 = 0;
 
         slot_id = spdm_context->local_context.provisioned_slot_id;
     } else {
-        spdm_response->header.param2 = (1 << slot_id);
+        libspdm_get_cert_slot_mask_and_count(spdm_context, &slot_mask, &slot_count);
+        if (slot_count == spdm_context->local_context.slot_count) {
+            LIBSPDM_ASSERT(slot_mask);
```
jyao1, line 114: this is encap request. what if the slot is not provisioned.
Should it return some internal error?

steven-bellock: If we've gotten to this point in the code then the (Requester) integrator has set `CHAL_CAP=1` and `CERT_CAP=1 ^^ PUB_KEY_ID_CAP=1`, so if the integrator hasn't provisioned a certificate chain or public key then they've messed up and I don't see much point in returning an error to a buggy integrator. If they're paranoid about this they can turn on assertions.

library/spdm_common_lib/libspdm_com_crypto_service.c
```diff
@@ -492,6 +492,26 @@ bool libspdm_generate_cert_chain_hash(libspdm_context_t *spdm_context,
...
 
+/**
+ * Get the certificate slot mask
+ *
+ * @param[in]   context              A pointer to the SPDM context.
+ * @param[out]  slot_mask            address of get slot mask
+ *
+ **/
+void libspdm_get_cert_slot_mask_and_count(libspdm_context_t *spdm_context,
+                                          uint8_t *slot_mask)
```
jyao1, line 503: I think it should be:

```
uint8_t libspdm_get_cert_slot_mask (libspdm_context_t *spdm_context);
```

Wenxing-hou: Thanks. I forgot change the function name before. I have changed it.

include/internal/libspdm_common_lib.h
```diff
@@ -1132,4 +1132,15 @@ void libspdm_get_receiver_buffer (
...
     size_t *receiver_buffer_size);
 
+/**
+ * Get the certificate slot mask
+ *
+ * @param[in]   context              A pointer to the SPDM context.
+ * @param[out]  slot_mask            address of get slot mask
+ *
+ **/
+void libspdm_get_cert_slot_mask(
```
jyao1, line 1142: No need to return via parameter. please follow my suggestion:

```
uint8_t libspdm_get_cert_slot_mask (libspdm_context_t *spdm_context);
```

Wenxing-hou: I have changed the code. Thanks.

jyao1 (approved):

PR #1048, opened 2022-07-15
Fix and add fuzzy test cases to improve code coverage

Xiaohanjlll:


jyao1:
What is purpose of this patch?
Why you need change seed?

Xiaohanjlll:
> What is purpose of this patch? Why you need change seed?

1.Because the test cases have not been updated for a long time, resulting in insufficient coverage, they need to be updated
2.I think the old seed is wrong, but in fact it is correct. Both the old and new seeds are correct

Issue #1050, opened 2022-07-17
Debugging Issues


steven-bellock:
1. libspdm has debugging capabilities enabled by default that are disabled by defining the `MDEPKG_NDEBUG` macro. I'd like to see debugging disabled by default, and we should have a more libspdm-specific macro to enable debugging.
2. Currently the build system doesn't define `MDEPKG_NDEBUG` for Release targets.
3. When `MDEPKG_NDEBUG` is defined there are many `unused-but-set-variable` warning / errors. For example https://github.com/DMTF/libspdm/blob/main/library/spdm_common_lib/libspdm_com_context_data.c#L1776-L1783 When debugging is disabled `LIBSPDM_ASSERT` goes away and `finished_key_ready` is not checked.

steven-bellock:
Also
4. `LIBSPDM_DEBUG_CODE_BEGIN` says "Otherwise, the source lines between LIBSPDM_LIBSPDM_DEBUG_CODE_BEGIN() and LIBSPDM_LIBSPDM_DEBUG_CODE_END() are not included in a module." but that doesn't appear to be true. The code is included in a module.

PR #1051, opened 2022-07-17
Disable debugging on Release builds

steven-bellock:
and fix warnings / errors caused by that.

Fixes 1 and 3 of #1050.


steven-bellock:
@jyao1 @Zhiqiang520 This pull request is encountering a stack overflow in https://github.com/DMTF/libspdm/runs/7381028883?check_suite_focus=true and https://github.com/DMTF/libspdm/runs/7381029200?check_suite_focus=true when running `test_spdm_requester_challenge`. The main change in this pull request is defining the `MDEPKG_NDEBUG` macro. If possible can you have a look? Not sure if it is a false positive. Thanks.

Zhiqiang520:
> @jyao1 @Zhiqiang520 This pull request is encountering a stack overflow in https://github.com/DMTF/libspdm/runs/7381028883?check_suite_focus=true and https://github.com/DMTF/libspdm/runs/7381029200?check_suite_focus=true when running `test_spdm_requester_challenge`. The main change in this pull request is defining the `MDEPKG_NDEBUG` macro. If possible can you have a look? Not sure if it is a false positive. Thanks.

@steven-bellock Sorry for late reply. We have reproduced this stack overflow in our local Ubuntu OS, now we are debugging and try to find the root cause.

liuw1:
> @jyao1 @Zhiqiang520 This pull request is encountering a stack overflow in https://github.com/DMTF/libspdm/runs/7381028883?check_suite_focus=true and https://github.com/DMTF/libspdm/runs/7381029200?check_suite_focus=true when running `test_spdm_requester_challenge`. The main change in this pull request is defining the `MDEPKG_NDEBUG` macro. If possible can you have a look? Not sure if it is a false positive. Thanks.

@steven-bellock I am not sure if it is a positive false, but I tracked the the code line caused this issue https://github.com/DMTF/libspdm/blob/b2c86cc6493179642e1f7ba319a592887d1203d8/library/spdm_crypt_lib/libspdm_crypt_crypt.c#L1705. If we redefine the function and modify type of "context_size" to size_t. This issue can be solved.

steven-bellock:
@liuw1 So the main effect of this change is that all the `LIBSPDM_ASSERT`s go away. I wonder if this is a latent issue that wasn't detected by the `LIBSPDM_ASSERT`s due to a change in layout.

liuw1:
> So the main effect of this change is that all the `LIBSPDM_ASSERT`s go away. I wonder if this is a latent issue that wasn't detected by the `LIBSPDM_ASSERT`s due to a change in layout.

This is a strange issue. I redefine the code and the issue cannot be reproduced now. Please help to review the PR #1072.

steven-bellock:
Will return to this at the beginning of August. Closing for now.

PR #1062, opened 2022-07-19
Cleanup base.h and processor_bind

steven-bellock:


steven-bellock:
UEFI does not like stdarg.h. Will make a new pull request.

PR #1064, opened 2022-07-20
Fix libspdm_create_measurement_opaque address error

Wenxing-hou:


steven-bellock:
What error does this fix?

Wenxing-hou:
Because  `libspdm_create_measurement_signature` will use `fill_response_ptr`  after `libspdm_create_measurement_opaque` change `fill_response_ptr`.
https://github.com/DMTF/libspdm/blob/e288c30df793be6cdbd798cfff440a5a2a09d980/library/spdm_responder_lib/libspdm_rsp_measurements.c#L59-L94

steven-bellock (approved):

PR #1072, opened 2022-07-21
Fix stack buffer overflow issue of test_spdm_requester_challenge

liuw1:
Issue details:
https://github.com/DMTF/libspdm/pull/1051


jyao1:
I wont treat it as a fix. It is more like a work-around.

@steven-bellock , what is your opinion?

liuw1:
Agree, it's a work-around.

steven-bellock:
I'm starting to think it's a false positive. But I'm out next week so let's table this for now and I can return to investigating it at the beginning of August. Feel free to close this pull request.

Issue #1077, opened 2022-07-21
[Code Inspection] Dynamic memory stored in 'x509_req' allocated through function 'X509_REQ_new' at line 2440 can be lost at line 2480


Zhiqiang520:
Dynamic memory stored in `x509_req` allocated through function `X509_REQ_new` at line 2440 can be lost at line 2480.
https://github.com/DMTF/libspdm/blob/e288c30df793be6cdbd798cfff440a5a2a09d980/os_stub/cryptlib_openssl/pk/x509.c#L2438-L2443
https://github.com/DMTF/libspdm/blob/e288c30df793be6cdbd798cfff440a5a2a09d980/os_stub/cryptlib_openssl/pk/x509.c#L2476-L2482

PR #1081, opened 2022-07-22
fix memory lost in function libspdm_gen_x509_csr.

Zhiqiang520:
fix: #1077


jyao1:
I dont think this is right fix. Why you change status code?

Zhiqiang520:
> I dont think this is right fix. Why you change status code?

Refer to the following Line 2545(original code),  force set `ret = 0;` before Line 2547 `goto free_all;`
I think this force set is to make sure the Line 2555 `return (ret != 0);` return false when error occur.
But I found some code directly `goto free_all;` without force set `ret = 0;` when error occur.
https://github.com/DMTF/libspdm/blob/cf1cbcb108efa06d4a7c4177275e5cf89f352e06/os_stub/cryptlib_openssl/pk/x509.c#L2544-L2555

jyao1:
but why you can resolve memory lost issue by enforcing ret to 0?
The memory leak is not related to ret status.

jyao1 (approved):

Issue #1082, opened 2022-07-22
[Code Inspection] Missing restore spdm_context->transcript.digest_context_m1m2 to NULL in function libspdm_append_message_c


Zhiqiang520:
in function `libspdm_append_message_c`
1. Before running Line 1196 `return LIBSPDM_STATUS_CRYPTO_ERROR;`, first run Line 1193~1195 to restore `spdm_context->transcript.digest_context_m1m2` to `NULL`.
```
     libspdm_hash_free (spdm_context->connection_info.algorithm.base_hash_algo, 
                        spdm_context->transcript.digest_context_m1m2); 
     spdm_context->transcript.digest_context_m1m2 = NULL; 
```
https://github.com/DMTF/libspdm/blob/cf1cbcb108efa06d4a7c4177275e5cf89f352e06/library/spdm_common_lib/libspdm_com_context_data.c#L1186-L1197
2.  But Line 1204 directly `return LIBSPDM_STATUS_CRYPTO_ERROR;`, not run the above Line 1193~1195 to restore `spdm_context->transcript.digest_context_m1m2` to `NULL`.
https://github.com/DMTF/libspdm/blob/cf1cbcb108efa06d4a7c4177275e5cf89f352e06/library/spdm_common_lib/libspdm_com_context_data.c#L1200-L1205
3. same in function `libspdm_append_message_b`, `libspdm_append_message_mut_b` and `libspdm_append_message_mut_c`.

rw8896:
@Zhiqiang520 

I am not 100% sure but it looks intended to not reset m1m2 in Line 1204.

"Before line 1196" is to initialize transcript.digest_context_m1m2 if it's originally NULL.
In case of initialization failure, it sets transcript.digest_context_m1m2 back to NULL.

line 1204 can't simply set transcript.digest_context_m1m2 to NULL, which will erase the previous message history.
Current error handling looks okay if libspdm_hash_update won't modify m1m2 in case of any error.

liuw1:
@rw8896 

Agree with you, we cannot simply set transcript.digest_context_m1m2 to NULL in line 1204. We need to add two scenarios for line 1200-1205. If the default transcript.digest_context_m1m2 is NULL, we need set transcript.digest_context_m1m2 to NULL if the result is false. Otherwise we should keep transcript.digest_context_m1m2.

PR #1089, opened 2022-07-26
Fix the missing restore in function libspdm_append_message_b, libspdm…

Zhiqiang520:
…_append_message_c, libspdm_append_message_mut_b and libspdm_append_message_mut_c.

Fix: #1082


jyao1:
Would you please clarify how you perform the test, to cover all possible flow?

Zhiqiang520:
> Would you please clarify how you perform the test, to cover all possible flow?
Currently it is hard to implement a system level test case to cover the branch ( result == False),  so this PR may have potential issue, so I suggest to close this PR firstly.

steven-bellock:
We ultimately need a test double for the crypto library that can return test-controlled failures to libspdm.

PR #1099, opened 2022-07-29
Improve fuzzy test coverage and support version 1.2

Xiaohanjlll:


jyao1:
what is current code coverage with this patch?

Xiaohanjlll:
> what is current code coverage with this patch?

The current code coverage of patches should be above 75%.And cover version 1.2

Issue #1102, opened 2022-08-01
add HKDF.Extract API and use that to replace HMAC in KeyDerivation to eliminate confusing.


jyao1:
Current SPDM key derivation uses hmac_all. The meaning is actually hkdf_extract. I think we should use hkdf_extract explicitly to avoid confusing.

1) Define libspdm_hkdf_extract() in spdm_crypt_lib.h.

2) Use libspdm_hkdf_extract() to replace libspdm_hmac_all() in libspdm_generate_session_handshake_key() and libspdm_generate_session_data_key().

PR #1103, opened 2022-08-01
Define and use hkdf_extract.

jyao1:
Fix: https://github.com/DMTF/libspdm/issues/1102

### DIFF
--- a/library/spdm_crypt_lib/libspdm_crypt_crypt.c
+++ b/library/spdm_crypt_lib/libspdm_crypt_crypt.c
@@ -1543,6 +1543,98 @@ bool libspdm_hmac_all(uint32_t base_hash_algo, const void *data,
 }
 
 /**
+ * Return HKDF extract function, based upon the negotiated HKDF algorithm.
+ *
+ * @param  base_hash_algo                 SPDM base_hash_algo
+ *
+ * @return HKDF extract function
+ **/
+libspdm_hkdf_extract_func get_spdm_hkdf_extract_func(uint32_t base_hash_algo)
+{
+    switch (base_hash_algo) {
+    case SPDM_ALGORITHMS_BASE_HASH_ALGO_TPM_ALG_SHA_256:
+#if LIBSPDM_SHA256_SUPPORT == 1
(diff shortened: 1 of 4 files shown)


steven-bellock (changes requested):

steven-bellock:
I filed https://github.com/DMTF/Security-TF/issues/2133 on the specification side.

library/spdm_crypt_lib/libspdm_crypt_crypt.c
```diff
@@ -1542,6 +1542,98 @@ bool libspdm_hmac_all(uint32_t base_hash_algo, const void *data,
...
+ * @param  base_hash_algo                 SPDM base_hash_algo
+ *
+ * @return HKDF extract function
+ **/
+libspdm_hkdf_extract_func get_spdm_hkdf_extract_func(uint32_t base_hash_algo)
+{
+    switch (base_hash_algo) {
+    case SPDM_ALGORITHMS_BASE_HASH_ALGO_TPM_ALG_SHA_256:
+#if LIBSPDM_SHA256_SUPPORT == 1
+        return libspdm_hkdf_sha256_extract;
```
steven-bellock, line 1557: Does `cryptlib.h` really need to implement, for example, `libspdm_hkdf_sha256_extract`? Why can't these functions live in `spdm_crypt_lib.h` and call, for example, `libspdm_hmac_sha256_all`?

jyao1: We should follow the crypto API. And hmac will be called by the crypto lib API, such as https://github.com/Mbed-TLS/mbedtls/blob/f71e2878084126737cc39083e1e15afc459bd93d/library/hkdf.c#L49.

I don't see the need to call hmac directly.

steven-bellock: Ok that'll be a separate issue on the dividing line between cryptography that libspdm manages and cryptography that `cryptlib.h` manages.

steven-bellock (approved):

Issue #1122, opened 2022-08-09
consider add CodeQL analysis


jyao1:
reference: https://github.com/github/codeql

liuw1:
I will take it

steven-bellock:
It looks like the scan did not find anything particularly interesting, probably because we've already run Coverity. We are able to automatically generate issues from the scan, like in #1132, so that is convenient.

PR #1125, opened 2022-08-10
Enable CodeQL scan in CI

liuw1:
Fix: #1122 


liuw1:
@jyao1 Please help to open CodeQL in actions.

steven-bellock:
Do you know how long it takes to run this? If it takes a long time, to start this may be something we run manually, or every week or so.

liuw1:
> Do you know how long it takes to run this? If it takes a long time, to start this may be something we run manually, or every week or so.

@steven-bellock About 28 minutes. What's your proposal?

jyao1:
Maybe schedule and run once per week? E.g. Saturday night.

jyao1:
Please consider also add a CodeQL section in https://github.com/DMTF/libspdm/blob/main/doc/test.md.
And tell people how to view result.

steven-bellock:
So when you say it's 28 minutes it looks like it's also scanning OpenSSL, MbedTLS, and Cmocka, which we shouldn't be doing. It's the responsibility of those projects to scan for issues. We should just be scanning `libspdm/include`, `libspdm/library` , and files in `libspdm/os_stub` that we've created. Maybe `libspdm/unit_test` too if it helps find bugs.

liuw1:
> So when you say it's 28 minutes it looks like it's also scanning OpenSSL, MbedTLS, and Cmocka, which we shouldn't be doing. It's the responsibility of those projects to scan for issues. We should just be scanning `libspdm/include`, `libspdm/library` , and files in `libspdm/os_stub` that we've created. Maybe `libspdm/unit_test` too if it helps find bugs.

Agree with you, I will update it later.

liuw1:
Currently, CodeQL need 5 minutes to run one round.

liuw1:
Need to resolve the alerts in report link: https://github.com/liuw1/libspdm/pull/2/checks?check_run_id=7848870526

Issue #1121, opened 2022-08-09
Test failed when requester (DATA_SUPPORT = 0) <=> responder (DATA_SUPPORT = 1) 


Wenxing-hou:
When open `LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT` in requester and close `LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT` in responder, the emu will fail.

Both `requester (DATA_SUPPORT = 0) <=> responder (DATA_SUPPORT = 1)`   and   `requester (DATA_SUPPORT = 1) <=> responder (DATA_SUPPORT = 0) ` case will fail.

The failed reason is The failed reason is gen_sign and verify_sign message is different.

![image](https://user-images.githubusercontent.com/87060980/183571280-b0135b0d-cfef-4242-8e64-97988413d009.png)

jyao1:
There is pareameter - `is_data_hash`. Have you checked that?

Wenxing-hou:
Thanks. I have checked the `is_data_hash` and find the reason: the `mut_hash` is different between requester and responder.

PR #1128, opened 2022-08-10
Fix the mut_hash inconsistent

Wenxing-hou:
Fix: #1121 

### DIFF
--- a/library/spdm_common_lib/libspdm_com_crypto_service.c
+++ b/library/spdm_common_lib/libspdm_com_crypto_service.c
@@ -176,6 +176,23 @@ static bool libspdm_calculate_m1m2(void *context, bool is_mut,
         spdm_context->connection_info.algorithm.base_hash_algo);
 
     if (is_mut) {
+        if ((spdm_context->connection_info.version >> SPDM_VERSION_NUMBER_SHIFT_BIT) >
+            SPDM_MESSAGE_VERSION_11) {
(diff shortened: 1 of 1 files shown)


jyao1:
I am not sure if this is right fix.
message_a is not needed for mut_auth according to the spec.

Wenxing-hou:
> I am not sure if this is right fix. message_a is not needed for mut_auth according to the spec.

I will check code again. If the fix is not right, I will change it.

library/spdm_common_lib/libspdm_com_crypto_service.c
```diff
@@ -175,6 +175,20 @@ static bool libspdm_calculate_m1m2(void *context, bool is_mut,
...
+        libspdm_get_managed_buffer_size(
+            &spdm_context->transcript.message_a));
+    status = libspdm_append_managed_buffer(
+        m1m2,
+        libspdm_get_managed_buffer(&spdm_context->transcript.message_a),
+        libspdm_get_managed_buffer_size(
+            &spdm_context->transcript.message_a));
+    if (LIBSPDM_STATUS_IS_ERROR(status)) {
+        return false;
+    }
```
jyao1, line 190: I think we need add SPDM version check and only add mut auth for SPDM 1.2 and above.

jyao1 (approved):

Issue #1140, opened 2022-08-16
Code scan alert - Potentially uninitialized local variable "hash_size" in library/spdm_common_lib/libspdm_com_context_data.c


liuw1:
Alert: 
https://github.com/liuw1/libspdm/security/code-scanning/12
https://github.com/liuw1/libspdm/security/code-scanning/13
https://github.com/liuw1/libspdm/security/code-scanning/14
https://github.com/liuw1/libspdm/security/code-scanning/15

Xiaohanjlll:
I will check it.

PR #1149, opened 2022-08-22
Fix uninitialized variable hash_size in library/spdm_common_lib/libspdm_com_context_data.c

Xiaohanjlll:
Fix: #1140 


jyao1 (changes requested):
Please dont do init in definition.

Issue #1141, opened 2022-08-16
Code scan alert - Potentially uninitialized local variable in os_stub/cryptlib_mbedtls/pk/x509.c


liuw1:
Alerts:
https://github.com/liuw1/libspdm/security/code-scanning/16
https://github.com/liuw1/libspdm/security/code-scanning/17
https://github.com/liuw1/libspdm/security/code-scanning/18
https://github.com/liuw1/libspdm/security/code-scanning/19
https://github.com/liuw1/libspdm/security/code-scanning/20

Xiaohanjlll:
I will check it.

PR #1150, opened 2022-08-22
fix uninitialized local variable in os_stub/cryptlib_mbedtls/pk/x509.c 

Xiaohanjlll:
Fix : #1141 


jyao1 (changes requested):
Please dont init in definition

Issue #1139, opened 2022-08-16
Code scan alert - Use of potentially dangerous function "gmtime"


liuw1:
https://github.com/liuw1/libspdm/security/code-scanning/27

steven-bellock:
I am not able to access that page.

Xiaohanjlll:
I will check it.

PR #1151, opened 2022-08-22
Fix the potential hazards of gmtime

Xiaohanjlll:
Fix: #1139 

### DIFF
--- a/os_stub/cryptlib_mbedtls/sys_call/timer_wrapper_host.c
+++ b/os_stub/cryptlib_mbedtls/sys_call/timer_wrapper_host.c
@@ -7,6 +7,7 @@
 /** @file
  * C Run-Time Libraries (CRT) Time Management Routines Wrapper Implementation.
  **/
+#define _POSIX_C_SOURCE 200112L
 
 #include <base.h>
 #include "library/memlib.h"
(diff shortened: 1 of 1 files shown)


liuw1 (commented):
Why did you handle it for windows?

liuw1 (approved):

os_stub/cryptlib_mbedtls/sys_call/timer_wrapper_host.c
```diff
@@ -15,13 +16,17 @@
 struct tm *mbedtls_platform_gmtime_r(const mbedtls_time_t *tt,
                                      struct tm *tm_buf)
 {
-    struct tm *lt;
-
-    lt = gmtime(tt);
+#ifdef _MSC_VER
+    if (gmtime_s(tm_buf, tt) != 0) {
```
steven-bellock, line 20: Should be `gmtime_s(tm_buf, tt) == NULL`, like below.

Xiaohanjlll: > Should be `gmtime_s(tm_buf, tt) == NULL`, like below.
errno_t gmtime_s(
   struct tm* tmDest,
   const __time_t* sourceTime
);
typedef int                           errno_t;
C4047: "errno_t" and "void *" have different indirection levels

basis:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/gmtime-s-gmtime32-s-gmtime64-s?view=msvc-170

steven-bellock: C11 defines it as `struct tm *gmtime_s(const time_t * restrict timer, struct tm * restrict result);` so I guess this compiler isn't following the standard? I guess in the short-term we'll go with the compiler.

steven-bellock (approved):

Issue #1136, opened 2022-08-15
ResponderVerifyData / RequesterVerifyData calculation should use hash of transcript


xiaoyuruan:
From how libspdm_append_managed_buffer() and libspdm_hmac_update_with_response_finished_key() are implemented and invoked, it seems the libspdm is calculating ResponderVerifyData / RequesterVerifyData as HMAC(finished_key, transcript) instead of HMAC(finished_key, hash(transcript)).

SPDM 1.2 `ResponderVerifyData` says:

>If the Session Handshake Phase is encrypted and/or message authenticated, then this field shall be of length H and it shall equal the HMAC of the **transcript hash**, using finished_key as the secret key and using the negotiated hash algorithm as the hash function. The transcript hash shall be the **hash of the transcript** for KEY_EXCHANGE_RSP HMAC as [Transcript for KEY_EXCHANGE_RSP HMAC](https://github.com/DMTF/SPDM-WG/blob/master/docs/DSP0274/SPDMSpecification.md#definition_th_ke_resp_vf) shows.

jyao1:
SPDM 1.1 is same.

PR #1152, opened 2022-08-22
ResponderVerifyData / RequesterVerifyData should use hash of transcript

jyao1:
Fix https://github.com/DMTF/libspdm/issues/1136.

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -183,9 +183,9 @@ typedef struct {
     libspdm_small_managed_buffer_t message_mut_c;
     libspdm_large_managed_buffer_t message_m;
 #else
-    void                   *digest_context_m1m2;
-    void                   *digest_context_mut_m1m2;
-    void                   *digest_context_l1l2;
+    void *digest_context_m1m2;
+    void *digest_context_mut_m1m2;
+    void *digest_context_l1l2;
 #endif
 } libspdm_transcript_t;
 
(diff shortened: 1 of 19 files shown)


Wenxing-hou:
I think `temp_message_k` can be deleted. 
Because `temp_message_k` is used to store transcript for HMAC input before. Now the HMAC input is `digest_context_th`.

Wenxing-hou:
The follow code in `libspdm_calculate_th_hmac_for_exchange_rsp`  may be not needed: 
```
    if (session_info->session_transcript.digest_context_th == NULL) {
        /* trigger message_k to initialize context after finished_key is ready.*/
        status = libspdm_append_message_k (context, spdm_session_info, is_requester, NULL, 0);
            return false;
        }
        LIBSPDM_ASSERT(session_info->session_transcript.digest_context_th != NULL);
    }
```
I have passed the unit_test without above code.

Because`libspdm_generate_key_exchange_rsp_hmac`/`libspdm_verify_key_exchange_rsp_hmac`/`libspdm_generate_psk_exchange_rsp_hmac`/`libspdm_verify_psk_exchange_rsp_hmac` will call  `libspdm_calculate_th_hmac_for_exchange_rsp` .

And `libspdm_append_message_k` is called before above four funciton to ensure `digest_context_th` isn't null.

jyao1:
> I think temp_message_k can be deleted.

Good catch. Agree. Fixed.

jyao1:
> The follow code in libspdm_calculate_th_hmac_for_exchange_rsp may be not needed:

Good catch. Agree. Fixed.

steven-bellock:
This pull request is too large for me to comprehend, so feel free to merge it when you see fit. I can present ideas on how to prevent an issue like this from happening again in our next meeting.

PR #1164, opened 2022-08-25
Initial draft for message logging

steven-bellock:
This commit is to solicit feedback on the API (spdm_requester_lib.h) and backend design. Currently only supports
* GET_MEASUREMENTS / MEASUREMENTS
* Non-wrapping writes that don't clamp.

I am not sure how this will be tested.


steven-bellock:
@PrithviAPai @vbodired Please review the API in `spdm_requester_lib.h`.

library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -672,6 +681,18 @@ libspdm_return_t libspdm_receive_spdm_response(libspdm_context_t *spdm_context,
...
     }
 
+    #if LIBSPDM_ENABLE_MSG_LOG
+    if ((status == LIBSPDM_STATUS_SUCCESS) && (spdm_context->msg_log.mode &
+        LIBSPDM_MSG_LOG_MODE_ENABLE) != 0) {
+        if (spdm_context->msg_log.offset + *response_size > spdm_context->msg_log.buffer_size) {
+        } else {
+            libspdm_copy_mem((uint8_t *)spdm_context->msg_log.buffer + spdm_context->msg_log.offset,
+                             spdm_context->msg_log.buffer_size, spdm_response, *response_size);
+            //spdm_context->msg_log.offset += *response_size;
```
taprinz, line 691: Why not increment here? or is this for development purposes

steven-bellock: It gets incremented in `libspdm_req_get_measurements.c` since, on transports that don't have 1-byte resolution, the true size of the response is only known once its been parsed.

jyao1:
> I am not sure how this will be tested.

My recommendation is to add unit_test. Just use get_measurement.
The test case can pre-calculate the expected L1L2 and compare the returned L1L2.

include/internal/libspdm_common_lib.h
```diff
@@ -305,9 +304,23 @@ typedef struct {
...
 #endif /* LIBSPDM_ENABLE_CHUNK_CAP */
 
+#if LIBSPDM_ENABLE_MSG_LOG
+typedef struct {
+    void *buffer;
+    size_t buffer_size;
+    uint32_t mode;
+    size_t offset;
+    uint32_t status;
+} libspdm_msg_log_t;
```
jyao1, line 316: After read the code, I realize that buffer_size means max_buffer_size, while offset means real buffer_size. Suggest to change the name.

library/spdm_requester_lib/libspdm_req_get_measurements.c
```diff
@@ -525,6 +525,10 @@ static libspdm_return_t libspdm_try_get_measurement(void *context, const uint32_
 
     status = LIBSPDM_STATUS_SUCCESS;
 
+    #if LIBSPDM_ENABLE_MSG_LOG
+    spdm_context->msg_log.offset += spdm_response_size;
+    #endif /* LIBSPDM_ENABLE_MSG_LOG */
```
jyao1, line 530: It looks weird to me that this function suddenly add the offset...

I suggest we have a consistent programing mode:
 The caller should use a function to append the data to msg_log. The data appending and size adjustment should be done in one function. Not separately.

I think we need have an internal function : libspdm_append_msg_log()

steven-bellock: That will work if, for example, `libspdm_try_get_measurement` doesn't modify the response before it is appended. I'll add it in the next commit.

include/internal/libspdm_common_lib.h
```diff
@@ -305,9 +304,23 @@ typedef struct {
...
+
+#define LIBSPDM_MSG_LOG_MODE_ENABLE 1
+#define LIBSPDM_MSG_LOG_MODE_WRAP 2
+#define LIBSPDM_MSG_LOG_MODE_INC_ERR 4
+#define LIBSPDM_MSG_LOG_MODE_INC_CHUNK 8
+#endif /* LIBSPDM_ENABLE_MSG_LOG */
```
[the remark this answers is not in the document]
steven-bellock: They're in `spdm_common_lib` now.

library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -672,6 +675,19 @@ libspdm_return_t libspdm_receive_spdm_response(libspdm_context_t *spdm_context,
         }
     }
 
+    #if LIBSPDM_ENABLE_MSG_LOG
```
steven-bellock, line 678: This can be deleted in the next commit.

steven-bellock:
Need to add `VCA` support as well.

library/spdm_common_lib/libspdm_com_msg_log.c
```diff
@@ -0,0 +1,100 @@
...
+            libspdm_copy_mem((uint8_t *)spdm_context->msg_log.buffer +
+                             spdm_context->msg_log.buffer_size,
+                             spdm_context->msg_log.max_buffer_size, message, message_size);
+            spdm_context->msg_log.buffer_size += message_size;
+        }
+
+        LIBSPDM_DEBUG((LIBSPDM_DEBUG_INFO, "Message Logging Status = [%x] Buffer Size = [%x] "
+                      "Max Buffer Size = [%x]\n", spdm_context->msg_log.status,
+                      spdm_context->msg_log.buffer_size, spdm_context->msg_log.max_buffer_size));
+        libspdm_internal_dump_hex(spdm_context->msg_log.buffer, spdm_context->msg_log.buffer_size);
```
steven-bellock, line 97: Apparently dumping the message log is causing test failures only on Windows (even GCC / Mingw). @jyao1, I can leave this debug info out, but off the top of your head might you know why this would only fail on Windows?

jyao1: I am surprised on that. Does LIBSPDM_DEBUG() cause failure? or does libspdm_internal_dump_hex() cause failure?

steven-bellock: `libspdm_internal_dump_hex` prints out the entire message log, which gets pretty big, so it might be memory corruption.

steven-bellock: Will leave this unresolved for now.

steven-bellock:
@jyao1 @taprinz I've marked this ready for review. I decided to leave off the wrapping behavior until someone requests it, so right now when the buffer is full then it will stop writing to it. It currently only logs `VCA` and `GET_MEASUREMENTS` but that can be expanded. I'll have a follow-up CL that includes more tests.

library/spdm_common_lib/libspdm_com_msg_log.c
```diff
@@ -0,0 +1,101 @@
...
+
+    if (((spdm_context->msg_log.mode & LIBSPDM_MSG_LOG_MODE_ENABLE) != 0) &&
+        ((spdm_context->msg_log.status & LIBSPDM_MSG_LOG_STATUS_BUFFER_FULL) == 0)) {
+        if (spdm_context->msg_log.buffer_size + message_size >
+            spdm_context->msg_log.max_buffer_size) {
+            libspdm_copy_mem((uint8_t *)spdm_context->msg_log.buffer +
+                             spdm_context->msg_log.buffer_size,
+                             spdm_context->msg_log.max_buffer_size, message,
+                             spdm_context->msg_log.max_buffer_size -
+                             spdm_context->msg_log.buffer_size);
```
steven-bellock, line 86: Need to set `buffer_size` to `max_buffer_size`.

jyao1: please do it.

steven-bellock: It is done.

jyao1:
Ref: https://github.com/DMTF/libspdm/discussions/1049
Ref: https://github.com/DMTF/libspdm/discussions/929

PR #1167, opened 2022-08-28
Remove debug code in libspdm_calculate_m1m2 & libspdm_calculate_l1l2

lsun100:
Remove the debug code in these two functions since they update the
m1m2 or l1l2 content thus affect the actual code logic once enabled
in the case when LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT is set.

### DIFF
--- a/library/spdm_common_lib/libspdm_com_crypto_service.c
+++ b/library/spdm_common_lib/libspdm_com_crypto_service.c
@@ -165,15 +165,10 @@ static bool libspdm_calculate_m1m2(void *context, bool is_mut,
 {
     libspdm_context_t *spdm_context;
     libspdm_return_t status;
-    uint32_t hash_size;
-    uint8_t hash_data[LIBSPDM_MAX_HASH_SIZE];
 
     spdm_context = context;
 
     libspdm_init_managed_buffer(m1m2, LIBSPDM_MAX_MESSAGE_BUFFER_SIZE);
-
-    hash_size = libspdm_get_hash_size(
-        spdm_context->connection_info.algorithm.base_hash_algo);
 
     if (is_mut) {
         if ((spdm_context->connection_info.version >> SPDM_VERSION_NUMBER_SHIFT_BIT) >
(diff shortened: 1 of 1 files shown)


steven-bellock:
@lsun100 I attempted to resolve this in #1051, however the fuzzing tests detected a stack overflow due to the changes, so they didn't go in. If possible that would be the preferable approach, since the debug code would be removed in a release build.

jyao1:
@lsun100, I dont understand how could it happen that "since they update the m1m2 or l1l2 content ".
Would you please elaborate?

lsun100:
> 

For example, libspdm_calculate_m1m2() is called in libspdm_generate_challenge_auth_signature() with the m1m2 as output (line 745 below).  Later the m1m2 is used to calculate the signature (line 789 below) with 'the false argument (line 793) which expects the m1m2 to be data instead of hash.

The removed 'LIBSPDM_DEBUG_CODE' code calls the libspdm_hash_all() to update the m1m2 buffer.  That's to say, it will happen to work when LIBSPDM_DEBUG_CODE is compiled out, but it will break signature verification when LIBSPDM_DEBUG_CODE is compiled in. Usually I would assume that debugging code shouldn't cause any behavior change no matter it's compiled on or off.  Please correct me if I misunderstand it.

library/spdm_common_lib/libspdm_com_crypto_service.c

 744 #if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT
 745     result = libspdm_calculate_m1m2(spdm_context, is_requester, &m1m2);
 746     m1m2_buffer = libspdm_get_managed_buffer(&m1m2);
 747     m1m2_buffer_size = libspdm_get_managed_buffer_size(&m1m2);
 748 #else

 788 #if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT
 789         result = libspdm_responder_data_sign(
 790             spdm_context->connection_info.version, SPDM_CHALLENGE_AUTH,
 791             spdm_context->connection_info.algorithm.base_asym_algo,
 792             spdm_context->connection_info.algorithm.base_hash_algo,
 793             false, m1m2_buffer, m1m2_buffer_size, signature,
 794             &signature_size);

lsun100:
> 
Sorry, looks like my misunderstanding... I'll withdraw the code review.

lsun100:
> 

That's great. I'll close this one.

lsun100:
Thanks @steven-bellock and @jyao1 for the comments. I double checked that it's my misunderstanding. It actually works without the change.

PR #1169, opened 2022-08-28
Fix potential unaligned access

lsun100:
Fix the issue: #70


lsun100:
Looks like the fix is already in. close it.

Issue #1124, opened 2022-08-09
DEADCODE in function "libspdm_rsa_set_key"


yaohuixguo:
Line 158 `switch()` is executed when "key_tag" is `LIBSPDM_RSA_KEY_N` or `LIBSPDM_RSA_KEY_E` or `LIBSPDM_RSA_KEY_D`
line 168 cannot reach this statement: "default:"
https://github.com/DMTF/libspdm/blob/30bcde1b71b1eeb8d703c52da2398e957c006906/os_stub/cryptlib_openssl/pk/rsa_basic.c#L158-L171
line 205 cannot reach this statement: "default:"
https://github.com/DMTF/libspdm/blob/30bcde1b71b1eeb8d703c52da2398e957c006906/os_stub/cryptlib_openssl/pk/rsa_basic.c#L198-L208
line 252 cannot reach this statement: "default:"
https://github.com/DMTF/libspdm/blob/30bcde1b71b1eeb8d703c52da2398e957c006906/os_stub/cryptlib_openssl/pk/rsa_basic.c#L252-L254

steven-bellock:
Note that the entire function `libspdm_rsa_set_key` is essentially dead code, as nothing of substance calls it, besides the unit test.

jyao1:
I think it will be useful after we add raw public key. We can result default unreachable at first.

PR #1171, opened 2022-08-29
Fix 'switch' unreached 'default' path

yaohuixguo:
Removing `default` directly will cause compilation errors. Replace `switch` with `if`
Fix: #1124
Signed-off-by: yaohuixguo <yaohuix.guo@intel.com>


steven-bellock:
@yaohuixguo How did you detect this issue? Does this cause a warning with `-Wextra`?

steven-bellock (approved):

yaohuixguo:
> @yaohuixguo How did you detect this issue? Does this cause a warning with `-Wextra`?

Removing `default` directly will cause `-Werror=switch`
`enumeration value ‘LIBSPDM_RSA_KEY_N’ not handled in switch [-Werror=switch]`

jyao1:
This seems not a fix, but a work-around.

PR #1173, opened 2022-08-29
Add CXL IDE_KM message in CXL 3.0

jyao1:
the spdm-emu patch is at https://github.com/DMTF/spdm-emu/pull/144
the spdm-dump patch is at https://github.com/DMTF/spdm-dump/pull/54

### DIFF
--- a/include/industry_standard/cxl_idekm.h
+++ b/include/industry_standard/cxl_idekm.h
@@ -1,0 +1,175 @@
+/**
+ *  Copyright Notice:
+ *  Copyright 2021-2022 DMTF. All rights reserved.
+ *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
+ **/
+
+/** @file
+ * Definitions in CXL 3.0 specification.
+ **/
+
+#ifndef CXL_IDE_KM_H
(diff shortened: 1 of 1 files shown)


include/industry_standard/cxl_idekm.h
```diff
@@ -0,0 +1,175 @@
...
+/** @file
+ * Definitions in CXL 3.0 specification.
+ **/
+
+#ifndef CXL_IDE_KM_H
+#define CXL_IDE_KM_H
+
+/* Standard ID and Vendor ID for CXL*/
+
+#define SPDM_STANDARD_ID_CXL SPDM_REGISTRY_ID_PCISIG
```
steven-bellock, line 16: These should go in `spdm.h`.

jyao1: I dont think so.
We already have SPDM_REGISTRY_ID_CXL defined in spdm.h, but CXL does not adopt it.

Instead CXL reuses the PCISIG as standard ID, and a special vendor_id (0x1E98).

All those are defined in CXL spec.

It is not related to SPDM specification.

steven-bellock: Ah ok.

steven-bellock (approved):

Issue #1176, opened 2022-08-29
add LIBSPDM_DATA_LOCAL_SLOT_COUNT back to keep compatibility


jyao1:
https://github.com/DMTF/libspdm/pull/1095 deleted LIBSPDM_DATA_LOCAL_SLOT_COUNT.

We should keep the enum to provide compatibility.

PR #1184, opened 2022-09-01
Add slot_count back in enum for compatibility

Wenxing-hou:
Fix: #1176 


include/library/spdm_common_lib.h
```diff
@@ -66,6 +66,7 @@ typedef enum {
     /* Certificate info*/
 
     LIBSPDM_DATA_LOCAL_PUBLIC_CERT_CHAIN,
+    LIBSPDM_DATA_LOCAL_SLOT_COUNT,
```
jyao1, line 69: please add comment: // reserved.

Wenxing-hou: Thanks. I have added it.

jyao1: thank you.
Since reserved is not put in the same line, I feel we should say "LIBSPDM_DATA_LOCAL_SLOT_COUNT is deprecated. Do not use it. Reserved only."

Wenxing-hou: Thanks. I have added it.

PR #1187, opened 2022-09-01
Rename capability macros

### LINKED ISSUE #1028: Rename LIBSPDM_ENABLE_SET_CERTIFICATE_CAP and LIBSPDM_ENABLE_CHUNK_CAP
so that they are consistent with the other capability macros.

steven-bellock:
Fixes #1028.

If integrator exclusively uses the old macro then the compiler will print a message / warning about its use, but still allow its use. See `library/spdm_common_lib.h`.


include/library/spdm_common_lib.h
```diff
@@ -13,6 +13,29 @@
...
 
+#if defined(LIBSPDM_ENABLE_SET_CERTIFICATE_CAP) && \
+    !defined(LIBSPDM_ENABLE_CAPABILITY_SET_CERTIFICATE_CAP)
+#ifdef _MSC_VER
+#pragma message("LIBSPDM_ENABLE_SET_CERTIFICATE_CAP is deprecated. Use " \
+    "LIBSPDM_ENABLE_CAPABILITY_SET_CERTIFICATE_CAP instead. This warning will be removed with " \
+    "the next release.")
+#else
+#warning "LIBSPDM_ENABLE_SET_CERTIFICATE_CAP is deprecated. Use" \
+    "LIBSPDM_ENABLE_CAPABILITY_SET_CERTIFICATE_CAP instead. This warning will be removed with " \
```
steven-bellock, line 24: @ekohandel uncrustify wants me to indent the `#warning` string. However that indentation shows up when the warning is printed, which is ugly. Note that MSVC concatenates the string literal components and ignores the indented white space. Any tips or trick?

steven-bellock: Actually looks like this has been resolved.

jyao1 (approved):

steven-bellock:
> @steven-bellock , do we also need change spdm-emu and spdm-dump?

Yes I'll have those done tomorrow.

PR #1190, opened 2022-09-02
Remove libspdm_random_seed

### LINKED ISSUE #1112: Random number generation
`cryptlib.h` specifies two random number functions : `libspdm_random_seed` and `libspdm_random_bytes`. There are multiple issues with these functions.

1. `libspdm_random_seed` requires a (truly) random seed that libspdm by itself cannot provide.
2. `libspdm_random_bytes` states that it `Generates a pseudorandom byte stream of the specified size.` However it goes on to say `Pseudorandom number generator fails to generate due to lack of entropy.` PRNGs cannot fail due to lack of entropy as they are deterministic based on the random seed.

What I think `cryptlib.h` needs is to get rid of `libspdm_random_seed` and libspdm will just use `libspdm_random_bytes` without any mention of true or pseudo randomness. The quality of the random numbers will be up to the integrator and their security posture.

steven-bellock:
Fixes #1112

### DIFF
--- a/include/hal/library/cryptlib.h
+++ b/include/hal/library/cryptlib.h
@@ -194,9 +194,6 @@ bool libspdm_rsa_get_key(void *rsa_context, const libspdm_rsa_key_tag_t key_tag,
  * length in bits of RSA modulus N as input, and generates all key components.
  * If public_exponent is NULL, the default RSA public exponent (0x10001) will be used.
  *
- * Before this function can be invoked, pseudorandom number generator must be correctly
- * initialized by libspdm_random_seed().
- *
  * If rsa_context is NULL, then return false.
  * If this interface is not supported, then return false.
  *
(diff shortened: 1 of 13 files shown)


include/hal/library/cryptlib.h
```diff
@@ -1942,42 +1936,21 @@ bool libspdm_sm2_dsa_verify(const void *sm2_context, size_t hash_nid,
...
- * @retval true   Pseudorandom number generator has enough entropy for random generation.
- * @retval false  Pseudorandom number generator does not have enough entropy for random generation.
- * @retval false  This interface is not supported.
- *
- **/
-bool libspdm_random_seed(const uint8_t *seed, size_t seed_size);
-
-/**
- * Generates a pseudorandom byte stream of the specified size.
+ * Generates a random byte stream of the specified size.
```
jyao1, line 1943: can we add some comment, such as the library provide or integrator should initialize the random seed to ensure the randomness.

steven-bellock: It is done.

Issue #184, opened 2021-10-09
split crytplib.h to smaller ones.


jyao1:
from @steven-bellock, cryptlib.h is starting to get huge. We might want to split it up until multiple .h files. Something like (hashing algorithms, symmetric key algorithms, public key algorithms, x509, etc)

PR #1191, opened 2022-09-04
Cleanup certificate functions in cryptlib

steven-bellock:
This is work towards completing #184.


steven-bellock:
@jyao1 There are a bunch of functions that should be moved to `unit_test` / `os_stub` or deleted. Let's bring this pull request in as-is and then go over what should stay or go in a subsequent issue / pull request.

Issue #1196, opened 2022-09-07
Need to session state when SPDM_FINISH_RSP is in Chunk.


qizhangz:
When SPDM_FINISH_RSP is in Chunk in secure session, the session state need to be set LIBSPDM_SESSION_STATE_ESTABLISHED.

PR #1197, opened 2022-09-07
need to session state when SPDM_FINISH_RSP is in Chunk.

qizhangz:
Fix: #1196 


library/spdm_responder_lib/libspdm_rsp_receive_send.c
```diff
@@ -662,6 +664,36 @@ libspdm_return_t libspdm_build_response(void *context, const uint32_t *session_i
                     LIBSPDM_SESSION_STATE_ESTABLISHED);
             }
             break;
+        #if LIBSPDM_ENABLE_CAPABILITY_CHUNK_CAP || LIBSPDM_ENABLE_CHUNK_CAP
```
jyao1, line 667: I think this should be global for both in-session and out-of-session.

Suggest to define a local variable request_response_code, to get the code in chunk message, then fall through the current handling process.

jyao1 (approved):

steven-bellock (approved):

PR #1199, opened 2022-09-08
Fix fuzzing error by adding check when release build

Wenxing-hou:


library/spdm_requester_lib/libspdm_req_get_measurements.c
```diff
@@ -147,7 +147,10 @@ static libspdm_return_t libspdm_try_get_measurement(void *context, const uint32_
     if (LIBSPDM_STATUS_IS_ERROR(status)) {
         return status;
     }
-    LIBSPDM_ASSERT (message_size >= transport_header_size);
+    if (message_size >= transport_header_size) {
+        LIBSPDM_ASSERT (false);
+        return LIBSPDM_STATUS_INVALID_MSG_SIZE;
+    }
```
jyao1, line 153: I dont understand the fix.

The message_size is for the send_buffer. It must >= transport_header_size.
Why return error???

PR #1205, opened 2022-09-13
Add chunking to CERTIFICATE response

### LINKED ISSUE #1018: support CHUNK_GET for CERTIFICATE
According to 10.8 GET_CERTIFICATE, the Responder shall fragment and serve CERTIFICATE message in chunks.

https://github.com/DMTF/libspdm/blob/e8c3632e1edcc9588714e84cb154705e8c908e97/library/spdm_responder_lib/libspdm_rsp_certificate.c#L114-L123

Current implementation does not support chunks.

steven-bellock:
Fixes #1018.

Still needs unit test coverage.

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_certificate.c
+++ b/library/spdm_responder_lib/libspdm_rsp_certificate.c
@@ -56,8 +56,7 @@ libspdm_return_t libspdm_get_response_certificate(void *context,
             spdm_request->header.request_response_code,
             response_size, response);
     }
-    if (spdm_context->connection_info.connection_state <
-        LIBSPDM_CONNECTION_STATE_NEGOTIATED) {
+    if (spdm_context->connection_info.connection_state < LIBSPDM_CONNECTION_STATE_NEGOTIATED) {
         return libspdm_generate_error_response(spdm_context,
                                                SPDM_ERROR_CODE_UNEXPECTED_REQUEST,
                                                0, response_size, response);
@@ -104,8 +103,7 @@ libspdm_return_t libspdm_get_response_certificate(void *context,
                                                response_size, response);
     }
 
-    if (spdm_context->local_context
-        .local_cert_chain_provision[slot_id] == NULL) {
+    if (spdm_context->local_context.local_cert_chain_provision[slot_id] == NULL) {
         return libspdm_generate_error_response(
             spdm_context, SPDM_ERROR_CODE_INVALID_REQUEST,
             0, response_size, response);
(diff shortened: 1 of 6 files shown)


library/spdm_responder_lib/libspdm_rsp_certificate.c
```diff
@@ -104,46 +103,49 @@ libspdm_return_t libspdm_get_response_certificate(void *context,
...
 
-    libspdm_reset_message_buffer_via_request_code(spdm_context, session_info,
-                                                  spdm_request->header.request_response_code);
+    if (!libspdm_is_capabilities_flag_supported(spdm_context, false,
+        SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_CHUNK_CAP,
+        SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_CHUNK_CAP)) {
+        if (length > LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN) {
+            length = LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN;
+        }
+    }
```
jyao1, line 132: Why you add check cap check for LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN ?

jyao1: Is there any test to cover the new case: (CHECK_CAP == 1) && (length > LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN) ?

steven-bellock: > Why you add check cap check for LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN ?

If chunking is not supported then it reverts to the old behavior.

jyao1: I am thinking below:
1) If LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN is ignored in CHUNK, then we should describe in spdm_config.h.
2) If we want to change, we should also change in encap_certificate.
3) We need have new test to cover this new corner case, to ensure it still works in all scenario.

steven-bellock: Looking through the code I don't think `LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN` is being used correctly / consistently. For example `libspdm_get_certificate` uses `LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN` as the maximum size of the entire certificate, rather than just a portion. It's possible it's not needed anymore now that we have the send and receive buffers. I'll file a separate issue.

jyao1: > For example libspdm_get_certificate uses LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN as the maximum size of the entire certificate, rather than just a portion.

Would you please point out where the code is? I think length means portion_length
```
    length = LIBSPDM_MIN(length, LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN);
```
Why it is entire certificate?

steven-bellock:
The https://github.com/DMTF/libspdm/blob/main/unit_test/test_spdm_requester/chunk_get.c#L552 test needs to be overhauled. We can discuss that in the meeting tomorrow.

jyao1:
question: Have you tested spdm-emu, with chunk enabled?

steven-bellock:
> question: Have you tested spdm-emu, with chunk enabled?

I synced to head of `spdm-emu` and ran `spdm_requester_emu` and `spdm_responder_emu` and they worked.

Issue #1203, opened 2022-09-12
Validate chunking support in libspdm in session.


jyao1:
Ref: https://github.com/DMTF/spdm-emu/issues/134

We notice chunking does not work in spdm-emu. Need investigate what issue is.

PR #1206, opened 2022-09-13
Validate chunking support in libspdm in session.

qizhangz:
Fix: #1203

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_send_receive.c
+++ b/library/spdm_requester_lib/libspdm_req_send_receive.c
@@ -656,19 +656,10 @@ libspdm_return_t libspdm_receive_spdm_response(libspdm_context_t *spdm_context,
             goto receive_done;
         }
 
-        /* These are the expected possible large response types.
-         * Anything else is an unexpected error. */
-        if (spdm_response->request_response_code != SPDM_DIGESTS &&
-            spdm_response->request_response_code != SPDM_CERTIFICATE &&
-            spdm_response->request_response_code != SPDM_CHALLENGE_AUTH &&
-            spdm_response->request_response_code != SPDM_MEASUREMENTS &&
-            spdm_response->request_response_code != SPDM_VENDOR_DEFINED_RESPONSE &&
-            spdm_response->request_response_code != SPDM_ALGORITHMS &&
-            spdm_response->request_response_code != SPDM_KEY_EXCHANGE_RSP &&
-            spdm_response->request_response_code != SPDM_FINISH_RSP &&
-            spdm_response->request_response_code != SPDM_PSK_EXCHANGE_RSP &&
-            spdm_response->request_response_code != SPDM_ENCAPSULATED_REQUEST &&
-            spdm_response->request_response_code != SPDM_ENCAPSULATED_RESPONSE_ACK
+        /* Per the spec, SPDM_VERSION and SPDM_CAPABILITIES shall not be chunked
+         * and should be an unexpected error. */
+        if (spdm_response->request_response_code == SPDM_VERSION ||
+            spdm_response->request_response_code == SPDM_CAPABILITIES
             ) {
             status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
             goto receive_done;
--- a/library/spdm_requester_lib/libspdm_req_send_receive.c
+++ b/library/spdm_requester_lib/libspdm_req_send_receive.c
@@ -644,7 +644,7 @@ libspdm_return_t libspdm_receive_spdm_response(libspdm_context_t *spdm_context,
         && spdm_response->param1 == SPDM_ERROR_CODE_LARGE_RESPONSE) {
 
         status = libspdm_handle_error_large_response(
-            spdm_context, NULL,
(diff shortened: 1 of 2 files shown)


library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -668,7 +668,8 @@ libspdm_return_t libspdm_receive_spdm_response(libspdm_context_t *spdm_context,
             spdm_response->request_response_code != SPDM_FINISH_RSP &&
             spdm_response->request_response_code != SPDM_PSK_EXCHANGE_RSP &&
             spdm_response->request_response_code != SPDM_ENCAPSULATED_REQUEST &&
-            spdm_response->request_response_code != SPDM_ENCAPSULATED_RESPONSE_ACK
+            spdm_response->request_response_code != SPDM_ENCAPSULATED_RESPONSE_ACK &&
+            spdm_response->request_response_code != SPDM_CSR
```
jyao1, line 672: Instead of check what is allowed, can we check what is not allowed?

For example, if request_response_code is GET_VERSION, GET_CAPABILITYIES, then return error.

That can avoid changing for new command, which will most likely support chunking

library/spdm_responder_lib/libspdm_rsp_receive_send.c
```diff
@@ -621,9 +621,11 @@ libspdm_return_t libspdm_build_response(void *context, const uint32_t *session_i
...
     switch (request_response_code) {
     case SPDM_CHUNK_SEND_ACK:
-        request_response_code =
-            ((spdm_message_header_t*)(my_response + sizeof(spdm_chunk_send_ack_response_t)))
-            ->request_response_code;
+        if (my_response_size > sizeof(spdm_chunk_send_ack_response_t)) {
+            request_response_code =
+                ((spdm_message_header_t*)(my_response + sizeof(spdm_chunk_send_ack_response_t)))
+                ->request_response_code;
+        }
```
jyao1, line 628: In which scenario, this check will fail?

qizhangz: When Chunk is not completed (send_info->chunk_bytes_transferred < send_info->large_message_size), the chunk_send response only contains the data of spdm_chunk_send_ack_response_t like (12 05 00 0f 02 00 ) which indicates chunk_seq_no.
When Chun is completed, the chunk_send response is like (12 05 00 0e 33 00 12 6e 00 00 ). So need to check the response size to get the response code in Chunk

jyao1 (approved):

library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -658,17 +658,8 @@ libspdm_return_t libspdm_receive_spdm_response(libspdm_context_t *spdm_context,
 
         /* These are the expected possible large response types.
```
steven-bellock, line 659: The comment needs to change now that this is a deny-list.

library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -658,17 +658,8 @@ libspdm_return_t libspdm_receive_spdm_response(libspdm_context_t *spdm_context,
...
-            spdm_response->request_response_code != SPDM_CHALLENGE_AUTH &&
-            spdm_response->request_response_code != SPDM_MEASUREMENTS &&
-            spdm_response->request_response_code != SPDM_VENDOR_DEFINED_RESPONSE &&
-            spdm_response->request_response_code != SPDM_ALGORITHMS &&
-            spdm_response->request_response_code != SPDM_KEY_EXCHANGE_RSP &&
-            spdm_response->request_response_code != SPDM_FINISH_RSP &&
-            spdm_response->request_response_code != SPDM_PSK_EXCHANGE_RSP &&
-            spdm_response->request_response_code != SPDM_ENCAPSULATED_REQUEST &&
-            spdm_response->request_response_code != SPDM_ENCAPSULATED_RESPONSE_ACK
+        if (spdm_response->request_response_code == SPDM_VERSION ||
```
steven-bellock, line 661: There are other responses that should not be chunked, like `HEARTBEAT_ACK`.

steven-bellock (changes requested):

library/spdm_requester_lib/libspdm_req_send_receive.c
jyao1: I don't think we need deal with HEATBEAT_ACK.

The check is based upon SPDM spec, which mentioned that VERSION and CAPABILITIES shall not be chunked. HEATBEAT_ACK is not listed there.

library/spdm_requester_lib/libspdm_req_send_receive.c
qizhangz: done

jyao1:
@steven-bellock , I merge it since we need the patch to test chunking.

Feel free to file new one, if you think any improvement is needed.

steven-bellock:
Ok I expanded https://github.com/DMTF/libspdm/issues/1207

Issue #1189, opened 2022-09-02
Base Hash Algorithm selected is mandated to be non-zero  


mct-sparky:
The following lines of code mandate that the base hash algorithm selected be non-zero, without any prior checks for capabilities:

https://github.com/DMTF/libspdm/blob/3ed56c7f11d5e405562d145418a7ae3ce0cc241b/library/spdm_responder_lib/libspdm_rsp_algorithms.c#L539-L545

According to the DMTF Validator Case 3.5, Assertion 3.5.10:

"Assertion 3.5.10: SpdmMessage.BaseHashSel only has one bit at most. if (CHAL_CAP == 1 || MEAS_CAP == 2) then (BaseHashSel == one of {TPM_ALG_SHA_256, TPM_ALG_SHA_384, TPM_ALG_SHA_512, TPM_ALG_SHA3_256, TPM_ALG_SHA3_384, TPM_ALG_SHA3_512}) if (CHAL_CAP == 0 && MEAS_CAP != 2) then (BaseHashSel == 0)"

The selected base hash algorithm should be allowable to be zero in the case that CHAL_CAP is 0 and MEAS_CAP !=2 (or other capabilities that require it be not supported).

jyao1:
I think we need update the assertion, to add the condition of CERT_CAP (because cert chain also requires digest for root cert) and KEY_EX_CAP/PSK_EX_CAP (requires HMAC for verification)

jyao1:
9/5/2022. Agree

PR #1208, opened 2022-09-14
Fix Condition for Base Hash and Base Asym Algorithm Selection

mct-sparky:
The base hash algorithm selected is incorrectly mandated to be non-zero, given the case that the responder only supports unsigned measurements. The base asymmetric algorithm selected should also check for this case, however only the challenge capability is checked.

Fix by checking for capabilities which require these algorithms:
- CERT_CAP, CHAL_CAP, MEAS_CAP_SIG, KEY_EX_CAP, PSK_CAP, PUB_KEY_ID_CAP

This fixes issue #1189

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_algorithms.c
+++ b/library/spdm_responder_lib/libspdm_rsp_algorithms.c
@@ -536,16 +536,47 @@ libspdm_return_t libspdm_get_response_algorithms(void *context,
                 0, response_size, response);
         }
     }
-    algo_size = libspdm_get_hash_size(
-        spdm_context->connection_info.algorithm.base_hash_algo);
-    if (algo_size == 0) {
-        return libspdm_generate_error_response(spdm_context,
-                                               SPDM_ERROR_CODE_INVALID_REQUEST, 0,
-                                               response_size, response);
-    }
+
     if (libspdm_is_capabilities_flag_supported(
             spdm_context, false, 0,
-            SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_CHAL_CAP)) {
+            SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_CERT_CAP) ||
+        libspdm_is_capabilities_flag_supported(
+            spdm_context, false, 0,
+            SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_CHAL_CAP) ||
+        libspdm_is_capabilities_flag_supported(
+            spdm_context, false, 0,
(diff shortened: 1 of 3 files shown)


library/spdm_responder_lib/libspdm_rsp_algorithms.c
```diff
@@ -536,16 +536,34 @@ libspdm_return_t libspdm_get_response_algorithms(void *context,
...
+            spdm_context, false,
+            SPDM_GET_CAPABILITIES_REQUEST_FLAGS_KEY_EX_CAP,
+            SPDM_GET_CAPABILITIES_REQUEST_FLAGS_KEY_EX_CAP) ||
+        libspdm_is_capabilities_flag_supported(
+            spdm_context, false,
+            SPDM_GET_CAPABILITIES_REQUEST_FLAGS_PSK_CAP,
+            SPDM_GET_CAPABILITIES_REQUEST_FLAGS_PSK_CAP) ||
+        libspdm_is_capabilities_flag_supported(
+            spdm_context, false, 0,
+            SPDM_GET_CAPABILITIES_REQUEST_FLAGS_PUB_KEY_ID_CAP)) {
```
jyao1, line 559: Would you please remove PUB_KEY_ID_CAP?
I dont think it is required.

Ref: https://github.com/DMTF/SPDM-Responder-Validator/pull/27

mct-sparky: Removed.

jyao1:
please also fix the unit test to make CI pass.

library/spdm_responder_lib/libspdm_rsp_algorithms.c
```diff
@@ -536,16 +536,32 @@ libspdm_return_t libspdm_get_response_algorithms(void *context,
...
+            spdm_context->connection_info.algorithm.base_hash_algo);
+        if (algo_size == 0) {
+            return libspdm_generate_error_response(
+                spdm_context,
+                SPDM_ERROR_CODE_INVALID_REQUEST, 0,
+                response_size, response);
+        }
         algo_size = libspdm_get_asym_signature_size(
             spdm_context->connection_info.algorithm.base_asym_algo);
         if (algo_size == 0) {
```
jyao1, line 567: I just notice that this check also applied for base_asym_algo. However, PSK_CAP is not required for base_asym algo.

PKS_CAP only requires BaseHash, but not BaseAsym. I suggest we separate the check for BaseHash and BaseAsym.

mct-sparky: Done

jyao1 (approved):

PR #1210, opened 2022-09-14
reduce complie optimize option in Libfuuzer for work around

Wenxing-hou:
Fix: #1050 


Wenxing-hou:
Waiting local fuzzing test.

Issue #1194, opened 2022-09-06
Building with custom OpenSSL fails


antkorczak:
I build libspdm by providing OpenSSL with -DCMAKE_C_FLAGS=-Iopenssl_path/include -DENABLE_BINARY_BUILD=1 -DCOMPILED_LIBCRYPTO_PATH=openssl_path/libcrypto.a -DCOMPILED_LIBSSL_PATH=openssl_path/libssl.a 
This is enough to build libraries (so the resulting build is usable), but the process fails during tests: 
make[2]: *** No rule to make target 'openssl_path/libcrypto.a', needed by 'bin/test_spdm_requester'.  Stop.
make[2]: Leaving directory 'libspdm_path/build'
CMakeFiles/Makefile2:2110: recipe for target 'unit_test/test_spdm_requester/CMakeFiles/test_spdm_requester.dir/all' failed
make[1]: *** [unit_test/test_spdm_requester/CMakeFiles/test_spdm_requester.dir/all] Error 2
make[1]: Leaving directory 'libspdm_path/build'
Makefile:93: recipe for target 'all' failed
make: *** [all] Error 2

The libspdm_path and openssl_path are existing absolute paths 

Is there some better way to provide custom OpenSSL? Ideally there should be a way to provide e.g OPENSSL_ROOT_DIR variable in CMAKE.

jyao1:
9/12/2022 - This is to link with customer build openssl, not the system default openssl.

liuw1:
I will take this issue.

liuw1:
hi @antkorczak, I think yo can use $openssl_path and $libspdm_path in command and then the system can identify the values as variables.

antkorczak:
> hi @antkorczak, I think yo can use $openssl_path and $libspdm_path in command and then the system can identify the values as variables.

I don't quite understand what do You mean

In the request I just didn't want to include real paths that were used, so I just wrote -DCOMPILED_LIBCRYPTO_PATH=openssl_path/libcrypto.a
but in fact the path was valid e.g -DCOMPILED_LIBCRYPTO_PATH=/home/akorczak/Desktop/spdm_wrapper/dependencies/openssl/libcrypto.a
In my build script I use variables like : -DCOMPILED_LIBCRYPTO_PATH=${openssl_root_dir}/lib_static/libcrypto.a

liuw1:
OK, Cmake just need to find OPENSSL_PATH in the older version to support customize build.
[...]

PR #1222, opened 2022-09-21
Openssl customized build enhancement

liuw1:
Fix: #1194


jyao1:
It looks good to me.
@antkorczak, would you please review and feedback?

jyao1 (approved):

Issue #1214, opened 2022-09-19
test_spdm_responder_csr occur heap-buffer-overflow with the LibFuzzer generated specific mutation seed


Xiaohanjlll:
Verified on ce2f118f93d5f757b61aaa8bf7c6778985a05ec6
Ubuntu 20.04.4 LTS

cd libspdm
mkdir build_libfuzzer
cd build_libfuzzer
cmake -DARCH=x64 -DTOOLCHAIN=LIBFUZZER -DTARGET=Release -DCRYPTO=openssl..
make copy_sample_key
make
cd bin

./test_spdm_responder_csr ./../../test_spdm_responder_csr_crash 

> =================================================================
==2695961==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x62100000590f at pc 0x000000638e91 bp 0x7ffef5049e70 sp 0x7ffef5049e68
READ of size 1 at 0x62100000590f thread T0
    #0 0x638e90 in ASN1_get_object /home/test/libspdm_XiaoHan/os_stub/openssllib/openssl/crypto/asn1/asn1_lib.c:55:12
    #1 0x5ea684 in libspdm_asn1_get_tag /home/test/libspdm_XiaoHan/os_stub/cryptlib_openssl/pk/x509.c:269:5
    #2 0x60e0a9 in libspdm_verify_req_info /home/test/libspdm_XiaoHan/library/spdm_crypt_lib/libspdm_crypt_crypt.c:5172:11
    #3 0x5866df in libspdm_get_response_csr /home/test/libspdm_XiaoHan/library/spdm_responder_lib/libspdm_rsp_csr.c:131:14
    #4 0x581f89 in libspdm_test_responder_csr_case1 /home/test/libspdm_XiaoHan/unit_test/fuzzing/test_responder/test_spdm_responder_csr/csr.c:41:5
    #5 0x5829ac in libspdm_run_test_harness /home/test/libspdm_XiaoHan/unit_test/fuzzing/test_responder/test_spdm_responder_csr/csr.c:107:5
    #6 0x58492e in LLVMFuzzerTestOneInput /home/test/libspdm_XiaoHan/unit_test/fuzzing/spdm_unit_fuzzing_common/toolchain_harness.c:124:5
    #7 0x48c761 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (/home/test/libspdm_XiaoHan/build/bin/test_spdm_responder_csr+0x48c761)
[...]

PR #1224, opened 2022-09-22
Fix memory overflow in CSR

Xiaohanjlll:
FIx: #1214 

### DIFF
--- a/os_stub/cryptlib_openssl/pk/x509.c
+++ b/os_stub/cryptlib_openssl/pk/x509.c
@@ -21,6 +21,9 @@
 /*buffer size to store subject object*/
 #define MAX_SBUJECT_NAME_LEN 0x100
 
+/*see link:"https://man.openbsd.org/ASN1_get_object.3" */
+#define OPENSSL_ASN1_ERROR_MASK 0x80
+
 /* OID*/
 #define OID_EXT_KEY_USAGE     { 0x55, 0x1D, 0x25 }
 #define OID_BASIC_CONSTRAINTS { 0x55, 0x1D, 0x13 }
@@ -255,7 +258,7 @@ bool libspdm_asn1_get_tag(uint8_t **ptr, const uint8_t *end, size_t *length,
     int32_t obj_tag;
     int32_t obj_class;
     long obj_length;
-
+    int32_t ret;
 
     /* Save ptr position*/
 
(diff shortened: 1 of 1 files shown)


os_stub/cryptlib_openssl/pk/x509.c
```diff
@@ -266,8 +266,13 @@ bool libspdm_asn1_get_tag(uint8_t **ptr, const uint8_t *end, size_t *length,
         return false;
     }
 
-    ASN1_get_object((const uint8_t **)ptr, &obj_length, &obj_tag, &obj_class,
-                    (int32_t)(end - (*ptr)));
+    status = ASN1_get_object((const uint8_t **)ptr, &obj_length, &obj_tag, &obj_class,
+                    (int32_t)(end - (*ptr)));   
+    /* Either a primitive encoding with a valid tag and definite length, but the content octets won't fit into omax, or parsing failed. */
+    if (status == 0x80 || status == 0xa0){
```
jyao1, line 272: please define macro for 0x80 and 0xa0.

Hardcode value is not allowed.

include/hal/library/cryptlib.h
```diff
@@ -97,6 +97,8 @@
 #define LIBSPDM_CRYPTO_ASN1_TAG_CLASS_MASK 0xC0
 #define LIBSPDM_CRYPTO_ASN1_TAG_PC_MASK 0x20
 #define LIBSPDM_CRYPTO_ASN1_TAG_VALUE_MASK 0x1F
+#define LIBSPDM_CRYPTO_ASN1_PARSE_ERROR 0x80
+#define LIBSPDM_CRYPTO_ASN1_TOO_LONG 0xA0
```
jyao1, line 101: No need to put to cryptlib.h

just put it to x509.c

os_stub/cryptlib_openssl/pk/x509.c
```diff
@@ -21,6 +21,9 @@
 /*buffer size to store subject object*/
 #define MAX_SBUJECT_NAME_LEN 0x100
 
+#define LIBSPDM_CRYPTO_ASN1_PARSE_ERROR 0x80
```
steven-bellock, line 24: How do you know this? Is this defined by ASN or OpenSSL? If it's defined by OpenSSL why aren't we using the OpenSSL #defines?

jyao1:
NOTE: The case we are testing is to change  `LIBSPDM_DATA_TRANSFER_SIZE`  MACRO to `42` in spdm-emu.

Please ensure that configuration can work.

Xiaohanjlll:
> NOTE: The case we are testing is to change `LIBSPDM_DATA_TRANSFER_SIZE` MACRO to `42` in spdm-emu.
> 
> Please ensure that configuration can work.

I check it now

Xiaohanjlll:
> NOTE: The case we are testing is to change `LIBSPDM_DATA_TRANSFER_SIZE` MACRO to `42` in spdm-emu.
> 
> Please ensure that configuration can work.

LIBSPDM_DATA_TRANSFER_SIZE MACRO  changes to 42 and works normally.

jyao1:
> LIBSPDM_DATA_TRANSFER_SIZE MACRO changes to 42 and works normally.

Sorry, please ignore that. Wrong input window.

os_stub/cryptlib_openssl/pk/x509.c
Xiaohanjlll: > How do you know this? Is this defined by ASN or OpenSSL? If it's defined by OpenSSL why aren't we using the OpenSSL #defines?

Sorry, I can't find a suitable macro in OpenSSL

jyao1: 1) define a macro named "OPENSSL_ASN1_ERROR_MASK 0x80", and add link "https://man.openbsd.org/ASN1_get_object.3" to the comment for the MACRO.
2) use "ret & OPENSSL_ASN1_ERROR_MASK" in the code to judge if it is an error.

PR #1229, opened 2022-09-23
Fixes for chunking with secure sessions.

richkong88:
Fixes for chunking. 

@qizhangz Can you please double check spdm-emu. With full data transfer size, everything seems to pass (although I get output diffs from the baseline. I think there must be random nonces in there?) With a small data transfer size (42), the test runs but exits early. The assert is fixed though.  I'm not sure how spdm-emu should behave. Think we should continue to look at after this PR, as this PR takes us further.

### DIFF
--- a/include/library/spdm_lib_config.h
+++ b/include/library/spdm_lib_config.h
@@ -316,43 +316,69 @@
 #endif
 
 
-/* Required scratch buffer size for libspdm internal usage.
- * It may be used to hold the encrypted/decrypted message and/or last sent/received message.
- * It may be used to hold the large request/response and intermediate send/receive buffer
- * in case of chunking.
+/* Scratch buffer defines:
+ * The scratch buffer is reserved for internal libspdm usage.
+ * It is a large buffer that is split into several smaller buffers, which are used as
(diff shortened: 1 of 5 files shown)


jyao1:
@richkong88 , I think there are more problems.

I have some sample fix to pass spdm-emu with 42 as transport size.
https://github.com/jyao1/libspdm/tree/temp_chunk
https://github.com/jyao1/spdm-emu/tree/temp_chunk

They are not final yet. I need run unit test and do more clean up.

richkong88:
> @richkong88 , I think there are more problems. (The requester is using sender buffer to hold spdm request message. But if we set to 42, the sender buffer is too small. We have to use another bigger buffer.)
> 
> I have some sample/draft fix to pass spdm-emu with 42 as transport size. https://github.com/jyao1/libspdm/tree/temp_chunk https://github.com/jyao1/spdm-emu/tree/temp_chunk
> 
> They are not final yet. I need run unit test and do more clean up.

Yeah, I saw the requester bug too, but figured I'd stick with smaller number of changes.

jyao1:
@richkong88 , I created https://github.com/DMTF/libspdm/pull/1230, please review that.

Issue #1213, opened 2022-09-16
Buffer Management Refactor?


richkong88:
With the addition of chunking, the buffer management has gotten quite complex. Currently, there are many buffers with different names:
sender buffer, receiver buffer, scratch space (currently divided into three buffers), and the last_spdm_request.  After being away for a few months, I find it difficult to follow and somewhat error prone.  I'm wondering if we should look into a refactoring of all of this to make it more manageable and easier to understand?

jyao1:
Here is my thought: The buffer usage is below:
1) Sender Buffer: The buffer is to send the a message in transport layer from local. This is defined for optimization, because the device may allocate DMA memory directly to avoid copy.

The "SenderBufferSize" might be bigger or smaller than remote SPDM `DataTransferSize`. However, "SenderMessageSize" shall be smaller than or equal to remote SPDM `DataTransferSize`.

2) Receiver Buffer: The buffer is to receive the a message in transport layer from remote/peer. This is defined for optimization, because the device may allocate DMA memory directly to avoid copy.

To follow SPDM spec, "ReceiverBufferSize" shall be bigger than or equal to local SPDM `DataTransferSize`.

3) Scratch Space: It is used for libspdm internal usage. There is no requirement for the device driver be aware it. The integrator can allocate any space (heap or global data) for it.

The last_spdm_request is to cache the last_spdm_request, as such the spdm responder can handle it later.
I think "last_spdm_request" should be part of scratch space. It is not there, just because we defined it before we introduce scratch space.

In the future, I think we need define "last_spdm_response" to support retry. - But we can hold it until we really want to add support.

For this issue, I suggest 
1) Merge "last_spdm_request" into "scratch buffer".
2) Document the three buffers' usage as above. (sender buffer, receiver buffer, scratch space)

steven-bellock:
Concerning compile-time macros in general, if a macro isn't used to allocate space at compile-time, or if it doesn't confer optimizations, then it should probably be made run-time assignable.

jyao1:
The design without chunking is documented at https://github.com/DMTF/libspdm/discussions/702.

We should document the chunking case.
requester/responder, session/non-session, large message send/receive

jyao1:
9/19/2022 - Make it Q4 stable task.

richkong88:
Sorry, I don't have time to work on this this quarter and probably not next. If someone else would like to try, please do. Or just close since it was just a suggestion for a refactor and not functionality.

[...]

PR #1230, opened 2022-09-23
Make the chunk work with transfer_size 42 in spdm-emu.

jyao1:

with change: `#define LIBSPDM_DATA_TRANSFER_SIZE 42`

1) libspdm unit test pass 
2) spdm-emu test pass

### DIFF
--- a/library/spdm_common_lib/libspdm_com_context_data.c
+++ b/library/spdm_common_lib/libspdm_com_context_data.c
@@ -1981,6 +1981,12 @@ libspdm_return_t libspdm_acquire_sender_buffer (
     }
     spdm_context->sender_buffer = *msg_buf_ptr;
     spdm_context->sender_buffer_size = *max_msg_size;
+    #if LIBSPDM_ENABLE_CAPABILITY_CHUNK_CAP || LIBSPDM_ENABLE_CHUNK_CAP
+    /* it return scratch buffer, because the requester need build message there.*/
+    *msg_buf_ptr = spdm_context->scratch_buffer +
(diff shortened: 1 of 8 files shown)


library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -1981,6 +1981,12 @@ libspdm_return_t libspdm_acquire_sender_buffer (
     }
     spdm_context->sender_buffer = *msg_buf_ptr;
     spdm_context->sender_buffer_size = *max_msg_size;
+    #if LIBSPDM_ENABLE_CAPABILITY_CHUNK_CAP
```
richkong88, line 1984: Ok. I can understand why put this here, so that you wouldn't need to modify all the different requester command functions to use either the sender buffer or the scratch space. I think it's ok to get it working like this, but may be confusing to follow for someone first looking at the code or revisiting it after a while. Maybe in the future, as part of a refactoring or something, it would be more straight forward to have the decision in each of the requester functions. Just a thought.

jyao1: that is right. I only do necessary change to make it work in Q3, due to limited time.
We may do refactoring at https://github.com/DMTF/libspdm/issues/1213 in Q4.

richkong88: Sounds good.

jyao1:
Thank you! @richkong88 .

@steven-bellock , we need merge it soon to perform test in spdm-emu and spdm-dump to see if there is any other gap.
Feel free to comment or submit issue, and we will follow up.

PR #1231, opened 2022-09-23
Document send and receive buffer requirements

### LINKED ISSUE #1180: Document libspdm's expectations / requirements for the send and receive buffers
In particular : 
- If libspdm never writes to the receive buffer then it should be ok for it to be a read-only buffer. I'm not sure if there are any write-only buffers out there.
- libspdm does not copy the receive buffer to a more secure region of memory. So if the receive buffer could possibly be written to by an attacker while libspdm is processing the message then it's up to the integrator to copy the data to a more secure region before handing it off to libspdm.

steven-bellock:
Fixes #1180.


steven-bellock:
@jyao1 Please make sure these statements are correct.

jyao1 (approved):

PR #1236, opened 2022-09-26
Fix Building with no RSA Support

mct-sparky:
With both LIBSPDM_RSA_SSA_SUPPORT and LIBSPDM_RSA_PSS_SUPPORT switched off, any file which references the cryptlib_rsa.h header file will have an undefined reference.

Fix by switching out all functions which depend on RSA support

### DIFF
--- a/os_stub/cryptlib_mbedtls/pem/pem.c
+++ b/os_stub/cryptlib_mbedtls/pem/pem.c
@@ -31,6 +31,7 @@ static size_t ascii_str_len(const char *string)
     return length;
 }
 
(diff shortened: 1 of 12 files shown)


steven-bellock:
Thanks. I guess I checked with one on and one off, but didn't check with both of them off. We'll want to add disablement of algorithms to CI/CD, probably in the case where all the capabilities have been disabled.

mct-sparky:
> Thanks. I guess I checked with one on and one off, but didn't check with both of them off. We'll want to add disablement of algorithms to CI/CD, probably in the case where all the capabilities have been disabled.

No problem, I've checked building with all algorithm support off and there are a number of compilation issues, mostly regarding unit tests.

Issue #1237, opened 2022-09-27
SPDM emu fails when RSP trans_size is 42 and REQ trans_size is 4096


jyao1:
The root cause is https://github.com/DMTF/libspdm/blob/main/library/spdm_requester_lib/libspdm_req_send_receive.c#L465

when responder returns chunk_send_ack with valid response, the send_info->large_message_size is override by response in chunk_send_ack. It is invalid to use send_info->chunk_bytes_transferred to check.

PR #1238, opened 2022-09-27
Fix the chunk issue.

jyao1:
Fix: https://github.com/DMTF/libspdm/issues/1237

When req trans size is 4096 and rsp trans size is 42, req fails.

The root cause is that send_info->large_message_size is override by response in chunk_send_ack. It is invalid to use send_info->chunk_bytes_transferred to check.


jyao1:
> With the way the loop is structured you can get rid of stop_chunk_send and just use break. 

Right. Updated.

PR #1240, opened 2022-09-27
Remove minor compiler warnings

steven-bellock:

### DIFF
--- a/library/spdm_crypt_lib/libspdm_crypt_crypt.c
+++ b/library/spdm_crypt_lib/libspdm_crypt_crypt.c
@@ -1740,17 +1740,17 @@ typedef struct {
     size_t zero_pad_size;
 } libspdm_signing_context_str_t;
 
-libspdm_signing_context_str_t m_libspdm_signing_context_str_table[]={
-    {false, SPDM_CHALLENGE_AUTH, SPDM_CHALLENGE_AUTH_SIGN_CONTEXT,
+const libspdm_signing_context_str_t m_libspdm_signing_context_str_table[]={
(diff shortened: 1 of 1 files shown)


steven-bellock:
@ekohandel I think these warnings should have been caught by your compiler flag changes in #859. Not sure why they weren't reported.

library/spdm_crypt_lib/libspdm_crypt_crypt.c
```diff
@@ -3336,10 +3308,6 @@ bool libspdm_req_asym_sign_hash(
         /* Need use SPDM 1.2 signing*/
 
         switch (req_base_asym_alg) {
-        case SPDM_ALGORITHMS_BASE_ASYM_ALGO_TPM_ALG_SM2_ECC_SM2_P256:
-            param = "";
```
jyao1, line 3340: @steven-bellock , have you verified the different between NULL and "" in SM2?

steven-bellock: `param` would map to the `id_a` parameter in that function. But I see what you mean. Let me try this a different way.

steven-bellock:
Will start a new pull request.

ekohandel:
> @ekohandel I think these warnings should have been caught by your compiler flag changes in #859. Not sure why they weren't reported.

@steven-bellock if I understand correctly the change here is adding a `const` qualifier for the `context` which is a string. This is then used as a parameter to `libspdm_asym_verify_func` which has that parameter defined as a `const uint8_t *` type.

The change in #859 only generates a warning if a qualifier is removed but in this case a qualifier is added. The intention here would be to catch trying to mutate memory that is immutable but it would be ok to mark mutable memory as immutable.

Issue #1271, opened 2022-10-10
Add ARM CLANG tool chain


jyao1:
Currently, we only have ARM-GCC tool chain.

We need add ARM-CLANG (https://developer.arm.com/documentation/100067/0612/armclang-Command-line-Options)

PR #1280, opened 2022-10-12
Enable armclang tool chain for windows

Wenxing-hou:
Fix: #1271 

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@ MESSAGE("## Build Configuration ##")
 MESSAGE("#########################")
 
 SET(ARCH ${ARCH} CACHE STRING "Choose the arch of build: ia32 x64 arm aarch64 riscv32 riscv64 arc" FORCE)
-SET(TOOLCHAIN ${TOOLCHAIN} CACHE STRING "Choose the toolchain of build: Windows: VS2015 VS2019 VS2022 CLANG LIBFUZZER Linux: GCC ARM_GCC AARCH64_GCC RISCV32_GCC RISCV64_GCC ARC_GCC CLANG CBMC AFL KLEE LIBFUZZER" FORCE)
+SET(TOOLCHAIN ${TOOLCHAIN} CACHE STRING "Choose the toolchain of build: Windows: VS2015 VS2019 VS2022 CLANG LIBFUZZER ARM_CLANG Linux: GCC ARM_GCC AARCH64_GCC RISCV32_GCC RISCV64_GCC ARC_GCC CLANG CBMC AFL KLEE LIBFUZZER" FORCE)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -553,6 +555,32 @@ elseif(CMAKE_SYSTEM_NAME MATCHES "Windows")
 
         SET(CMAKE_C_FLAGS_RELEASE     "/GL /O1b2s")
         SET(CMAKE_C_FLAGS_DEBUG       "/Od")
+
+    elseif(TOOLCHAIN STREQUAL "ARM_CLANG")
+        set(TOOLCHAIN_PATH C:/ArmStudio/sw/ARMCompiler6.18)
```
jyao1, line 560: Please don't hardcode it.

I have installed Development Studio 2022.1, then it is at:
C:\Program Files\Arm\Development Studio 2022.1\sw\ARMCompiler6.18

I dont check we need TOOLCHAIN_PATH at all. We can require the user open command prompt with ARMCLANG in the PATH.

CMakeLists.txt
```diff
@@ -553,6 +555,32 @@ elseif(CMAKE_SYSTEM_NAME MATCHES "Windows")
...
+        set(CMAKE_ASM_COMPILER ${ARM_ASM} CACHE STRING "")
+        set(CMAKE_LINKER ${ARM_LINK} CACHE STRING "")
+
+        set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
+        set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
+        set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
+        set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
+        set(CMAKE_FIND_ROOT_PATH ${TOOLCHAIN_PATH})
+
+        ADD_COMPILE_OPTIONS(--target=arm-arm-none-eabi)
```
jyao1, line 579: This is 32bit ARM.
I think we also need support AARCH64, which will use "--target=aarch64-arm-none-eabi"

unit_test/test_size/intrinsiclib/CMakeLists.txt
```diff
@@ -18,7 +15,7 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
         memory_intrinsics.c        
     )
 elseif(CMAKE_SYSTEM_NAME MATCHES "Windows")
-    if(ARCH STREQUAL "x64")
+    if(ARCH STREQUAL "x64" OR ARCH STREQUAL "arm")
```
jyao1, line 18: I think we need "aarch64" as well.

PR #1284, opened 2022-10-13
Update crt_support.h

### LINKED ISSUE #1267: openssllib - include folder' crt_support.h
hi
you need to change params in line 323:
char *strncpy(char *destinin, const char *source, int maxlen);
the last parameter should be size_t instead of int.

you can see definition:
https://cplusplus.com/reference/cstring/strncpy/

not all compilers support int as size_t.

moshe8931:
Changed declaration of char *strncpy to use size_t instead of int. https://cplusplus.com/reference/cstring/strncpy/

Fixes #1267
Signed-off-by: Moshe Hershkovitz <mhershkovitz@marvell.com>


jyao1:
@moshe8931, please fix DCO

`Signed-off-by: Moshe Hershkovitz <mhershkovitz@marvell.com>`

PR #1294, opened 2022-10-14
Move functions from public common_lib to internal

### LINKED ISSUE #1292: Many integrator-visible functions in common_lib can be moved to the internal header
For example integrator doesn't need to call `libspdm_calculate_th_hash_for_finish`.

steven-bellock:
The functions that have been moved should not be visible to the integrator.

Fixes #1292.

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -845,4 +845,410 @@ uint8_t libspdm_get_cert_slot_count(libspdm_context_t *spdm_context);
 void libspdm_append_msg_log(void *context, void *message, size_t message_size);
 #endif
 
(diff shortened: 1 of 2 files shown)


jyao1:
> Please confirm that the functions I moved aren't needed by the integrator.

@steven-bellock , would you please try spdm-emu, spdm-dump, and SPDM-Responder-Validator?
If you can pass build for above 3, then they are not needed.

steven-bellock:
This pull request is fine, but a previous pull request I did broke SPDM-Responder-Validator due to making some functions `static`. I'll bring this pull request in and then fix the `static` issues in the SPDM-Responder-Validator and libspdm.

PR #1300, opened 2022-10-17
Fix build break in SPDM-Responder-Validator

steven-bellock:
#1281 moved functions from internal common_lib to the Requester / Responder .c files and made them static. However the SPDM-Responder-Validator needs to access some of those functions and #1281 caused a build break.

This pull request moves function declarations that are needed by SPDM-Responder-Validator to the internal requester_lib and responder_lib libraries. The function definitions remain in the requester_lib and responder_lib .c files.


steven-bellock:
Merging now since this fixes a build break.

PR #1309, opened 2022-10-27
Update the unit tests of spdm_responder_version.

Zhiqiang520:
Reset the connection state if the GET_VERSION request is valid. 
Not reset the connection state if the GET_VERSION request is invalid.


unit_test/test_spdm_responder/version.c
```diff
@@ -223,6 +230,8 @@ void libspdm_test_responder_version_case6(void **state)
     assert_int_equal(spdm_response->header.param1,
```
steven-bellock, line 230: This should probably be a `VersionMismatch` error response, but I'll file a separate issue for that.

steven-bellock (approved):

PR #1315, opened 2022-10-29
fix build error when some cap disabled.

jyao1:


steven-bellock:
It would be good to sweep through all the capability combinations every week or so, just to make sure it builds. There are 256 combinations so far.

jyao1:
yeah, you may file a new issue on that.

jyao1:
I am thinking to support some typical use cases, to CI.
1. SPDM 1.0 only (no session) device
2. SPDM 1.1 only (no chunk, provisioning) device with KEY_EX, but no mutual auth, no encapsulated message.
3. SPDM 1.1 only (no chunk, provisioning) device with PSK

Just need build in one configuration.

Issue #1327, opened 2022-11-01
One typo in spdm_return_status.h


qizhangz:
spdm_return_status.h
#define LIBSPSM_SOURCE_TRANSPORT 0x04

should be LIBSPDM_ not LIBSPSM_

PR #1330, opened 2022-11-01
Fix typo error

Wenxing-hou:
Fix: #1327 


include/library/spdm_return_status.h
```diff
@@ -42,7 +42,7 @@ typedef uint32_t libspdm_return_t;
 #define LIBSPDM_SOURCE_CORE 0x01
 #define LIBSPDM_SOURCE_CRYPTO 0x02
 #define LIBSPDM_SOURCE_CERT_PARSE 0x03
-#define LIBSPSM_SOURCE_TRANSPORT 0x04
+#define LIBSPDM_SOURCE_TRANSPORT 0x04
```
steven-bellock, line 45: You also need to keep the `#define` for the misspelled macro. We can get rid of it in a future release.

steven-bellock (changes requested):

include/library/spdm_return_status.h
Wenxing-hou: Thanks. I have made it back and added comment.

steven-bellock (approved):

PR #1345, opened 2022-11-03
Add test_crypt in CI

Wenxing-hou:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -149,6 +149,11 @@ jobs:
         run: |
           cd build/bin
           ./test_spdm_crypt
+      - name: Test_crypt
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -149,6 +149,11 @@ jobs:
         run: |
           cd build/bin
           ./test_spdm_crypt
+      - name: Test_crypt
```
steven-bellock, line 152: `test_crypt` always returns success to CI/CD.
```
int main(void)
{
    libspdm_cryptest_main();
    return 0;
}
```

Wenxing-hou: I have submitted a new PR #1350 to fix it.

PR #1346, opened 2022-11-03
Fix dh_compute_key copy mem overlap

Wenxing-hou:

### DIFF
--- a/os_stub/cryptlib_mbedtls/pk/dh.c
+++ b/os_stub/cryptlib_mbedtls/pk/dh.c
@@ -13,6 +13,7 @@
 #include "internal_crypt_lib.h"
 #include <mbedtls/dhm.h>
 #include <mbedtls/bignum.h>
+#include <string.h>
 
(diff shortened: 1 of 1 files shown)


os_stub/cryptlib_mbedtls/pk/dh.c
```diff
@@ -322,10 +322,24 @@ bool libspdm_dh_compute_key(void *dh_context, const uint8_t *peer_public_key,
...
+            return false;
+        }
+
+        libspdm_copy_mem(store_actual_key + dh_key_size - return_size,
+                         return_size,
                          key, return_size);
-        libspdm_zero_mem(key, dh_key_size - return_size);
+        libspdm_zero_mem(store_actual_key, dh_key_size - return_size);
+        libspdm_copy_mem(key, dh_key_size, store_actual_key, dh_key_size);
+
```
jyao1, line 340: Why not just use memmove() ?

PR #1350, opened 2022-11-04
Add test_crypt and test_spdm_crypt return for CI

Wenxing-hou:


jyao1 (approved):

steven-bellock (approved):
We can merge this pull request. But more work needs to be done on the individual tests that always return success to CI/CD. For example : https://github.com/DMTF/libspdm/blob/beee7bd85a4f0961a5773aefc4dcab0408d538d1/unit_test/test_crypt/hash_verify.c#L420-L439

PR #1361, opened 2022-11-05
Enhance and refactor debugging

### LINKED ISSUE #1351: Debug enhancements and refactoring
Follow-up to #1050.

- Define (at least) three new macros
  - `LIBSPDM_DEBUG_ASSERT_ENABLE`
  - `LIBSPDM_DEBUG_PRINT_ENABLE`
  - `LIBSPDM_DEBUG_ENABLE` - enables the previous two macros (and possibly others) even if they are explicitly defined and disabled. This has the same behavior as the current `MDEPKG_NDEBUG` macro.
- Produce a warning during compilation if `MDEPKG_NDEBUG` is defined that it will be removed in a future release.
- Remove the `LIBSPDM_DEBUG_VERBOSE` macro.
- The `libspdm_internal_dump_*` functions should have macro wrappers so that they can be removed if `LIBSPDM_DEBUG_PRINT_ENABLE` is disabled, similar to `LIBSPDM_DEBUG`.
- Add `extern` to the declarations for `libspdm_debug_print` and `libspdm_debug_assert`.

### LINKED ISSUE #1352: LIBSPDM_DEBUG_ERROR prints are not printing
See https://github.com/DMTF/libspdm/blob/beee7bd85a4f0961a5773aefc4dcab0408d538d1/os_stub/debuglib/debuglib.c#L31

Should be `(LIBSPDM_DEBUG_INFO | LIBSPDM_DEBUG_ERROR)`

steven-bellock:
Fixes #1351.
Fixes #1352.


steven-bellock:
This will break the other repositories so I'll obviously have follow-up pull requests for them.

jyao1:
Please resolve conflict.

jyao1:
Question: Why to create include/internal/hal/debuglib_internal.h?

If the external function uses the MACRO, then it should be in public, not internal.

steven-bellock:
> Question: Why to create include/internal/hal/debuglib_internal.h?
> 
> If the external function uses the MACRO, then it should be in public, not internal.

`libspdm_debug_print` only needs the `LIBSPDM_DEBUG_INFO` and `LIBSPDM_DEBUG_ERROR` macros. `debuglib_internal.h` is for the macros that libspdm needs, `LIBSPDM_DEBUG`, `LIBSPDM_ASSERT`, etc.

jyao1:
> > Question: Why to create include/internal/hal/debuglib_internal.h?
> > If the external function uses the MACRO, then it should be in public, not internal.
> 
> `libspdm_debug_print` only needs the `LIBSPDM_DEBUG_INFO` and `LIBSPDM_DEBUG_ERROR` macros. `debuglib_internal.h` is for the macros that libspdm needs, `LIBSPDM_DEBUG`, `LIBSPDM_ASSERT`, etc.

I think the integrator need `LIBSPDM_DEBUG`, `LIBSPDM_ASSERT`, right?
That is why I think those should be in public.

jyao1:
I feel it is unnecessary to create include/internal/hal/debuglib_internal.h. I suggest to keep one include hal/library/debuglib.h, because the libspdm integrator/consumer need use hal/library/debuglib.h. However, the integrator should not use include/internal/hal/debuglib_internal.h.


I am OK on all other changes.

include/internal/libspdm_lib_config.h
```diff
@@ -13,4 +13,55 @@
...
+#define LIBSPDM_DEBUG_PRINT_ENABLE (LIBSPDM_DEBUG_ENABLE)
+#define LIBSPDM_DEBUG_BLOCK_ENABLE (LIBSPDM_DEBUG_ENABLE)
+#elif defined(MDEPKG_NDEBUG)
+#undef LIBSPDM_DEBUG_ASSERT_ENABLE
+#undef LIBSPDM_DEBUG_PRINT_ENABLE
+#undef LIBSPDM_DEBUG_BLOCK_ENABLE
+
+#define LIBSPDM_DEBUG_ASSERT_ENABLE 0
+#define LIBSPDM_DEBUG_PRINT_ENABLE 0
+#define LIBSPDM_DEBUG_BLOCK_ENABLE 0
```
taprinz, line 64: Should we maybe explicitly print out in the make process that debug config is enabled or disabled? Especially if we're potentially overriding user defines here

steven-bellock:
Ok, I'll get rid of `internal/hal/debuglib_internal.h` but I think the `LIBSPDM_DEBUG` / `LIBSPDM_ASSERT` macros would need to come through `library/spdm_common_lib.h`.

include/internal/libspdm_lib_config.h
steven-bellock: I'll add more documentation to `spdm_lib_config.h`. When CMake is set to `DTARGET=Release` then all debug will be disabled, regardless of the value of `LIBSPDM_DEBUG_ENABLE` in `spdm_lib_config.h`.

steven-bellock:
Looks like it can't be merged. I'll just close this pull request and start a fresh one.

PR #1367, opened 2022-11-06
Summarize the arch and compiler.

jyao1:
This only includes popular X86/ARM/RISC-V.

Some other targets (LIBFUZZ, AFL, ARC, NIOS-II, etc) are not covered.


README.md
```diff
@@ -55,6 +55,28 @@
 
    Support to be included in [OpenBMC](https://github.com/openbmc). It is in planning, see [SPDM Integration](https://www.youtube.com/watch?v=PmgXkLJYI-E).
 
+7) Supported Architecture and Compiler
```
steven-bellock, line 58: Need to specify that these tables are for cross-compilation on x86 host.

jyao1: Yes, please let me know if the update is OK.

Or, please suggest the text directly.

steven-bellock (approved):

Issue #1360, opened 2022-11-05
Add AARCH64 arch build to CI


jyao1:
Current CI toolchains are all X86 arch. We need add AARCH64 arch, such as `ARM_GNU` or `AARCH64_GCC` toolchain.

Only one configuration is enough.

Xiaohanjlll:
I will check this

PR #1383, opened 2022-11-10
Add ARM_GNU arch build to CI 

Xiaohanjlll:
Fix: #1360 


.github/workflows/build.yml
```diff
@@ -30,6 +30,7 @@ jobs:
         arch:
           - x64
           - ia32
+          - arm
```
jyao1, line 33: I suggest to use aarch64

Issue #1311, opened 2022-10-28
Allow Integrator to specify location of the secured message context


steven-bellock:
Currently the secured message context is appended to the SPDM context. However as the secured message context contains things like keys, an Integrator may want to separate it from the SPDM context. For example the SPDM context may reside in pageable memory while the secured message context resides in unpageable / tightly coupled memory.

steven-bellock:
This was fixed by #1400.

Issue #1273, opened 2022-10-10
Delete Doxygen comments from .c files where appropriate


steven-bellock:
Most of the function definitions in the codebase contain Doxygen comments that are identical to the ones found in the header. These should be deleted as IDEs know to look at the function declaration in the header. For those functions that don't have a declaration, ie `static` functions, the Doxygen comments can remain.

steven-bellock:
This is also causing warnings in Doxygen. For example
>warning: return value 'false' of libspdm_asym_sign_hash has multiple documentation sections

PR #1400, opened 2022-11-16
Allow user to specify the location of secured message context

taprinz:
Implements #1311. The motivation behind this is that the secured message context(s) hold key and key-related material, which some integrators may want to store separately in sensitive data storage. A split is desired, since this storage is often limited - we would like to avoid having to support the entire libspdm_context_t struct which can be quite large.

To do this, introduces new APIs which allows the user to allocated SPDM context and secured message contexts separately. Decided to add new APIs, rather than update the existing ones, to avoid extensive breaking change both internally and externally.

Existing APIs are now wrappers around the new APIs.

Updated documentation to alert users of this API.

Adds test as well, just to double check this code does what it says it does.

### DIFF
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -246,11 +246,12 @@ void libspdm_set_last_spdm_error_struct(void *spdm_context,
  *                               the location of a secured message context.
  * @param  num_secured_contexts  Number of secured message contexts to initialize.
  *                               Currently, only LIBSPDM_MAX_SESSION_COUNT is supported.
+ *                               In future releases, lesser values may be supported.
  *
  * @retval RETURN_SUCCESS        Contexts are initialized.
  * @retval RETURN_DEVICE_ERROR   Context initialization failed.
  */
-libspdm_return_t libspdm_init_context_with_secured_context(void *context,
+libspdm_return_t libspdm_init_context_with_secured_context(void *spdm_context,
                                                            void **secured_contexts,
                                                            size_t num_secured_contexts);
 
@@ -266,7 +267,7 @@ libspdm_return_t libspdm_init_context_with_secured_context(void *context,
  * @retval RETURN_SUCCESS       context is initialized.
  * @retval RETURN_DEVICE_ERROR  context initialization failed.
  */
-libspdm_return_t libspdm_init_context(void *context);
+libspdm_return_t libspdm_init_context(void *spdm_context);
 
 /**
  * Reset an SPDM context.
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -232,16 +232,38 @@ void libspdm_set_last_spdm_error_struct(void *spdm_context,
                                         libspdm_error_struct_t *last_spdm_error);
 
 /**
- * Initialize an SPDM context.
+ * Initialize an SPDM context, as well as all secured message contexts,
+ * in the specified locations.
  *
- * The size in bytes of the spdm_context can be returned by libspdm_get_context_size.
+ * The size in bytes of the spdm_context can be returned by
+ * libspdm_get_context_size_without_secured_context.
+ * 
+ * The size in bytes of the secured message context can be returned by
+ * libspdm_secured_message_get_total_context_size.
+ * 
+ * @param  spdm_context              A pointer to the SPDM context.
+ * @param  secured_message_contexts  A pointer to the secured message context buffer.
+ *                                   This will hold all secured message contexts.
  *
- * @param  spdm_context  A pointer to the SPDM context.
+ * @retval RETURN_SUCCESS       Contexts are initialized.
+ * @retval RETURN_DEVICE_ERROR  Context initialization failed.
+ */
+libspdm_return_t libspdm_init_context_with_secure_data_location(void *context,
+    void *secured_message_contexts);
+
+/**
+ * Initialize an SPDM context, as well as secured message contexts.
+ * The secured message contexts are appended to the context structure.
+ * 
+ * The total size in bytes of the spdm_context and all secured message
+ * contexts can be returned by libspdm_get_context_size().
+ *
+ * @param  spdm_context         A pointer to the SPDM context.
  *
  * @retval RETURN_SUCCESS       context is initialized.
  * @retval RETURN_DEVICE_ERROR  context initialization failed.
  */
-libspdm_return_t libspdm_init_context(void *spdm_context);
+libspdm_return_t libspdm_init_context(void *context);
 
 /**
  * Reset an SPDM context.
@@ -253,9 +275,21 @@ libspdm_return_t libspdm_init_context(void *spdm_context);
 void libspdm_reset_context(void *spdm_context);
 
 /**
- * Return the size in bytes of the SPDM context.
+ * Return the size in bytes of just the SPDM context, without secured message context.
+ * 
+ * For the complete context size, use libspdm_get_context_size.
  *
  * @return the size in bytes of the SPDM context.
+ **/
+size_t libspdm_get_context_size_without_secured_context(void);
+
+/**
+ * Return the size in bytes of the SPDM context. This includes all secured message
+ * context data as well.
+ * 
+ * For just the SPDM context size, use libspdm_get_context_size_without_secured_context.
+ *
+ * @return the size in bytes of the SPDM context and secured message contexts.
  **/
 size_t libspdm_get_context_size(void);
 
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -238,18 +238,21 @@ void libspdm_set_last_spdm_error_struct(void *spdm_context,
  * The size in bytes of the spdm_context can be returned by
  * libspdm_get_context_size_without_secured_context.
  *
- * The size in bytes of the secured message context can be returned by
- * libspdm_secured_message_get_total_context_size.
+ * The size in bytes of a single secured message context can be returned by
+ * libspdm_secured_message_get_context_size.
  *
- * @param  spdm_context              A pointer to the SPDM context.
- * @param  secured_message_contexts  A pointer to the secured message context buffer.
- *                                   This will hold all secured message contexts.
+ * @param  spdm_context          A pointer to the SPDM context.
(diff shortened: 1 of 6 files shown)


include/library/spdm_common_lib.h
```diff
@@ -232,16 +232,38 @@ void libspdm_set_last_spdm_error_struct(void *spdm_context,
...
+ * 
+ * @param  spdm_context              A pointer to the SPDM context.
+ * @param  secured_message_contexts  A pointer to the secured message context buffer.
+ *                                   This will hold all secured message contexts.
+ *
+ * @retval RETURN_SUCCESS       Contexts are initialized.
+ * @retval RETURN_DEVICE_ERROR  Context initialization failed.
+ */
+libspdm_return_t libspdm_init_context_with_secure_data_location(void *context,
+    void *secured_message_contexts);
```
taprinz, line 252: Open to a name change (for parameters and APIs) if we think secured_message_contexts exposes too much information about internal implementation.

taprinz:
Another approach considered was to simply add an API or a case in libspdm_set_data() that would allow the user to move/update the location of the secured message contexts after the SPDM context has been initialized.

I have a copy of this approach implemented as well - however, this was not preferred, as it would still require the user to allocate space for the secured message contexts appended to the SPDM contexts that would go unused.

taprinz:
Yeah, I think there's quite a bit more flexibility we can include with the context initialization, especially since so much of it is abstracted away from the user again.

With the above, I'm not sure whether the `libspdm_init_context2()` use case is necessary. We could either swallow it into `libspdm_init_context1()` (but then break the API), or expect the user to limit the amount of sessions supported via LIBSPDM_MAX_SESSION_COUNT at compile time, rather than runtime.

I see the reasoning for it, mostly I'm unsure how to decide between how many interfaces we want to support proactively versus support when requested.

jyao1:
I assume this patch will maintain the compatibility - the spdm_emu does not need to change with exiting APIs. Am I right?

I can understand why we have `libspdm_init_context_with_secure_data_location()` and `libspdm_get_context_size_without_secured_context()`. I suggest to make them consistent: change `libspdm_init_context_with_secure_data_location()` to `libspdm_init_context_with_secured_context()`.

I dont understand why we need `libspdm_secured_message_get_total_context_size()` . Why not use `libspdm_secured_message_get_context_size() * LIBSPDM_MAX_SESSION_COUNT` ? I remember libspdm code has assumption on where the next secure_context is. How does it work if `libspdm_secured_message_get_total_context_size() != libspdm_secured_message_get_context_size() * LIBSPDM_MAX_SESSION_COUNT` ?

include/library/spdm_common_lib.h
jyao1: I suggest `libspdm_init_context_with_secured_context`. Detail feedback at https://github.com/DMTF/libspdm/pull/1400#issuecomment-1321018164

taprinz:
Updated according to review comments. Updated naming for better consistency. Changed allocation from complete block allocation to array, to allow for different locations for each secured message context.

Included number of contexts parameter to future-proof the API a bit - this way, we can change so that the API supports a smaller number of contexts than LIBSPDM_MAX_SESSION_COUNT without breakages. Since no use case for now, we only support one value.

include/library/spdm_common_lib.h
```diff
@@ -232,16 +232,41 @@ void libspdm_set_last_spdm_error_struct(void *spdm_context,
...
+ * @param  spdm_context          A pointer to the SPDM context.
+ * @param  secured_contexts      An array of pointers, with each entry containing
+ *                               the location of a secured message context.
+ * @param  num_secured_contexts  Number of secured message contexts to initialize.
+ *                               Currently, only LIBSPDM_MAX_SESSION_COUNT is supported.
+ *
+ * @retval RETURN_SUCCESS        Contexts are initialized.
+ * @retval RETURN_DEVICE_ERROR   Context initialization failed.
+ */
+libspdm_return_t libspdm_init_context_with_secured_context(void *context,
```
steven-bellock, line 253: `context` -> `spdm_context`

include/library/spdm_common_lib.h
```diff
@@ -232,16 +232,41 @@ void libspdm_set_last_spdm_error_struct(void *spdm_context,
...
+ * The total size in bytes of the spdm_context and all secured message
+ * contexts can be returned by libspdm_get_context_size().
+ *
+ * @param  spdm_context         A pointer to the SPDM context.
  *
  * @retval RETURN_SUCCESS       context is initialized.
  * @retval RETURN_DEVICE_ERROR  context initialization failed.
  */
-libspdm_return_t libspdm_init_context(void *spdm_context);
+libspdm_return_t libspdm_init_context(void *context);
```
steven-bellock, line 269: `context` -> `spdm_context`

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -2046,22 +2046,37 @@ void libspdm_set_last_spdm_error_struct(void *context, libspdm_error_struct_t *l
 }
 
 /**
- * Initialize an SPDM context.
+ * Initialize an SPDM context, as well as all secured message contexts,
```
steven-bellock, line 2049: You can delete all the Doxygen comments in .c files if they're present in a header. See #1273.

include/library/spdm_common_lib.h
```diff
@@ -232,16 +232,41 @@ void libspdm_set_last_spdm_error_struct(void *spdm_context,
...
  *
- * @param  spdm_context  A pointer to the SPDM context.
+ * The size in bytes of a single secured message context can be returned by
+ * libspdm_secured_message_get_context_size.
+ *
+ * @param  spdm_context          A pointer to the SPDM context.
+ * @param  secured_contexts      An array of pointers, with each entry containing
+ *                               the location of a secured message context.
+ * @param  num_secured_contexts  Number of secured message contexts to initialize.
+ *                               Currently, only LIBSPDM_MAX_SESSION_COUNT is supported.
```
steven-bellock, line 248: Also state that `num_secured_contexts` may be less than `LIBSPDM_MAX_SESSION_COUNT` in a future release.

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -2046,22 +2046,37 @@ void libspdm_set_last_spdm_error_struct(void *context, libspdm_error_struct_t *l
...
+libspdm_return_t libspdm_init_context_with_secured_context(void *context,
+                                                           void **secured_contexts,
+                                                           size_t num_secured_contexts)
 {
     libspdm_context_t *spdm_context;
-    void *secured_message_context;
-    size_t SecuredMessageContextSize;
     size_t index;
 
+    if (context == NULL || secured_contexts == NULL ||
```
steven-bellock, line 2074: These should be `LIBSPDM_ASSERT`s.

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -2115,6 +2130,48 @@ libspdm_return_t libspdm_init_context(void *context)
...
+ */
+libspdm_return_t libspdm_init_context(void *context)
+{
+    libspdm_context_t *spdm_context;
+    void *secured_context;
+    void *secured_contexts[LIBSPDM_MAX_SESSION_COUNT];
+    size_t secured_context_size;
+    size_t index;
+
+    if (context == NULL)
```
steven-bellock, line 2153: This should be a `LIBSPDM_ASSERT`.

steven-bellock (changes requested):

library/spdm_common_lib/libspdm_com_context_data.c
taprinz: Thought to keep as runtime check was because this is input from the integrator, rather than the library itself. In the library itself, we can control to ensure with proper configuration, this is never called with invalid args. We can't necessarily guarantee that from integrator.

If we move to `LIBSPDM_ASSERT`, I should clarify in comments the expectations.

include/library/spdm_common_lib.h
taprinz: I noticed that in a lot of these, the .h file has `void *spdm_context`, but the equivalent .c function has `void *context`, usually done as to avoid a naming conflict with a `libspdm_context_t spdm_context` variable.

Should we keep the current split, or clean this up for consistency between .c/.h files?

steven-bellock: The main requirement is that for publicly visible functions in the header the name is `spdm_context` so that it's not to be confused with any other context out there, and it's `void *` so the type is opaque. In the implementation we can do whatever is convenient. @jyao1 may have an opinion on the implementation side.

library/spdm_common_lib/libspdm_com_context_data.c
steven-bellock: The asserts allow an Integrator who knows what they're doing to not have to pay a performance or code size penalty. If an Integrator is unsure of their use of the API then they can enable asserts in debug or release.

taprinz: Fair enough, will change

jyao1 (approved):

include/library/spdm_common_lib.h
taprinz: Updated

include/library/spdm_common_lib.h
taprinz: Updated as requested

library/spdm_common_lib/libspdm_com_context_data.c
taprinz: Discussed, will leave Doxygen comments in .c files for widespread cleanup.

include/library/spdm_common_lib.h
taprinz: Added

library/spdm_common_lib/libspdm_com_context_data.c
taprinz: Updated

library/spdm_common_lib/libspdm_com_context_data.c
taprinz: Updated

steven-bellock (approved):

Issue #1402, opened 2022-11-17
libspdm_req_key_exchange.c memory leak with the LIBFUZZER generated specific mutation seed


Xiaohanjlll:
1. Verfied on : 37e58c70fb213c73d2572b1846dbeb87133b79d4
Ubuntu 20.04.4 LTS

> cd libspdm
cd libspdm
mkdir build_libfuzzer
cd build_libfuzzer
cmake -DARCH=x64 -DTOOLCHAIN=LIBFUZZER -DTARGET=Debug -DCRYPTO=openssl..
make copy_sample_key
make
cd bin
./test_spdm_requester_key_exchange ./../../test_spdm_requester_key_exchange_leak 

`LIBSPDM_ASSERT: /home/test/libspdm_XiaoHan/unit_test/fuzzing/spdm_unit_fuzzing_common/common.c(19): !m_send_receive_buffer_acquired`

2.  the LIBFUZZER generated specific mutation seed files
[test_spdm_requester_key_exchange_leak.zip](https://github.com/DMTF/libspdm/files/10028448/test_spdm_requester_key_exchange_leak.zip)

Xiaohanjlll:
I will check this

PR #1403, opened 2022-11-17
Fix the memory leak that occurred in req_key_exchange 

Xiaohanjlll:
Fix : #1402 


library/spdm_requester_lib/libspdm_req_key_exchange.c
```diff
@@ -491,7 +491,10 @@ static libspdm_return_t libspdm_try_send_receive_key_exchange(
                 spdm_context, true,
                 SPDM_GET_CAPABILITIES_REQUEST_FLAGS_MUT_AUTH_CAP,
                 SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_MUT_AUTH_CAP)) {
-            return LIBSPDM_STATUS_INVALID_MSG_FIELD;
+            status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
+            libspdm_secured_message_dhe_free(
+                spdm_context->connection_info.algorithm.dhe_named_group, dhe_context);
+            goto receive_done;
```
jyao1, line 497: Can you follow other code to move "status = " just before "goto" ?

Issue #1312, opened 2022-10-28
KEY_UPDATE should have callbacks similar to session_state_callbacks


taprinz:
The existence of the libspdm_trigger_session_state_callback allows an integrator to take action whenever the SPDM session keys are derived, changed, or deleted (handshake, app phase, and end session).

This would allow an integrator to take action in securing, updating, or deleting these keys - especially if they are desired to be used in secure key storage.

To follow along with the above usage - libspdm should provide a similar set of callbacks for KEY_UPDATE functionality, whether a new set of states, or an extension of libspdm_trigger_session_state_callback.

For example, the following callbacks/states could be set:
- Update Key
- Update Finished
- Rollback Key

In the case of the integrator storing the SPDM keys in secure key storage, it could allow for the following behavior:
- Update Key callback tells integrator to put new updated key in secure storage, and backup previous key
- Update Finished callback tells integrator they can discard backup key
- Rollback Key callback tells integrator to replace key in secure storage with backup key.

Open to modifications on the above - but in general, would like some way for the integrator to be aware that the keys are being updated.

steven-bellock:
This issue was fixed by #1463.

PR #1408, opened 2022-11-17
Add KEY_UPDATE callbacks for SPDM Responder

taprinz:
Implements #1312.

Draft pull request for now for initial feedback on the approach itself, noting that I need to more thoroughly test this. Wanted to seek feedback on whether the approach was acceptable first.

This adds Responder-only callbacks for KEY_UPDATE functionality. This is useful for any integrator who wants/needs to know whenever the keys themselves are updated - for example, if they are loading keys into secure key storage before use.

Adds relevant states for each of these updates, including "create, commit, rollback", to alert the integrator when:
- a new key has been derived and backup of old saved
- new key committed, backup discarded
- backup restored, new key discarded. 

I attempted a few ways to implement this, including:
- Putting the callback in `libspdm_create_session_data_key ` and `libspdm_activate_update_session_data_key ` , issue is no access to requester-only functions or base spdm_context.
- Making the callback available to both Requester and Responder, did not match up with other callback convention.

It also gets rid of the recursion in `libspdm_decode_secured_message`, replacing it with retries whenever we attempt to call transport_decode_message and fail with specific error.


taprinz:
Despite my best efforts, the commit history is not as clean as I would like. Can leave this draft for initial review of approach, then port to new branch with clean history for final review.

include/library/spdm_return_status.h
```diff
@@ -128,9 +128,13 @@ typedef uint32_t libspdm_return_t;
...
     LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_ERROR, LIBSPDM_SOURCE_CORE, 0x000f)
 
+/* Re-triable decrypt error from peer - must rollback to backup keys. */
+#define LIBSPDM_STATUS_SESSION_MSG_ERROR_TRY_DISCARD_KEY_UPDATE \
+    LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_ERROR, LIBSPDM_SOURCE_CORE, 0x0010)
+
 /* Unable to acquire resource. */
 #define LIBSPDM_STATUS_ACQUIRE_FAIL \
-    LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_ERROR, LIBSPDM_SOURCE_CORE, 0x0010)
+    LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_ERROR, LIBSPDM_SOURCE_CORE, 0x0011)
```
jyao1, line 137: I dont think we can change this.

Please append new status definition after 0x0010.

taprinz:
closing and moving to #1463, which has clean commit history.

Issue #1366, opened 2022-11-06
need a simple version of memory allocation lib for test_size


jyao1:
Current version is https://github.com/DMTF/libspdm/tree/main/unit_test/test_size/malloclib_null. But it does not have any functionality.

We need a simple version of memory allocation lib for test_size, to make data more accurately.

Reference: https://github.com/tianocore/edk2/blob/master/CryptoPkg/Library/BaseCryptLib/SysCall/RuntimeMemAllocation.c

PR #1416, opened 2022-11-21
Implement simple memory allocation

Wenxing-hou:
Fix: #1366 

Test for the patch:

1. Use `malloc_null` lib for `test_crypt` in unit_test to ensure all enable memory feature is right;
2. Use `malloc_null` lib for `spdm-emu` to ensure feature right again and test the malloc used size is 46KB;


jyao1:
The code seems good. I think we need to name it as - malloclib_simple.

Issue #1413, opened 2022-11-20
investigate RISCV-XPACK tool chain


jyao1:
Currently, we support RISCV-GNU toolchain from https://github.com/riscv-collab/riscv-gnu-toolchain.

Need investigate how to enable https://github.com/xpack-dev-tools/riscv-none-elf-gcc-xpack, which is used by NIOSV, such as release - https://github.com/xpack-dev-tools/riscv-none-embed-gcc-xpack/releases/tag/v10.2.0-1.2.

reference:
https://www.intel.com/content/www/us/en/products/docs/programmable/niosv-opensource-tools-video.html  
https://www.intel.com/content/www/us/en/docs/programmable/726952/22-1-21-2-0/setting-up-open-source-tools.html  
https://www.intel.com/content/www/us/en/docs/programmable/683472/22-1/downloading-individual-executable-files.html  

Need enable test_size_* for size evaluation.

PR #1438, opened 2022-11-29
Enable RISCV_XPACK tool_chain for libspdm

Wenxing-hou:
Fix: #1413 

1. The riscv32 arch has passed all build. And the test_size image size is follow:
   image

2.  The riscv64 arch is not supported now. 


jyao1 (approved):

steven-bellock (approved):
My only comment is that `README.md` is getting too large, but that's a separate issue.

PR #1445, opened 2022-11-30
Move PRIVATE_KEY_USE_PEM Macro to config

Wenxing-hou:

### DIFF
--- a/include/library/spdm_lib_config.h
+++ b/include/library/spdm_lib_config.h
@@ -179,6 +179,11 @@
 #define LIBSPDM_SM3_256_SUPPORT 1
 #endif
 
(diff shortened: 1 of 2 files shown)


include/library/spdm_lib_config.h
```diff
@@ -179,6 +179,11 @@
 #define LIBSPDM_SM3_256_SUPPORT 1
 #endif
 
+/*Private key use mode: PEM or Raw data*/
+#ifndef LIBSPDM_PRIVATE_KEY_USE_PEM
+#define LIBSPDM_PRIVATE_KEY_USE_PEM 1
+#endif
```
jyao1, line 185: why we need move this ???

PR #1449, opened 2022-11-30
Enhance slot_id checking

### LINKED ISSUE #1425: Checking of KEY_EXCHANGE_RSP's SlotIDParam is incorrect for SPDM 1.2
And maybe 1.1. SPDM 1.2 specification says
>Bit [3:0]. Shall be SlotID. Slot number of the Requester certificate chain that shall be used for mutual authentication, if MutAuthRequested Bit 0 is set. The value in this field shall be between 0 and 7 inclusive, or 0xF if the public key of the Requester was provisioned to the Responder through other means. All other values Reserved.
For any other value of MutAuthRequested this field shall be set to 0 and ignored by the Requester.

while SPDM 1.1 says
>Bit[3:0] = SlotID. The slot number of the certificate chain of the Requester to be used for mutual authentication. The value in this field shall be between 0 and 7 inclusive, or 0xF if the public key of the Requester was provisioned to the Responder through other
means. All other values Reserved. If MutAuthRequested = 0x00 this field shall be set to 0 and ignored by the Requester.

libspdm currently implements the 1.1 check : https://github.com/DMTF/libspdm/blob/8a5f88b4425539a0b3b5aa22bdd8c91c89fd6068/library/spdm_requester_lib/libspdm_req_key_exchange.c#L494-L519 so it needs to be updated for 1.2. However I think the 1.1 wording is a bug in the specification and that the 1.2 wording should be in 1.1. I'll file an issue against the specification.

### LINKED ISSUE #1426: KEY_EXCHANGE_RSP's SlotIDParam is checked against reserved bits
The specification defines the upper four bits of `KEY_EXCHANGE_RSP.SlotIDParam` as Reserved. However libspdm assumes they are zero and checks them. See https://github.com/DMTF/libspdm/blob/8a5f88b4425539a0b3b5aa22bdd8c91c89fd6068/library/spdm_requester_lib/libspdm_req_key_exchange.c#L504-L512 The upper four bits need to be masked before checking the lower four bits.

steven-bellock:
Fixes #1425.
Fixes #1426.

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_key_exchange.c
+++ b/library/spdm_requester_lib/libspdm_req_key_exchange.c
@@ -501,10 +501,10 @@ static libspdm_return_t libspdm_try_send_receive_key_exchange(
         }
         if ((mut_auth_requested != SPDM_KEY_EXCHANGE_RESPONSE_MUT_AUTH_REQUESTED) &&
             (*req_slot_id_param != 0)) {
-                libspdm_secured_message_dhe_free(
-                    spdm_context->connection_info.algorithm.dhe_named_group, dhe_context);
-                status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
-                goto receive_done;
+            libspdm_secured_message_dhe_free(
+                spdm_context->connection_info.algorithm.dhe_named_group, dhe_context);
+            status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
+            goto receive_done;
         }
     } else if (*req_slot_id_param != 0) {
         libspdm_secured_message_dhe_free(
--- a/library/spdm_requester_lib/libspdm_req_key_exchange.c
+++ b/library/spdm_requester_lib/libspdm_req_key_exchange.c
@@ -286,6 +286,7 @@ static libspdm_return_t libspdm_try_send_receive_key_exchange(
     uint8_t *message;
     size_t message_size;
     size_t transport_header_size;
+    uint8_t mut_auth_requested;
 
     /* -=[Check Parameters Phase]=- */
     LIBSPDM_ASSERT((slot_id < SPDM_MAX_SLOT_COUNT) || (slot_id == 0xff));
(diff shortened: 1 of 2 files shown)


library/spdm_requester_lib/libspdm_req_key_exchange.c
```diff
@@ -468,8 +469,11 @@ static libspdm_return_t libspdm_try_send_receive_key_exchange(
     if (heartbeat_period != NULL) {
         *heartbeat_period = spdm_response->header.param1;
     }
-    *req_slot_id_param = spdm_response->req_slot_id_param;
-    if (spdm_response->mut_auth_requested != 0) {
+
+    *req_slot_id_param = spdm_response->req_slot_id_param & 0xf;
+    mut_auth_requested = spdm_response->mut_auth_requested & 0xf;
```
jyao1, line 474: Should it be `& 0x7` ? Only BIT0, BIT1, BIT2 are valid.

steven-bellock: Indeed. Good catch.

steven-bellock:
Not sure why it's complaining about DCO. I've signed-off each commit.
```
commit 0fcb0216d7a69b10a2bd388b8c99c447fa9acd9c (HEAD -> fix-1425, origin/fix-1425)
Author: Steven Bellock <sbellock@nvidia.com>
Date:   Thu Dec 1 07:42:41 2022 -0800

    Fix bitmask bug

    `mut_auth_requested` is three bits, not four.

commit 8f55050d73f4e4e47b692a0fb43eb583ade08b90
Author: Steven Bellock <sbellock@nvidia.com>
Date:   Wed Nov 30 15:12:08 2022 -0800

    Fix style compliance

commit 5afd771c12c41dff3b9ab60b184a2aa09e6e6f8e
Author: Steven Bellock <sbellock@nvidia.com>
Date:   Wed Nov 30 15:09:57 2022 -0800

    Enhance slot_id checking

    Fixes #1425.
    Fixes #1426.

```

jyao1 (approved):

Issue #1424, opened 2022-11-23
SessionID creation is backwards


steven-bellock:
The specification says
>The final session ID (SessionID) = Concatenate (ReqSessionID, RspSessionID).

where presumably `ReqSessionID` occupies the least-significant bytes and `RspSessionID` occupies the most-significant bytes. However in libspdm it is the opposite : https://github.com/DMTF/libspdm/blob/8a5f88b4425539a0b3b5aa22bdd8c91c89fd6068/library/spdm_requester_lib/libspdm_req_key_exchange.c#L530

steven-bellock:
Additionally the specification doesn't formally define `Concatenate()`, which it should. I'll file an issue against the specification.

steven-bellock:
Issue has been filed at https://github.com/DMTF/SPDM-WG/issues/2281

steven-bellock:
`(req_session_id << 16) | rsp_session_id` is in enough places (four) that it should be its own function. 
```
uint32_t libspdm_generate_session_id (uint16_t req_session_id, uint16_t rsp_session_id)
```

PR #1450, opened 2022-12-01
Fix session_id generate

Wenxing-hou:
Fix: #1424 
Make session_id generate right.


jyao1 (approved):

steven-bellock (approved):
I edited the text in the header file.

Issue #390, opened 2021-12-10
Remove is_requester parameter from publicly visible functions


steven-bellock:
The integrator should only have to specify, once, with a #define that they're building a Requester or a Responder. This would be accomplished by compiling libspdm_secured_message_lib and libspdm_common_lib directly into the Requester and Responder libraries.

jyao1:
This is only applicable when this is_requester parameter is used consistently by one entity.

If one entity need call is_requester=TRUE in one place and is_requester=FALSE is another place, then it is hard to remove.

I forget if we have such case, we may need carefully review all places.

steven-bellock:
I will provide a list of functions that would benefit from this. One solution would be add a new setting to the spdm context where, at the beginning of time, the integrator specifies whether the context is for a Responder or Requester.

steven-bellock:
I'd like to see this in the Q1 2023 release. If the Integrator specifies at runtime in the `spdm_context` whether its for a Requester or Responder then we can do away with the `is_requester` parameter from all functions that take `spdm_context` as an input.

PR #1451, opened 2022-12-01
Add deinit_context function to free memory

Wenxing-hou:
Fix: #1460 

### DIFF
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -279,6 +279,16 @@ libspdm_return_t libspdm_init_context(void *spdm_context);
 void libspdm_reset_context(void *spdm_context);
 
 /**
+ * Free the memory of contexts within the SPDM context.
+ * These are typically contexts whose memory has been allocated by the cryptography library.
+ * This function does not free the SPDM context itself.
+ *
+ * @param[in]  spdm_context         A pointer to the SPDM context.
+ *
+ */
+void libspdm_deinit_context(void *spdm_context);
+
+/**
  * Return the size in bytes of the SPDM context. This includes all
  * secured message context data as well.
  *
--- a/library/spdm_common_lib/libspdm_com_context_data.c
+++ b/library/spdm_common_lib/libspdm_com_context_data.c
@@ -2221,6 +2221,47 @@ void libspdm_reset_context(void *context)
 }
 
 /**
+ * Free the memory of contexts within the SPDM context.
+ * These are typically contexts whose memory has been allocated by the cryptography library.
+ * This function does not free the SPDM context itself.
+ *
+ * @param[in]  spdm_context         A pointer to the SPDM context.
+ *
+ */
+void libspdm_deinit_context(void *context)
+{
+#if !(LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT)
+    libspdm_context_t *spdm_context;
+    void *pubkey_context;
+    bool is_requester;
+    uint8_t slot_index;
+
+    spdm_context = context;
+    is_requester = spdm_context->local_context.is_requester;
+
+    for (slot_index = 0; slot_index < SPDM_MAX_SLOT_COUNT; slot_index++) {
+        pubkey_context = spdm_context->connection_info.peer_used_cert_chain[slot_index].
+                         leaf_cert_public_key;
+
+        if (pubkey_context != NULL) {
+            if (is_requester) {
+                libspdm_asym_free(
+                    spdm_context->connection_info.algorithm.base_asym_algo, pubkey_context);
+            } else {
+                libspdm_req_asym_free(
+                    spdm_context->connection_info.algorithm.req_base_asym_alg, pubkey_context);
+            }
+
+            pubkey_context = NULL;
+            spdm_context->connection_info.peer_used_cert_chain[slot_index].
+            leaf_cert_public_key = NULL;
+        }
+    }
+
+#endif
+}
+
+/**
  * Return the size in bytes of the SPDM context. This includes all
  * secured message context data as well.
  *
--- a/doc/user_guide.md
+++ b/doc/user_guide.md
@@ -237,6 +237,12 @@ Refer to spdm_client_init() in [spdm_requester.c](https://github.com/DMTF/spdm-e
    libspdm_send_receive_data (spdm_context, &session_id, TRUE, &request, request_size, &response, &response_size);
    ```
 
+7. Free the memory of contexts within the SPDM context when all flow is over.
+   This function doesn't free the SPDM context itself.
+   ```
+   libspdm_deinit_context(spdm_context);
+   ```
+
 ## SPDM Responder
 
 Refer to spdm_server_init() in [spdm_responder.c](https://github.com/DMTF/spdm-emu/blob/main/spdm_emu/spdm_responder_emu/spdm_responder_spdm.c)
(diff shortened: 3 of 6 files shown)


include/library/spdm_common_lib.h
```diff
@@ -278,6 +278,15 @@ libspdm_return_t libspdm_init_context(void *spdm_context);
...
 void libspdm_reset_context(void *spdm_context);
 
+/**
+ * Free the SPDM context: if there are some memory need free, the function will free them.
+ *
+ * @param[in]  spdm_context         A pointer to the SPDM context.
+ * @param[in]  is_requester        Indicate of the key generation for a requester or a responder.
+ *
+ */
+void libspdm_deinit_context(void *context, bool is_requester);
```
steven-bellock, line 288: `*context` -> `*spdm_context`

include/library/spdm_common_lib.h
```diff
@@ -278,6 +278,15 @@ libspdm_return_t libspdm_init_context(void *spdm_context);
  */
 void libspdm_reset_context(void *spdm_context);
 
+/**
+ * Free the SPDM context: if there are some memory need free, the function will free them.
```
steven-bellock, line 282: > Free the memory of contexts within the SPDM context. These are typically contexts whose memory has been allocated by the cryptography library. This function does not free the SPDM context itself.

include/library/spdm_common_lib.h
```diff
@@ -278,6 +278,15 @@ libspdm_return_t libspdm_init_context(void *spdm_context);
...
+ *
+ * @param[in]  spdm_context         A pointer to the SPDM context.
+ * @param[in]  is_requester        Indicate of the key generation for a requester or a responder.
```
steven-bellock, line 285: >Is the function called from a requester.

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -2220,6 +2220,39 @@ void libspdm_reset_context(void *context)
...
+/**
+ * Free the SPDM context: if there are some memory need free, the function will free them.
+ *
+ * @param[in]  spdm_context         A pointer to the SPDM context.
+ * @param[in]  is_requester        Indicate of the key generation for a requester or a responder.
+ *
+ */
+void libspdm_deinit_context(void *context, bool is_requester)
+{
+#if !LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT
```
steven-bellock, line 2232: Even if `LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT` is enabled this would still be needed right?

steven-bellock (changes requested):

steven-bellock:
This pull request has rekindled my desire for #380. ie, at runtime the Integrator specifies whether an SPDM context is for a Requester or Responder, and then we can get rid of the `is_requester` parameter whenever `spdm_context` is passed to a function. But that will be a 3.0 change.

include/library/spdm_common_lib.h
Wenxing-hou: Thanks. I have changed the function description.

library/spdm_common_lib/libspdm_com_context_data.c
Wenxing-hou: When `LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT` is enabled, the `spdm_context->connection_info.peer_used_cert_chain[slot_id].leaf_cert_public_key` will be not exist. But I think we can make the `LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT` in the `libspdm_deinit_context`, because integrator can use the function easily: Just call `libspdm_deinit_context`, no matter weather `LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT` is existed.

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -2220,6 +2220,41 @@ void libspdm_reset_context(void *context)
...
+/**
+ * Free the memory of contexts within the SPDM context.
+ * These are typically contexts whose memory has been allocated by the cryptography library.
+ * This function does not free the SPDM context itself.
+ *
+ * @param[in]  spdm_context         A pointer to the SPDM context.
+ * @param[in]  is_requester         Is the function called from a requester.
+ *
+ */
+void libspdm_deinit_context(void *context, bool is_requester)
```
jyao1, line 2232: Can we add `is_requester` to the spdm_context?

The requester set it in get_version is called.

Then we dont need to bother the integrate to input the value.

steven-bellock: See #390.

steven-bellock (approved):

library/spdm_common_lib/libspdm_com_context_data.c
jyao1: I do not suggest change all functions.

I just request to not add more in `libspdm_deinit_context()`.

jyao1 (changes requested):
Please avoid adding "is_requester" in new function.

No need to change other old function.

library/spdm_common_lib/libspdm_com_context_data.c
steven-bellock: Let's discuss this at the Monday meeting.

Wenxing-hou:
> Please avoid adding "is_requester" in new function.
> 
> No need to change other old function.

Thanks. I have move `is_requester` to `local_context`.

library/spdm_common_lib/libspdm_com_context_data.c
steven-bellock: @jyao1 explained it at the meeting. Need to make sure that the internal `is_requester` bit gets set properly and robustly.

jyao1: Agree. I think we need set is_requester in VCA cache in spdm_emu.

steven-bellock (changes requested):

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -2220,6 +2220,42 @@ void libspdm_reset_context(void *context)
...
+ * Free the memory of contexts within the SPDM context.
+ * These are typically contexts whose memory has been allocated by the cryptography library.
+ * This function does not free the SPDM context itself.
+ *
+ * @param[in]  spdm_context         A pointer to the SPDM context.
+ *
+ */
+void libspdm_deinit_context(void *context)
+{
+#if !LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT
```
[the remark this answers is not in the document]
Wenxing-hou: Thanks. I have changed it.

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -2220,6 +2220,42 @@ void libspdm_reset_context(void *context)
...
+{
+#if !(LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT)
+    libspdm_context_t *spdm_context;
+    void *pubkey_context;
+    bool is_requester;
+
+    spdm_context = context;
+    is_requester = spdm_context->local_context.is_requester;
+    pubkey_context = spdm_context->connection_info.peer_used_cert_chain
+                     [spdm_context->connection_info.peer_used_cert_chain_slot_id].
```
jyao1, line 2241: I think we need use a for-loop to free all 8 slots. (not only the used one)

Thinking about the case:
Step 1: requester gets all 8 slots (from 0 to 7).
Step 2: requester does challenge on slot 1.
Step 3: requester does key exchange on slot 2.

jyao1 (approved):

PR #1463, opened 2022-12-02
Adding callbacks for Key Update scenarios on Responder side

taprinz:
Implements https://github.com/DMTF/libspdm/issues/1312. Clean github commit history of pull request #1408.

This adds Responder-only callbacks for KEY_UPDATE functionality. This is useful for any integrator who wants/needs to know whenever the keys themselves are updated - for example, if they are loading keys into secure key storage before use.

Adds relevant states for each of these updates, including "create, commit, rollback", to alert the integrator when:
- a new key has been derived and backup of old saved
- new key committed, backup discarded
- backup restored, new key discarded.

I attempted a few ways to implement this, including:

- Putting the callback in `libspdm_create_session_data_key` and `libspdm_activate_update_session_data_key`, issue is no access to requester-only functions or base spdm_context.
- Making the callback available to both Requester and Responder, did not match up with other callback convention.

It also gets rid of the recursion in `libspdm_decode_secured_message`, replacing it with retries whenever we attempt to call `transport_decode_message` and fail with specific error.


taprinz:
Build failures appear to be unrelated openssl issue

steven-bellock:
> Build failures appear to be openssl issue unchanged in this changelist - unsure whether upstream is broken?

Maybe CLANG / LLVM got updated in CI/CD, but I would have expected old CLANG to flag this issue. I'll look into it.

include/library/spdm_return_status.h
```diff
@@ -132,6 +132,10 @@ typedef uint32_t libspdm_return_t;
 #define LIBSPDM_STATUS_ACQUIRE_FAIL \
     LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_ERROR, LIBSPDM_SOURCE_CORE, 0x0010)
 
+/* Re-triable decrypt error from peer - must rollback to backup keys. */
+#define LIBSPDM_STATUS_SESSION_MSG_ERROR_TRY_DISCARD_KEY_UPDATE \
```
steven-bellock, line 136: I would call this `LIBSPDM_STATUS_SESSION_TRY_DISCARD_KEY_UPDATE`

steven-bellock (approved):

PR #1466, opened 2022-12-03
Disable most CLANG warnings when compiling OpenSSL

### LINKED ISSUE #1465: New CLANG / LLVM warnings
CLANG is throwing new warnings that weren't there yesterday. See : https://github.com/DMTF/libspdm/actions/runs/3605366524/jobs/6075691594
```
/home/runner/work/libspdm/libspdm/os_stub/openssllib/openssl/crypto/conf/conf_sap.c:46:17: error: variable 'filename' set but not used [-Werror,-Wunused-but-set-variable]
    const char *filename;
                ^
```
Looking at the toolchain setup it looks like Github has transitioned `ubuntu-latest` from `20.04` to `22.04` and perhaps CLANG got a little smarter. Note that this was fixed upstream three months ago by https://github.com/openssl/openssl/commit/8ae080bf851a25187b93803b8c6a93e82dd97437

steven-bellock:
Fixes #1465.

Note that we should do this with other libraries in `os_stub/`.


steven-bellock:
One of the tests fails due to #1467. I'll merge this pull request so it can compile, and we can look at the test failure later.

jyao1:
Hi Steven
This P-R will cause CI fail.

I dont believe we should merge now. We should merge after we fix other issue, to ensure CI can always pass.

 We need rollback to unblock other patch merge.

Issue #1423, opened 2022-11-22
MeasurementSpecification in NEGOTIATE_ALGORITHM


hiption:
Hi all,
I have a question about MeasurementSpecification field in NEGOTIATE_ALGORITHM.

https://github.com/DMTF/libspdm/blob/dc93a39396efaf083815d57c46abc92fd063e568/library/spdm_requester_lib/libspdm_req_negotiate_algorithms.c#L303-L317

According to above code, the requester returns INVALID_MSG_FIELD status as follows;
1. Responder has a MEAS_CAP.
2. Requester does not want to use measurement.
3. Requester sets MeasurementSpecification as zero in NEGOTIATE_ALGORITHM.
4. Responder sets MeasurementSpecificationSel as zero in ALGORITMS.
5. Requester returns INVALID_MSG_FIELD status because the MeasurementSpecificationSel is not SPDM_MEASUREMENT_BLOCK_HEADER_SPECIFICATION_DMTF, but the Responder has a MEAS_CAP.

In some case, even if the responder has the measurement capability, the requester may not want to use measurement.

steven-bellock:
I thought this was fixed in https://github.com/DMTF/libspdm/pull/876 but that looks like it was all on the Responder side.

hiption:
BTW, how to access [https://github.com/DMTF/Security-TF](https://github.com/DMTF/Security-TF/issues/2014)?
Can only DMTF members access the Security-TF?

steven-bellock:
> Can only DMTF members access the Security-TF?

Correct. If you are not a DMTF member and have questions / issues about the specification you can ask on the public http://spdmforum.com/.

hiption:
I think one more condition should be added as follows;

if (libspdm_is_capabilities_flag_supported( 
         spdm_context, true, 0, 
         SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_MEAS_CAP)
         && (spdm_request->measurement_specification != 0)) {

PR #1468, opened 2022-12-04
Add more check for measurement_spec

Wenxing-hou:
Fix: #1423 

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_get_measurements.c
+++ b/library/spdm_requester_lib/libspdm_req_get_measurements.c
@@ -212,6 +212,8 @@ static libspdm_return_t libspdm_try_get_measurement(void *context, const uint32_
         return LIBSPDM_STATUS_UNSUPPORTED_CAP;
     }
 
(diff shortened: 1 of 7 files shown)


library/spdm_requester_lib/libspdm_req_get_measurements.c
```diff
@@ -212,6 +212,10 @@ static libspdm_return_t libspdm_try_get_measurement(void *context, const uint32_
         return LIBSPDM_STATUS_UNSUPPORTED_CAP;
     }
 
+    if (spdm_context->local_context.algorithm.measurement_spec == 0) {
```
steven-bellock, line 215: This should be a `LIBSPDM_ASSERT`.

steven-bellock (changes requested):

library/spdm_requester_lib/libspdm_req_get_measurements.c
Wenxing-hou: Thanks. I have changed the code.

steven-bellock (approved):

Issue #928, opened 2022-06-07
Question about ALGORITHMS for multiple tables


hiption:
https://github.com/DMTF/libspdm/blob/b42d264dd770d6f14c2a2278c9942806e2c554fd/unit_test/test_spdm_responder/algorithms.c#L1139-L1204

Above TC (libspdm_test_responder_algorithms_case7) is for nonmal behavior with multiple tables.

https://github.com/DMTF/libspdm/blob/b42d264dd770d6f14c2a2278c9942806e2c554fd/library/spdm_responder_lib/libspdm_rsp_algorithms.c#L408-L439

In the current implementation, the same type of algorithm structure is overwritten when it comes.
Therefore, when multiple tables are used, only the algorithm described last of the same type  is set.
I wonder it is correct.

steven-bellock:
The specification says
>AlgType shall monotonically increase for subsequent entries.

and that is something that the Requester / Responder can check, and for the Responder's case it can return an `ERROR` response if the Requester has generated an algorithm structure with repeated entries or entries that are in reverse order.

steven-bellock:
@hiption Are you interested in creating a pull request to resolve this issue? It entails creating a counter to ensure that entries aren't repeated and are always increasing.

hiption:
I agree with your comment that "it can return an ERROR response if the Requester has generated an algorithm structure with repeated entries or entries that are in reverse order".

I am interested in participating in libSpdm. 
However, PR is not possible due to the restrictions of my development environment.

I am sorry.

PR #1483, opened 2022-12-06
Add AlgType checker

hiption:
Fix the issue: #928

add a conditional to check if the AlgType monotonically incerease for subsequent entries or not.

Signed-off-by: Sungho Yoon <sungho.yoon@samsung.com>

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_negotiate_algorithms.c
+++ b/library/spdm_requester_lib/libspdm_req_negotiate_algorithms.c
@@ -79,6 +79,7 @@ static libspdm_return_t libspdm_try_negotiate_algorithms(libspdm_context_t *spdm
     uint8_t *message;
     size_t message_size;
     size_t transport_header_size;
+    uint8_t alg_type_pre;
 
     /* -=[Verify State Phase]=- */
     if (spdm_context->connection_info.connection_state !=
@@ -226,6 +227,7 @@ static libspdm_return_t libspdm_try_negotiate_algorithms(libspdm_context_t *spdm
                  sizeof(uint32_t) * spdm_response->ext_asym_sel_count +
                  sizeof(uint32_t) * spdm_response->ext_hash_sel_count);
     if (spdm_response->header.spdm_version >= SPDM_MESSAGE_VERSION_11) {
+        alg_type_pre = struct_table->alg_type;
         /* header.param1 is implictly checked through spdm_response_size. */
         for (index = 0; index < spdm_response->header.param1; index++) {
             if ((size_t)spdm_response + spdm_response_size < (size_t)struct_table) {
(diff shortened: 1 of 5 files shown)


library/spdm_responder_lib/libspdm_rsp_algorithms.c
```diff
@@ -383,7 +383,16 @@ libspdm_return_t libspdm_get_response_algorithms(void *context,
                      sizeof(spdm_negotiate_algorithms_request_t) +
                      sizeof(uint32_t) * spdm_request->ext_asym_count +
                      sizeof(uint32_t) * spdm_request->ext_hash_count);
+        uint8_t alg_type_pre = struct_table->alg_type;
```
steven-bellock, line 386: This declaration needs to be at line 381. Even though libspdm is a C99 project we try to make it as compatible with C89 as possible.

steven-bellock (changes requested):
The Requester will also need this check. Also we'll need at least one error test so that
```
            if ((index != 0) && (struct_table->alg_type <= alg_type_pre)) {
                    spdm_context,
                    SPDM_ERROR_CODE_INVALID_REQUEST, 0,
                    response_size, response);
            }
```
is hit. Note that on the Requester side the error tests have moved to https://github.com/DMTF/libspdm/blob/main/unit_test/test_spdm_requester/error_test/negotiate_algorithms_err.c

hiption:
@steven-bellock 
Thank you for your feedback.

library/spdm_requester_lib/libspdm_req_negotiate_algorithms.c
```diff
@@ -378,7 +379,13 @@ static libspdm_return_t libspdm_try_negotiate_algorithms(libspdm_context_t *spdm
                      sizeof(spdm_algorithms_response_t) +
                      sizeof(uint32_t) * spdm_response->ext_asym_sel_count +
                      sizeof(uint32_t) * spdm_response->ext_hash_sel_count);
+        alg_type_pre = struct_table->alg_type;
         for (index = 0; index < spdm_response->header.param1; index++) {
+            /* AlgType shall monotonically increase for subsequent entries. */
```
steven-bellock, line 384: 1. `alg_type_pre` needs to be updated inside the for-loop.
2. This code should be moved to the `-=[Validate Response Phase]=-` section at line `249`.

library/spdm_responder_lib/libspdm_rsp_algorithms.c
```diff
@@ -383,7 +384,16 @@ libspdm_return_t libspdm_get_response_algorithms(void *context,
                      sizeof(spdm_negotiate_algorithms_request_t) +
                      sizeof(uint32_t) * spdm_request->ext_asym_count +
                      sizeof(uint32_t) * spdm_request->ext_hash_count);
+        alg_type_pre = struct_table->alg_type;
         for (index = 0; index < spdm_request->header.param1; index++) {
+            /* AlgType shall monotonically increase for subsequent entries. */
+            if ((index != 0) && (struct_table->alg_type <= alg_type_pre)) {
```
steven-bellock, line 390: 1. `alg_type_pre` needs to be updated inside the for-loop.
2. This code should be moved to the `-=[Validate Response Phase]=-` section at line `287`.

steven-bellock (changes requested):

library/spdm_requester_lib/libspdm_req_negotiate_algorithms.c
hiption: 1. I made a mistake of omitting alg_type_pre update at line 389. To prevent this mistake, I will add  not only redundant case but also decreasing case.
2. I'll move this checker to line 249.

library/spdm_responder_lib/libspdm_rsp_algorithms.c
hiption: 1. The update code is already written at line 396.
2. ok

library/spdm_requester_lib/libspdm_req_negotiate_algorithms.c
```diff
@@ -237,14 +239,20 @@ static libspdm_return_t libspdm_try_negotiate_algorithms(libspdm_context_t *spdm
...
                 goto receive_done;
             }
+            /* AlgType shall monotonically increase for subsequent entries. */
+            if ((index != 0) && (struct_table->alg_type <= alg_type_pre)) {
+                status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
+                goto receive_done;
+            }
+            alg_type_pre = struct_table->alg_type;
+            fixed_alg_size = (struct_table->alg_count >> 4) & 0xF;
+            ext_alg_count = struct_table->alg_count & 0xF;
```
hiption, line 255: I have moved these two lines behind checking `alg_type`

steven-bellock (approved):

Issue #1475, opened 2022-12-06
Pointer 'session_info' checked for NULL at line 41 may be dereferenced at line 54.


Zhiqiang520:
Pointer `'session_info'` checked for NULL at line 41 may be dereferenced at line 54.

The line 41 `LIBSPDM_ASSERT (session_info != NULL);` only workable when debug build, 
release build will disable `LIBSPDM_ASSERT`.
So I think maybe the line 41 `LIBSPDM_ASSERT (session_info != NULL);` should be changed like following:
```
    if (session_info == NULL) {
        return LIBSPDM_STATUS_INVALID_PARAMETER;
    }
```
https://github.com/DMTF/libspdm/blob/3337da5d52c32d4bf3585506c67d8c92b587b256/library/spdm_requester_lib/libspdm_req_heartbeat.c#L40-L54

jyao1:
The issue is reported by static analysis tool, when we build release image, where the ASSERT is nothing.

PR #1485, opened 2022-12-07
Fix pointer 'session_info' checked for NULL may be dereferenced.

Zhiqiang520:
Fix: #1475 
Fix: #1476 


jyao1:
The issue is reported by static analysis tool, when we build release image, where the ASSERT is nothing.

steven-bellock (approved):

Issue #1328, opened 2022-11-01
Raw public key provisioning is using certificate chain


xiangfeima:
According to SPDM ver 1.2.1 section 7.2.2 Raw public keys

> Alternatively to certificate chains, the vendor can provision the raw public key of the Responder to the Requester in a
trusted environment; for example, during the secure manufacturing process. In this case, trust of the public key of the
Responder is established without the need for a certificate-based public key infrastructure. The definition of a trusted
environment is outside of the scope of this specification.
The format of the provisioned public key is out of scope of this specification. Vendors can use proprietary formats or
public key formats that other standards define, such as RFC7250 and RFC4716.

Libspdm uses certificate chain even in the raw public key provisioning (slot 0xFF) scenario.
Would you mind to let me know your opinion about this?

steven-bellock:
> Libspdm uses certificate chain

Only the leaf certificate is required.

My opinion is that until the SPDM specification mandates a single raw public key format it will be difficult for a general purpose library like libspdm to provide support. I would be fine if it was a simple ASN.1 data structure that came with a simple parser and not an entire certificate chain parsing library. Maybe the specification could have its blessed format that it recommends everyone, including libspdm, use, but allow implementers to use their own format if, for example, ASN.1 doesn't work for them.

xiangfeima:
> Only the leaf certificate is required.

Yes though, it is still a certificate instead of the raw public key.
How about change to use just ASN.1 SubjectPublicKeyInfo of a certificate, as the specification claims (RFC7250)?
as you mentioned:
> a simple ASN.1 data structure that came with a simple parser and not an entire certificate chain parsing library.

jyao1:
11/7/2022 - One possible option - libspdm can include an ASN.1 parser, knowing it will only parse the raw public key. Not a full parser. Hope it is small portion of code. Also, we assume the provisioned public key is good, not malicious.

@xiangfeima, will you please take this issue and submit patch?

xiangfeima:
@jyao1 

> One possible option - libspdm can include an ASN.1 parser, knowing it will only parse the raw public key. Not a full parser.

I think we don't need to introduce such "raw public key parser", my idea is to add interface in `spdm_crypt_ext_lib.h` to read PEM public key, which would be implemented by mbedtls/openssl/null underlaying (similar to the private key reading process).
The PEM public key for testing can be extracted from existing requester/responder key file under unit_test/sample_key.

PR #1486, opened 2022-12-07
Adding raw public key provisioning

xiangfeima:
Fix the issue:#1328

### DIFF
--- a/library/spdm_common_lib/libspdm_com_context_data.c
+++ b/library/spdm_common_lib/libspdm_com_context_data.c
@@ -1407,6 +1407,7 @@ libspdm_return_t libspdm_append_message_k(void *context, void *session_info,
 
         if (spdm_session_info->session_transcript.digest_context_th == NULL) {
             if (!spdm_session_info->use_psk) {
+                LIBSPDM_ASSERT(slot_id < SPDM_MAX_SLOT_COUNT);
                 if (is_requester) {
                     if(spdm_context->connection_info.peer_used_cert_chain[slot_id].buffer_hash_size
                        != 0) {
@@ -1575,6 +1576,7 @@ libspdm_return_t libspdm_append_message_f(void *context, void *session_info,
                         return LIBSPDM_STATUS_CRYPTO_ERROR;
                     }
                 } else {
+                    LIBSPDM_ASSERT(slot_id < SPDM_MAX_SLOT_COUNT);
                     if (spdm_context->connection_info.peer_used_cert_chain[slot_id].buffer_hash_size
                         != 0) {
                         hash_size =
--- a/library/spdm_common_lib/libspdm_com_crypto_service.c
+++ b/library/spdm_common_lib/libspdm_com_crypto_service.c
@@ -28,6 +28,7 @@ bool libspdm_get_peer_cert_chain_buffer(void *context,
     spdm_context = context;
 #if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT
     slot_id = spdm_context->connection_info.peer_used_cert_chain_slot_id;
+    LIBSPDM_ASSERT(slot_id < SPDM_MAX_SLOT_COUNT);
     if (spdm_context->connection_info.peer_used_cert_chain[slot_id].buffer_size != 0) {
         *cert_chain_buffer = spdm_context->connection_info.peer_used_cert_chain[slot_id].buffer;
         *cert_chain_buffer_size = spdm_context->connection_info
@@ -867,6 +868,7 @@ bool libspdm_verify_certificate_chain_hash(libspdm_context_t *spdm_context,
     }
 #else
     slot_id = spdm_context->connection_info.peer_used_cert_chain_slot_id;
+    LIBSPDM_ASSERT(slot_id < SPDM_MAX_SLOT_COUNT);
     if (spdm_context->connection_info.peer_used_cert_chain[slot_id].buffer_hash_size != 0) {
         if (spdm_context->connection_info.peer_used_cert_chain[slot_id].buffer_hash_size !=
             certificate_chain_hash_size) {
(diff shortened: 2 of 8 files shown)


include/library/spdm_common_lib.h
```diff
@@ -57,7 +57,8 @@ typedef enum {
     /* LIBSPDM_DATA_LOCAL_SLOT_COUNT is deprecated. Do not use it. Reserved only. */
     LIBSPDM_DATA_LOCAL_SLOT_COUNT,
     LIBSPDM_DATA_PEER_PUBLIC_ROOT_CERT,
-    LIBSPDM_DATA_PEER_PUBLIC_CERT_CHAIN,
+    LIBSPDM_DATA_LOCAL_PUBLIC_KEY,
```
jyao1, line 60: This issue only "add raw pub support". As such, it should not remove `LIBSPDM_DATA_PEER_PUBLIC_CERT_CHAIN`.

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -1575,32 +1611,28 @@ libspdm_return_t libspdm_append_message_f(void *context, void *session_info,
...
-                    } else {
-                        result = libspdm_get_peer_cert_chain_buffer(
+                    if (slot_id == 0xFF) {
+                        result = libspdm_get_peer_public_key_buffer(
                             spdm_context,
                             (const void **)&mut_cert_chain_buffer,
                             &mut_cert_chain_buffer_size);
                         if (!result) {
-                            return LIBSPDM_STATUS_INVALID_STATE_PEER;
+                            return false;
```
jyao1, line 1620: why `return false`?

include/internal/libspdm_common_lib.h
```diff
@@ -63,22 +63,22 @@ typedef struct {
     /* My Certificate */
     const void *local_cert_chain_provision[SPDM_MAX_SLOT_COUNT];
     size_t local_cert_chain_provision_size[SPDM_MAX_SLOT_COUNT];
-    /* My provisioned certificate (for slot_id - 0xFF, default 0)*/
-    uint8_t provisioned_slot_id;
```
jyao1, line 67: Why remove `provisioned_slot_id`?

steven-bellock:
Yes, these changes would have to go in after the Q4 release. And I think it would help to split it up into smaller pull requests once we digest the overall design. In particular the changes to `cryptlib` are effectively standalone and testable and would make a good standalone pull request.

include/library/spdm_common_lib.h
xiangfeima: My understanding from SPDM specification is that the certificate chain of peer should be and only be retrieved from `GET_CERTIFICATE` process, and the `LIBSPDM_DATA_PEER_PUBLIC_ROOT_CERT` should be the only one to be set base upon needs to help the verification of peer certificate chain.
Another thought is, it looks there is no place for the idea of "provisioned certificate chain of peer" because the leaf cert of the chain would be new-generated every time from the device secure-booting if it uses `alias_cert` mode - I guess that means a fixed, pre-set certificate chain of peer device may not be applicable.

Besides, currently the `LIBSPDM_DATA_PEER_PUBLIC_CERT_CHAIN` seems to be only for the slot id 0xFF (raw public key provisioning) related process. The only code that refers to `LIBSPDM_DATA_PEER_PUBLIC_CERT_CHAIN` I can find is in spdm-emu it sets `LIBSPDM_DATA_PEER_PUBLIC_CERT_CHAIN` for slot id 0xFF testing.
https://github.com/DMTF/spdm-emu/blob/5c85a16499ae4d4aa915b864727ba11572693291/spdm_emu/spdm_requester_emu/spdm_requester_spdm.c#L350-L369

This PR provides more appropriate type `LIBSPDM_DATA_LOCAL_PUBLIC_KEY` for raw public key process so I think `LIBSPDM_DATA_PEER_PUBLIC_CERT_CHAIN` can be removed.

library/spdm_common_lib/libspdm_com_context_data.c
xiangfeima: I'll fix it.

include/internal/libspdm_common_lib.h
xiangfeima: `provisioned_slot_id` was just for the synchronization from slot id 0xFF of the request message to the one of slots (default is slot 0) libspdm holding internally in its context. We were doing so because the raw public key of slot 0xFF was using certificate process routine.
From this PR the 0xFF slot id can be processed directly so the `provisioned_slot_id` is no longer needed.

library/spdm_common_lib/libspdm_com_context_data.c
xiangfeima: Fixed in https://github.com/DMTF/libspdm/pull/1486/commits/4e2a94599a97f52a050c15b133b1ccdb1b224336 .

jyao1:
> Yes, these changes would have to go in after the Q4 2022 release.

I second that. This change is too big. Hard to review in limited time. And we need fix other planned items.

@xiangfeima, any concern if we push to after the Q4 2022 release.

jyao1:
> I will update new PR for spdm-emu once this one is considered to be good to merge.

Usually, what we do is to create a PR for spdm-emu as well, as draft.
It will fail the CI as expected, but allow people to review the code and feedback.

Then once libspdm merge, we create a new PR to sync libspdm. Once it is merged, the CI will pass.

xiangfeima:
Sorry for late, I left office for few days.

> any concern if we push to after the Q4 2022 release.

It's okay to me. I understand we need more time to review this big change.

> Usually, what we do is to create a PR for spdm-emu as well, as draft.
> It will fail the CI as expected, but allow people to review the code and feedback.
> 
> Then once libspdm merge, we create a new PR to sync libspdm. Once it is merged, the CI will pass.

I will do this.

PR #1494, opened 2022-12-07
Allow export of export master secret to target small destinations

### LINKED ISSUE #1488: Allow libspdm_secured_message_export_master_secret to target small destinations
Follow up to #1303.

`libspdm_secured_message_export_master_secret` checks if the size of the destination buffer is less than the size of the Export Master Secret and returns an error if it is. However the Integrator may only want a portion of the Export Master Secret. We should remove the check and only write to `export_master_secret_size` if the destination size is larger than the size of the Export Master Secret.

https://github.com/DMTF/libspdm/blob/31c6d38582af9dfdfd8b34e9a9e0a65913eca07c/library/spdm_secured_message_lib/libspdm_secmes_context_data.c#L198-L214

steven-bellock:
Fixes #1488.

### DIFF
--- a/library/spdm_secured_message_lib/libspdm_secmes_context_data.c
+++ b/library/spdm_secured_message_lib/libspdm_secmes_context_data.c
@@ -200,7 +200,7 @@ bool libspdm_secured_message_export_master_secret(
 
     libspdm_copy_mem(export_master_secret, *export_master_secret_size,
                      secured_message_context->export_master_secret,
-                     secured_message_context->hash_size);
+                     *export_master_secret_size);
 
     return true;
 }
--- a/library/spdm_secured_message_lib/libspdm_secmes_context_data.c
+++ b/library/spdm_secured_message_lib/libspdm_secmes_context_data.c
@@ -186,15 +186,6 @@ bool libspdm_secured_message_import_dhe_secret(void *spdm_secured_message_contex
     return true;
 }
 
-/**
- * Export the export_master_secret from an SPDM secured message context.
- *
- * @param  spdm_secured_message_context    A pointer to the SPDM secured message context.
- * @param  export_master_secret           Indicate the buffer to store the export_master_secret.
- * @param  export_master_secret_size       The size in bytes of the export_master_secret.
- *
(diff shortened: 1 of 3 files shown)


library/spdm_secured_message_lib/libspdm_secmes_context_data.c
```diff
@@ -186,30 +186,22 @@ bool libspdm_secured_message_import_dhe_secret(void *spdm_secured_message_contex
...
-        return false;
-    }
+
     libspdm_copy_mem(export_master_secret, *export_master_secret_size,
                      secured_message_context->export_master_secret,
-                     secured_message_context->hash_size);
-    *export_master_secret_size = secured_message_context->hash_size;
+                     sizeof(secured_message_context->export_master_secret));
+
+    if (*export_master_secret_size > secured_message_context->hash_size) {
```
steven-bellock, line 201: This needs to go before the `libspdm_copy_mem`.

library/spdm_secured_message_lib/libspdm_secmes_context_data.c
```diff
@@ -186,30 +186,22 @@ bool libspdm_secured_message_import_dhe_secret(void *spdm_secured_message_contex
...
 
     secured_message_context = spdm_secured_message_context;
-    if (*export_master_secret_size < secured_message_context->hash_size) {
+
+    if (*export_master_secret_size > secured_message_context->hash_size) {
         *export_master_secret_size = secured_message_context->hash_size;
-        return false;
     }
+
     libspdm_copy_mem(export_master_secret, *export_master_secret_size,
```
jyao1, line 201: have you tested this?

```
 * Asserts and returns a non-zero value if any of the following are true:
 *   3) "src_len" is greater than "dst_len".
```
https://github.com/jyao1/libspdm/blob/readme_build/include/hal/library/memlib.h#L26

steven-bellock: Heh I did not. That's what I get for trying to rush this through before the end of the workday. I'll fix it and make sure it's tested.

jyao1 (approved):

Issue #1473, opened 2022-12-05
test_crypt does not function as a test


steven-bellock:
Follow-up to my comment in #1350. `test_crypt` has multiple issues.
1. Tests fail if the backing cryptography library does not support the algorithm. For example, currently (our) MbedTLS does not support HMAC-SHA3 which is reported as a failure rather than skipped.
2. Some of the tests do not return a status on pass / fail, rendering them useless in CI/CD. For example : https://github.com/DMTF/libspdm/blob/beee7bd85a4f0961a5773aefc4dcab0408d538d1/unit_test/test_crypt/hash_verify.c#L420-L439
3. Return status confusion. In particular https://github.com/DMTF/libspdm/blob/3337da5d52c32d4bf3585506c67d8c92b587b256/unit_test/test_crypt/test_crypt.c#L149-L158 is incorrect. `libspdm_cryptest_main` returns `true (1)` if all the tests pass and `false (0)` if one of them fails. `main` should return `0` if all tests pass and `!= 0` if one of them fails.

Due to 3, `test_crypt` has failing tests but it is being reported as a pass to CI/CD. We should disable `test_crypt` in CI/CD until these issues are resolved.

jyao1:
once you fix it issue, please re-enable test_crypt in CI.

PR #1501, opened 2022-12-08
Make test crypt be test function

Wenxing-hou:
Fix: #1473

The PR has tested pass by checking local log(mbedtls/openssl)  and running on my branch.

For commit [Make test_crypt be standard test function](https://github.com/DMTF/libspdm/commit/1197308950cd81a9e5197714b77fdd391c780c0b) :

- Add some macro like LIBSPDM_RSA_SSA_SUPPORT_TEST;
- For some only supported crypt function in opesnssl, skip the test when mbedtls build.
- For some crypt function which both mbedtls and openssl don't support, skip them.
- Make all return right.

For commit [Enable test_crypt in CI again](https://github.com/DMTF/libspdm/commit/85eca38152b742ea65ecc3a2bf4271822af575cb):

- enable CI.


steven-bellock (approved):

unit_test/test_crypt/aead_verify.c
```diff
@@ -252,57 +252,8 @@ bool libspdm_validate_crypt_aead_cipher(void)
...
-    if (libspdm_const_compare_mem(OutBuffer, m_libspdm_sm4_gcm_pt,
-                                  sizeof(m_libspdm_sm4_gcm_pt)) != 0) {
-        libspdm_my_print("[Fail]");
-        goto Exit;
-    }
-
-    libspdm_my_print("[Pass]");
-Exit:
+    /* the sm4_gcm is not support now*/
+    libspdm_my_print("\n- SM4-GCM Encryption is not support");
```
jyao1, line 256: why you **comment** unsupported sm2, but **delete** unsupported sm4 here?
We need use the consistent rule.

Maybe just comment the unsupported code.

unit_test/test_crypt/test_crypt.c
```diff
@@ -150,7 +153,7 @@ int main(void)
 {
     int return_value = 0;
 
-    if (libspdm_cryptest_main() != 0) {
+    if (libspdm_cryptest_main() == false) {
```
jyao1, line 156: Don't compare with bool (true/false).

just use `!libspdm_cryptest_main()`.

unit_test/test_crypt/CMakeLists.txt
```diff
@@ -38,6 +38,10 @@ SET(test_crypt_LIBRARY
     spdm_crypt_ext_lib
 )
 
+if(CRYPTO STREQUAL "openssl")
+    add_definitions(-DLIBSPDM_USE_OPENSSL=1)
```
jyao1, line 42: If `LIBSPDM_USE_OPENSSL` is false, there is no guarantee some crypto functions are enabled/disabled.

For consistency, I recommend:
1) we define a test_mask, for all functions, match - crypto configuration in spdm_lib_config.h
```
LIBSPDM_RSA_SSA_SUPPORT_TEST
LIBSPDM_RSA_PSS_SUPPORT_TEST
LIBSPDM_ECDSA_SUPPORT_TEST
LIBSPDM_SM2_DSA_SUPPORT_TEST
LIBSPDM_EDDSA_ED25519_SUPPORT_TEST
LIBSPDM_EDDSA_ED448_SUPPORT_TEST
LIBSPDM_FFDHE_SUPPORT_TEST
LIBSPDM_ECDHE_SUPPORT_TEST
LIBSPDM_SM2_KEY_EXCHANGE_SUPPORT_TEST
LIBSPDM_AEAD_GCM_SUPPORT_TEST
LIBSPDM_AEAD_CHACHA20_POLY1305_SUPPORT_TEST
LIBSPDM_AEAD_SM4_SUPPORT_TEST
LIBSPDM_SHA256_SUPPORT_TEST
LIBSPDM_SHA384_SUPPORT_TEST
LIBSPDM_SHA512_SUPPORT_TEST
LIBSPDM_SHA3_256_SUPPORT_TEST
LIBSPDM_SHA3_384_SUPPORT_TEST
LIBSPDM_SHA3_512_SUPPORT_TEST
LIBSPDM_SM3_256_SUPPORT_TEST
```

2) We may use `LIBSPDM_USE_OPENSSL` to configure above macro.
3) We may use `LIBSPDM_USE_MBEDTLS` to configure above macro as well.

unit_test/test_crypt/test_crypt.c
```diff
@@ -120,10 +120,13 @@ bool libspdm_cryptest_main(void)
...
-    if (!status) {
-        return status;
-    }
+
+    /* the Crypto SM2 key Exchange is not support now
+     * status = libspdm_validate_crypt_sm2();
+     * if (!status) {
+     *     return status;
+     * }
+     **/
```
jyao1, line 129: If we can add crypto configuration for test, then we dont need comment or delete.
We can just disable it by macro.

jyao1 (changes requested):
Please use crypto test configuration to isolate unsupported algo test.

unit_test/test_crypt/test_crypt.c
Wenxing-hou: Thanks. I have changed it.

unit_test/test_crypt/CMakeLists.txt
Wenxing-hou: Thanks. I have added many macro in CMakelists.

jyao1 (approved):

PR #1505, opened 2022-12-11
Add compile-time macro checking

### LINKED ISSUE #1261: Add compile-time checking of macros
For example if `LIBSPDM_ENABLE_CAPABILITY_KEY_EX_CAP` is enabled then at least one of `LIBSPDM_FFDHE_SUPPORT` / `LIBSPDM_ECDHE_SUPPORT` / etc needs to be enabled. I think this would be disabled by default, so if an Integrator wants to enable checking they would set `LIBSPDM_CHECK_MACROS` to `1`. 

This is related to #1027 which checks the context at runtime.

steven-bellock:
This pull request is just to get things started. We can add more checks as needed.

Fixes #1261.

### DIFF
--- a/include/library/spdm_lib_config.h
+++ b/include/library/spdm_lib_config.h
@@ -407,4 +407,8 @@
 #define LIBSPDM_ENABLE_MSG_LOG 1
 #endif
 
+#ifndef LIBSPDM_CHECK_MACRO
+#define LIBSPDM_CHECK_MACRO 0
+#endif
+
 #endif /* SPDM_LIB_CONFIG_H */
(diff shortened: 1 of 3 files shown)


include/library/spdm_lib_config.h
```diff
@@ -407,4 +407,8 @@
 #define LIBSPDM_ENABLE_MSG_LOG 1
 #endif
 
+#ifndef LIBSPDM_CHECK_MACRO
+#define LIBSPDM_CHECK_MACRO 0
```
jyao1, line 411: why disable? I think it is good to add them to find issue earlier.

steven-bellock: Since this check is in internal configuration header it will be `#include`d in every file and checked by the compiler. It could affect compile times.

Issue #1497, opened 2022-12-08
Pointer checked for NULL will be dereferenced in libspdm_com_msg_log.c


Zhiqiang520:
**1. Pointer 'spdm_context' checked for NULL at line 14 will be dereferenced at line 18.**

https://github.com/DMTF/libspdm/blob/b4b7474f64fedec25a0ce028f71c65bf3e936879/library/spdm_common_lib/libspdm_com_msg_log.c#L14-L18

**2. The following are similar as the above one.**
Pointer 'spdm_context' checked for NULL at line 29 will be dereferenced at line 32.
Pointer 'spdm_context' checked for NULL at line 39 will be dereferenced at line 43.
Pointer 'spdm_context' checked for NULL at line 50 will be dereferenced at line 54.
Pointer 'spdm_context' checked for NULL at line 61 will be dereferenced at line 65.
Pointer 'spdm_context' checked for NULL at line 74 will be dereferenced at line 78.

steven-bellock:
@Zhiqiang520 I think what you're looking for is the ability to enable assertions when the CMake target is Release, correct?

Zhiqiang520:
@steven-bellock No. The issue is reported by static analysis tool, I just report it.

steven-bellock:
Understood. Feel free to waive these issues (and #1495 and #1496) since these are pre-conditions that the Integrator needs to honor.

I do think that the Integrator should be able to enable assertions on a Release build. I'll file a separate issue.

jyao1:
I dont think we should "waive" the issue here.

Assuming the static analysis tool will check the final integrated image instead of individual function, we have two ways to fix the issue.
 * fix the caller
 * fix the callee

It seems the patch fixes the callee. But if we use ASSERT, then we should fix caller.

@Zhiqiang520 , would you please evaluate the fist option?

Zhiqiang520:
waive this issue since these are pre-conditions that the Integrator needs to honor.

Issue #1496, opened 2022-12-08
Pointer checked for NULL may be dereferenced in libspdm_req_key_exchange.c


Zhiqiang520:
**The issue is regression due to the PR** [#1428 ](https://github.com/DMTF/libspdm/pull/1428): _Improve code coverage for KEY_EXCHANGE_

- Pointer 'session_info' checked for NULL at line 522 may be dereferenced at line 606.

https://github.com/DMTF/libspdm/blob/b4b7474f64fedec25a0ce028f71c65bf3e936879/library/spdm_requester_lib/libspdm_req_key_exchange.c#L520-L522
https://github.com/DMTF/libspdm/blob/b4b7474f64fedec25a0ce028f71c65bf3e936879/library/spdm_requester_lib/libspdm_req_key_exchange.c#L605-L607

Issue #1495, opened 2022-12-08
Pointer checked for NULL may be dereferenced in libspdm_com_context_data.c


Zhiqiang520:
**The issue is regression due to the PR** [#1400](https://github.com/DMTF/libspdm/pull/1400): _Changing parameter checks to LIBSPDM_ASSERTS._

- Pointer 'spdm_context' checked for NULL at line 2074 will be dereferenced at line 2079.

https://github.com/DMTF/libspdm/blob/b4b7474f64fedec25a0ce028f71c65bf3e936879/library/spdm_common_lib/libspdm_com_context_data.c#L2074-L2079

- Pointer 'secured_contexts' checked for NULL at line 2075 may be dereferenced at line 2118.

https://github.com/DMTF/libspdm/blob/b4b7474f64fedec25a0ce028f71c65bf3e936879/library/spdm_common_lib/libspdm_com_context_data.c#L2117-L2120

- Pointer 'context' checked for NULL at line 2151 will be dereferenced at line 2165.

https://github.com/DMTF/libspdm/blob/b4b7474f64fedec25a0ce028f71c65bf3e936879/library/spdm_common_lib/libspdm_com_context_data.c#L2151-L2167

Zhiqiang520:
waive this issue since these are pre-conditions that the Integrator needs to honor.

PR #1506, opened 2022-12-12
Fix pointer checked for NULL may be dereferenced.

Zhiqiang520:
Fix: #1495 
Fix: #1496
Fix: #1497


library/spdm_common_lib/libspdm_com_msg_log.c
```diff
@@ -11,7 +11,9 @@ void libspdm_init_msg_log (void *context, void *msg_buffer, size_t msg_buffer_si
 {
     libspdm_context_t *spdm_context;
 
-    LIBSPDM_ASSERT((context != NULL) && (msg_buffer != NULL));
+    if (context == NULL || msg_buffer == NULL) {
```
jyao1, line 14: why we need check "context == NULL" here?

If we added check in init(), then that will guarantee all rest function calls will have context != NULL.

Please clarify in which path, you can still trigger NULL context input.

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -2071,9 +2071,10 @@ libspdm_return_t libspdm_init_context_with_secured_context(void *context,
     libspdm_context_t *spdm_context;
     size_t index;
 
-    LIBSPDM_ASSERT(context != NULL);
-    LIBSPDM_ASSERT(secured_contexts != NULL);
-    LIBSPDM_ASSERT(num_secured_contexts == LIBSPDM_MAX_SESSION_COUNT);
+    if (context == NULL || secured_contexts == NULL ||
```
jyao1, line 2074: The caller should guarantee those conditions are satisfied before call this function.

Can we fix the caller?

library/spdm_requester_lib/libspdm_req_key_exchange.c
```diff
@@ -517,9 +517,12 @@ static libspdm_return_t libspdm_try_send_receive_key_exchange(
     *session_id = libspdm_generate_session_id(req_session_id, rsp_session_id);
     session_info = libspdm_assign_session_id(spdm_context, *session_id, false);
 
-    /* session_info cannot be null as the check after libspdm_allocate_req_session_id assures
-     * that there is space for a new session ID and its initialization. */
-    LIBSPDM_ASSERT(session_info != NULL);
+    if (session_info == NULL) {
```
jyao1, line 520: Only this check seems reasonable. It align with https://github.com/DMTF/libspdm/pull/1485.

Can you please slip this specific one? We can merge this one at first, then debate the rest

Zhiqiang520: > Only this check seems reasonable. It align with #1485.
> 
> Can you please slip this specific one? We can merge this one at first, then debate the rest

Done! I have updated the PR to only fix this specific one.

Issue #1498, opened 2022-12-08
uint8_t rtt might be too small.


jyao1:
Current definition `uint8_t rtt;`. That means it only supports 255 microsecond as max value.

I think we need use `uint64_t rtt`.

PR #1507, opened 2022-12-12
Change rtt size

Wenxing-hou:
Fix: #1498 

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -22,7 +22,7 @@ typedef struct {
 
 typedef struct {
     uint8_t ct_exponent;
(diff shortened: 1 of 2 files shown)


jyao1:
why you use uint32_t instead of uint64_t?

jyao1 (changes requested):
I think you need move slot_id fix to a dedicated PR - https://github.com/DMTF/libspdm/pull/1516

Don't mix with RTT fix, please.

Wenxing-hou:
> I think you need move slot_id fix to a dedicated PR - #1516
> 
> Don't mix with RTT fix, please.

Thanks. This PR will pass the CI when the PR: #1516.

PR #1377, opened 2022-11-08
NEGOTIATE_ALGORITHM Coverage

steven-bellock:
This pull request moves `NEGOTIATE_ALGORITHM` error testing to its own file, and expands test coverage. This pull request also adds a new function, `libspdm_onehot0`, that checks that bitmasks have at most one bit set.

Signed-off-by: Steven Bellock <sbellock@nvidia.com>

PR #1509, opened 2022-12-12
Return error if CHALLENGE is sent in a secure sesion

steven-bellock:
Fixes #1377.

Note that a separate issue / pull request will be filed for other requests.


library/spdm_responder_lib/libspdm_rsp_challenge_auth.c
```diff
@@ -44,15 +44,19 @@ libspdm_return_t libspdm_get_response_challenge_auth(void *context,
             spdm_request->header.request_response_code,
             response_size, response);
     }
+    if (spdm_context->last_spdm_request_session_id_valid) {
```
steven-bellock, line 47: @jyao1 this is the main change of this pull request.

PR #1512, opened 2022-12-13
Cleanup VCA

steven-bellock:
This pull request weakens the strict request size checks done by the Responder in VCA. It also adds the "new" GET_CAPABILITIES checks from the Requester.

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_capabilities.c
+++ b/library/spdm_responder_lib/libspdm_rsp_capabilities.c
@@ -35,7 +35,7 @@ static bool libspdm_check_request_version_compability(libspdm_context_t *spdm_co
 }
 
 /**
- * This function checks the compability of the received CAPABILITES flag.
+ * This function checks the compability of the received GET_CAPABILITES flag.
  * Some flags are mutually inclusive/exclusive.
  *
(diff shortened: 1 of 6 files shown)


library/spdm_responder_lib/libspdm_rsp_capabilities.c
```diff
@@ -154,23 +145,20 @@ libspdm_return_t libspdm_get_response_capabilities(void *context,
                                                response_size, response);
     }
     if (spdm_request->header.spdm_version >= SPDM_MESSAGE_VERSION_12) {
-        if (request_size != sizeof(spdm_get_capabilities_request_t)) {
+        if (request_size > sizeof(spdm_get_capabilities_request_t)) {
```
jyao1, line 148: why bigger is problem?

I think the general flow is:
```
    if (size < sizeof(xxx_request_t)) {
        return libspdm_generate_error_response ();
    } else {
        size = sizeof(xxx_request_t);
    }
```

steven-bellock: Thanks. Got my direction wrong.

jyao1 (approved):

jyao1 (approved):

Issue #1515, opened 2022-12-14
When the slot_id is 0xFF, the local_context.provisioned_slot_id should be 0-7 in all code


Wenxing-hou:
When the slot_id is 0xFF, the `local_context.provisioned_slot_id` should be 0-7 and the `local_context.local_cert_chain_provision[local_context.provisioned_slot_id]` should not be NULL.

The example is:
https://github.com/DMTF/libspdm/blob/3d5f485b75c410a4e4176a836dc20beab89a1158/library/spdm_responder_lib/libspdm_rsp_key_exchange.c#L224-L243

PR #1516, opened 2022-12-14
Fix the provisioned slot_id

Wenxing-hou:
Fix: #1515


jyao1:
It seems spdm_context->local_context.provisioned_slot_id is provisioned in a wrong way.

It cannot be 0xFF by definition.
```
    /* My provisioned certificate (for slot_id - 0xFF, default 0)*/
    uint8_t provisioned_slot_id;
```

I think we need fix the test case libspdm_test_responder_key_exchange_case13() to not use 0xFF.

Wenxing-hou:
> It seems spdm_context->local_context.provisioned_slot_id is provisioned in a wrong way.
> 
> It cannot be 0xFF by definition.
> 
> ```
>     /* My provisioned certificate (for slot_id - 0xFF, default 0)*/
>     uint8_t provisioned_slot_id;
> ```
> 
> I think we need fix the test case libspdm_test_responder_key_exchange_case13() to not use 0xFF.

Thanks. I have deleted the `libspdm_test_responder_key_exchange_case13`.

Issue #1518, opened 2022-12-14
no way to provision provisioned_slot_id


jyao1:
Current libspdm uses cert_chain as public key.

The `provisioned_slot_id` is used when the slot_id 0xFF is required.

However, the code does not have a way to set `provisioned_slot_id` . It is also 0.

We may need add `LIBSPDM_DATA_LOCAL_PUBLIC_CERT_CHAIN_DEFAULT_SLOT_ID` to `libspdm_data_type_t` .
The valid value is uint8_t from 0 to 7, when local_cert_chain is provisioned.

That will match existing logic such as :
```
    if (slot_id == 0xFF) {
        spdm_response->header.param2 = 0;

        slot_id = spdm_context->local_context.provisioned_slot_id;
```

jyao1:
Also need fix spdm_emu to provision the default SLOT_ID.

steven-bellock:
When the Requester has the raw public key of the Responder, then ideally the `slot_id` parameter would be ignored. For example in https://github.com/DMTF/libspdm/blob/052ea76c8d7bae021e26f620901a4607c6d8a1b0/include/library/spdm_requester_lib.h#L468-L474 This would also prevent `0xf` versus `0xff` confusion.

PR #1521, opened 2022-12-15
Add libspdm_set_data for provisioned_slot_id

Wenxing-hou:
Fix: #1518

### DIFF
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -93,6 +93,8 @@ typedef enum {
     /* VCA cached for CACHE_CAP in 1.2 for transcript. */
     LIBSPDM_DATA_VCA_CACHE,
 
+    /* My provisioned certificate slot_id (for slot_id - 0xFF, default 0)*/
+    LIBSPDM_DATA_LOCAL_PUBLIC_CERT_CHAIN_DEFAULT_SLOT_ID,
     /* MAX */
     LIBSPDM_DATA_MAX
 } libspdm_data_type_t;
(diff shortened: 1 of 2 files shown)


doc/user_guide.md
```diff
@@ -386,7 +391,12 @@ Refer to spdm_server_init() in [spdm_responder.c](https://github.com/DMTF/spdm-e
    }
    ```
 
-   1.6, if PSK is required, optionally deploy PSK Hint.
+   1.6, if public key of responder was provisoned for requester, the provisioned_slot_id is which slot_id cert_chain was provsioned for requester.
+   ```
+   libspdm_set_data (spdm_context, LIBSPDM_DATA_LOCAL_PUBLIC_CERT_CHAIN_DEFAULT_SLOT_ID, NULL, provisioned_slot_id, sizeof(provisioned_slot_id));
+   ```
+
+   1.7, if PSK is required, optionally deploy PSK Hint.
```
jyao1, line 399: This text is confusing. I suggest no update, so far.

We have another big patch to support raw public key provision. Will discuss later.

Please drop the user_guide update in this patch.

Wenxing-hou: Thanks. I have deleted the user_guide change.

steven-bellock:
So before this change folks had to provision the raw public key (leaf certificate) in slot 0? Now they can choose any slot?

jyao1:
> So before this change folks had to provision the raw public key (leaf certificate) in slot 0? Now they can choose any slot?

Right. It is a bug.

steven-bellock:
Why's that a bug? Why would they want to provision the raw public key in, for example, slot 4?

jyao1:
how can you assume that public key is also from slot 0?

steven-bellock:
On the Responder's side, these slots aren't visible to the Requester, since `CERT_CAP==0`. When raw public keys are used libspdm is just leveraging slots to make things easier. So the Integrator shouldn't care which internal "slot" libspdm stores the public key in.

jyao1:
The integrator of requester does not care, which real slot is used by when slot 0xFF is used, because the public cert is already provisioned. That is correct.

But the integrator of the responder needs to know that, because it must choose a right public cert, when slot 0xFF is used, to **create the transcript**. It cannot choose arbitrary public cert, because it will fail the transcript hash.

steven-bellock (approved):

PR #1525, opened 2022-12-17
Fix the MACRO for ENCAP_CAP and MUT_AUTH_CAP

Wenxing-hou:
Skip the req_slot_id check when disable mut_cap and encap;


jyao1 (approved):

jyao1 (changes requested):
Should we move the check to above line 432?

PR #1532, opened 2022-12-19
add LIBSPDM_FFDHE_SUPPORT check

jyao1:


steven-bellock:
I can add checks for the other macros.

Issue #1523, opened 2022-12-15
Need a way to set is_requester in case of VCA cache


jyao1:
Currently `is_requester` is set in get_version() call. See https://github.com/DMTF/libspdm/pull/1451.

If VCA cache is enabled, then get_version is not invoked.
In this case, we need a way to set `is_requester` parameter.

It seems we need expose an interface to allow integrator setup `LIBSPDM_DATA_IS_REQUESTER`.

jyao1:
We also need fix spdm_emu to add is_requester for VCA cache.

PR #1539, opened 2022-12-20
Add is_requester config setting.

jyao1:
Fix: https://github.com/DMTF/libspdm/issues/1523

spdm_emu fix: https://github.com/DMTF/spdm-emu/pull/192

### DIFF
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -95,6 +95,14 @@ typedef enum {
 
     /* My provisioned certificate slot_id (for slot_id - 0xFF, default 0)*/
     LIBSPDM_DATA_LOCAL_PUBLIC_CERT_CHAIN_DEFAULT_SLOT_ID,
+
+    /* if the context is for a requester. It only needs to be set in VCA cache.
+     * In normal flow, the value is set in GET_VERSION or VERSION automatically.
+     * false means responder
+     * true means requester
+     **/
+    LIBSPDM_DATA_IS_REQUESTER,
+
     /* MAX */
     LIBSPDM_DATA_MAX
 } libspdm_data_type_t;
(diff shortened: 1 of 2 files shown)


jyao1:
> * Update documentation to say that this is required.

I have concern to make it required in normal case. That will become a incompatible change.

steven-bellock:
> > * Update documentation to say that this is required.
> 
> I have concern to make it required in normal case. That will become a incompatible change.

Ok we can save it for 3.0.

include/library/spdm_common_lib.h
```diff
@@ -95,6 +95,14 @@ typedef enum {
 
     /* My provisioned certificate slot_id (for slot_id - 0xFF, default 0)*/
     LIBSPDM_DATA_LOCAL_PUBLIC_CERT_CHAIN_DEFAULT_SLOT_ID,
+
+    /* if the context is for a requester. It only needs to be set in VCA cache.
+     * In normal flow, the value is set in GET_VERSION or VERSION automatically.
+     * 0x0 means responder
```
steven-bellock, line 101: `false means Responder`
`true means Requester`

jyao1:
> Ok we can save it for 3.0.

Agree.

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -465,6 +465,12 @@ libspdm_return_t libspdm_set_data(void *context, libspdm_data_type_t data_type,
                          sizeof(spdm_context->transcript.message_a.buffer),
                          data, data_size);
         break;
+    case LIBSPDM_DATA_IS_REQUESTER:
+        if (data_size != sizeof(uint8_t)) {
```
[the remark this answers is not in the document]
jyao1: Fixed

include/library/spdm_common_lib.h
jyao1: Fixed

steven-bellock (approved):

Issue #1541, opened 2022-12-20
SlotID 0xFF will trigger assert.


jyao1:
This is because there is no check for slot_id before it is used as index.

PR #1543, opened 2022-12-20
Fix slot 0xFF system crash issue.

jyao1:
Fix: https://github.com/DMTF/libspdm/issues/1541

Also need spdm-emu: https://github.com/DMTF/spdm-emu/pull/194

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_encap_response.c
+++ b/library/spdm_responder_lib/libspdm_rsp_encap_response.c
@@ -187,8 +187,6 @@ void libspdm_init_mut_auth_encap_state(libspdm_context_t *spdm_context, uint8_t 
     libspdm_zero_mem(&spdm_context->encap_context.last_encap_request_header,
                      sizeof(spdm_context->encap_context.last_encap_request_header));
     spdm_context->encap_context.certificate_chain_buffer.buffer_size = 0;
-    spdm_context->response_state = LIBSPDM_RESPONSE_STATE_PROCESSING_ENCAP;
-
 
     /* Clear Cache*/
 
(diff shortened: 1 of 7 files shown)


steven-bellock (approved):

library/spdm_responder_lib/libspdm_rsp_key_exchange.c
```diff
@@ -362,10 +362,16 @@ libspdm_return_t libspdm_get_response_key_exchange(void *context,
...
-        libspdm_init_mut_auth_encap_state(
-            context, spdm_response->mut_auth_requested);
-        spdm_response->req_slot_id_param =
-            (spdm_context->encap_context.req_slot_id & 0xF);
+        if (spdm_response->mut_auth_requested == SPDM_KEY_EXCHANGE_RESPONSE_MUT_AUTH_REQUESTED) {
+            /* no need to libspdm_init_mut_auth_encap_state() because of no ENCAP message */
+            spdm_response->req_slot_id_param =
+                (spdm_context->encap_context.req_slot_id & 0xF);
+        } else {
+            libspdm_init_mut_auth_encap_state(
```
xiangfeima, line 370: Would you please confirm if this should be updated as well?
As far as I can see, the possible sequences should be only
`     *    2.1 GET_DIGEST/GET_CERTIFICATE (spdm_context->encap_context.req_slot_id != 0xFF) */`
`     *    2.2 N/A (SPDM_GET_CAPABILITIES_REQUEST_FLAGS_PUB_KEY_ID_CAP, spdm_context->encap_context.req_slot_id must be 0xFF)`

https://github.com/DMTF/libspdm/blob/befd3a8c41427feb28719db4324ec2f82ed62790/library/spdm_responder_lib/libspdm_rsp_encap_response.c#L199-L226

jyao1: you are right. Will update as well.

jyao1: @xiangfeima, please review the update if possible.

library/spdm_responder_lib/libspdm_rsp_encap_response.c
```diff
@@ -198,30 +196,47 @@ void libspdm_init_mut_auth_encap_state(libspdm_context_t *spdm_context, uint8_t
 
     /* Possible Sequence:
      * 2. Session Mutual Auth: (spdm_context->last_spdm_request_session_id_valid)
-     *    2.1 GET_DIGEST/GET_CERTIFICATE (spdm_context->encap_context.req_slot_id != 0xFF)
-     *    2.2 GET_DIGEST (spdm_context->encap_context.req_slot_id == 0xFF)
-     *    2.3 N/A (SPDM_GET_CAPABILITIES_REQUEST_FLAGS_PUB_KEY_ID_CAP)*/
+     *    2.1 GET_DIGEST/GET_CERTIFICATE (MUT_AUTH_REQUESTED_WITH_ENCAP_REQUEST, encap_context.req_slot_id must not be 0xFF)
+     *    2.2 GET_DIGEST (MUT_AUTH_REQUESTED_WITH_GET_DIGESTS, encap_context.req_slot_id must not be 0xFF)
```
xiangfeima, line 200: Hi @jyao1 , I don't see how this 2.2 sequence comes here.
According to SPDM specification v1.2, section 10.22.2 Optimized encapsulated request flow, paragraph 610, the `Optimized encapsulated request flow` (case `SPDM_KEY_EXCHANGE_RESPONSE_MUT_AUTH_REQUESTED_WITH_GET_DIGESTS`) just makes the Requester be able to skip the initial `GET_ENCAPSULATED_REQUEST` to get encapsulated `GET_DIGESTS` request from the Responder. And the encapsulated `GET_CERTIFICATE` flow afterwards may still needed, as the Figure 22 — Optimized session-based mutual authentication example shows.

Though the SDPM specification dosen't have clear description about that encapsulated `GET_CERTIFICATE` flow can be skipped in mutual authentication process if encapsulated `GET_DIGESTS` result is good, just as same as the one-way authentication flow does, I'm just guessing the idea here is to handle such scenario. However I think we would never know it at here at this moment but after the encapsulated `DIGESTS` response is processed by the Responder.

jyao1: you are right.  my mistake. Flow is updated.

xiangfeima:
Looks all good now 👍

Issue #1542, opened 2022-12-20
Add check no more than one bit shall be set for the opaque data format in ALGORITHMS response message.


Zhiqiang520:
1. Refer to https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf
At paragraph 261 
Table 21 — Successful **ALGORITHMS response** message format
**OtherParamsSelection** field :
Selected Parameter Bit Mask. The Responder shall select one of the opaque data
formats supported by the Requester. Thus, **no more than one bit shall be set for
the opaque data format.**
• Bit [3:0]. See Opaque Data Format Support and Selection Table.
• Bit [7:4]. Reserved.

2. I think we should add check for **OtherParamsSelection** field like the following **MeasurementSpecificationSel** at line 192.
https://github.com/DMTF/libspdm/blob/a732e90c4df1863e4fe6e8c75a8559a036fefa5c/library/spdm_requester_lib/libspdm_req_negotiate_algorithms.c#L192-L195

steven-bellock:
This also needs a unit test to trigger the check in https://github.com/DMTF/libspdm/blob/main/unit_test/test_spdm_requester/error_test/negotiate_algorithms_err.c

Zhiqiang520:
I will check it.

PR #1545, opened 2022-12-21
Add unit test for OtherParamsSelection has more than one bit set.

Zhiqiang520:
Fix: #1542

### DIFF
--- a/unit_test/test_spdm_requester/error_test/negotiate_algorithms_err.c
+++ b/unit_test/test_spdm_requester/error_test/negotiate_algorithms_err.c
@@ -123,6 +123,8 @@ static libspdm_return_t libspdm_requester_negotiate_algorithms_test_send_message
         return LIBSPDM_STATUS_SUCCESS;
     case 0x29:
         return LIBSPDM_STATUS_SUCCESS;
+    case 0x2A:
+        return LIBSPDM_STATUS_SUCCESS;
     default:
         return LIBSPDM_STATUS_SEND_FAIL;
     }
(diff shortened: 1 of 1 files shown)


library/spdm_requester_lib/libspdm_req_negotiate_algorithms.c
```diff
@@ -193,6 +193,12 @@ static libspdm_return_t libspdm_try_negotiate_algorithms(libspdm_context_t *spdm
         status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
         goto receive_done;
     }
+    if (spdm_response->header.spdm_version >= SPDM_MESSAGE_VERSION_12) {
+        if (!libspdm_onehot0(spdm_response->other_params_support)) {
```
steven-bellock, line 197: The name of the response field should be `other_params_selection` but I'll file a separate issue for that.

steven-bellock (approved):

library/spdm_requester_lib/libspdm_req_negotiate_algorithms.c
jyao1, line 197: it should be `if (!libspdm_onehot0(spdm_response->other_params_support & 0xF)) {`

The upper bit are reserved and should be ignored.

Zhiqiang520: The line 486~488 check the opaque data format must be **OpaqueDataFmt1**, but this PR is to check **no more than one bit** shall be set for the opaque data format.
https://github.com/DMTF/libspdm/blob/befd3a8c41427feb28719db4324ec2f82ed62790/library/spdm_requester_lib/libspdm_req_negotiate_algorithms.c#L485-L488

jyao1: If it is only Fmt1, that already meets no more than one bit requirement.

I dont think more check is needed.

Zhiqiang520: ok, no more check is needed, and just add unit test for OtherParamsSelection has more than one bit set.

PR #1556, opened 2022-12-26
Add raw public key provisioning.

xiangfeima:
Fix: #1328
This pull request implements DER encoded raw public key parsing and identity authentication (one-way or mutual) when raw public key is provisioned instead of using chertificate chain.


include/internal/libspdm_common_lib.h
```diff
@@ -630,6 +641,20 @@ bool libspdm_verify_certificate_chain_hash(libspdm_context_t *spdm_context,
...
+ * @param  spdm_context            A pointer to the SPDM context.
+ * @param  public_key_hash         The public key hash data buffer.
+ * @param  public_key_hash_size    size in bytes of the public key hash data buffer.
+ *
+ * @retval true  hash verification pass.
+ * @retval false hash verification fail.
+ **/
+bool libspdm_verify_public_key_hash(libspdm_context_t *spdm_context,
+                                    const void *public_key_hash,
+                                    size_t public_key_hash_size);
```
steven-bellock, line 656: `public_key_hash_size` isn't needed since this info is in the `spdm_context`.

steven-bellock (changes requested):

include/internal/libspdm_common_lib.h
```diff
@@ -569,18 +568,30 @@ bool libspdm_is_capabilities_flag_supported(const libspdm_context_t *spdm_contex
...
+/**
+ * This function generates the public key hash.
+ *
+ * @param  spdm_context               A pointer to the SPDM context.
+ * @param  hash                       The buffer to store the public key hash.
+ *
+ * @retval true  public key hash is generated.
+ * @retval false public key hash is not generated.
+ **/
+bool libspdm_generate_public_key_hash(libspdm_context_t *spdm_context,
```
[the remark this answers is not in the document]
xiangfeima: The name was designed to be aligned with the function we already have for certificate chain hash generation/verification.
`bool libspdm_generate_cert_chain_hash()` <-> `bool libspdm_generate_public_key_hash()`
`bool libspdm_verify_certificate_chain_hash()` <-> `bool libspdm_verify_public_key_hash()`
Since the public key hash generation/verification process has the same routine as certificate chain does (one only **generates local hash** and **verifies peer hash**), so I tend to keep the function name as the way it is right now.

include/hal/library/cryptlib/cryptlib_ec.h
```diff
@@ -22,6 +22,26 @@
  **/
 extern void *libspdm_ec_new_by_nid(size_t nid);
 
+/**
+ * Generates Elliptic Curve context from DER-encoded public key data.
+ *
+ * @param[in]  der_data    Pointer to the DER-encoded public key data.
+ * @param[in]  der_size    Size of the DER-encoded public key data in bytes.
+ * @param[out] ec_context  Pointer to new-generated EC context which contains the
```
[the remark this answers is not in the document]
xiangfeima: All "Pointer to new-generated" phrases will be chaged to "Pointer to the newly generated" as suggested by next patch set.

include/internal/libspdm_common_lib.h
xiangfeima: There is a comparison between `public_key_hash_size` and the `hash_size` from `spdm_context` in the hash verification process (just as same as certificate chain has verification process does), so the `public_key_hash_size` is necessary.

library/spdm_common_lib/libspdm_com_crypto_service.c
```diff
@@ -20,27 +20,19 @@ bool libspdm_get_peer_cert_chain_buffer(void *context,
...
-    libspdm_context_t *spdm_context;
 #if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT
+    libspdm_context_t *spdm_context;
     uint8_t slot_id;
-#endif
 
     spdm_context = context;
-#if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT
     slot_id = spdm_context->connection_info.peer_used_cert_chain_slot_id;
-    LIBSPDM_ASSERT(slot_id < SPDM_MAX_SLOT_COUNT);
```
jyao1, line 31: why you remove this ASSERT?

library/spdm_common_lib/libspdm_com_crypto_service.c
```diff
@@ -74,8 +70,8 @@ bool libspdm_get_peer_cert_chain_data(void *context,
 
     *cert_chain_data = (const uint8_t *)*cert_chain_data + sizeof(spdm_cert_chain_t) + hash_size;
     *cert_chain_data_size = *cert_chain_data_size - (sizeof(spdm_cert_chain_t) + hash_size);
-
-    return true;
+#endif
+    return result;
```
jyao1, line 74: The logic seems return data if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT.

If LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT is 0, then it return FALSE.

Is that possible to align libspdm_get_peer_cert_chain_data() to libspdm_get_peer_cert_chain_buffer()?

library/spdm_common_lib/libspdm_com_crypto_service.c
xiangfeima: It's a miss. Thanks for pointing it out.
I'll fix it by next patch set.

library/spdm_common_lib/libspdm_com_crypto_service.c
xiangfeima: > The logic seems return data if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT.
> If LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT is 0, then it return FALSE.

Yes indeed.
Because we store peer certificate chain data in context only if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT is 1.
https://github.com/DMTF/libspdm/blob/e9e1d0a37a42dd3b91fdf63e8e33f038e98e0d09/include/internal/libspdm_common_lib.h#L44-L54

> Is that possible to align libspdm_get_peer_cert_chain_data() to libspdm_get_peer_cert_chain_buffer()?

I lost you here. Would you please explain this comment more clearly?

library/spdm_common_lib/libspdm_com_crypto_service.c
```diff
@@ -517,54 +581,72 @@ uint8_t libspdm_get_cert_slot_count(libspdm_context_t *spdm_context)
...
-            cert_chain_buffer_hash);
-        if (!result) {
-            LIBSPDM_DEBUG((LIBSPDM_DEBUG_INFO,
-                           "!!! verify_peer_digests - FAIL (hash calculation) !!!\n"));
-            return false;
-        }
+    for (slot_id = 0; slot_id < SPDM_MAX_SLOT_COUNT; slot_id++) {
+        if (slot_mask & (1 << slot_id)) {
+#if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT
+            if (spdm_context->connection_info.peer_used_cert_chain[slot_id].buffer_size != 0) {
```
jyao1, line 610: Originally, this API is to check the runtime data v.s. local_context.

But, the local_context.peer_cert_chain_provision is moved. The check is change to runtime data v.s. connection_info.

I dont believe that is the expectation.

I recommend we remove libspdm_verify_peer_digests() API at all, because there is nothing to verify.

library/spdm_common_lib/libspdm_com_crypto_service.c
xiangfeima: I see, the `result` seems to be unnecessary. I'll change the way it reture to be aligned with `libspdm_get_peer_cert_chain_buffer()`.

jyao1: I mean, just change the 
```
libspdm_get_peer_cert_chain_data()
{
#if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT
    ......
#else
    return FALSE;
#endif
}
```

I dont see the need to define result and assign to FALSE, then return result, if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT==0

library/spdm_common_lib/libspdm_com_crypto_service.c
```diff
@@ -882,38 +941,53 @@ bool libspdm_verify_certificate_chain_hash(libspdm_context_t *spdm_context,
             LIBSPDM_DEBUG((LIBSPDM_DEBUG_INFO, "!!! verify_certificate_chain_hash - FAIL !!!\n"));
             return false;
         }
+    }
```
jyao1, line 944: It seems you removed "else" logic.
Do you assume buffer_hash_size is always !=0?
If yes, I suggest to add ASSERT().

xiangfeima: > It seems you removed "else" logic.

Yes. The "else" logic was trying to read provisioned peer certificate chain from context.
It is no longer needed since local_context.peer_cert_chain_provision is gone.

> Do you assume buffer_hash_size is always !=0?
> If yes, I suggest to add ASSERT().

Yes, buffer_hash_size must not be 0 here.
Because this function is called in CHALLENGE process,
the certificate chain hash must be there from GET_CERTIFICATE process or imported from context cache.
I'll add the assertion.

library/spdm_responder_lib/libspdm_rsp_key_exchange.c
```diff
@@ -238,11 +254,8 @@ libspdm_return_t libspdm_get_response_key_exchange(void *context,
         }
     }
```
jyao1, line 256: Since slot_id maybe 0xFF, do we need similar check above?
E.g. check to ensure spdm_context->local_context.local_public_key_provision

library/spdm_responder_lib/libspdm_rsp_challenge_auth.c
```diff
@@ -158,8 +158,6 @@ libspdm_return_t libspdm_get_response_challenge_auth(void *context,
 
     if (slot_id == 0xFF) {
         spdm_response->header.param2 = 0;
-
-        slot_id = spdm_context->local_context.provisioned_slot_id;
```
jyao1, line 162: Since slot_id maybe 0xFF, do we need similar check above line 89?
E.g. check to ensure spdm_context->local_context.local_public_key_provision

os_stub/cryptlib_mbedtls/der/der.c
```diff
@@ -0,0 +1,189 @@
+/**
+ *  Copyright Notice:
+ *  Copyright 2021-2022 DMTF. All rights reserved.
+ *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
+ **/
+
+/** @file
+ * DER (Distinguished Encoding Rules) format Handler Wrapper Implementation.
```
jyao1, line 8: Is that possible to add some crypto unit test for the new APIs in https://github.com/DMTF/libspdm/tree/main/unit_test/test_crypt?

unit_test/test_spdm_requester/encap_request.c
```diff
@@ -73,7 +73,7 @@ libspdm_return_t libspdm_requester_encap_request_test_receive_message(
         if (sub_index == 0)
         {
             temp_buf_size = sizeof(spdm_digest_response_t) +
-                            libspdm_get_hash_size(m_libspdm_use_hash_algo) * SPDM_MAX_SLOT_COUNT +
+                            libspdm_get_hash_size(m_libspdm_use_hash_algo) +
```
jyao1, line 76: why we need change this? Is it related to raw public key provisioning?

unit_test/test_spdm_requester/encap_request.c
```diff
@@ -93,12 +93,9 @@ libspdm_return_t libspdm_requester_encap_request_test_receive_message(
                             (uint8_t)(0xFF));
 
             digest = (void *)(spdm_response + 1);
-            /*send all eight certchains digest
-             * but only No.7 is right*/
-            digest += libspdm_get_hash_size(m_libspdm_use_hash_algo) * (SPDM_MAX_SLOT_COUNT - 2);
             libspdm_hash_all(m_libspdm_use_hash_algo, m_libspdm_local_certificate_chain,
                              LIBSPDM_MAX_MESSAGE_BUFFER_SIZE, &digest[0]);
-            spdm_response->header.param2 |= (0xFF << 0);
+            spdm_response->header.param2 |= (0x01 << 0);
```
jyao1, line 98: why we need change this? Is it related to raw public key provisioning?

jyao1:
@xiangfeima , thanks to contribute this patch. I think the overall design is clear.
I have finished 1st round review and provided all feedback for the implementation.


Since the code change is big, I would like to double check if there is enough validation for the feature.
1) enable spdm_emu to test 0xFF (I think you already have that, great)
2) enable spdm_dump (https://github.com/DMTF/spdm-dump) to test 0xFF (I dont see PR there).
3) test new spdm_requester_emu(slot - FF) with new spdm_responder_emu(slot - FF) - I assume you have done.
4) test new spdm_requester_emu(slot - non-FF) with new spdm_responder_emu(slot - non-FF) - I assume you have done.
5) test old spdm_requester_emu(slot - non-FF) with new spdm_responder_emu(slot - non-FF) - this is to ensure change will not break the existing transcript.
6) test new spdm_requester_emu(slot - non-FF) with old spdm_responder_emu(slot - non-FF) - this is to ensure change will not break the existing transcript.

library/spdm_common_lib/libspdm_com_crypto_service.c
xiangfeima: In my opinion the libspdm_verify_peer_digests() was supposed to be checking runtime retrieved digests data.
Refer to SPDM specification v1.2 `Figure 8 — Responder authentication: Example certificate retrieval flow`, 
> 2. Compare digests in DIGESTS response
message to cached digests. _**Continue if
no match is found**_.

The word **_Continue_** here in my understanding is to perform `GET_CERTIFICATE` afterwards.
Thus as I see the use cases of libspdm_verify_peer_digests() are still there:
1. First time `GET_DIGESTS`. The libspdm_verify_peer_digests() should return FALSE (since there is nothing in local context to verify with the retrieved digests), indicate people to perform `GET_CERTIFICATE` (Continue if
no match is found), and cache digest of certificate chain per slot in local context connection_info by the `GET_CERTIFICATE`.
2. `GET_DIGESTS` after `LIBSPDM_CONNECTION_STATE_AUTHENTICATED`. The libspdm_verify_peer_digests() should return actual check result (runtime retrieved digests data v.s. cached digest per slot in local context connection_info), to tell if the peer identity is changed (after a `SET_CERTIFICATE`) or tampered, and a `GET_CERTIFICATE` afterwards may be needed base upon the check result.

Currently `GET_CERTIFICATE` is called only if `GET_DIGESTS` is succeeded so I keep the libspdm_verify_peer_digests() to return TRUE at first time it is called (nothing is cached in local context connection_info). It can be filed to another issue and be resolved later.
https://github.com/DMTF/spdm-emu/blob/9202e51b57abba044387746a926832df8174114d/spdm_emu/spdm_requester_emu/spdm_requester_authentication.c#L42-L54

In summary, I think libspdm_verify_peer_digests() is needed.

library/spdm_responder_lib/libspdm_rsp_key_exchange.c
xiangfeima: No need to.
The slot_id is clearly checked just at the line L224-L239 above.

library/spdm_responder_lib/libspdm_rsp_challenge_auth.c
xiangfeima: No need to.
The slot_id is clearly checked just at the line L81-L95 above.

os_stub/cryptlib_mbedtls/der/der.c
xiangfeima: Sure. I will add TCs accordingly.

include/hal/library/cryptlib/cryptlib_rsa.h
```diff
@@ -35,6 +35,25 @@ typedef enum {
  **/
 extern void *libspdm_rsa_new(void);
 
+/**
+ * Generates RSA context from DER-encoded public key data.
```
jyao1, line 39: DER is just encoding.

I think we need clarify which standard we are following for the RSA public key format.

unit_test/test_spdm_requester/encap_request.c
xiangfeima: This is changed because both of
1. provisioned peer certificate chain in local context is gone
2. the way that libspdm_verify_peer_digests() is changed

Previously the way of testing
> send all eight certchains digest but only No.7 is right
would be succeeded because we were doing it in the way that to see if there is digest of one slot in retrieved digests data matches the hash of provisioned peer chetificate chain in local context.

Since provisioned peer certificate chain in local context is gone,
and we are caching peer certificate chain data/hash slot by slot in `GET_CERTIFICATE` process,
the way to check peer digests in libspdm_verify_peer_digests() is also changed to check the hash value of every slot according to given slot_mask.

And due to that, `send all eight certchains digest but only No.7 is right` way of testing won't be succeeded any more.
Now pass condition for `GET_DIGESTS` should be: hash value for every slot is correct against the given slot_mask.

unit_test/test_spdm_requester/encap_request.c
xiangfeima: This is changed because both of

provisioned peer certificate chain in local context is gone
the way that libspdm_verify_peer_digests() is changed
Previously the way of testing

send all eight certchains digest but only No.7 is right

would be succeeded because we were doing it in the way that to see if there is digest of one slot in retrieved digests data matches the hash of provisioned peer chetificate chain in local context.

Since provisioned peer certificate chain in local context is gone,
and we are caching peer certificate chain data/hash slot by slot in GET_CERTIFICATE process,
the way to check peer digests in libspdm_verify_peer_digests() is also changed to check the hash value of every slot according to given slot_mask.

And due to that, send all eight certchains digest but only No.7 is right way of testing won't be succeeded any more.
Now pass condition for GET_DIGESTS should be: hash value for every slot is correct against the given slot_mask

include/hal/library/cryptlib/cryptlib_rsa.h
xiangfeima: The description is added to doc/user_guide.md
> The public key is ASN.1 DER-encoded as [RFC7250](https://www.rfc-editor.org/rfc/rfc7250) describes,
    namely, the `SubjectPublicKeyInfo` structure of a X.509 certificate.

I can add such comment to here as well if you insist.

xiangfeima:
Hi @jyao1 , thanks for your time to go through such many code change.
About the 6 points you listed above for the feature validation, I will update the result here alone with I update next patch set to fix the comments from 1st round review.

library/spdm_responder_lib/libspdm_rsp_key_exchange.c
jyao1: See comment in https://github.com/DMTF/libspdm/pull/1556#discussion_r1062308490

xiangfeima: Point is taken, I will fix it by next patch set.

library/spdm_responder_lib/libspdm_rsp_challenge_auth.c
xiangfeima: Point is taken, I will fix it by next patch set.

library/spdm_common_lib/libspdm_com_crypto_service.c
jyao1: > First time GET_DIGESTS. The libspdm_verify_peer_digests() should return FALSE

This is not the purpose, when we designed libspdm_verify_peer_digests().
The expectation of verify_xxx to check if the data from the peer meets the basic requirement before return the consumer.
As such, we only verify if it matches the provisioned data. If no data is provisioned, this API should just return TRUE, because there is nothing to verify.

> GET_DIGESTS after LIBSPDM_CONNECTION_STATE_AUTHENTICATED. The libspdm_verify_peer_digests() should return actual check result

This is also not the purpose of verify_xxx API. Thinking of runtime update use case, the change of digest is a valid condition. That helps the consumer to know that the cached digest is no longer valid. The requester should send get_certificate again to get the latest result. 



To me, what you describe above is something like is_cached_data_valid() for runtime data, instead of verify_xxx() according to the provisioned data.
My suggestion is: Don't mix them in raw public key provision patch. Just remove libspdm_verify_peer_digests() in this patch. If you really want to add is_cached_data_valid(), let's use another issue and PR.

xiangfeima: This explanation helps a lot, thank you.
Now I fully understand the design of `libspdm_verify_peer_digests()` is only to verify if the retrieved digest matches provisioned data before return it to consumer.
I will remove `libspdm_verify_peer_digests()` by next patch set as you suggested since it is no longer needed.

Also, there's not much need to do such `is_cached_data_valid()` thing since it would be consumer's call to send get_certificate (or not) afterwards.

unit_test/test_spdm_requester/encap_request.c
xiangfeima: I will remain the way to generate digest in test routine as the way it is right now (only send correct digest value of one slot for most cases), because the previous way of digest generation
> send all eight certchains digest but only No.7 is right

since `libspdm_verify_peer_digests()` will be removed at all.
Refer to https://github.com/DMTF/libspdm/pull/1556#discussion_r1063003386

unit_test/test_spdm_requester/encap_request.c
xiangfeima: I will remain the way to generate digest in test routine as the way it is right now (only send correct digest value of one slot for most cases), because the previous way of digest generation
> send all eight certchains digest but only No.7 is right

is no longer needed, since `libspdm_verify_peer_digests()` will be removed at all.
Refer to https://github.com/DMTF/libspdm/pull/1556#discussion_r1063003386

xiangfeima:
@jyao1 , all comments from your 1st round review are fixed by https://github.com/DMTF/libspdm/commit/44dde488c819719ab6571450403f2cbd5d9539d7.

The status of items for feature validation are updated as below:
1. enable spdm_emu to test 0xFF  **-> https://github.com/DMTF/spdm-emu/pull/204**
2. enable spdm_dump (https://github.com/DMTF/spdm-dump) to test 0xFF **-> In progress**
3. test new spdm_requester_emu(slot - FF) with new spdm_responder_emu(slot - FF) **-> Tested and confirmed working well**
4. test new spdm_requester_emu(slot - non-FF) with new spdm_responder_emu(slot - non-FF) **-> Tested and confirmed working well**
5. test old spdm_requester_emu(slot - non-FF) with new spdm_responder_emu(slot - non-FF) - this is to ensure change will not break the existing transcript. **-> Tested and confirmed working well**
6. test new spdm_requester_emu(slot - non-FF) with old spdm_responder_emu(slot - non-FF) - this is to ensure change will not break the existing transcript. **-> Tested and confirmed working well**

jyao1 (approved):

jyao1:
Thanks @xiangfeima, I approved it.

Once you finish 2) the spdm_dump, we can merge it.

jyao1:
@xiangfeima, I will merge this now, since we have more pending P-R, which may cause conflict.

Please continue working on spdm_dump and submit patch there. Thank you!

xiangfeima:
@jyao1 thank you. I'll update the patch to spdm-dump soon.
(The patch itself is ready, the last thing I'm working on is to finish some internal compliance process.)

xiangfeima:
@jyao1 The patch for spdm-dump is update to https://github.com/DMTF/spdm-dump/pull/75

jyao1:
All merged

xiangfeima:
Thank you 👍

Issue #1555, opened 2022-12-26
libspdm_sleep() should use microsecond


jyao1:
Currently, libspdm_sleep() is used RDTExponent.

According to the spec, RDTExponent is microsecond unit.

But libspdm_sleep is milliseconds unit. It should be changed to microsecond unit.

For the implementation that only supports millisecond unit, it may use below:
```
void libspdm_sleep(uint64_t microsecond)
{
    uint64_t milliseconds;

    milliseconds = (microsecond + 999) / 1000;
    Sleep((DWORD)milliseconds);
}
```

steven-bellock:
At this time I think the main requirement is that libspdm can support sensible values of WT<sub>Max</sub>, which is `(RDT * RDTM) - RTT`. I can't imagine WT<sub>Max</sub> would ever be less than a millisecond, so for now it should be fine to keep `libspdm_sleep` at milliseconds.

PR #1557, opened 2022-12-26
define libspdm_sleep_in_us to sleep microseconds.

jyao1:
Deprecate libspdm_sleep which sleeps milliseconds.

Fix: https://github.com/DMTF/libspdm/issues/1555


jyao1:
@steven-bellock , can you please review this as well?

steven-bellock:
I already commented in #1555. We can discuss this at the meeting on Monday.

steven-bellock (approved):

PR #1565, opened 2023-01-02
Remove function pointers for asymmetric cryptography in spdm_crypt_lib

steven-bellock:


include/library/spdm_crypt_lib.h
```diff
@@ -346,7 +346,7 @@ void libspdm_asym_free(uint32_t base_asym_algo, void *context);
  * @retval  true   Valid asymmetric signature.
  * @retval  false  Invalid asymmetric signature or invalid asymmetric context.
  **/
-bool libspdm_asym_verify(
+bool libspdm_asym_verify_transcript(
```
jyao1, line 349: why change API name?

steven-bellock: Ok. We can save that for Q1 2023.

PR #1569, opened 2023-01-03
Remove function pointers from DHE cryptography in spdm_crypt_lib

### LINKED ISSUE #1452: Remove function pointers from spdm_crypt_lib
Follow-up to #259.

Using AEAD as an example below, the logic in `libspdm_get_aead_enc_func` can be moved to `libspdm_aead_encryption`.

https://github.com/DMTF/libspdm/blob/c9aed77adfc8eaf1d15e4514acb8a3a58a70c4bb/library/spdm_crypt_lib/libspdm_crypt_aead.c#L107-L170

steven-bellock:
Fixes #1452.

### DIFF
--- a/library/spdm_crypt_lib/libspdm_crypt_dhe.c
+++ b/library/spdm_crypt_lib/libspdm_crypt_dhe.c
@@ -7,65 +7,6 @@
 #include "internal/libspdm_crypt_lib.h"
 
 /**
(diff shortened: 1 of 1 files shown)


steven-bellock:
@jyao1 Please check that the sm2 code looks good.

jyao1 (approved):

Issue #1260, opened 2022-10-05
[Feature request] Support FIPS 140-3 level 1 compliance


xiaoyuruan:
FIPS 140-3 is a US government government standard for crypto and security modules.  Some open source libraries are certified for compliance with FIPS 140-3 level 1.  For example, OpenSSL 3.0 achieved FIPS 140-3 level 1 certificate (https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/4282).  

Though libspdm does not have to acquire FIPS 140-3 certificate, compliance with FIPS 140-3 makes easier for adopters of the libspdm to achieve FIPS 140-3 certificates for their products.

Detailed requirements to make libspdm FIPS compliant and design suggestions:

1. **FIPS SKU**: Create a "FIPS" configuration for compile / build and hence a FIPS SKU of libspdm.  Following requirements apply only for the FIPS SKU of libspdm.
2. **zeroization**: explicitly zeroize (memset_s to 0) within the function as as soon as a variable holding key (including public key) is no longer needed.  libspdm_reset_context() API should zeroize all keys in context (I believe it is already doing so).
3. **Only Approved algorithms**: Only allow use of NIST-Approved algorithms, listed at https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-140Cr1.pdf.
4. **module_id API**: libspdm exposes an API that returns version number of the libspdm build.
5. **run_selftests API**: libspdm exposes an API that triggers running of all selftests and returns the results.
6. **selftest_results API**: libspdm exposes an API that returns results of last selftests.
7. **integrity selftest**: at libspdm load (e.g., DLLmain()), read the binary of libspdm and underlying crypto library (if applicable) that's loaded to memory and calculate HMAC-SHA256 (using the function in crypto lib) of it, with a hardcoded key (could be 32-bytes of 0x00's).  Compare the resulted 32-byte HMAC with a reference value, which is calculated at build and shipped in a separate file shipped alongside libspdm.  If they do not match, then disable all APIs except module_id and selftest_results (which shows the integrity selftest failure).
8. **pre-operational selftest**: libspdm performs pre-operational selftests before first use of the algorithm (or at load).  see Section 10 of FIPS Implementation Guidance (https://csrc.nist.gov/CSRC/media/Projects/cryptographic-module-validation-program/documents/fips%20140-3/FIPS%20140-3%20IG.pdf).
[...]

PR #1577, opened 2023-01-04
Make only approved algo when in FIPS mode

Wenxing-hou:
Reference: #1260

The approved algo is listed at: https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-140Cr1.pdf


steven-bellock:
This logic should be in https://github.com/DMTF/libspdm/blob/main/include/internal/libspdm_lib_config.h

Wenxing-hou:
> This logic should be in https://github.com/DMTF/libspdm/blob/main/include/internal/libspdm_lib_config.h In addition https://github.com/DMTF/libspdm/blob/main/include/internal/libspdm_macro_check.h should check that at least one FIPS algorithm is supported.

Thanks. I have move the code to internal/libspdm_lib_config.h. 
And I have submit the PR https://github.com/DMTF/libspdm/pull/1581 to check that at least one FIPS algorithm is supported.

steven-bellock (approved):

PR #1590, opened 2023-01-06
Zero key for FIPS

Wenxing-hou:
Reference: #1260

**2023/1/7**  Commit 1: Zero key for FIPS when encode and decode.
**2023/1/13**  Commit 2: Fix unit test because of the salt change.
**2023/1/9**   Commit 3: Zero salt1 in `libspdm_generate_session_data_key`.

### DIFF
--- a/library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
+++ b/library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
@@ -49,8 +49,8 @@ libspdm_return_t libspdm_encode_secured_message(
     size_t record_header_size;
     spdm_secured_message_cipher_header_t *enc_msg_header;
     bool result;
-    uint8_t key[LIBSPDM_MAX_AEAD_KEY_SIZE];
-    uint8_t salt[LIBSPDM_MAX_AEAD_IV_SIZE];
+    const uint8_t *key;
+    uint8_t *salt;
     uint64_t sequence_number;
     uint64_t sequence_num_in_header;
     uint8_t sequence_num_in_header_size;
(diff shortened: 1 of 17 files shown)


library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
```diff
@@ -266,6 +266,11 @@ libspdm_return_t libspdm_encode_secured_message(
         LIBSPDM_ASSERT(false);
         return LIBSPDM_STATUS_UNSUPPORTED_CAP;
     }
+
+    /*zero key for security*/
```
jyao1, line 270: we need zero key in *any* return path.

library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
```diff
@@ -576,5 +581,9 @@ libspdm_return_t libspdm_decode_secured_message(
         return LIBSPDM_STATUS_UNSUPPORTED_CAP;
     }
 
+    /*zero key for security*/
```
jyao1, line 584: we need zero key in *any* return path.

steven-bellock:
Why does the key need to be on the stack at all? Why not use 
```
secured_message_context->handshake_secret.request_handshake_encryption_key
```
directly?

jyao1:
Yes, I think we can change to `uint8_t *key`.

library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
Wenxing-hou: Thanks. I have added the salt zero in every return path.

Wenxing-hou:
> Yes, I think we can change to `uint8_t *key`.

Thanks. I have changed the key from array to pointer.

library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
```diff
@@ -573,8 +607,13 @@ libspdm_return_t libspdm_decode_secured_message(
 
     default:
         LIBSPDM_ASSERT(false);
+        /*zero salt for security*/
+        libspdm_zero_mem(salt, LIBSPDM_MAX_AEAD_IV_SIZE);
         return LIBSPDM_STATUS_UNSUPPORTED_CAP;
     }
 
+    /*zero salt for security*/
```
jyao1, line 615: Please remove `/*zero salt for security*/`. No need to say it every where.

Wenxing-hou: Thanks. I have removed the comments except the first comment.

library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
```diff
@@ -49,7 +49,7 @@ libspdm_return_t libspdm_encode_secured_message(
     size_t record_header_size;
     spdm_secured_message_cipher_header_t *enc_msg_header;
     bool result;
-    uint8_t key[LIBSPDM_MAX_AEAD_KEY_SIZE];
+    const uint8_t *key;
     uint8_t salt[LIBSPDM_MAX_AEAD_IV_SIZE];
```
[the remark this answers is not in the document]
jyao1: If so, the XOR sequence_number need to be: `*(uint64_t *)salt ^ (sequence_number - 1) ^ sequence_number`

Please test it between old version and new version.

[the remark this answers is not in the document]
Wenxing-hou: Thanks. I have changed the salt to pointer and added a new commit to fix the unit test.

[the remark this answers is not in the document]
Wenxing-hou: And I have tested pass the emu.
the array salt is old version; the pointer salt is new version. The following test has passed.

```
new version requester <---> new version responder
old version requester <---> new version responder
new version requester <---> old version responder
```

jyao1 (approved):

library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
```diff
@@ -49,8 +49,8 @@ libspdm_return_t libspdm_encode_secured_message(
     size_t record_header_size;
     spdm_secured_message_cipher_header_t *enc_msg_header;
     bool result;
-    uint8_t key[LIBSPDM_MAX_AEAD_KEY_SIZE];
-    uint8_t salt[LIBSPDM_MAX_AEAD_IV_SIZE];
+    const uint8_t *key;
+    uint8_t *salt;
```
steven-bellock, line 53: The salt should stay `const`, like the key.
Also we'll need to declare `uint8_t iv[LIBSPDM_MAX_AEAD_IV_SIZE]`, that actually gets fed into `libspdm_aead_encryption`.

library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
```diff
@@ -129,7 +117,9 @@ libspdm_return_t libspdm_encode_secured_message(
         return LIBSPDM_STATUS_SEQUENCE_NUMBER_OVERFLOW;
     }
 
-    *(uint64_t *)salt = *(uint64_t *)salt ^ sequence_number;
+    if (sequence_number > 0) {
+        *(uint64_t *)salt = *(uint64_t *)salt ^ (sequence_number - 1) ^ sequence_number;
```
steven-bellock, line 121: This (and the 2.3 version) is misleading as once the assignment occurs this value is no longer the `salt`, it's the `IV`. So here I would zero-extend `sequence_number` and XOR it with `salt` and assign it (libspdm_copy_mem) to `iv[]`.

library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
jyao1: I don't think we need define `uint8_t iv[LIBSPDM_MAX_AEAD_IV_SIZE]`. That will go back to original problem.

library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
jyao1: If the name is confusing, I think we can rename to `uint8_t *iv`. I don't see the need to cache the `iv[]`.

steven-bellock:
Ok. Feel free to merge this and I'll file an issue.

steven-bellock (approved):

PR #1593, opened 2023-01-06
Remove the checking of extended_key_usage

### LINKED ISSUE #1592: The checking of extended_key_usage shouldn't be mandatory
Currently the code seems checking extended_key_usage and requires that it exists, which seems not necessary according to the latest spec.

https://github.com/DMTF/libspdm/blob/702be53a73bd1cf962bf89ede58a16607e289a1b/library/spdm_crypt_lib/libspdm_crypt_cert.c#L822-L828

lsun100:
Fixes #1592

extended_key_usage was required in spec 1.1 but has been fixed in 1.1.1. It appears more like an 'errta' fix. This commit removes such checking to be compatible with all the different versions.

### DIFF
--- a/library/spdm_crypt_lib/libspdm_crypt_cert.c
+++ b/library/spdm_crypt_lib/libspdm_crypt_cert.c
@@ -819,15 +819,6 @@ bool libspdm_x509_certificate_check(const uint8_t *cert, size_t cert_size,
         goto cleanup;
     }
 
-    /* 11. extended_key_usage*/
-    value = 0;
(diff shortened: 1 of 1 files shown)


steven-bellock:
@lsun100 Add 
```
Fixes #1592.
```
to the description, so that it closes the issue when the pull request is merged.

lsun100:
> @lsun100 Add
> 
> ```
> Fixes #1592.
> ```
> 
> to the description, so that it closes the issue when the pull request is merged.

Done. Thanks!

PR #1595, opened 2023-01-07
Check context before libspdm_asym_free().

jyao1:


steven-bellock:
Alternatively we could specify that calling a `*free` function with `NULL` pointer is a NOP. Then the caller doesn't have check every time, and that is consistent with the C `free` function and MbedTLS and OpenSSL. 

https://github.com/openssl/openssl/blob/accd3bdd11bd4a69fdba42bbeead28945fe50e56/crypto/rsa/rsa_lib.c#L131-L136
https://github.com/Mbed-TLS/mbedtls/blob/b17803682e268aae4f6fe0a9634693b13a8c95d9/library/rsa.c#L2249-L2252

PR #1596, opened 2023-01-08
Remove multiple callback functions

### LINKED ISSUE #1582: Remove multiple callback function registration
libspdm allows an Integrator to register multiple callback functions for a single callback event. For example : https://github.com/DMTF/libspdm/blob/bd23982247f0e52f5c948f381364bcfebfdb2649/library/spdm_responder_lib/libspdm_rsp_receive_send.c#L844-L862

Unless there's a good reason to keep multiple function registration and execution, we should get rid of that capability and its associated macros, `LIBSPDM_MAX_SESSION_STATE_CALLBACK_NUM`, `LIBSPDM_MAX_CONNECTION_STATE_CALLBACK_NUM`, etc.

steven-bellock:
Fixes #1582.


steven-bellock:
> Could you please squash them into one patch? No need to use 2 commit.

It is done.

PR #1599, opened 2023-01-08
Initial draft for programming environment document

steven-bellock:
Note that assumptions about endianness and alignment have been omitted. If libspdm can support arbitrary endianness and alignment then we can leave that out of this document, else we'll have to document it. Also, I'll be adding more to this document. When I am done we can close the issue.

### DIFF
--- a/doc/programming_environment.md
+++ b/doc/programming_environment.md
@@ -1,0 +1,26 @@
+# Programming Environment
+
+## Core Libraries
+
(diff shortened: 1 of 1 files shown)


jyao1:
Could you please squash them into one patch? No need to use 2 commit.

steven-bellock:
> Could you please squash them into one patch? No need to use 2 commit.

It is done.

PR #1602, opened 2023-01-09
Initial commit for libspdm_check_context

### LINKED ISSUE #1027: Provide a function that validates spdm_context
This is a follow-up to #145.

Populating the `spdm_context` is a multi-step process. Rather than validate inputs as they are entered the flow should be for the integrator to populate the `spdm_context` and then at the end call a `libspdm_context_validator` function that determines if the `spdm_context` is well-formed.

steven-bellock:
More tests can be added as needed.

Fixes #1027.


jyao1:
Could you please squash them into one patch? No need to use 2 commit.

steven-bellock:
It is done.

Issue #1605, opened 2023-01-09
private_pem should explicitly zeroize before free


Zhiqiang520:
1. in `libspdm_responder_data_sign` function, `private_pem` at line 1769 and 1784 should explicitly zeroize before free.
https://github.com/DMTF/libspdm/blob/178bb19c58e40747def7f494d03a62b57d406d06/os_stub/spdm_device_secret_lib_sample/lib.c#L1766-L1771
https://github.com/DMTF/libspdm/blob/178bb19c58e40747def7f494d03a62b57d406d06/os_stub/spdm_device_secret_lib_sample/lib.c#L1784
2. similar case in `libspdm_requester_data_sign` function, `private_pem` at line 1704 and 1719 should explicitly zeroize before free.
https://github.com/DMTF/libspdm/blob/178bb19c58e40747def7f494d03a62b57d406d06/os_stub/spdm_device_secret_lib_sample/lib.c#L1699-L1706
https://github.com/DMTF/libspdm/blob/178bb19c58e40747def7f494d03a62b57d406d06/os_stub/spdm_device_secret_lib_sample/lib.c#L1719
3. similar case in `libspdm_gen_csr` function, `prikey` at line 1099 and 1112 should explicitly zeroize before free.
https://github.com/DMTF/libspdm/blob/178bb19c58e40747def7f494d03a62b57d406d06/os_stub/spdm_device_secret_lib_sample/lib.c#L1096-L1101
https://github.com/DMTF/libspdm/blob/178bb19c58e40747def7f494d03a62b57d406d06/os_stub/spdm_device_secret_lib_sample/lib.c#L1112

Wenxing-hou:
I will check it.

PR #1606, opened 2023-01-09
Zero private key before free for FIPS

Wenxing-hou:
Fix: #1605


os_stub/spdm_device_secret_lib_sample/lib.c
```diff
@@ -1096,6 +1096,8 @@ bool libspdm_gen_csr(uint32_t base_hash_algo, uint32_t base_asym_algo, bool *nee
         result = libspdm_asym_get_private_key_from_pem(
             base_asym_algo, prikey, prikey_size, NULL, &context);
         if (!result) {
+            /*zero private key before free for security*/
```
steven-bellock, line 1099: I would just delete the comments. There's no "security" in `spdm_device_secret_lib_sample/lib.c`.

Wenxing-hou: Thanks. I have deleted the comments.

PR #1608, opened 2023-01-09
Fix bug when checking Requester's capabilities

steven-bellock:
Increment patch number and delete a unit test that used a "random" capabilities flags. The test needs to be replaced with individual and targeted tests.

### DIFF
--- a/VERSION.md
+++ b/VERSION.md
@@ -1,1 +1,1 @@
-libspdm version 2.3.0 (released)
+libspdm version 2.3.1 (released)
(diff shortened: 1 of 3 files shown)


steven-bellock:
https://github.com/steven-bellock/libspdm/actions/runs/3876801191 has the CI/CD run. Currently CI/CD is configured to only run on the `main` branch.

steven-bellock:
@mcsmonk as well.

jyao1 (approved):

jyao1:
we need patch for master as well.

Issue #1617, opened 2023-01-11
Need delete check for new dhe_context in libspdm_try_send_receive_key_exchange


Wenxing-hou:
Need delete check for new `dhe_context` in `libspdm_try_send_receive_key_exchange`  to enhance the code performance.

The follow check need be deleted, because the NULL dhe_context will cause the `libspdm_secured_message_dhe_generate_key` fail. 

We can just check the `libspdm_secured_message_dhe_generate_key` result.
```
    if (dhe_context == NULL) {
        libspdm_release_sender_buffer (spdm_context);
        return LIBSPDM_STATUS_CRYPTO_ERROR;
    }
```
https://github.com/DMTF/libspdm/blob/f1ffe7e094b3b7daf85c4fae2fae6b9d98212d11/library/spdm_requester_lib/libspdm_req_key_exchange.c#L361-L377

Wenxing-hou:
I will check it.

jyao1:
I dont understand why we need it.

Whenever you talk about "performance", please show me the data - how many seconds you have saved?

steven-bellock:
The security reviewers will also wonder why `libspdm_secured_message_dhe_new` isn't checked.

Wenxing-hou:
Thanks. I will close the issue and PR. And I have submitted a new issue:  https://github.com/DMTF/libspdm/issues/1624 to add check in responder lib.

PR #1618, opened 2023-01-11
Delete new dhe_context check for enhance code performance

Wenxing-hou:
Fix: #1617


jyao1:
I dont understand why we need it.

Whenever you talk about "performance", please show me the data - how many seconds you have saved?

Wenxing-hou:
Thanks. I will close the PR, because the PR will make others confused.

Issue #1573, opened 2023-01-03
Allow libspdm's version to be queried at compile / run time


steven-bellock:
This is derived from #1260, and would be good to have even if libspdm doesn't support FIPS 140-3.

jyao1:
This is to expose the module_id API even in non-FIPS mode.

PR #1623, opened 2023-01-12
Add fips_module_id api

Wenxing-hou:
Reference: #1260

Fix: #1573

### DIFF
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -12,6 +12,11 @@
 #include "library/spdm_secured_message_lib.h"
 #include "library/spdm_return_status.h"
 
+#define LIBSPDM_MAJOR_VERSION 0x03
+#define LIBSPDM_MINOR_VERSION 0x00
+#define LIBSPDM_PATCH_VERSION 0x00
+#define LIBSPDM_ALPHA         0xff
+
 /* Connection: When a host sends messgages to a device, they create a connection.
  *             The host can and only can create one connection with one device.
  *             The host may create multiple connections with multiple devices at same time.
@@ -841,5 +846,17 @@ bool libspdm_get_peer_public_key_buffer(void *spdm_context,
 bool libspdm_get_local_public_key_buffer(void *spdm_context,
                                          const void **local_public_key_buffer,
                                          size_t *local_public_key_buffer_size);
+
+/**
+ * byte3 - libspdm major version
+ * byte2 - libspdm minor version
+ * byte1 - libspdm patch version
+ * byte0 - libspdm alpha
+ *         (office release with tag: 0, release candidate with tag: 1, non official release: 0xFF)
+ **/
+uint32_t libspdm_module_version(void);
+
+/*true: FIPS enabled, false: FIPS disabled*/
+bool libspdm_get_fips_mode(void);
 
 #endif /* SPDM_COMMON_LIB_H */
--- a/library/spdm_common_lib/libspdm_com_support.c
+++ b/library/spdm_common_lib/libspdm_com_support.c
@@ -229,3 +229,28 @@ void libspdm_init_managed_buffer(void *m_buffer, size_t max_buffer_size)
 
     libspdm_reset_managed_buffer(m_buffer);
 }
+
+/**
+ * byte3 - libspdm major version
+ * byte2 - libspdm minor version
+ * byte1 - libspdm patch version
+ * byte0 - libspdm alpha
+ *         (office release with tag: 0, release candidate with tag: 1, non official release: 0xFF)
+ **/
+uint32_t libspdm_module_version(void)
+{
+    return (LIBSPDM_MAJOR_VERSION << 24) |
+           (LIBSPDM_MINOR_VERSION << 16) |
+           (LIBSPDM_PATCH_VERSION << 8) |
+           (LIBSPDM_ALPHA);
+}
+
+/*true: FIPS enabled, false: FIPS disabled*/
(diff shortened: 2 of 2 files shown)


include/internal/libspdm_common_lib.h
```diff
@@ -1295,4 +1295,8 @@ static inline bool libspdm_onehot0(uint32_t mask)
     return !mask || !(mask & (mask - 1));
 }
 
+#if LIBSPDM_FIPS_MODE
```
jyao1, line 1298: According to https://github.com/DMTF/libspdm/issues/1573,
1) we need to expose this API even without FIPS mode.
2) we need to expose publicly, not in internal.

library/spdm_common_lib/libspdm_com_support.c
```diff
@@ -229,3 +233,12 @@ void libspdm_init_managed_buffer(void *m_buffer, size_t max_buffer_size)
 
     libspdm_reset_managed_buffer(m_buffer);
 }
+
+#if LIBSPDM_FIPS_MODE
+
+size_t libspdm_module_version(void)
```
jyao1, line 239: @steven-bellock , I think we can use the semantic version, align with libspdm release. 4 bytes in total.

byte3 - libspdm major version
byte2 - libspdm minor version
byte1 - libspdm patch version
byte0 - FIPS mode enable/disable

Thought?

include/internal/libspdm_common_lib.h
steven-bellock: It should return a `uint32_t`.

library/spdm_common_lib/libspdm_com_support.c
steven-bellock: Byte0 should be the release status.
0x00 - unreleased and pending
0x10 - pre-release
0x20 - released

There should be a separate mechanism to determine if FIPS is enabled or not.

library/spdm_common_lib/libspdm_com_support.c
```diff
@@ -6,6 +6,10 @@
 
 #include "internal/libspdm_common_lib.h"
 
+#if LIBSPDM_FIPS_MODE
+#define LIBSPDM_MODULE_VERSION_NUMBER 0x01
```
steven-bellock, line 10: This needs to be available to the Integrator at compile time, so it will need to reside in `include/library/spdm_common_lib.h`.

steven-bellock (changes requested):

library/spdm_common_lib/libspdm_com_support.c
jyao1: Good idea to include tag release. But I dont think we need distinguish unrelease v.s. pre-release. They should be same. One bit is enough. I suggest:
BIT[0] - official release tag or no tag.
BIT[1:5] - reserved.
BIT[6] - debug build or release debug
BIT[7] - FIPS mode or not

Why we need use different mechanism for reporting?

steven-bellock: The library's version is completely orthogonal to whether FIPS is enabled or not.

jyao1: I would say: the Byte0 is to provide additional binary info.
What we provide is 
* official release tag or not
* release build or debug build
* FIPS mode or not

That helps people to understand what this binary is.

If you dont like FIPS info, then we should make it `reserved to 0`.

steven-bellock: I don't like the FIPS information in `libspdm_module_version`. So maybe we should have something more generic like
```
typedef enum {
  LIBSPDM_INFO_VERSION,
  LIBSPDM_INF_BUILD,
  LIBSPDM_INFO_FIPS,
 } libspdm_info_t;

uint32_t libspdm_get_module_info (libspdm_info_t info);
```
Then we can expand the enum as needed.

Wenxing-hou: Thanks. I am fixing for it.

jyao1: I think we can use below APIs:
```
/*
byte3 - libspdm major version
byte2 - libspdm minor version
byte1 - libspdm patch version
byte0 - libspdm alpha (office release: 0, non official release: non-zero)
*/
uint32_t libspdm_module_version(void);
```

jyao1:
Please 1. resolve conflict. 2. use the final conclusion below:

```
/*
byte3 - libspdm major version
byte2 - libspdm minor version
byte1 - libspdm patch version
byte0 - libspdm alpha (office release with tag: 0, release candidate with tag: 1, non official release: 0xFF)
*/
uint32_t libspdm_module_version(void);

/* true: FIPS enabled, false: FIPS disabled*/
bool libspdm_get_fips_mode(void);
```

include/library/spdm_common_lib.h
```diff
@@ -12,6 +12,11 @@
 #include "library/spdm_secured_message_lib.h"
 #include "library/spdm_return_status.h"
 
+#define LIBSPDM_MAJOR_VERSION 0x01
+#define LIBSPDM_MINOR_VERSION 0x02
+#define LIBSPDM_PATCH_VERSION 0x00
+#define LIBSPDM_ALPHA         0x00
```
jyao1, line 18: please change to 0x03.0x00.0x00.0xff

Wenxing-hou: Thanks. I have changed it.

jyao1 (approved):

Issue #1607, opened 2023-01-09
add algo type check for responder


jyao1:
Currently, requester has algo type check https://github.com/DMTF/libspdm/blob/main/library/spdm_requester_lib/libspdm_req_negotiate_algorithms.c#L242-L246

```
            if ((struct_table->alg_type < SPDM_NEGOTIATE_ALGORITHMS_STRUCT_TABLE_ALG_TYPE_DHE) ||
                (struct_table->alg_type >
                 SPDM_NEGOTIATE_ALGORITHMS_STRUCT_TABLE_ALG_TYPE_KEY_SCHEDULE)) {
                status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
                goto receive_done;
            }
```

responder need add that check as well.

PR #1628, opened 2023-01-12
Add alg_type check in algorithm response.

jyao1:
Fix: https://github.com/DMTF/libspdm/issues/1607


steven-bellock (changes requested):
Please also include a unit test that hits this check.

jyao1:
> Please also include a unit test that hits this check.

Done.
Add both requester and responder.
Without check, both tests fail.
With check, both tests pass.

steven-bellock (approved):

Issue #1588, opened 2023-01-06
Convert LIBSPDM_MAX_REQUEST_RETRY_TIMES to a runtime parameter


steven-bellock:
It does not specify a statically allocated size or impart an optimization benefit.

jyao1:
remove the MACRO - LIBSPDM_MAX_REQUEST_RETRY_TIMES , and expose it via set_data.

Default value is 0 - no retry.

xiangfeima:
@steven-bellock @jyao1 I can take this one.

PR #1633, opened 2023-01-13
Change request retry times to a runtime parameter

xiangfeima:
Fix: #1588
Remove the macro `LIBSPDM_MAX_REQUEST_RETRY_TIMES` and expose it via `libspdm_set_data()`.
Default value is 0 (no retry).

### DIFF
--- a/library/spdm_common_lib/libspdm_com_context_data.c
+++ b/library/spdm_common_lib/libspdm_com_context_data.c
@@ -467,6 +467,12 @@ libspdm_return_t libspdm_set_data(void *context, libspdm_data_type_t data_type,
         }
         spdm_context->local_context.is_requester = *(bool *)data;
         break;
+    case LIBSPDM_DATA_REQUEST_RETRY_TIMES:
+        if (data_size != sizeof(uint8_t)) {
+            return LIBSPDM_STATUS_INVALID_PARAMETER;
+        }
+        spdm_context->retry_times = *(uint8_t *)data;
+        break;
     default:
         return LIBSPDM_STATUS_UNSUPPORTED_CAP;
         break;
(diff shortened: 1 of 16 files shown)


library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -467,6 +467,12 @@ libspdm_return_t libspdm_set_data(void *context, libspdm_data_type_t data_type,
         }
         spdm_context->local_context.is_requester = *(bool *)data;
         break;
+    case LIBSPDM_DATA_REQUEST_RETRY_TIMES:
+        if (data_size != sizeof(uint8_t)) {
+            return LIBSPDM_STATUS_INVALID_PARAMETER;
+        }
+        spdm_context->retry_times = *(uint8_t *)data;
+        break;
```
xiangfeima, line 475: Though it is requester-only, I think we don't need to add a `is_requester` check here.
1. if it is checked, maybe a mechanism should be introduced along with it, that libspdm or integrator to guarantee the set data sequence (such as, set is_requester first, then set retry times only for requester), it would make things more complicated IMO.
2. Responder never uses retry times (so far), no harm.

xiangfeima:
CI is failed due to network issue, I will give it a re-run later.

library/spdm_common_lib/libspdm_com_context_data.c
steven-bellock: Yeah that's fine. 

As an aside, we don't impose any ordering on `set_data`, such that it would check `is_requester`. The expected flow is the Integrator populates the SPDM context and once that is complete they can call `libspdm_check_context` to check that all the different parameters are legal together.

steven-bellock (approved):

PR #1637, opened 2023-01-15
Delete unneeded assignment to 0

### LINKED ISSUE #1636: Delete unneeded assignment to 0
The entire context is already zeroized (line 2142) so the assignment to `0` at line 2159 is not needed.

https://github.com/DMTF/libspdm/blob/af2afce708271b50bee57fff3c0cbd21a5f157a4/library/spdm_common_lib/libspdm_com_context_data.c#L2142-L2159

steven-bellock:
Fixes #1636.


jyao1:
In general, if it is reserved, I agree to remove it.

But if it has real meaning, I prefer to keep it. The reason is that people will know, it is purposely assigned to 0, not the situation that the developer forgets to assign the value.

Issue #1580, opened 2023-01-04
Remove Requester's _in_session public functions


steven-bellock:
Remove all the `*__in_session` functions and move the `session_id` to the appropriate function, similar to how `libspdm_get_measurement` is defined.

xiangfeima:
@steven-bellock @jyao1 I can take this one.

PR #1639, opened 2023-01-16
Remove *_in_session() public functions

xiangfeima:
Fix: #1580
Remove all `*_in_session()` functions
and add `session_id` parameter to the appropriate functions.


jyao1:
The patch itself looks good to me.

I am just thinking if we need add session_id for `libspdm_challenge` and just ensure it is NULL now
Just in case it is added later.

steven-bellock:
Rather than calling it `session_id` we can call it `reserved`. But we can leave that to a separate issue / pull request.

steven-bellock (approved):

Issue #748, opened 2022-03-18
libspdm Context Parameter Consistency


steven-bellock:
libspdm should be consistent with how spdm_context is used as a parameter, both it's name and it's type.

Name:
1. spdm_context OR
2. context

Type:
1. void
2. libspdm_context_t

jyao1:
Existing Rule:
* The external API must use "void"

The rest is flexible:
* Internal API may use "void" or "libspdm_context_t"
* The name may be context or spdm_context.


Maybe we can define rule below:
* external API - "void *context".
* internal API - "libspdm_context_t *spdm_context"

jyao1:
agree

steven-bellock:
I'd reverse that.
* External - `void *spdm_context`
* Internal - `libspdm_context_t *context`

PR #1646, opened 2023-01-17
spdm_common_lib: Context Parameter Consistency.

qizhangz:
Fix: #748 


steven-bellock:
The Doxygen comments need to be updated, but it would be good to merge this soon since it touches so many functions.

steven-bellock (approved):

Issue #1642, opened 2023-01-16
add reserved for session_id in libspdm_challenge 


jyao1:
According to the discussion in https://github.com/DMTF/libspdm/pull/1639, we would add void *reserved for session_id in libspdm_challenge  function.

PR #1652, opened 2023-01-19
Add reserved parameter to libspdm_challenge function

xiangfeima:
Fix: #1642
add `void *reserved` parameter for session_id to libspdm_challenge function.

### DIFF
--- a/include/library/spdm_requester_lib.h
+++ b/include/library/spdm_requester_lib.h
@@ -254,6 +254,7 @@ libspdm_return_t libspdm_get_certificate_choose_length_ex(void *spdm_context,
  * this function also perform the basic mutual authentication.
  *
  * @param  spdm_context           A pointer to the SPDM context.
+ * @param  reserved               Reserved for session_id and is ignored.
  * @param  slot_id                The number of slot for the challenge.
  * @param  measurement_hash_type  The type of the measurement hash.
  * @param  measurement_hash       A pointer to a destination buffer to store the measurement hash.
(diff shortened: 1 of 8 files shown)


library/spdm_requester_lib/libspdm_req_challenge.c
```diff
@@ -29,6 +29,7 @@ typedef struct {
  * this function also perform the basic mutual authentication.
  *
  * @param  spdm_context           A pointer to the SPDM context.
+ * @param  reserved               Reserved for session_id.
```
steven-bellock, line 32: The internal functions don't need it yet. Just the public ones.

include/library/spdm_requester_lib.h
```diff
@@ -254,6 +254,7 @@ libspdm_return_t libspdm_get_certificate_choose_length_ex(void *spdm_context,
  * this function also perform the basic mutual authentication.
  *
  * @param  spdm_context           A pointer to the SPDM context.
+ * @param  reserved               Reserved for session_id.
```
steven-bellock, line 257: I would say "Reserved for session_id and is ignored." Ie, they can pass a pointer or `NULL` but it won't have any effect.

steven-bellock (changes requested):

library/spdm_requester_lib/libspdm_req_challenge.c
xiangfeima: The change for internal function is restored.

include/library/spdm_requester_lib.h
xiangfeima: That is more appropriate indeed, the doxygen comment is updated.

steven-bellock (approved):

PR #1677, opened 2023-01-31
Move dump functions to debuglib

### LINKED ISSUE #1455: Move dump functions to debug_lib?
It would useful for integrators to reuse the existing platform dump functions or customize the dump functions (e.g. to only dump the first N bytes to avoid flushing the logs).

https://github.com/DMTF/libspdm/blob/4f8f99005368cb0e0091ef07b693631593af9b97/include/internal/libspdm_common_lib.h#L399-L431

rw8896:
Resolves #1455

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -141,6 +141,7 @@ jobs:
           export PATH=~/gcc-arm-11.2-2022.02-x86_64-aarch64-none-linux-gnu/bin:$PATH
           mkdir build
           cd build
+          cmake --version
           cmake -E env CFLAGS="${{ matrix.configurations }} -DLIBSPDM_DEBUG_LIBSPDM_ASSERT_CONFIG=0" cmake -DARCH=${{ matrix.arch }} -DTOOLCHAIN=${{ matrix.toolchain }} -DTARGET=${{ matrix.target }} -DCRYPTO=${{ matrix.crypto }} ..
           make copy_sample_key
           make copy_seed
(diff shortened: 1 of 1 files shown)


steven-bellock:
I was thinking of something more like
```
#ifdef LIBSPDM_INTERNAL_DUMP_HEX_STR_OVERRIDE
void LIBSPDM_INTERNAL_DUMP_HEX_STR_OVERRIDE(const uint8_t *data, size_t size);
#define LIBSPDM_INTERNAL_DUMP_HEX_STR(data, size) LIBSPDM_INTERNAL_DUMP_HEX_STR_OVERRIDE(data, size)
#else
void libspdm_internal_dump_hex_str(const uint8_t *data, size_t size);
#define LIBSPDM_INTERNAL_DUMP_HEX_STR(data, size) libspdm_internal_dump_hex_str(data, size)
#endif /* LIBSPDM_INTERNAL_DUMP_HEX_STR_OVERRIDE */
```

rw8896:
> I was thinking of something more like
> 
> ```
> #ifdef LIBSPDM_INTERNAL_DUMP_HEX_STR_OVERRIDE
> void LIBSPDM_INTERNAL_DUMP_HEX_STR_OVERRIDE(const uint8_t *data, size_t size);
> #define LIBSPDM_INTERNAL_DUMP_HEX_STR(data, size) LIBSPDM_INTERNAL_DUMP_HEX_STR_OVERRIDE(data, size)
> #else
> void libspdm_internal_dump_hex_str(const uint8_t *data, size_t size);
> #define LIBSPDM_INTERNAL_DUMP_HEX_STR(data, size) libspdm_internal_dump_hex_str(data, size)
> #endif /* LIBSPDM_INTERNAL_DUMP_HEX_STR_OVERRIDE */
> ```

That's less work to do, but I was wondering why dump functions were "internal".

Understood that integrators who use their own debug lib would have to implement dump functions after dump functions get moved to debug lib. Is there other reason we should keep these functions internal?

steven-bellock:
The `include/hal/` functions are mainly for functionality that the core libspdm library can't provide itself. In this case the core library can do that, and I think what is there now would be the default implementation for most Integrators of the library. The override macros would allow the small percentage of folks who need it the ability to customize those functions without burdening others with providing ``include/hal/` implementations. If you're concerned about the API stability of those functions, we can make a provision that overridable internal functions have the same API guarantees as publicly available functions.

rw8896:
Got it, thanks for the clarification.
Okay, I will change it to the default-override style.

PR #1684, opened 2023-02-01
Remove return value from opaque build function

### LINKED ISSUE #1683: Multiple opaque processing functions should not return status
For example : https://github.com/DMTF/libspdm/blob/4853e80f8aa14f4893941f1986f758353541be1d/library/spdm_requester_lib/libspdm_req_common.c#L48-L51

https://github.com/DMTF/libspdm/blob/4853e80f8aa14f4893941f1986f758353541be1d/library/spdm_requester_lib/libspdm_req_key_exchange.c#L386-L388

steven-bellock:
Fixes #1683.

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_common.c
+++ b/library/spdm_requester_lib/libspdm_req_common.c
@@ -24,32 +24,26 @@ uint16_t libspdm_allocate_req_session_id(libspdm_context_t *spdm_context)
     return (INVALID_SESSION_ID & 0xFFFF);
 }
 
-libspdm_return_t libspdm_build_opaque_data_supported_version_data(libspdm_context_t *spdm_context,
-                                                                  size_t *data_out_size,
-                                                                  void *data_out)
+void libspdm_build_opaque_data_supported_version_data(libspdm_context_t *spdm_context,
+                                                      size_t *data_out_size,
+                                                      void *data_out)
 {
     size_t final_data_size;
-    secured_message_general_opaque_data_table_header_t
-    *general_opaque_data_table_header;
-    spdm_general_opaque_data_table_header_t
-    *spdm_general_opaque_data_table_header;
-    secured_message_opaque_element_table_header_t
-    *opaque_element_table_header;
-    secured_message_opaque_element_supported_version_t
-    *opaque_element_support_version;
+    secured_message_general_opaque_data_table_header_t *general_opaque_data_table_header;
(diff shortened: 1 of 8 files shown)


library/spdm_requester_lib/libspdm_req_common.c
```diff
@@ -24,11 +24,11 @@ uint16_t libspdm_allocate_req_session_id(libspdm_context_t *spdm_context)
...
 
-libspdm_return_t libspdm_build_opaque_data_supported_version_data(libspdm_context_t *spdm_context,
-                                                                  size_t *data_out_size,
-                                                                  void *data_out)
+void libspdm_build_opaque_data_supported_version_data(libspdm_context_t *spdm_context,
+                                                      size_t *data_out_size,
+                                                      void *data_out)
 {
-    size_t final_data_size;
+    size_t final_data_size = libspdm_get_opaque_data_supported_version_data_size(spdm_context);
```
jyao1, line 31: I don't recommend we do assignment in definition.

Can we move it just before ASSERT?

```
final_data_size = libspdm_get_opaque_data_supported_version_data_size(spdm_context);
LIBSPDM_ASSERT(*data_out_size >= final_data_size);
```

library/spdm_responder_lib/libspdm_rsp_common.c
```diff
@@ -24,12 +24,11 @@ uint16_t libspdm_allocate_rsp_session_id(const libspdm_context_t *spdm_context)
...
-libspdm_return_t
-libspdm_build_opaque_data_version_selection_data(const libspdm_context_t *spdm_context,
-                                                 size_t *data_out_size,
-                                                 void *data_out)
+void libspdm_build_opaque_data_version_selection_data(const libspdm_context_t *spdm_context,
+                                                      size_t *data_out_size,
+                                                      void *data_out)
 {
-    size_t final_data_size;
+    size_t final_data_size = libspdm_get_opaque_data_version_selection_data_size(spdm_context);
```
jyao1, line 31: I don't recommend we do assignment in definition.

Can we move it just before ASSERT?

library/spdm_requester_lib/libspdm_req_common.c
steven-bellock: Will do.

library/spdm_responder_lib/libspdm_rsp_common.c
steven-bellock: Will do.

PR #1693, opened 2023-02-04
Fix potential alignment and byte-order issue

### LINKED ISSUE #1669: Fix possible unaligned access errors in `libspdm/library`
For example : https://github.com/DMTF/libspdm/blob/89074616f96d08b17ccb899ecbc49aaeedc676f6/library/spdm_requester_lib/libspdm_req_challenge.c#L249

lsun100:
Direct usage of '*(uint32_t *)' or '*(uint16_t *)' could cause potential alignment issue, or byte-ordering issue if the data buffer is transmitted from another host with different endiness. This commit converts them into libspdm_read_uintxx()/libspdm_write_uintxx().

Code in os_stub and unit_test are not covered in this commit for now.

Resolves #1669

### DIFF
--- a/library/spdm_common_lib/libspdm_com_context_data.c
+++ b/library/spdm_common_lib/libspdm_com_context_data.c
@@ -49,10 +49,12 @@ libspdm_return_t libspdm_set_data(void *spdm_context, libspdm_data_type_t data_t
 {
     libspdm_context_t *context;
     uint32_t session_id;
+    uint32_t data32;
     libspdm_session_info_t *session_info;
     uint8_t slot_id;
     uint8_t mut_auth_requested;
     uint8_t root_cert_index;
+    uint16_t data16;
 #if !(LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT)
     bool status;
 #endif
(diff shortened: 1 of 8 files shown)


library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -48,11 +48,12 @@ libspdm_return_t libspdm_set_data(void *spdm_context, libspdm_data_type_t data_t
                                   size_t data_size)
 {
     libspdm_context_t *context;
-    uint32_t session_id;
+    uint32_t session_id, data32;
```
jyao1, line 51: Would you please split it to two line to make it clear?

```
uint32_t session_id;
uint32_t data32;
```

steven-bellock (changes requested):
This pull request does not resolve any endianness issues, only alignment, so please edit the description. In addition, there are instances of 64-bit access, such as `context->local_context.capability.rtt = *(uint64_t *)data;`.

library/spdm_common_lib/libspdm_com_context_data.c
lsun100: Done. will be updated soon.

lsun100:
> This pull request does not resolve any endianness issue...

Done. Updated the commit message.

But for the endianness, for example, the code below will put 'opaque_key_exchange_rsp_size' into ptr in host order, where the byte order relies on host endianness.  On big-endian host (like some PPC), the uint16 will be in network order (big endian). When it's passed to the requestor, it'll be interpreted incorrectly.

--- a/library/spdm_responder_lib/libspdm_rsp_key_exchange.c
+++ b/library/spdm_responder_lib/libspdm_rsp_key_exchange.c
@@ -481,7 +481,7 @@ libspdm_return_t libspdm_get_response_key_exchange(libspdm_context_t *spdm_conte
     }
     ptr += measurement_summary_hash_size;

-    *(uint16_t *)ptr = (uint16_t)opaque_key_exchange_rsp_size;
+    libspdm_write_uint16(ptr, (uint16_t)opaque_key_exchange_rsp_size);

Or probably this 'ptr' is actually only for local use thus not relevant to byte-order? By the way, just a question, do we have test cases that run requestor and responder on hosts with different endianness?

steven-bellock (approved):

steven-bellock:
> By the way, just a question, do we have test cases that run requestor and responder on hosts with different endianness?

We do not as nobody is really clamoring for us to support big-endian. But it would be reasonably easy to add support by modifying the `libspdm_{read/write}_uint*` functions. To test it in CI/CD, as well as alignment, we'd need to use QEMU.

jyao1 (approved):

Issue #717, opened 2022-03-10
Add delay when retrying requests


steven-bellock:
If the Responder returns a `Busy` error to the Requester, the Requester immediately sends a retry request. Instead, there should be some integrator-configurable delay between retry requests.

jyao1:
agree

PR #1714, opened 2023-02-22
Add delay time between retry requests

xiangfeima:
Fix: #717
This patch adds integrator configurable
delay time in microseconds between retry requests.

### DIFF
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -114,6 +114,13 @@ typedef enum {
      **/
     LIBSPDM_DATA_REQUEST_RETRY_TIMES,
 
+    /* If the Responder replies with a Busy `ERROR` response to a request
+     * then the Requester is free to retry sending the request.
+     * This value specifies the delay time in microseconds between each retry requests.
+     * If its value is 0 then libspdm will send retry request immediately.
+     **/
+    LIBSPDM_DATA_REQUEST_RETRY_DELAY_TIME,
+
     /* MAX */
     LIBSPDM_DATA_MAX
 } libspdm_data_type_t;
--- a/library/spdm_requester_lib/libspdm_req_challenge.c
+++ b/library/spdm_requester_lib/libspdm_req_challenge.c
@@ -338,11 +338,13 @@ libspdm_return_t libspdm_challenge(void *spdm_context, void *reserved,
 {
     libspdm_context_t *context;
     size_t retry;
+    uint64_t retry_delay_time;
     libspdm_return_t status;
 
     context = spdm_context;
     context->crypto_request = true;
     retry = context->retry_times;
+    retry_delay_time = context->retry_delay_time;
     do {
         status = libspdm_try_challenge(context, slot_id,
                                        measurement_hash_type,
@@ -350,6 +352,8 @@ libspdm_return_t libspdm_challenge(void *spdm_context, void *reserved,
         if (LIBSPDM_STATUS_BUSY_PEER != status) {
             return status;
         }
+
+        libspdm_sleep_in_us(retry_delay_time);
     } while (retry-- != 0);
 
     return status;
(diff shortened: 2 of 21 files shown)


library/spdm_requester_lib/libspdm_req_challenge.c
```diff
@@ -338,18 +338,24 @@ libspdm_return_t libspdm_challenge(void *spdm_context, void *reserved,
...
     do {
         status = libspdm_try_challenge(context, slot_id,
                                        measurement_hash_type,
                                        measurement_hash, slot_mask, NULL, NULL, NULL);
         if (LIBSPDM_STATUS_BUSY_PEER != status) {
             return status;
         }
+
+        if (retry_delay_time > 0) {
+            libspdm_sleep_in_us(retry_delay_time);
```
jyao1, line 357: do we need such check `retry_delay_time > 0` ?

I think libspdm_sleep_in_us(0) should be OK ?

xiangfeima: It's just about the concern that sleep(0) would yield control back to the scheduler and lead to unexpected result.

jyao1: Good point. (https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-sleep)

I think we should document the expected behavior for 0 anyway. (https://github.com/DMTF/libspdm/blob/main/include/hal/library/platform_lib.h#L27)

xiangfeima: I will add appropriate notation.

include/hal/library/platform_lib.h
```diff
@@ -23,6 +23,11 @@ void libspdm_sleep(uint64_t milliseconds);
  *
  * @param microseconds     The time interval for which execution is to be suspended, in milliseconds.
  *
+ * Be aware if 0 is given as the sleep time interval, depends on the platform implementation though,
```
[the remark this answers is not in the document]
xiangfeima: I will change it as below:
1. remove such description in `include/hal/library/platform_lib.h`
2. add if `microseconds` is 0 then return immediately in windows implementation and explain why.
3. remove `microseconds > 0` check from all calling positions.

We're good on linux implementation since man page of `select()` says
> If both fields of the timeval structure are zero, then **select()** returns immediately.

include/library/spdm_common_lib.h
```diff
@@ -9,6 +9,7 @@
 
 #include "internal/libspdm_lib_config.h"
 #include "hal/base.h"
+#include "hal/library/platform_lib.h"
```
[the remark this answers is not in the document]
xiangfeima: Not really. I will move it to `include/internal/libspdm_requester_lib.h`

steven-bellock (approved):

os_stub/platform_lib/time_win.c
```diff
@@ -19,6 +19,13 @@ void libspdm_sleep_in_us(uint64_t microseconds)
 {
     uint64_t milliseconds;
 
+    /* If 0 is given as the sleep time interval, the thread will relinquish the remainder of its time slice.
+     * https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-sleep
+     * Return immediately to avoid unexpected re-scheduling. */
+    if (microseconds == 0) {
```
jyao1, line 25: Do we need change Linux version as well?

https://github.com/DMTF/libspdm/blob/main/os_stub/platform_lib/time_linux.c

xiangfeima: We're good on linux implementation since man page of `select()` says it will return immediately.

https://man7.org/linux/man-pages/man2/select.2.html
**Arguments** -> *timeout*
> If both fields of the timeval structure are zero, then select() returns immediately.

Issue #796, opened 2022-03-28
Need add transport length check libspdm_mctp_encode_message()


jyao1:
In libspdm_pci_doe_encode_message(), there is length check for *transport_message_size. But the code is missing in libspdm_mctp_encode_message().

```
    LIBSPDM_ASSERT(*transport_message_size >=
                   aligned_message_size + sizeof(pci_doe_data_object_header_t));
    if (*transport_message_size <
        aligned_message_size + sizeof(pci_doe_data_object_header_t)) {
        *transport_message_size = aligned_message_size +
                                  sizeof(pci_doe_data_object_header_t);
        return LIBSPDM_STATUS_BUFFER_TOO_SMALL;
    }
```

jyao1:
@xiangfeima, would you please take this one?

xiangfeima:
@jyao1 I'm all set in China office. Please assigne this one to me.

PR #1715, opened 2023-02-22
Check transport length in libspdm_mctp_encode_message()

xiangfeima:
Fix: #796


steven-bellock:
@jyao1 I don't have an opinion on this one. So merge as you see fit.

jyao1:
@xiangfeima, May I know what test you have done?  Do you run spdm-emu?

xiangfeima:
Yes, I confirmed that spdm-emu works well with this patch.

jyao1:
@xiangfeima , would you please take a look at https://github.com/DMTF/spdm-emu/pull/225 ?

Any hint on the failure?

Issue #1720, opened 2023-02-23
incorrect error data


lazineer:
Hi
There is incorrect assert in digests unittest case7.
Error data of error code 'unspecified' should be 0x00.
plz check it.

https://github.com/DMTF/libspdm/blob/db735a85ded19f6500133d4504351444bf87b3a3/unit_test/test_spdm_responder/digests.c#L377

Zhiqiang520:
There are similar issues in other codes, for example :
https://github.com/DMTF/libspdm/blob/444c268fe36583927386baf9bf3f1d2cf11d6910/library/spdm_responder_lib/libspdm_rsp_challenge_auth.c#L74-L75
https://github.com/DMTF/libspdm/blob/444c268fe36583927386baf9bf3f1d2cf11d6910/library/spdm_responder_lib/libspdm_rsp_key_exchange.c#L232-L234

PR #1728, opened 2023-02-24
Fix incorrect error data.

Zhiqiang520:
Fix: #1720


library/spdm_responder_lib/libspdm_rsp_measurements.c
```diff
@@ -257,7 +257,7 @@ libspdm_return_t libspdm_get_response_measurements(libspdm_context_t *spdm_conte
                 SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_MEAS_CAP_SIG)) {
             return libspdm_generate_error_response(
                 spdm_context, SPDM_ERROR_CODE_UNSUPPORTED_REQUEST,
-                0, response_size, response);
+                SPDM_GET_MEASUREMENTS, response_size, response);
```
steven-bellock, line 260: This is not the correct error code for this scenario, but I'll file a separate issue for that.

steven-bellock (approved):

library/spdm_responder_lib/libspdm_rsp_measurements.c
steven-bellock: It has been filed as #1730.

Issue #1709, opened 2023-02-17
Propose using AFL++ to instead AFL/AFL_TURBO for libspdm fuzzing


liuw1:
The fuzzer afl++ is afl with community patches, qemu 5.1 upgrade, collision-free coverage, enhanced laf-intel & redqueen, AFLfast++ power schedules, MOpt mutators, unicorn_mode, and a lot more!

AFL++ is a superior fork to Google's AFL - more speed, more and better mutations, more and better instrumentation, custom module support, etc. And it is well maintained by google.
Repository: https://github.com/AFLplusplus/AFLplusplus

liuw1:
[afl_turbo](https://github.com/sleicasper/aflturbo) is an experimental project, few people are using this tool for fuzzing test currently.

jyao1:
Would you please propose a patch to https://github.com/DMTF/libspdm/blob/main/doc/test.md?

Is there any data on code coverage?

liuw1:
@Xiaohanjlll Please help to submitted the patch and collect the data. Thanks!

PR #1740, opened 2023-02-27
added fuzz guide and AFL++ patch

Xiaohanjlll:
Fix: #1709 

In the test, on the same machine, the total  paths found by AFL++at the same time is faster than that by AFL

image

image


liuw1:
AFL hasn't been updated for a couple of years; while it should still work fine, a more complex fork with a variety of improvements and additional features, known as [AFL++](https://github.com/AFLplusplus/AFLplusplus), is available from other members of the community and is worth checking out. 
The fuzzer afl++ is afl with community patches, qemu 5.1 upgrade, collision-free coverage, enhanced laf-intel & redqueen, AFLfast++ power schedules, MOpt mutators, unicorn_mode, and a lot more! So AFL++ is a superior fork to Google's AFL - more speed, more and better mutations, more and better instrumentation, custom module support, etc.

jyao1 (approved):

Issue #1734, opened 2023-02-26
LIBFUZZER warning


steven-bellock:
```
/unit_test/fuzzing/spdm_unit_fuzzing_common/toolchain_harness.c:131:11: warning: variable 'file_name' set but not used [-Wunused-but-set-variable]
    char *file_name;
```
https://github.com/DMTF/libspdm/blob/e5537facd81704a16f056d01242de51171c6ee85/unit_test/fuzzing/spdm_unit_fuzzing_common/toolchain_harness.c#L129-L139

Xiaohanjlll:
I will check this

PR #1752, opened 2023-03-01
Fix LIBFUZZER warning

Xiaohanjlll:
Fix: #1734


steven-bellock:
This actually works? Why is the function is even there?

Xiaohanjlll:
> This actually works? Why is the function is even there?

Yes, LLVMFuzzerInitialize is no longer needed and does not affect

image

steven-bellock (approved):

Issue #1761, opened 2023-03-05
Add LIBSPDM_X509_SUPPORT macro to strip all X509 function in mbedtls for the responder


jyao1:
This can reduce the final image size from 160K to 120K.

```
157,696 spdm_device_responder.exe

119,296 spdm_device_responder.exe
```

PR #1762, opened 2023-03-05
Add LIBSPDM_CERT_PARSE_SUPPORT for device image size reduction.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/1761

### DIFF
--- a/include/library/spdm_lib_config.h
+++ b/include/library/spdm_lib_config.h
@@ -166,6 +166,11 @@
 #define LIBSPDM_SM3_256_SUPPORT 1
 #endif
 
+/* This can be set to 0 for the device which does not need X509 parser.*/
+#ifndef LIBSPDM_CERT_PARSE_SUPPORT
+#define LIBSPDM_CERT_PARSE_SUPPORT 1
+#endif
+
 /* Code space optimization for Optional request/response messages.*/
 
 /* Consumers of libspdm may wish to not fully implement all of the optional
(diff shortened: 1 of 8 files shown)


include/library/spdm_lib_config.h
```diff
@@ -166,6 +166,11 @@
 #define LIBSPDM_SM3_256_SUPPORT 1
 #endif
 
+/* This can be set to 0 for the device which does not need X509 parser.*/
+#ifndef LIBSPDM_X509_SUPPORT
```
steven-bellock, line 170: I would call this `LIBSPDM_CERT_PARSE_SUPPORT` since there may come a time where we support non-x509 certificates.

steven-bellock (commented):
This should also extend to `crytlib_cert.h` and `spdm_crypt_lib`?

include/library/spdm_lib_config.h
jyao1: done. Use LIBSPDM_CERT_PARSE_SUPPORT.

jyao1:
@steven-bellock , any additional feedback?

steven-bellock:
Either as part of this pull request, or a subsequent pull request, update `libspdm_macro_check.h`  to check what capabilities are needed for `LIBSPDM_CERT_PARSE_SUPPORT` to be `1` (or `0`).

steven-bellock (approved):

jyao1:
> Either as part of this pull request, or a subsequent pull request, update `libspdm_macro_check.h` to check what capabilities are needed for `LIBSPDM_CERT_PARSE_SUPPORT` to be `1` (or `0`).

Current design is: responder may set it to 0 (no mutual auth) or 1 (mutual auth). requester should set it 1.

Issue #1585, opened 2023-01-05
memlib enhancements


steven-bellock:
Proposal for the following changes to `memlib`.

```
int libspdm_copy_mem(void *dst_buf, size_t dst_len, const void *src_buf, size_t src_len);

becomes

void libspdm_copy_mem(void *dst_buf, size_t dst_len, const void *src_buf, size_t src_len);
```
libspdm does not currently check the return value. 
```
void *libspdm_zero_mem(void *buffer, size_t length);

becomes

void libspdm_zero_mem(void *buffer, size_t length);
```
libspdm has no need for the return value.
```
int32_t libspdm_const_compare_mem(const void *destination_buffer, const void *source_buffer, size_t length);

becomes

bool libspdm_const_compare_mem(const void *destination_buffer, const void *source_buffer, size_t length);
```
bool is the appropriate return type.
```
void *libspdm_set_mem(void *buffer, size_t length, uint8_t value);

becomes

...nothing...
```
`libspdm_set_mem` is not used by any of the core libraries.

steven-bellock:
@richkong88 @xiaoyuruan

jyao1:
libspdm_copy_mem() -> need assert and deadloop in the function.
libspdm_zero_mem() -> OK
libspdm_const_compare_mem() -> comply with linux. OK.
libspdm_set_mem() -> 1.3 may use it. keep it.

steven-bellock:
For `libspdm_const_compare_mem`
- BSD returns `1` if the strings are the same and `0` if they are not. https://man.freebsd.org/cgi/man.cgi?query=consttime_memequal&manpath=NetBSD+7.0
- Linux returns `0` if the strings are the same and `1` if they are not. https://elixir.bootlin.com/linux/latest/source/include/crypto/algapi.h#L280
- And regular C standard library `memcmp` is `0` if they are the same.

So a `bool` doesn't make sense for Linux and C standard. But we can change the `int32_t` to an `int`.

jyao1:
libspdm_const_compare_mem() -> that is too confusing. Do we have to change?
This will bring big impact - all existing code must be changed. Now I would prefer to keep it as is.

jyao1:
3/13/2023 - to avoid misuse of `compare_mem`, we can introduce `bool libspdm_const_is_mem_equal()`.
Then `true` means the memory is equal.

PR #1777, opened 2023-03-09
Change return type of `libspdm_const_compare_mem`

steven-bellock:
libspdm does not need to know the number of bytes that are the same, as is with `memcmp`. It just needs to know if the contents of the two buffers are equal or not.

### DIFF
--- a/include/hal/library/memlib.h
+++ b/include/hal/library/memlib.h
@@ -87,20 +87,17 @@ void *libspdm_set_mem(void *buffer, size_t length, uint8_t value);
 void *libspdm_zero_mem(void *buffer, size_t length);
 
 /**
- * Compares the contents of two buffers in const time.
+ * Compares the contents of two buffers in constant time.
  *
- * This function compares length bytes of source_buffer to length bytes of destination_buffer.
- * If all length bytes of the two buffers are identical, then 0 is returned.  Otherwise, the
- * value returned is the first mismatched byte in source_buffer subtracted from the first
- * mismatched byte in destination_buffer.
- *
+ * For a given length, the time to complete the comparison is always the same regardless of the
+ * contents of the two buffers.
  *
  * @param  destination_buffer  A pointer to the destination buffer to compare.
  * @param  source_buffer       A pointer to the source buffer to compare.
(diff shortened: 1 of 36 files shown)


include/hal/library/memlib.h
```diff
@@ -87,27 +87,19 @@ void *libspdm_set_mem(void *buffer, size_t length, uint8_t value);
...
+ * @param  destination_buffer  A pointer to the destination buffer to compare.
+ * @param  source_buffer       A pointer to the source buffer to compare.
+ * @param  length              The number of bytes to compare.
  *
+ * @return true   The contents of the two buffers are the same.
+ * @retval false  The contents of the two buffers are not the same.
  **/
-int32_t libspdm_const_compare_mem(const void *destination_buffer,
-                                  const void *source_buffer, size_t length);
+bool libspdm_const_compare_mem(const void *destination_buffer,
```
jyao1, line 102: Why we make it different from std `memcmp` ?

I don't understand why we have to make this change.

jyao1:
reference: https://github.com/DMTF/libspdm/issues/1585

include/hal/library/memlib.h
steven-bellock: Resolved in original issue.

jyao1 (approved):

library/spdm_crypt_lib/libspdm_crypt_cert.c
```diff
@@ -681,8 +682,8 @@ static bool libspdm_verify_leaf_cert_eku_spdm_OID(const uint8_t *cert, size_t ce
     /*find the spdm hardware identity OID*/
     find_sucessful = false;
     for(index = 0; index <= len - sizeof(hardware_identity_oid); index++) {
-        if (!libspdm_const_compare_mem(spdm_extension + index, hardware_identity_oid,
-                                       sizeof(hardware_identity_oid))) {
+        if (!libspdm_consttime_is_mem_equal(spdm_extension + index, hardware_identity_oid,
```
steven-bellock, line 685: (From the future) This is a bug. See #1949.

Issue #1526, opened 2022-12-18
checking the size of the request


hiption:
The logic for checking the size of the request in the recent PR (https://github.com/DMTF/libspdm/pull/1512) seems to have been changed as follows.

![image](https://user-images.githubusercontent.com/6868066/208321217-60f9eb63-c94c-43a2-be0d-4379b3119bfc.png)

Why bigger is not problem?
I am curious about the reason for the change.
Do you have a plan to add some fields of `GET_VERSION` and `GET_CAPABILITIES` in version 1.3?

jyao1:
> Why bigger is not problem?

Because the transport layer may add padding. For example, PCI_DOE requires 4 Bytes alignment.

hiption:
Thank you for your reply.
In my understanding, SPDM can use MCTP or PCI_DOE as a transport layer.
In terms of `layered architecture`, each layer should complete processing their own messages.
Although PCI_DOE uses 4 bytes alignment, the message could be re-aligned before transmission to SPDM. 
why does SPDM depends on the certain lower layer?

hiption:
Thank you for your reply.

BTW, I don't understand why DOE doesn't have padding length field.

steven-bellock:
Also in #397, for unencrypted messages the transport does not need to know the size of the message, since SPDM allows the message to be parsed from top to bottom. We rely on the transport to perform basic size checks.

steven-bellock:
Yes, those need to be fixed as well.

hiption:
OK. I will fix it.

xiangfeima:
Fix #1778 is proposed because I'm now facing the failure due to strict message size check while using PCI-DOE transport.

Issue #1422, opened 2022-11-22
Incorrect check of response size after RESPOND_IF_READY


steven-bellock:
See https://github.com/DMTF/libspdm/blob/dc93a39396efaf083815d57c46abc92fd063e568/library/spdm_requester_lib/libspdm_req_handle_error_response.c#L100-L103 The response size as returned by the transport may be larger than the actual response message.

Issue #1183, opened 2022-08-31
Response size compared to exact value with inequality operator


steven-bellock:
There are multiple instances in the Requester library where the response size is exactly checked, for instance
https://github.com/DMTF/libspdm/blob/0e2079143f733856fac404d1b37af94f44679caf/library/spdm_requester_lib/libspdm_req_finish.c#L245-L248

However the response size as returned by the transport layer may be larger than the actual SPDM message size.

jyao1:
This is in-session message. NO padding is expected.

steven-bellock:
@jyao1 This seems to be everywhere on the Responder side. For example : https://github.com/DMTF/libspdm/blob/68cbe6df93cc0c06480e783e0b3f965a41d359d4/library/spdm_responder_lib/libspdm_rsp_certificate.c#L93-L97 Is this expected?

jyao1:
> Is this expected?
No. It should be equal or greater than.

jyao1:
9/19/2022 - not need to put to Q3 stable, per meeting.

PR #1778, opened 2023-03-09
Allow message size check to accept transport layer padding

xiangfeima:
This pull request change the strict message size check to that
message size won't be less that the minimum expected size
so that message with transport layer padding can be properly handled.

For example, PCI-DOE requires 4 bytes alignment and may have at most
3 bytes padding data.

Messages can only be in secure session remains the strict size check
since no transport layer padding is involved in such case.

Fix: #1183
Fix: #1422
Fix: #1526
Related to: #1512

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_key_update.c
+++ b/library/spdm_requester_lib/libspdm_req_key_update.c
@@ -140,16 +140,17 @@ static libspdm_return_t libspdm_try_key_update(libspdm_context_t *spdm_context,
             spdm_context, &session_id, &spdm_response_size, (void **)&spdm_response);
 
         if (!LIBSPDM_STATUS_IS_ERROR(status)) {
-            if (spdm_response_size < sizeof(spdm_message_header_t)) {
-                status = LIBSPDM_STATUS_INVALID_MSG_SIZE;
-            } else if (spdm_response->header.spdm_version != spdm_request->header.spdm_version) {
-                status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
-            } else if (spdm_response->header.request_response_code == SPDM_ERROR) {
+            if (spdm_response->header.request_response_code == SPDM_ERROR) {
                 status = libspdm_handle_error_response_main(
                     spdm_context, &session_id,
                     &spdm_response_size, (void **)&spdm_response,
-                    SPDM_KEY_UPDATE, SPDM_KEY_UPDATE_ACK,
-                    sizeof(libspdm_key_update_response_mine_t));
+                    SPDM_KEY_UPDATE, SPDM_KEY_UPDATE_ACK);
+            } else if (spdm_response_size != sizeof(spdm_key_update_response_t)) {
+                /* this message can only be in secured session thus
+                 * don't need to consider transport layer padding, just check its exact size */
+                status = LIBSPDM_STATUS_INVALID_MSG_SIZE;
+            } else if (spdm_response->header.spdm_version != spdm_request->header.spdm_version) {
+                status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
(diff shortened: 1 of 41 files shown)


xiangfeima:
Thanks, Steven, some how I missed it (https://github.com/DMTF/libspdm/issues/1183#issuecomment-1237097228).
I will check again and make sure the messages that can only be in secure session checks with its exact size.

xiangfeima:
Hi @steven-bellock, 
latest SPDM specification [DSP0274 1.2.1](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf) seems having wrong description on `FINISH/FINISH_RSP` issuance allowance in paragraph 174, **Table 6 — SPDM request and response messages validity**, as we know they an be out-of-session when `HANDSHAKE_IN_THE_CLEAR_CAP` is set for both requester and responder.
Would help to confirm this with SPDM work group?

library/spdm_requester_lib/libspdm_req_key_update.c
```diff
@@ -140,16 +140,17 @@ static libspdm_return_t libspdm_try_key_update(libspdm_context_t *spdm_context,
             spdm_context, &session_id, &spdm_response_size, (void **)&spdm_response);
 
         if (!LIBSPDM_STATUS_IS_ERROR(status)) {
-            if (spdm_response_size < sizeof(spdm_message_header_t)) {
-                status = LIBSPDM_STATUS_INVALID_MSG_SIZE;
-            } else if (spdm_response->header.spdm_version != spdm_request->header.spdm_version) {
-                status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
-            } else if (spdm_response->header.request_response_code == SPDM_ERROR) {
+            if (spdm_response->header.request_response_code == SPDM_ERROR) {
```
xiangfeima, line 143: Here the check order is slightly changed to make sure ERROR response especially `RESPONSE_IF_READY` can be handled before the exact `spdm_key_update_response_t` response size check.

library/spdm_requester_lib/libspdm_req_key_update.c
```diff
@@ -268,29 +269,32 @@ static libspdm_return_t libspdm_try_key_update(libspdm_context_t *spdm_context,
...
-            sizeof(libspdm_key_update_response_mine_t));
+            SPDM_KEY_UPDATE, SPDM_KEY_UPDATE_ACK);
         if (LIBSPDM_STATUS_IS_ERROR(status)) {
             LIBSPDM_DEBUG((LIBSPDM_DEBUG_INFO, "SpdmVerifyKey[%x] Failed\n", session_id));
             libspdm_release_receiver_buffer (spdm_context);
             return status;
         }
     }
 
+    /* this message can only be in secured session
```
xiangfeima, line 285: same reason to check exact `spdm_key_update_response_t` response size after `libspdm_handle_error_response_main`

steven-bellock:
> Hi @steven-bellock, latest SPDM specification [DSP0274 1.2.1](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf) seems having wrong description on `FINISH/FINISH_RSP` issuance allowance in paragraph 174, **Table 6 — SPDM request and response messages validity**, as we know they can be out-of-session when `HANDSHAKE_IN_THE_CLEAR_CAP` is set for both requester and responder. Would help to confirm this with SPDM work group?

Yes this is a bug in the specification. I will file an issue against it.

jyao1 (approved):

steven-bellock (approved):

jyao1:
@xiangfeima , would please take a look at https://github.com/DMTF/spdm-emu/pull/225 ?

Any hint on the failure?

PR #1788, opened 2023-03-13
spdm_responder_lib: add code name to response dump

lironhimi:
This patch extend the request (and response) msg with command name.
check reference below using spdm-emu

Before:
SpdmReceiveRequest[.] ...
SpdmReceiveRequest[0] (0x14):
0000: 12 e1 00 00 00 00 00 00 c6 f7 02 00 00 12 00 00 00 12 00 00
SpdmSendResponse[0] ...
SpdmSendResponse[0] (0x14):
0000: 12 61 00 00 00 00 00 00 f7 fb 1a 00 00 12 00 00 00 12 00 00

After:
SpdmReceiveRequest[.] ...
SpdmReceiveRequest[0] **msg SPDM_GET_CAPABILITIES(0xe1)**, size (0x14):
0000: 12 e1 00 00 00 00 00 00 c6 f7 02 00 00 12 00 00 00 12 00 00
SpdmSendResponse[0] ...
SpdmSendResponse[0]: **msg SPDM_CAPABILITIES(0x61)**, size (0x14):
0000: 12 61 00 00 00 00 00 00 f7 fb 1a 00 00 12 00 00 00 12 00 00

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -409,6 +409,15 @@ typedef struct {
                                   LIBSPDM_SECURED_MESSAGE_CONTEXT_SIZE * LIBSPDM_MAX_SESSION_COUNT)
 
 #if LIBSPDM_DEBUG_PRINT_ENABLE
+/**
+ * Return the request code name based on given request code.
+ *
+ * @param  request_code                  The SPDM request code.
+ *
+ * @return request code name according to the request code.
+ **/
+const char *libspdm_get_code_str(uint8_t request_code);
+
 #ifdef LIBSPDM_INTERNAL_DUMP_HEX_STR_OVERRIDE
 extern void LIBSPDM_INTERNAL_DUMP_HEX_STR_OVERRIDE(const uint8_t *data, size_t size);
 #define LIBSPDM_INTERNAL_DUMP_HEX_STR(data, size) LIBSPDM_INTERNAL_DUMP_HEX_STR_OVERRIDE(data, size)
--- a/library/spdm_common_lib/libspdm_com_support.c
+++ b/library/spdm_common_lib/libspdm_com_support.c
@@ -7,6 +7,79 @@
 #include "internal/libspdm_common_lib.h"
 
 #if LIBSPDM_DEBUG_PRINT_ENABLE
+typedef struct {
+    uint8_t code;
+    const char *code_str;
+} libspdm_code_str_struct_t;
+
+const char *libspdm_get_code_str(uint8_t request_code)
+{
+    size_t index;
+
+    static libspdm_code_str_struct_t code_str_struct[] = {
+        /* SPDM response code (1.0) */
+        { SPDM_DIGESTS, "SPDM_DIGESTS" },
+        { SPDM_CERTIFICATE, "SPDM_CERTIFICATE" },
+        { SPDM_CHALLENGE_AUTH, "SPDM_CHALLENGE_AUTH" },
+        { SPDM_VERSION, "SPDM_VERSION" },
+        { SPDM_MEASUREMENTS, "SPDM_MEASUREMENTS" },
+        { SPDM_CAPABILITIES, "SPDM_CAPABILITIES" },
+        { SPDM_ALGORITHMS, "SPDM_ALGORITHMS" },
+        { SPDM_VENDOR_DEFINED_RESPONSE, "SPDM_VENDOR_DEFINED_RESPONSE" },
+        { SPDM_ERROR, "SPDM_ERROR" },
+        /* SPDM response code (1.1) */
+        { SPDM_KEY_EXCHANGE_RSP, "SPDM_KEY_EXCHANGE_RSP" },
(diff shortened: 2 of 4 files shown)


library/spdm_responder_lib/libspdm_rsp_receive_send.c
```diff
@@ -12,6 +12,87 @@ typedef struct {
...
+} libspdm_code_str_struct_t;
+
+/**
+ * Return the GET_SPDM_RESPONSE function via last request.
+ *
+ * @param  spdm_context                  The SPDM context for the device.
+ *
+ * @return GET_SPDM_RESPONSE function according to the last request.
+ **/
+const static char *libspdm_get_code_str(uint8_t code)
```
steven-bellock, line 27: This needs to be wrapped in `LIBSPDM_DEBUG_PRINT_ENABLE` so that it's not present in release code.

lironhimi: thanks, fixing it

lironhimi:
how to rerun the checks?

lironhimi:
should i open an issue as well?

steven-bellock:
> should i open an issue as well?

No, for something like this a pull request is fine.

steven-bellock (approved):

library/spdm_responder_lib/libspdm_rsp_receive_send.c
```diff
@@ -12,6 +12,89 @@ typedef struct {
     libspdm_get_spdm_response_func get_response_func;
 } libspdm_get_response_struct_t;
 
+#if LIBSPDM_DEBUG_PRINT_ENABLE
```
jyao1, line 15: This piece of code is generic for both requester and response.
Why not put to common lib?

lironhimi: correct. i was focusing on the responder side. i will make the change to be applicable for both

library/spdm_common_lib/libspdm_com_support.c
```diff
@@ -7,6 +7,80 @@
...
+    };
+
+    for (index = 0; index < LIBSPDM_ARRAY_SIZE(code_str_struct); index++) {
+        if (request_code == code_str_struct[index].code) {
+            return code_str_struct[index].code_str;
+        }
+    }
+
+    LIBSPDM_DEBUG((LIBSPDM_DEBUG_ERROR, "msg code 0x%x not found!!!\n", request_code));
+    LIBSPDM_ASSERT(0);
```
jyao1, line 79: Invalid message is a legal case.

I dont think we should assert here, because the message might be invalid.

I recommend just return "<unknown>" string

jyao1: The attacker may send invalid message to an entity at any time.

That is something MIGHT happen - we cannot use ASSERT.

ASSERT is only for something should NEVER happen.

lironhimi: i agree with you about the attacker. so isn't the validity of the message be confirmed before calling this function?
maybe the current use of this function is wrong, and should be moved to later stage after the message was validated if you say it isn't be done in libspdm_receive_response.

lironhimi: will do the required changes

include/internal/libspdm_common_lib.h
```diff
@@ -409,6 +409,14 @@ typedef struct {
...
 
 #if LIBSPDM_DEBUG_PRINT_ENABLE
+/**
+ * Return the request code name based on given request code.
+ *
+ * @param  request_code                  The SPDM request code.
+ *
+ * @return request code name according to the request code.
+ **/
+const char *libspdm_get_code_str(uint8_t request_code);
```
steven-bellock, line 419: Add a space after this line.

steven-bellock (changes requested):

include/internal/libspdm_common_lib.h
lironhimi: done

steven-bellock (approved):

library/spdm_common_lib/libspdm_com_support.c
```diff
@@ -7,6 +7,79 @@
...
+
+    for (index = 0; index < LIBSPDM_ARRAY_SIZE(code_str_struct); index++) {
+        if (request_code == code_str_struct[index].code) {
+            return code_str_struct[index].code_str;
+        }
+    }
+
+    LIBSPDM_DEBUG((LIBSPDM_DEBUG_ERROR, "msg code 0x%x not found!!!\n", request_code));
+
+    return "";
```
jyao1, line 80: I suggest to return `<unknown>` string in my previous comment.
As such, people can search the `<unknown>` to see if there is anything unexpected.

lironhimi: sorry. now fixed

xiangfeima:
Would you also remove `spdm_responder_lib:` from the title for consistency? Thank you!

PR #1810, opened 2023-03-16
Fix incorrect SPDM_RESPOND_IF_READY

### LINKED ISSUE #1780: RESPONSE_IF_READY is not handled correctly in responder
It observe that libspdm assert get hit as soon as it receive response if ready message.

https://github.com/DMTF/libspdm/blob/66b87eec0dab821fbe877ffe8615d437371a307c/library/spdm_responder_lib/libspdm_rsp_receive_send.c#L89

jyao1:
Fix #1780

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_receive_send.c
+++ b/library/spdm_responder_lib/libspdm_rsp_receive_send.c
@@ -86,7 +86,6 @@ libspdm_get_spdm_response_func libspdm_get_response_func_via_request_code(uint8_
         #endif /* LIBSPDM_ENABLE_CAPABILITY_CHUNK_CAP */
     };
 
-    LIBSPDM_ASSERT(request_code != SPDM_RESPOND_IF_READY);
     for (index = 0; index < sizeof(get_response_struct) / sizeof(get_response_struct[0]); index++) {
         if (request_code == get_response_struct[index].request_response_code) {
             return get_response_struct[index].get_response_func;


library/spdm_responder_lib/libspdm_rsp_receive_send.c
```diff
@@ -86,7 +86,6 @@ libspdm_get_spdm_response_func libspdm_get_response_func_via_request_code(uint8_
         #endif /* LIBSPDM_ENABLE_CAPABILITY_CHUNK_CAP */
     };
 
-    LIBSPDM_ASSERT(request_code != SPDM_RESPOND_IF_READY);
```
steven-bellock, line 89: How can this be? How are the unit tests passing?

jyao1: No idea on why this is written in this.
Responder unit test does not cover this function, because the main purpose is to test the requester message parsing, but not the state machine.


I tested the IF_READY function by setting one time "spdm_context->response_state = LIBSPDM_RESPONSE_STATE_NOT_READY" in responder.

The SPDM_RESPOND_IF_READY is issued from requester and the sequence can continue.
So I believe the rest flow works.

steven-bellock: Ok I filed https://github.com/DMTF/spdm-emu/issues/234 against spdm-emu to provide some coverage.

steven-bellock (approved):

PR #1811, opened 2023-03-16
rename LIBSPDM_ENABLE_CAPABILITY_SET_CERT_CAP

### LINKED ISSUE #1641: rename LIBSPDM_ENABLE_CAPABILITY_SET_CERTIFICATE_CAP to LIBSPDM_ENABLE_CAPABILITY_SET_CERT_CAP
This is to align all CAP definition in SPDM specification.

jyao1:
Fix #1641


steven-bellock (changes requested):
This also needs to be renamed in CI/CD.

jyao1:
> This also needs to be renamed in CI/CD.

Good catch. Update.

steven-bellock (approved):

PR #1819, opened 2023-03-20
Check integer overflow.

### LINKED ISSUE #1816: Overflow addition when calculating size of certificate chain
https://github.com/DMTF/libspdm/blob/1e2350efd9c08b8be9d8624fdaf592a8176243bb/library/spdm_requester_lib/libspdm_req_get_certificate.c#L242-L244

`total_responder_cert_chain_buffer_length`, `portion_length`, and `remainder_length` are all of type `uint16_t` so if the Responder returns a `CERTIFICATE` response with, for example, `portion_length == 1 && remainder_length == 0xffff` then `total_responder_cert_chain_buffer_length == 0`. Fortunately this cannot be exploited to keep the Requester in an infinite loop, as the managed buffer that holds the certificate chain is always growing. But the Requester should check that the combination of `portion_length` and `remainder_length` is sensible and will fit in the managed buffer. In particular the maximum size of a certificate chain should be 64 KiB; however the SPDM specification does not explicitly state that. I've filed an issue against the specification as https://github.com/DMTF/SPDM-WG/issues/2754.

This issue was discovered by the NVIDIA Offensive Security Research team.

jyao1:
Fix #1816

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_get_certificate.c
+++ b/library/spdm_requester_lib/libspdm_req_get_certificate.c
@@ -239,6 +239,17 @@ static libspdm_return_t libspdm_try_get_certificate(libspdm_context_t *spdm_cont
             status = LIBSPDM_STATUS_INVALID_MSG_SIZE;
             goto done;
         }
+        if (spdm_response->portion_length > 0xFFFF - spdm_request->offset) {
+            libspdm_release_receiver_buffer (spdm_context);
+            status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
+            goto done;
+        }
+        if (spdm_response->remainder_length > 0xFFFF - spdm_request->offset -
+            spdm_response->portion_length) {
+            libspdm_release_receiver_buffer (spdm_context);
+            status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
+            goto done;
+        }
         if (spdm_request->offset == 0) {
             total_responder_cert_chain_buffer_length = spdm_response->portion_length +
                                                        spdm_response->remainder_length;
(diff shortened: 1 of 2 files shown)


mcsmonk:
Don't we need to check whether overflow is occur or not in `libspdm_get_response_certificate` of `lipspdm_rsp_certificate.c`?

jyao1:
I cannot find anything need to check in libspdm_get_response_certificate().

Would you please be specific on where the check is needed?

mcsmonk:
Sorry, I missed the part. Thanks.

BTW, I guess that the code must be changed little
``` C
if ((size_t)(offset + length) > spdm_context->local_context.
    local_cert_chain_provision_size[slot_id]) {
    length = (uint16_t)(spdm_context->local_context.local_cert_chain_provision_size[slot_id] -
                        offset);
}
remainder_length = spdm_context->local_context.local_cert_chain_provision_size[slot_id] -
                   (length + offset);
```
into
``` C
if (((size_t)offset + length) > spdm_context->local_context.
    local_cert_chain_provision_size[slot_id]) {
    length = (uint16_t)(spdm_context->local_context.local_cert_chain_provision_size[slot_id] -
                        offset);
}
remainder_length = spdm_context->local_context.local_cert_chain_provision_size[slot_id]
                   - length - offset;
```
due to C Operator Precedence

jyao1:
@mcsmonk, I am not sure if I understand the change.
Would you please clarify what problem you are trying to resolve by changing code in this way?

mcsmonk:
Because `offset` and `length` are also `uint16_t`, overflow can occur when castring after sum.

mcsmonk (approved):

PR #1828, opened 2023-03-21
Move secured message functions

steven-bellock:
Move appropriate secured message functions from the public header to the internal header.


steven-bellock:
@jyao1 Please check that these functions aren't needed by the Integrator.

jyao1 (approved):

PR #1835, opened 2023-03-26
Require Integrator to provide certificate chain buffer

### LINKED ISSUE #1454: Relocate full certificate chain from stack to somewhere else
`libspdm_try_get_certificate` currently allocates space for the entire certificate chain on its stack. See : https://github.com/DMTF/libspdm/blob/4f8f99005368cb0e0091ef07b693631593af9b97/library/spdm_requester_lib/libspdm_req_get_certificate.c#L78

Since certificate chains can be large this may exhaust the stack, especially in embedded environments. Some options to alleviate this are
1. If Requester supports `GET_CERTIFICATE` then allocate a portion of the scratch space to store the entire certificate chain.
2. Use the Integrator-provided certificate chain buffer to store the certificate chain when verifying it. https://github.com/DMTF/libspdm/blob/4f8f99005368cb0e0091ef07b693631593af9b97/library/spdm_requester_lib/libspdm_req_get_certificate.c#L68

steven-bellock:
Fix #1454.

### DIFF
--- a/include/industry_standard/spdm.h
+++ b/include/industry_standard/spdm.h
@@ -425,7 +425,7 @@ typedef struct {
 } spdm_cert_chain_t;
 
 /* Maximum size, in bytes, of a certificate chain. */
-#define SPDM_MAX_CERTIFICATE_CHAIN_SIZE 65536
+#define SPDM_MAX_CERTIFICATE_CHAIN_SIZE 65535
 
 /* SPDM CHALLENGE request */
 typedef struct {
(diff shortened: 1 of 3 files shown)


include/industry_standard/spdm.h
```diff
@@ -424,6 +424,9 @@ typedef struct {
     /*uint8_t    certificates[length - 4 - hash_size];*/
 } spdm_cert_chain_t;
 
+/* Maximum size, in bytes, of a certificate chain. */
+#define SPDM_MAX_CERTIFICATE_CHAIN_SIZE 65536
```
jyao1, line 428: it should be 65535.

PR #1846, opened 2023-03-29
library: spdm_requester: Pass along reset required error messages

alistair23:
If a device provides a ResetRequired error message we want to report that up the stack so that applications can reset the device. This patch exposes that reset error information to users.

### DIFF
--- a/include/library/spdm_return_status.h
+++ b/include/library/spdm_return_status.h
@@ -137,6 +137,10 @@ typedef uint32_t libspdm_return_t;
 #define LIBSPDM_STATUS_SESSION_TRY_DISCARD_KEY_UPDATE \
     LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_ERROR, LIBSPDM_SOURCE_CORE, 0x0011)
 
+/* The device needs a reset after the operation (such as set certificate). */
+#define LIBSPDM_STATUS_RESET_REQUIRED_PEER \
+    LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_ERROR, LIBSPDM_SOURCE_CORE, 0x0012)
+
 /* - Cryptography Errors - */
 
 /* Generic failure originating from the cryptography module. */
--- a/library/spdm_requester_lib/libspdm_req_handle_error_response.c
+++ b/library/spdm_requester_lib/libspdm_req_handle_error_response.c
@@ -103,12 +103,22 @@ static libspdm_return_t libspdm_requester_respond_if_ready(libspdm_context_t *sp
 libspdm_return_t libspdm_handle_simple_error_response(libspdm_context_t *spdm_context,
                                                       uint8_t error_code)
 {
+    spdm_set_certificate_request_t *last_spdm_request;
+
     if (error_code == SPDM_ERROR_CODE_RESPONSE_NOT_READY) {
         return LIBSPDM_STATUS_NOT_READY_PEER;
     }
 
     if (error_code == SPDM_ERROR_CODE_BUSY) {
         return LIBSPDM_STATUS_BUSY_PEER;
+    }
+
+    last_spdm_request = (void *)spdm_context->last_spdm_request;
+    if (last_spdm_request->header.request_response_code == SPDM_SET_CERTIFICATE ||
+        last_spdm_request->header.request_response_code == SPDM_GET_CSR) {
+        if (error_code == SPDM_ERROR_CODE_RESET_REQUIRED) {
+            return LIBSPDM_STATUS_RESET_REQUIRED_PEER;
+        }
     }
 
     if (error_code == SPDM_ERROR_CODE_REQUEST_RESYNCH) {
--- a/unit_test/test_spdm_requester/get_csr.c
+++ b/unit_test/test_spdm_requester/get_csr.c
@@ -43,6 +43,8 @@ libspdm_return_t libspdm_requester_get_csr_test_send_message(
         return LIBSPDM_STATUS_SEND_FAIL;
     case 0x2:
         return LIBSPDM_STATUS_SUCCESS;
+    case 0x3:
+        return LIBSPDM_STATUS_SUCCESS;
 
     default:
         return LIBSPDM_STATUS_SEND_FAIL;
@@ -54,6 +56,7 @@ libspdm_return_t libspdm_requester_get_csr_test_receive_message(
     void **response, uint64_t timeout)
 {
     libspdm_test_context_t *spdm_test_context;
+    libspdm_context_t *context;
 
     spdm_test_context = libspdm_get_test_context();
     switch (spdm_test_context->case_id) {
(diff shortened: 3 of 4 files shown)


include/library/spdm_return_status.h
```diff
@@ -137,6 +137,10 @@ typedef uint32_t libspdm_return_t;
 #define LIBSPDM_STATUS_SESSION_TRY_DISCARD_KEY_UPDATE \
     LIBSPDM_STATUS_CONSTRUCT(LIBSPDM_SEVERITY_ERROR, LIBSPDM_SOURCE_CORE, 0x0011)
 
+/* The device needs a reset after the operation (such as set certificate). */
+#define LIBSPDM_STATUS_RESET_REQUIRED \
```
steven-bellock, line 141: `LIBSPDM_STATUS_RESET_REQUIRED` -> `LIBSPDM_STATUS_RESET_REQUIRED_PEER`

steven-bellock (changes requested):

alistair23:
I'm a little confused on why this causes so many test failures. Does anyone have any ideas?

steven-bellock:
The Requester should really be validating `ResetRequired`, since it can only come from `SET_CERTIFICATE` and `GET_CSR`. If the Responder replies with `ResetRequired` to a request that is not one of those two then libspdm should return `LIBSPDM_STATUS_ERROR_PEER`. That should cut down on the number of tests you need to modify.

alistair23:
Thanks! With that updated the tests are now all passing :)

steven-bellock:
That means the `SET_CERTIFICATE` and `GET_CSR` tests need more coverage.

library/spdm_requester_lib/libspdm_req_handle_error_response.c
```diff
@@ -103,6 +103,8 @@ static libspdm_return_t libspdm_requester_respond_if_ready(libspdm_context_t *sp
 libspdm_return_t libspdm_handle_simple_error_response(libspdm_context_t *spdm_context,
                                                       uint8_t error_code)
 {
+    spdm_set_certificate_request_t *last_spdm_request = (void *)spdm_context->last_spdm_request;
```
jyao1, line 106: we do not do assignment when define the variable.

Please split to two lines.

It had better assign the value just before use at line 116.

alistair23:
I have added unit tests and the tests are now all passing

unit_test/fuzzing/test_requester/test_spdm_requester_get_csr/get_csr.c
```diff
@@ -248,6 +248,64 @@ void libspdm_test_requester_get_csr_case2(void **State)
...
+    spdm_context->connection_info.capability.flags |=
+        SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_CSR_CAP;
+
+    session_id = 0xFFFFFFFF;
+    session_info = &spdm_context->session_info[0];
+    libspdm_session_info_init(spdm_context, session_info, session_id, true);
+    libspdm_secured_message_set_session_state(session_info->secured_message_context,
+                                              LIBSPDM_SESSION_STATE_ESTABLISHED);
+
+#if LIBSPDM_DEBUG_ASSERT_ENABLE
```
steven-bellock, line 298: Please use the cmock macros. For example https://github.com/DMTF/libspdm/blob/e2f08f1cac9aa66970be5bf1f6d4b125deacc054/unit_test/test_spdm_requester/get_version.c#L557-L558 Those are present regardless of the value of `LIBSPDM_DEBUG_ASSERT_ENABLE`.

unit_test/fuzzing/test_requester/test_spdm_requester_set_certificate/set_certificate.c
```diff
@@ -247,6 +247,64 @@ void libspdm_test_requester_set_certificate_case2(void **State)
...
+                                                    m_libspdm_use_asym_algo, &data,
+                                                    &data_size, NULL, NULL);
+
+    session_id = 0xFFFFFFFF;
+    session_info = &spdm_context->session_info[0];
+    libspdm_session_info_init(spdm_context, session_info, session_id, true);
+    libspdm_secured_message_set_session_state(session_info->secured_message_context,
+                                              LIBSPDM_SESSION_STATE_ESTABLISHED);
+
+#if LIBSPDM_DEBUG_ASSERT_ENABLE
```
steven-bellock, line 299: Please use the cmock macros. For example https://github.com/DMTF/libspdm/blob/e2f08f1cac9aa66970be5bf1f6d4b125deacc054/unit_test/test_spdm_requester/get_version.c#L557-L558 Those are present regardless of the value of `LIBSPDM_DEBUG_ASSERT_ENABLE`.

steven-bellock (changes requested):

unit_test/fuzzing/test_requester/test_spdm_requester_get_csr/get_csr.c
```diff
@@ -248,6 +248,64 @@ void libspdm_test_requester_get_csr_case2(void **State)
...
+    LIBSPDM_ASSERT(libspdm_get_csr(spdm_context, &session_id, NULL, 0, NULL, 0,
```
jyao1, line 299: Please dont put function in ASSERT. That is bad coding style.

unit_test/fuzzing/test_requester/test_spdm_requester_get_csr/get_csr.c
alistair23: `assert_int_equal()` isn't exposed in the fuzzer, so that doesn't work. Do you also want me to add support for cmock macros to the fuzzer?

unit_test/fuzzing/test_requester/test_spdm_requester_set_certificate/set_certificate.c
alistair23: See above, they aren't exposed in the fuzzer

unit_test/fuzzing/test_requester/test_spdm_requester_get_csr/get_csr.c
alistair23: Yeah, it's ugly. But the other option is to define a variable inside a `#if` which I don't love either

unit_test/fuzzing/test_requester/test_spdm_requester_get_csr/get_csr.c
alistair23: Or I can drop the tests from the fuzzer, up to you

alistair23:
Ok! I've dropped the fuzzer tests, everything should be ready to go now

library/spdm_requester_lib/libspdm_req_get_csr.c
```diff
@@ -143,7 +143,7 @@ static libspdm_return_t libspdm_try_get_csr(libspdm_context_t *spdm_context,
             spdm_context, session_id,
             &spdm_response_size,
             (void **)&spdm_response, SPDM_GET_CSR, SPDM_CSR);
-        if (LIBSPDM_STATUS_IS_ERROR(status)) {
+        if (LIBSPDM_STATUS_IS_ERROR(status) && status != LIBSPDM_STATUS_RESET_REQUIRED_PEER) {
```
steven-bellock, line 146: This addition isn't needed. If the Responder returns an error the flow should go to `receive_done`.

unit_test/test_spdm_requester/get_csr.c
```diff
@@ -158,6 +187,43 @@ void libspdm_test_requester_get_csr_case2(void **state)
     assert_memory_equal(csr_form_get, csr_data_pointer, global_csr_len);
 }
 
+/**
+ * Test 3: Successful response to set certificate for slot 0,
```
steven-bellock, line 191: Update test description.

unit_test/test_spdm_requester/set_certificate.c
```diff
@@ -388,6 +412,42 @@ void libspdm_test_requester_set_certificate_case5(void **state)
     free(data);
 }
 
+/**
+ * Test 6: Successful response to set certificate for slot 0,
```
steven-bellock, line 416: Update test description.

unit_test/test_spdm_requester/get_csr.c
```diff
@@ -158,6 +187,43 @@ void libspdm_test_requester_get_csr_case2(void **state)
...
+    spdm_context->connection_info.connection_state =
+        LIBSPDM_CONNECTION_STATE_NEGOTIATED;
+    spdm_context->connection_info.capability.flags |=
+        SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_CSR_CAP |
+        SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_CERT_INSTALL_RESET_CAP;
+
+    status = libspdm_get_csr(spdm_context, NULL, NULL, 0, NULL, 0, (void *)&csr_form_get,
+                             &csr_len);
+
+    assert_int_equal(status, LIBSPDM_STATUS_RESET_REQUIRED_PEER);
```
steven-bellock, line 221: `assert_int_equal(status, LIBSPDM_STATUS_RESET_REQUIRED_PEER);` is the only check that is needed.

steven-bellock (changes requested):

steven-bellock:
> Ok! I've dropped the fuzzer tests, everything should be ready to go now

Almost there.

library/spdm_requester_lib/libspdm_req_get_csr.c
alistair23: Ah you are right. I misunderstood something. I'll update this

alistair23:
I have pushed an update to the unit tests to trigger a reset and then continue with the `get_csr` request. Now all of the checks are correct.

I also believe the test descriptions are correct. The tests are successful cases, the responder is just returning that a reset is required

steven-bellock (approved):

PR #1848, opened 2023-03-29
Fix message logging bugs

### LINKED ISSUE #1842: `GET_VERSION` resets the message logger
`libspdm_try_get_version` calls https://github.com/DMTF/libspdm/blob/798ef4739ad5c1648f6d90a62774da2e7330795c/library/spdm_requester_lib/libspdm_req_get_version.c#L59 which in turn has https://github.com/DMTF/libspdm/blob/2217563eacff504b2af2a9cd7524fac512290bc4/library/spdm_common_lib/libspdm_com_context_data.c#L2254-L2260

An Integrator may want to keep message logging active through connection reset and so I'll probably remove all the message log code from `libspdm_reset_context`. If they want to reset the message log they'll need to call `libspdm_reset_msg_log`.

### LINKED ISSUE #1843: Message logger does not handle encrypted requests correctly
The message logging of a request message is effectively the last thing to happen before returning to the calling function. https://github.com/DMTF/libspdm/blob/798ef4739ad5c1648f6d90a62774da2e7330795c/library/spdm_requester_lib/libspdm_req_send_receive.c#L585-L589 This is because if any errors occur during the preparation of the request message it won't be logged. However encryption is effectively "in-place" https://github.com/DMTF/libspdm/blob/798ef4739ad5c1648f6d90a62774da2e7330795c/library/spdm_requester_lib/libspdm_req_send_receive.c#L71-L79 and so the ciphertext is logged to the message log instead of the plaintext.

steven-bellock:
Fix #1842.
Fix #1843.

### DIFF
--- a/library/spdm_common_lib/libspdm_com_context_data.c
+++ b/library/spdm_common_lib/libspdm_com_context_data.c
@@ -2652,13 +2652,6 @@ libspdm_return_t libspdm_init_context(void *spdm_context)
                                                      LIBSPDM_MAX_SESSION_COUNT);
 }
 
-/**
- * Reset an SPDM context.
- *
- * The size in bytes of the spdm_context can be returned by libspdm_get_context_size.
- *
- * @param  spdm_context                  A pointer to the SPDM context.
- */
 void libspdm_reset_context(void *spdm_context)
 {
     libspdm_context_t *context;
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -401,7 +401,8 @@ libspdm_return_t libspdm_init_context(void *spdm_context);
 /**
  * Reset an SPDM context.
  *
- * The size in bytes of the spdm_context can be returned by libspdm_get_context_size.
+ * Note that message logging is not reset by this function. To reset the message logger call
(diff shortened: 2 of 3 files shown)


library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -2250,14 +2243,6 @@ void libspdm_reset_context(void *spdm_context)
     context->last_spdm_request_size = 0;
     context->encap_context.certificate_chain_buffer.max_buffer_size =
         sizeof(context->encap_context.certificate_chain_buffer.buffer);
-
-    #if LIBSPDM_ENABLE_MSG_LOG
-    context->msg_log.buffer = NULL;
```
taprinz, line 2255: We can keep the buffer_size and status though, right?

steven-bellock: Nothing in the message logger gets reset by `libspdm_reset_context`.

jyao1:
@steven-bellock , what is status of this issue? Still draft?

steven-bellock:
Actually now that I look at it it has gotten stale. Let me do a new pull request.

PR #1857, opened 2023-04-01
Move `spdm_device_secret_lib` to `hal/library`

### LINKED ISSUE #1039: spdm_device_secret_lib.h should probably live in include\hal\library
Functions that `libspdm` requires the integrator to implement usually live in `include\hal\library`.

steven-bellock:
This commit also splits the platform library into Requester and Responder libraries.

Fix #1039.


steven-bellock:
`spdm-emu` will need to be updated after this change.

include/hal/library/secretlib/secretlib.h
```diff
@@ -0,0 +1,75 @@
...
+ * Writes a certificate chain to non-volatile memory.
+ *
+ * @param[in]  slot_id          The number of slot for the certificate chain.
+ * @param[in]  cert_chain       The pointer for the certificate chain to set.
+ * @param[in]  cert_chain_size  The size of the certificate chain to set.
+ *
+ * @retval true   Success write to non-volatile memory.
+ * @retval false  Unsuccessful write to non-volatile memory.
+ **/
+extern bool libspdm_write_certificate_to_nvm(uint8_t slot_id, const void *cert_chain,
```
jyao1, line 71: this should be in responder lib only?

include/hal/library/secretlib/secretlib.h
```diff
@@ -0,0 +1,75 @@
...
+ * @param  psk_hint_size   PSK Hint size in bytes.
+ * @param  info            Pointer to the application specific info.
+ * @param  info_size       Info size in bytes.
+ * @param  out             Pointer to buffer to receive HKDF value.
+ * @param  out_size        Size of HKDF bytes to generate.
+ *
+ * @retval true   HKDF generated successfully.
+ * @retval false  HKDF generation failed.
+ **/
+extern bool libspdm_psk_handshake_secret_hkdf_expand(
```
jyao1, line 31: I am thinking if we really need this standalone secret lib.

Can we duplicate that function in both requester lib and responder lib, since they are only 2 PSK optional functions?

include/hal/library/secretlib/secretlib.h
steven-bellock: It should be in Responder platformlib.

include/hal/library/secretlib/secretlib.h
steven-bellock: I will try that out.

jyao1 (approved):

Issue #1862, opened 2023-04-01
consider to one more API in transport lib


jyao1:
Problem statement:
Currently we only have MCTP and PCI_DOE binding for DSP0277. And we are seeing more specification defining secure message binding. Such as MIPI and TPM.
At same time, we keep adding new features to DSP0277, and increase version of DSP0277.

The current secured message version is defined to be binding spec specified version, instead of DSP0277 version. 
E.g. It is possible that the secured message version 1.0 means binding spec 1.0, which indicates DSP0277 version 1.2.
That makes hard to use the secured message version to do generic operation, such as AEAD operation.

Proposal:
I think we can define a secured message version <-> DSP0277 version translation function.

```
/*
  secured_message_version == the version specified in binding specification and transported in KEY_EXCHANGE/KEY_EXCHANGE_RSP
  secured_spdm_version == the DSP0277 version specified in binding specification, which is bound to secured_message_version.
*/
typedef void (*libspdm_secured_message_get_secured_spdm_version)(
    libspdm_device_version_t secured_message_version,
    libspdm_device_version_t *secured_spdm_version);
```

Then, we can write common code according to DSP0277 (secured SPDM).

PR #1863, opened 2023-04-01
add libspdm_secured_message_get_secured_spdm_version in transport lib.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/1862


include/library/spdm_secured_message_lib.h
```diff
@@ -207,12 +207,25 @@ typedef uint8_t (*libspdm_secured_message_get_sequence_number_func)(
...
+/**
+ * This function translates the negotiated secured_message_version to a DSP0277 version.
+ *
+ * @param  secured_message_version  The version specified in binding specification and
+ *                                  negotiated in KEY_EXCHANGE/KEY_EXCHANGE_RSP.
+ *
+ * @return The DSP0277 version specified in binding specification,
+ *         which is bound to secured_message_version.
+ */
+typedef spdm_version_number_t (*libspdm_secured_message_get_secured_spdm_version)(
```
steven-bellock, line 219: These should be defined in `spdm.h` or renamed to `libspdm_`.

jyao1: I am confused.
Which should be defined in spdm.h? or renamed to libspdm\_?

include/library/spdm_secured_message_lib.h
```diff
@@ -207,12 +207,25 @@ typedef uint8_t (*libspdm_secured_message_get_sequence_number_func)(
...
+    spdm_version_number_t secured_message_version);
+
+#define SPDM_SECURED_MESSAGE_CALLBACKS_VERSION 2
```
steven-bellock, line 222: Should be `LIBSPDM_`.

jyao1: done

PR #1864, opened 2023-04-01
Add capabilities flags to spdm_secured_message_context.

jyao1:
The purpose is to prepare the future enhancement for secured message.


jyao1:
seems not needed.

Issue #1865, opened 2023-04-03
Missing check the transmit buffer size of the sending SPDM endpoint when transferring large SPDM messages


Zhiqiang520:
The current code just check the `DataTransferSize` of the **receiving** SPDM endpoint, but it also should check the transmit buffer size of the **sending** SPDM endpoint when transferring large SPDM messages.

1. Refer to paragraph 681 in [DSP0274_1.2.1](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf).

> A **large SPDM message** is an SPDM message whose size is greater than the DataTransferSize of the **receiving**
> SPDM endpoint **or** greater than the transmit buffer size of the **sending** SPDM endpoint.
2. I think:
>     if ((((context->connection_info.capability.data_transfer_size != 0) &&
>           (my_response_size > context->connection_info.capability.data_transfer_size)) ||
>          ((context->local_context.capability.data_transfer_size != 0) &&
>           (my_response_size > context->local_context.capability.data_transfer_size))) &&
>         libspdm_is_capabilities_flag_supported(
>             context, false, 0,
>             SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_CHUNK_CAP)) {
Maybe the code at the following line 596~600 should be changed as the above.
https://github.com/DMTF/libspdm/blob/e2f08f1cac9aa66970be5bf1f6d4b125deacc054/library/spdm_responder_lib/libspdm_rsp_receive_send.c#L595-L600

3. The same issue at the following line 555~556.
https://github.com/DMTF/libspdm/blob/e2f08f1cac9aa66970be5bf1f6d4b125deacc054/library/spdm_requester_lib/libspdm_req_send_receive.c#L552-L557

jyao1:
now we have LIBSPDM_SENDER_DATA_TRANSFER_SIZE and LIBSPDM_RECEIVER_DATA_TRANSFER_SIZE

PR #1866, opened 2023-04-03
If a SPDM message size is greater than the transmit buffer siz…

Zhiqiang520:
…e of the sending SPDM endpoint, then it is also a large SPDM message.

Fix: #1865

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_receive_send.c
+++ b/library/spdm_responder_lib/libspdm_rsp_receive_send.c
@@ -592,9 +592,13 @@ libspdm_return_t libspdm_build_response(void *spdm_context, const uint32_t *sess
         }
     }
 
+    /* large SPDM message is the SPDM message whose size is greater than the DataTransferSize of the receiving
+     * SPDM endpoint or greater than the transmit buffer size of the sending SPDM endpoint */
     if ((status == LIBSPDM_STATUS_SUCCESS) &&
-        (context->connection_info.capability.data_transfer_size != 0) &&
-        (my_response_size > context->connection_info.capability.data_transfer_size) &&
+        (((context->connection_info.capability.data_transfer_size != 0) &&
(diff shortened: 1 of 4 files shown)


library/spdm_responder_lib/libspdm_rsp_receive_send.c
```diff
@@ -593,8 +593,12 @@ libspdm_return_t libspdm_build_response(void *spdm_context, const uint32_t *sess
         }
     }
 
-    if ((context->connection_info.capability.data_transfer_size != 0) &&
-        (my_response_size > context->connection_info.capability.data_transfer_size) &&
+    /* large SPDM message is the SPDM message whose size is greater than the DataTransferSize of the receiving
+     * SPDM endpoint or greater than the transmit buffer size of the sending SPDM endpoint */
+    if ((((context->connection_info.capability.data_transfer_size != 0) &&
+          (my_response_size > context->connection_info.capability.data_transfer_size)) ||
+         ((context->local_context.capability.data_transfer_size != 0) &&
```
jyao1, line 600: I dont understand the logic.
Why you check local receive buffer size?

Zhiqiang520: currently one macro `LIBSPDM_SENDER_RECEIVE_BUFFER_SIZE` defined both of the receiver buffer size and the sender buffer size for one same SPDM endpoint, so I take the local receive buffer size as its send buffer size here.

jyao1: now we have LIBSPDM_SENDER_DATA_TRANSFER_SIZE and LIBSPDM_RECEIVER_DATA_TRANSFER_SIZE

Zhiqiang520: I have updated the pull request, please check.

jyao1 (approved):

Issue #1887, opened 2023-04-06
req_session_id  and rsp_session_id check should be reversed.


jyao1:
Current code
```
    req_session_id = libspdm_allocate_req_session_id(spdm_context);
    if (req_session_id == ((INVALID_SESSION_ID & 0xFFFF0000) >> 16))
```
should be:
```
    req_session_id = libspdm_allocate_req_session_id(spdm_context);
    if (req_session_id == (INVALID_SESSION_ID & 0xFFFF))
```

Same error for `libspdm_allocate_rsp_session_id`.

Issue #1860, opened 2023-04-01
split LIBSPDM_MAX_SESSION_COUNT to DHE session and PSK session


jyao1:
Current LIBSPDM_MAX_SESSION_COUNT means the sum of DHE session and PSK session.

We need consider the case that a device may want to allow n DHE session and m PSK session.

As such, we need define two macro LIBSPDM_MAX_DHE_SESSION_COUNT and LIBSPDM_MAX_PSK_SESSION_COUNT, 
where LIBSPDM_MAX_SESSION_COUNT = LIBSPDM_MAX_DHE_SESSION_COUNT + LIBSPDM_MAX_PSK_SESSION_COUNT.

steven-bellock:
Since the size of a secured message context is the same for both PSK and DHE, they don't need to be macros, and can be set in the `spdm_context` at runtime.

jyao1:
We can add LIBSPDM_DATA_MAX_DHE_SESSION_COUNT and LIBSPDM_DATA_MAX_PSK_SESSION_COUNT.

PR #1888, opened 2023-04-06
Add max_dhe/psk_session_count support

jyao1:
Fix https://github.com/DMTF/libspdm/issues/1860
Fix https://github.com/DMTF/libspdm/issues/1887

### DIFF
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -121,12 +121,14 @@ typedef enum {
      **/
     LIBSPDM_DATA_REQUEST_RETRY_DELAY_TIME,
 
-    /* Below two entries are used to limite the number of DHE session and PSK session separately.
+    /* Below two entries are used to limit the number of DHE session and PSK session separately.
      * When set a new value, below rule is applied:
      *     new MaxDheSessionCount <= LIBSPDM_MAX_SESSION_COUNT - current MaxPskSessionCount
      *     new MaxPskSessionCount <= LIBSPDM_MAX_SESSION_COUNT - current MaxDheSessionCount
      * 0 means no limiation for the specific DHE or PSK session, as long as
      *     PskSessionCount + DheSessionCount <= LIBSPDM_MAX_SESSION_COUNT.
+     * If these values are modified while there are active sessions then the active sessions
+     * aren't terminated.
      **/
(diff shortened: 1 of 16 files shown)


include/library/spdm_common_lib.h
```diff
@@ -121,6 +121,16 @@ typedef enum {
      **/
     LIBSPDM_DATA_REQUEST_RETRY_DELAY_TIME,
 
+    /* Below two entries are used to limite the number of DHE session and PSK session separately.
```
steven-bellock, line 124: `limite` -> `limit`

include/library/spdm_common_lib.h
```diff
@@ -121,6 +121,16 @@ typedef enum {
...
+     * When set a new value, below rule is applied:
```
steven-bellock, line 125: This should also state that if these values are modified while there are active sessions then the active sessions aren't terminated. For example if there are two PSK sessions and `LIBSPDM_DATA_MAX_PSK_SESSION_COUNT` is changed to `1` then one of the sessions won't be terminated.

library/spdm_common_lib/libspdm_com_context_data_session.c
```diff
@@ -19,6 +19,32 @@ void libspdm_session_info_init(libspdm_context_t *spdm_context,
     libspdm_session_type_t session_type;
     uint32_t capabilities_flag;
 
+    if (session_id != INVALID_SESSION_ID) {
+        if (use_psk) {
+            if (spdm_context->max_psk_session_count != 0) {
```
steven-bellock, line 24: This might be picked up by a security scan / compiler warning if the `LIBSPDM_ASSERT` isn't present, as it's just an empty `if` statement.

steven-bellock (changes requested):

include/library/spdm_common_lib.h
jyao1: done

include/library/spdm_common_lib.h
jyao1: done

library/spdm_common_lib/libspdm_com_context_data_session.c
jyao1: done

steven-bellock (approved):

Issue #1878, opened 2023-04-05
Need allow different PSK_HINT for different sessions.


jyao1:
Current LIBSPDM_DATA_PSK_HINT is for the SPDM connection.

We need allow different PSK_HINT for different sessions.

Maybe add PskHint to libspdm_send_receive_psk_exchange_ex() to override default connection setting.

PR #1891, opened 2023-04-06
Add PSK in libspdm_start_session(), remove PSK in context provision

jyao1:
Fix https://github.com/DMTF/libspdm/issues/1878

### DIFF
--- a/include/library/spdm_requester_lib.h
+++ b/include/library/spdm_requester_lib.h
@@ -273,6 +273,8 @@ libspdm_return_t libspdm_get_measurement_ex(void *spdm_context, const uint32_t *
  * @param  spdm_context           A pointer to the SPDM context.
  * @param  use_psk                False means to use KEY_EXCHANGE/FINISH to start a session.
  *                                True means to use PSK_EXCHANGE/PSK_FINISH to start a session.
+ * @param  psk_hint               The psk_hint in PSK_EXCHANGE. It is ignored if use_psk is false.
+ * @param  psk_hint_size          The size in bytes of psk_hint. It is ignored if use_psk is false.
  * @param  measurement_hash_type  The type of the measurement hash.
  * @param  slot_id                The number of slot for the certificate chain.
  * @param  session_policy         The policy for the session.
@@ -303,6 +305,8 @@ libspdm_return_t libspdm_start_session(void *spdm_context, bool use_psk,
  * @param  spdm_context              A pointer to the SPDM context.
  * @param  use_psk                   False means to use KEY_EXCHANGE/FINISH to start a session.
  *                                   True means to use PSK_EXCHANGE/PSK_FINISH to start a session.
+ * @param  psk_hint                  The psk_hint in PSK_EXCHANGE. It is ignored if use_psk is false.
+ * @param  psk_hint_size             The size in bytes of psk_hint. It is ignored if use_psk is false.
  * @param  measurement_hash_type     The type of the measurement hash.
  * @param  slot_id                   The number of slot for the certificate chain.
  * @param  session_policy            The policy for the session.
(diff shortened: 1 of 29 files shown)


library/spdm_requester_lib/libspdm_req_communication.c
```diff
@@ -53,6 +55,7 @@ libspdm_return_t libspdm_start_session(void *spdm_context, bool use_psk,
 
     if (!use_psk) {
     #if LIBSPDM_ENABLE_CAPABILITY_KEY_EX_CAP
+        LIBSPDM_ASSERT((psk_hint == NULL) && (psk_hint_size == 0));
```
steven-bellock, line 58: These parameters can just be ignored if it's not a PSK session. No need to force Integrator to set their values.

steven-bellock (changes requested):
If `PskHint` is a Requester parameter then `OpaqueData` should be as well. But that can be a separate pull request.

library/spdm_requester_lib/libspdm_req_communication.c
jyao1: why?
There is no meaning to input PSK_HINT for DHE session. I would treat it as user mistake.

jyao1:
> If PskHint is a Requester parameter then OpaqueData should be as well. But that can be a separate pull request.

This issue/PR only handles PskHint.
Please do file separate one for PSK or KeyExchange.

library/spdm_requester_lib/libspdm_req_communication.c
jyao1: No.  libspdm does not evaluate psk_hint or psk_hint_size if the session is not PSK.

My intention is:
1. For DHE session, PSK_HINT must be NULL.
2. For PSK session, PSK_HINT could be NULL or non-NULL.

I am not sure why we need allow DHE with PSK_HINT non-NULL, and ignore them.
What is the use case we want to support?

steven-bellock: >What is the use case we want to support?

No use case. Just in the documentation I would like it to say
> If `use_psk` is false then the `psk_hint` and `psk_hint_size` parameters are ignored.

jyao1: updated.

include/library/spdm_requester_lib.h
```diff
@@ -285,6 +285,8 @@ libspdm_return_t libspdm_get_measurement_ex(void *spdm_context, const uint32_t *
  * @retval RETURN_SECURITY_VIOLATION    Any verification fails.
  **/
 libspdm_return_t libspdm_start_session(void *spdm_context, bool use_psk,
+                                       const void *psk_hint,
```
steven-bellock, line 288: These parameters need to be added to the Doxygen comment above. I can update the API documentation.

include/library/spdm_requester_lib.h
```diff
@@ -328,6 +330,8 @@ libspdm_return_t libspdm_start_session(void *spdm_context, bool use_psk,
  * @retval RETURN_SECURITY_VIOLATION    Any verification fails.
  **/
 libspdm_return_t libspdm_start_session_ex(void *spdm_context, bool use_psk,
+                                          const void *psk_hint,
```
steven-bellock, line 333: These parameters need to be added to the Doxygen comment above. I can update the API documentation.

steven-bellock (changes requested):

include/library/spdm_requester_lib.h
jyao1: Added.

include/library/spdm_requester_lib.h
jyao1: Added.

jyao1:
I merge it to avoid conflict later. Feel free to file new issue, if you find anything need improvement.

Issue #1894, opened 2023-04-06
assign crypto LIBSPDM_MAX_DHE/ASYM_KEY_SIZE based upon capability.


jyao1:
Currently, we have 
```
#define LIBSPDM_MAX_DHE_KEY_SIZE 512
#define LIBSPDM_MAX_ASYM_KEY_SIZE 512
#define LIBSPDM_MAX_HASH_SIZE 64
```

The max size could be reduced if the only ECC is supported. We can update them according to capability bit.

jyao1:
once we fix this one, we can move the LIBSPDM_MAX_MESSAGE_x_BUFFER_SIZE to internal.

PR #1898, opened 2023-04-07
define fine granularity control of crypto algo

jyao1:
Fix https://github.com/DMTF/libspdm/issues/1894


include/internal/libspdm_lib_config.h
```diff
@@ -107,22 +107,65 @@
...
-#define LIBSPDM_AEAD_SM4_SUPPORT 0
+#undef LIBSPDM_AEAD_SM4_128_GCM_SUPPORT
+#define LIBSPDM_AEAD_SM4_128_GCM_SUPPORT 0
 
 #undef LIBSPDM_SM3_256_SUPPORT
 #define LIBSPDM_SM3_256_SUPPORT 0
 #endif /*LIBSPDM_FIPS_MODE*/
 
+/* define crypto algorithm without parameter */
+#define LIBSPDM_RSA_SSA_SUPPORT (LIBSPDM_RSA_SSA_2048_SUPPORT || \
```
steven-bellock, line 127: All the macros used in expressions need `()` around them.

library/spdm_crypt_lib/libspdm_crypt_aead.c
```diff
@@ -67,6 +115,10 @@ bool libspdm_aead_encryption(const spdm_version_number_t secured_message_version
     case SPDM_ALGORITHMS_AEAD_CIPHER_SUITE_AES_128_GCM:
     case SPDM_ALGORITHMS_AEAD_CIPHER_SUITE_AES_256_GCM:
 #if LIBSPDM_AEAD_GCM_SUPPORT
+        LIBSPDM_ASSERT((aead_cipher_suite != SPDM_ALGORITHMS_AEAD_CIPHER_SUITE_AES_128_GCM) ||
```
steven-bellock, line 118: What are these assertions for?

steven-bellock (changes requested):

include/internal/libspdm_lib_config.h
jyao1: Added.

library/spdm_crypt_lib/libspdm_crypt_aead.c
jyao1: The purpose of this check is to ensure if an algo is not supported in MACRO, the algo shall not be input from the caller.

The original version of assert is as follows:
```
#if !LIBSPDM_AEAD_AES_128_GCM_SUPPORT
    LIBSPDM_ASSERT(aead_cipher_suite != SPDM_ALGORITHMS_AEAD_CIPHER_SUITE_AES_128_GCM);
#endif
```

steven-bellock: Original version is good.

jyao1: Done. Change to original version.

Issue #1895, opened 2023-04-06
Add LIBSPDM_DATA_SENDER_DATA_TRANSFER_SIZE to support runtime sender buffer adjustment.


jyao1:
This is to support case that: Integrator has one context for use over I2C whose send/receive buffer is 2 KiB, and another context for use over PCIE whose send/receive buffer is 2 MiB.

jyao1:
If we can assume that every time acquire_sender_buffer() returns same size, we can get it in init_function. Then we dont need LIBSPDM_DATA_SENDER_DATA_TRANSFER_SIZE.

@steven-bellock , what do you think?

jyao1:
I think `max_msg_size` is still needed to handle chunk/no-chunk cases.

jyao1:
I still keep LIBSPDM_DATA_CAPABILITY_DATA_TRANSFER_SIZE and LIBSPDM_DATA_CAPABILITY_SENDER_DATA_TRANSFER_SIZE.
The reason is that `data_transfer_size` means SPDM message size **without** transport data, while acquire_x_buffer ()r returns the size **with** transport data.
As such, we need calculate it with `data_transfer_size = data_buffer_size - LIBSPDM_TRANSPORT_ADDITIONAL_SIZE` by default.

In case LIBSPDM_TRANSPORT_ADDITIONAL_SIZE is too large, we give chance to let integrator fix the final `data_transfer_size`.

steven-bellock:
Since it sounds like we are moving to (or it's always been this way) a model where
1. The Integrator specifies to libspdm the size of these buffers AND
2. This information can be configured at runtime

I don't see why libspdm needs `LIBSPDM_TRANSPORT_ADDITIONAL_SIZE`, `LIBSPDM_MAX_SPDM_MSG_SIZE`, `LIBSPDM_DATA_CAPABILITY_DATA_TRANSFER_SIZE`, as macros. These can be set at runtime.

jyao1:
`LIBSPDM_MAX_SPDM_MSG_SIZE` and `LIBSPDM_TRANSPORT_ADDITIONAL_SIZE` must be there, because libspdm need prepare scratch buffer.

`LIBSPDM_DATA_CAPABILITY_DATA_TRANSFER_SIZE` is removed in the patch.

PR #1905, opened 2023-04-08
Add dynamic sender buffer assignement

jyao1:
Fix https://github.com/DMTF/libspdm/issues/1895

It also cleans up sender/receiver data buffer size from config.h


steven-bellock (changes requested):
The data transfer size (both send and receive) can be calculated from the size of the send / receive buffer and the additional transport overhead, so you can remove the set.

jyao1:
> The data transfer size (both send and receive) can be calculated from the size of the send / receive buffer and the additional transport overhead, so you can remove the set.

Done

Issue #1909, opened 2023-04-10
Size of scratch buffer


steven-bellock:
Follow-up to https://github.com/DMTF/libspdm/issues/1895. 

The size of the scratch buffer mainly depends on two conditions
1. Whether chunking is enabled AND/OR
2. Whether secure sessions with encryption is enabled 

Currently the size of the scratch buffer is computed at compile time and is the same for all SPDM contexts that use the compiled library. However, in a scenario where Context A has both chunking and secure sessions enabled and Context B does not, Context B's scratch buffer size is still the same as Context A's, which may be a waste of memory. Ideally libspdm should support runtime calculation of the scratch buffer's size. Note that there are two times the size can be calculated. 
1. Before `VCA`. This is the worst-case calculation that assumes everything is enabled AND
2. Immediately after `VCA` when the endpoints' capabilities are known.

To illustrate 2, the Requester may support chunking and key exchange and the Responder only supports chunking (and not key exchange). In such a case the Requester should not have to provide memory for encrypted messages with the Responder.

PR #1917, opened 2023-04-11
Support runtime scratch buffer size based upon max_spdm_msg_size

jyao1:
Fix https://github.com/DMTF/libspdm/issues/1909

The main idea is as follows. The integrator input `max_spdm_msg_size` and `libspdm_get_sizeof_required_scratch_buffer`.

```
size_t libspdm_get_sizeof_required_scratch_buffer (
    void *spdm_context, uint32_t max_spdm_msg_size)
{
    libspdm_context_t *context;
    size_t scratch_buffer_size;

    context = spdm_context;
    context->local_context.capability.max_spdm_msg_size = max_spdm_msg_size;
    scratch_buffer_size = libspdm_get_scratch_buffer_capacity(context);
    return scratch_buffer_size;
}
```

Or the integrator can pre-calculate the size based upon `libspdm_get_scratch_buffer_capacity`, then set `LIBSPDM_DATA_CAPABILITY_MAX_SPDM_MSG_SIZE`.

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -522,6 +522,52 @@ void libspdm_internal_dump_hex(const uint8_t *data, size_t size);
 #define LIBSPDM_INTERNAL_DUMP_DATA(data, size)
 #endif /* LIBSPDM_DEBUG_PRINT_ENABLE */
 
+/* Required scratch buffer size for libspdm internal usage.
+ * It may be used to hold the encrypted/decrypted message and/or last sent/received message.
+ * It may be used to hold the large request/response and intermediate send/receive buffer
+ * in case of chunking.
+ *
+ * If chunking is not supported, it should be at least below.
+ * +--------------------------+
+ * |    SENDER_RECEIVER       |
+ * +--------------------------+
+ * |<-Snd/Rcv buf for chunk ->|
+ *
+ *
+ * If chunking is supported, it should be at least below.
+ * +---------------+--------------+--------------------------+------------------------------+
+ * |SECURE_MESSAGE |LARGE_MESSAGE |    SENDER_RECEIVER       | LARGE SENDER_RECEIVER        |
+ * +---------------+--------------+--------------------------+------------------------------+
+ * |<-Secure msg ->|<-Large msg ->|<-Snd/Rcv buf for chunk ->|<-Snd/Rcv buf for large msg ->|
+ *
+ *
+ * The value is NOT configurable.
+ * The value MAY be changed in different libspdm version.
+ * It is exposed here, just in case the libspdm consumer wants to configure the setting at build time.
+ */
+#if LIBSPDM_ENABLE_CAPABILITY_CHUNK_CAP
+/* first section */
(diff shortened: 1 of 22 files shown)


include/internal/libspdm_common_lib.h
```diff
@@ -522,6 +524,62 @@ void libspdm_internal_dump_hex(const uint8_t *data, size_t size);
...
+ * |<-Secure msg ->|<-Large msg ->|<-Snd/Rcv buf for chunk ->|<-Snd/Rcv buf for large msg ->|<-last request ->|<-cache request->|
+ *
+ *
+ * The value is NOT configurable.
+ * The value MAY be changed in different libspdm version.
+ * It is exposed here, just in case the libspdm consumer wants to configure the setting at build time.
+ */
+#if LIBSPDM_ENABLE_CAPABILITY_CHUNK_CAP
+/* first section */
+uint32_t libspdm_get_scratch_buffer_secure_message_offset(libspdm_context_t *spdm_context);
```
steven-bellock, line 552: Easier and more scalable to just have `libspdm_get_scratch_buffer_offset` and `libspdm_get_scratch_buffer_capacity` functions that take an enum as input for what you want (`LIBSPDM_SCRATCH_SECURE_MESSAGE`, `LIBSPDM_SCRATCH_LARGE_MESSAGE`, etc)

jyao1: Technically, I can implement it in this way. See last 2 patches "adopt review feedback ..."

However, I think the original style is better, because it is simple and obvious.
Using switch/case to check enum is unnecessary to me.

I am not sure about what scalability issue we have with original style, since we can always add new internal functions for more section in scratch buffer.

steven-bellock: > I prefer we keep the original style.

That's fine.

include/internal/libspdm_common_lib.h
```diff
@@ -522,6 +524,52 @@ void libspdm_internal_dump_hex(const uint8_t *data, size_t size);
...
+ *
+ *
+ * If chunking is supported, it should be at least below.
+ * +---------------+--------------+--------------------------+------------------------------+-----------------+-----------------+
+ * |SECURE_MESSAGE |LARGE_MESSAGE |    SENDER_RECEIVER       | LARGE SENDER_RECEIVER        |MAX_SPDM_MSG_SIZE|MAX_SPDM_MSG_SIZE|
+ * +---------------+--------------+--------------------------+------------------------------+-----------------+-----------------+
+ * |<-Secure msg ->|<-Large msg ->|<-Snd/Rcv buf for chunk ->|<-Snd/Rcv buf for large msg ->|<-last request ->|<-cache request->|
+ *
+ *
+ * The value is NOT configurable.
```
steven-bellock, line 546: These comments are no longer relevant.

jyao1: fixed.

include/internal/libspdm_common_lib.h
jyao1: Then I drop the last 2 commit.

PR #1923, opened 2023-04-12
add check for transport for test.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/1924


Wenxing-hou:
The PR will pass CI  when the PR: https://github.com/DMTF/libspdm/pull/1979 is merged.

Issue #1928, opened 2023-04-15
CMake minimum required version


xiangfeima:
Is there specific reason to remain `cmake_minimum_required(VERSION 2.6)` at following places?
https://github.com/DMTF/libspdm/blob/113743fd0bb137a3b2b651b9022f39e7c118a23c/os_stub/armbuild_lib/CMakeLists.txt#L1
https://github.com/DMTF/libspdm/blob/113743fd0bb137a3b2b651b9022f39e7c118a23c/os_stub/platform_lib/CMakeLists.txt#L1
https://github.com/DMTF/libspdm/blob/113743fd0bb137a3b2b651b9022f39e7c118a23c/os_stub/platform_lib_null/CMakeLists.txt#L1

If no I would suggest to update them to 2.8.12 as other modules do to avoid warning as below:

```
D:\Work\Projects\maxf\libspdm\build>cmake -G"NMake Makefiles" -DARCH=x64 -DTOOLCHAIN=VS2022 -DTARGET=Debug -DCRYPTO=openssl ..
#########################
## Build Configuration ##
#########################
CMAKE_GENERATOR = NMake Makefiles
ARCH = x64
TOOLCHAIN = VS2022
TARGET = Debug
CRYPTO = openssl
ENABLE_BINARY_BUILD=0; Building openssl library from source.
CMake Deprecation Warning at os_stub/platform_lib/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


CMake Deprecation Warning at os_stub/platform_lib_null/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


-- Configuring done
-- Generating done
-- Build files have been written to: D:/Work/Projects/maxf/libspdm/build
```

PR #1929, opened 2023-04-15
Update CMake minimum required version

xiangfeima:
Fix: #1928
Update few remaining CMake minimum required version
from 2.6 to 2.8.12 to avoid CMake deprecation warning.


steven-bellock:
Thanks. We really haven't done much of an analysis on what the minimum CMake version should be. Apparently folks are using fairly "old" versions, maybe due to their Linux distribution. I don't think we really need any of the fancy new CMake features so we should be striving for "as old as comfortably possible".

steven-bellock (approved):

Issue #1942, opened 2023-04-16
Setting `LIBSPDM_DATA_PEER_USED_CERT_CHAIN_BUFFER` is incorrect


steven-bellock:
When setting `LIBSPDM_DATA_PEER_USED_CERT_CHAIN_BUFFER` a full certificate chain, with SPDM header and root certificate hash, is provided. If `LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT==0` then the certificate chain is hashed and the leaf certificate of the public key is extracted via the `libspdm_*_get_public_key_from_x509` functions. However those functions do not process the SPDM header and hash. For example https://github.com/DMTF/libspdm/blob/771d4c826ff35f0a27c21d904330070a5f05504b/os_stub/cryptlib_mbedtls/pk/x509.c#L432-L443

jyao1:
Or I think we can remove it, if it is only used by spdm-dump.

jyao1:
related issue: https://github.com/DMTF/libspdm/issues/1952

PR #1946, opened 2023-04-17
Add spdm cert_chain format process when set data

Wenxing-hou:
Fix: #1942

### DIFF
--- a/library/spdm_common_lib/libspdm_com_context_data.c
+++ b/library/spdm_common_lib/libspdm_com_context_data.c
@@ -159,6 +159,8 @@ libspdm_return_t libspdm_set_data(void *spdm_context, libspdm_data_type_t data_t
     uint16_t data16;
 #if !(LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT)
     bool status;
+    const uint8_t *cert_buffer;
+    size_t cert_buffer_size;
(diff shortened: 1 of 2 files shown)


library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -425,6 +425,13 @@ libspdm_return_t libspdm_set_data(void *spdm_context, libspdm_data_type_t data_t
                          sizeof(context->connection_info.peer_used_cert_chain[slot_id].buffer),
                          data, data_size);
 #else
+        /*process the SPDM cert header and hash*/
+        data = (uint8_t *)data + sizeof(spdm_cert_chain_t) +
+               libspdm_get_hash_size(context->connection_info.algorithm.base_hash_algo);
+        data_size = data_size -
+                    (sizeof(spdm_cert_chain_t) +
+                     libspdm_get_hash_size(context->connection_info.algorithm.base_hash_algo));
```
jyao1, line 433: Why we need skip header before hash?

That violates the SPDM spec.

Wenxing-hou: I have updated the code. And the unit_test is added to check the code. Thanks.

jyao1 (approved):

Issue #1934, opened 2023-04-15
Number of secured message versions need to be #defined


steven-bellock:
The SPDM secured message versions are defined in
https://github.com/DMTF/libspdm/blob/6684f43117784b12e25e665df65d1d122540f11f/include/industry_standard/spdm_secured_message.h#L54-L56

Similar to `SPDM_MAX_VERSION_COUNT` there needs to be a `SECURED_SPDM_MAX_VERSION_COUNT`. In particular https://github.com/DMTF/libspdm/blob/6684f43117784b12e25e665df65d1d122540f11f/library/spdm_common_lib/libspdm_com_context_data.c#L203-L204 is incorrect.

PR #1947, opened 2023-04-17
Add SPDM_MAX_SECURED_VERSION_COUNT MACRO

Wenxing-hou:
Fix: #1934

### DIFF
--- a/include/industry_standard/spdm_secured_message.h
+++ b/include/industry_standard/spdm_secured_message.h
@@ -51,6 +51,8 @@
  *                AssociatedData           AeadTag
  */
 
+/* 2 means SPDM secured message version 1.0, 1.1 */
+#define SECURED_SPDM_MAX_VERSION_COUNT 2
 /* the DSP0277 specification version */
 #define SECURED_SPDM_VERSION_10 0x10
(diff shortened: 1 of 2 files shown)


include/industry_standard/spdm.h
```diff
@@ -16,6 +16,8 @@
 
 /* 3 means SPDM spec 1.0, 1.1, 1.2 */
 #define SPDM_MAX_VERSION_COUNT 3
+/* 2 means SPDM secured message version 1.0, 1.1 */
+#define SPDM_MAX_SECURED_VERSION_COUNT 2
```
steven-bellock, line 20: This should go in https://github.com/DMTF/libspdm/blob/main/include/industry_standard/spdm_secured_message.h

steven-bellock: And please name it `SECURED_SPDM_MAX_VERSION_COUNT`.

steven-bellock (changes requested):

include/industry_standard/spdm.h
Wenxing-hou: Thanks. I have changed the code.

jyao1 (approved):

Issue #1952, opened 2023-04-18
Remove `LIBSPDM_DATA_LOCAL_USED_CERT_CHAIN_BUFFER`


steven-bellock:
This is used by https://github.com/DMTF/spdm-dump but an Integrator doesn't need to use it. After talking with @jyao1 let's get rid of this and `spdm-dump` can access the `spdm_context` directly.

jyao1:
agree to remove to eliminate confusing.

PR #1958, opened 2023-04-19
Remove LIBSPDM_DATA_LOCAL_USED_CERT_CHAIN_BUFFER

Wenxing-hou:
Fix: #1952


include/library/spdm_common_lib.h
```diff
@@ -70,6 +70,8 @@ typedef enum {
     LIBSPDM_DATA_HEARTBEAT_PERIOD,
 
     /* Negotiated result */
+
+    /* LIBSPDM_DATA_LOCAL_USED_CERT_CHAIN_BUFFER is deprecated. Do not use it. Reserved only. */
```
steven-bellock, line 74: You can just delete it, since this will be for the 3.0 release.

steven-bellock (changes requested):

include/library/spdm_common_lib.h
Wenxing-hou: Thanks. I have deleted it.

steven-bellock (approved):

Issue #1919, opened 2023-04-12
Can we not build latest LIBSPDM in Raspberry PI?


PrakashK94:
I ran these commands to build,
1. cmake -DARCH=arm -DTOOLCHAIN=GCC -DTARGET=Release -DCRYPTO=openssl ..
2. make copy_sample_key
3. make

Build output,
root@raspberrypi:/home/pi/Desktop/Hari/ResApp.bak20230403-1616/libspdm/build# make
[  0%] Building C object os_stub/cryptlib_openssl/CMakeFiles/cryptlib_openssl.dir/cipher/aead_aes_gcm.c.o
In file included from /home/pi/Desktop/Hari/ResApp.bak20230403-1616/libspdm/os_stub/cryptlib_openssl/internal_crypt_lib.h:20,
                 from /home/pi/Desktop/Hari/ResApp.bak20230403-1616/libspdm/os_stub/cryptlib_openssl/cipher/aead_aes_gcm.c:14:
/home/pi/Desktop/Hari/ResApp.bak20230403-1616/libspdm/os_stub/openssllib/include/crt_support.h:61:2: error: #error Unknown target architecture
 #error Unknown target architecture
  ^~~~~
In file included from /home/pi/Desktop/Hari/ResApp.bak20230403-1616/libspdm/os_stub/openssllib/openssl/include/openssl/aes.h:13,
                 from /home/pi/Desktop/Hari/ResApp.bak20230403-1616/libspdm/os_stub/cryptlib_openssl/cipher/aead_aes_gcm.c:15:
/home/pi/Desktop/Hari/ResApp.bak20230403-1616/libspdm/os_stub/openssllib/include/openssl/opensslconf.h:46:2: error: #error Unknown target architecture
 #error Unknown target architecture
  ^~~~~
In file included from /home/pi/Desktop/Hari/ResApp.bak20230403-1616/libspdm/os_stub/openssllib/openssl/include/openssl/evp.h:13,
                 from /home/pi/Desktop/Hari/ResApp.bak20230403-1616/libspdm/os_stub/cryptlib_openssl/cipher/aead_aes_gcm.c:16:
/home/pi/Desktop/Hari/ResApp.bak20230403-1616/libspdm/os_stub/openssllib/include/openssl/opensslconf.h:46:2: error: #error Unknown target architecture
 #error Unknown target architecture
  ^~~~~
[...]

PR #1959, opened 2023-04-19
Fix arm build error when openssl is updated

Wenxing-hou:
Fix: #1919


Wenxing-hou:
> @Wenxing-hou Did you verify this change locally?

Yes. I verify the change locally by using:
`cmake -DARCH=arm -DTOOLCHAIN=ARM_DS2022 -DTARGET=Release -DCRYPTO=openssl ..`.

The build result  changed from failure to success after the fix.

jyao1:
Now I think we need to add this (ARM) to CI.

Also add RISC-V to CI.

Issue #1976, opened 2023-04-23
According to FIPS 140-3, the salt needs to be cleared


Wenxing-hou:
Ref: https://github.com/DMTF/libspdm/issues/1260, key and salt need to be cleared.

The `data64` needs to be set 0, after the follow code:
https://github.com/DMTF/libspdm/blob/391ccd56d2df0574964005deec1769d9d2d24773/library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c#L128-L132

jyao1:
Why we need zero salt ?

Wenxing-hou:
Salt is the input of AEAD. When the salt is known, the attacker may accelerate the decoding work.

steven-bellock:
I'd say it qualifies as a critical security parameter, especially since its lifetime is the same as the secret key.

jyao1:
which document says that IV must be secret?

steven-bellock:
From [NIST 800-56B](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Br2.pdf):
> Security-related information whose disclosure or modification can compromise the security of a cryptographic module. Domain parameters, secret or private keys, shared secrets, key-derivation keys, intermediate values and **secret salts** are examples of quantities that may be considered critical security parameters in this Recommendation.

While the salts we're talking about aren't strictly secret, it shouldn't hurt to clear them.

steven-bellock:
Actually, since this salt value is derived from `* Direction Handshake Secret` or `* Direction Data Secret` one could make a case that it should be kept secret. In particular an attacker could create a lookup table of `* Direction * Secret`s and the corresponding salt values derived from that. If the attacker gets lucky and finds a salt value that is in their table then they know the value of the `* Direction * Secret` and can derive the key.

jyao1:
I am not convinced that IV/salt should be kept as secret based upon info above.
1) it shouldn't hurt to clear them != we must clear them.
2) lookup table attack is mitigated by nonce. I am not sure how that can work in this case.

On the other hand, if there is a way to attack the crypto with IV, then we have bigger problem, because the security space is downgrade from key length (256 bit or 128 bit) to IV length (96 bit).

steven-bellock:
We can ask SPDM WG if the salt is considered a secret value. I can also ask internally at my company.

jyao1:
We agree it is not secret. Close it.

PR #1977, opened 2023-04-23
Zero local variable slat

Wenxing-hou:
Fix: #1976

### DIFF
--- a/library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
+++ b/library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
@@ -131,6 +131,9 @@ libspdm_return_t libspdm_encode_secured_message(
         libspdm_write_uint64(salt, data64);
     }
 
+    /*clear for security*/
+    data64 = 0;
+
     sequence_num_in_header = 0;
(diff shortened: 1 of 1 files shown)


jyao1:
I dont understand:

1) Why we need zero salt? Salt is usually not secret.

2) How does this patch work? Compiler may optimize the zero value assignment.

library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
```diff
@@ -402,6 +405,9 @@ libspdm_return_t libspdm_decode_secured_message(
         libspdm_write_uint64(salt, data64);
     }
 
+    /*clear for security*/
+    data64 = 0;
```
jyao1, line 409: How that is useful ?

The compiler may optimize it.

library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
```diff
@@ -131,6 +131,9 @@ libspdm_return_t libspdm_encode_secured_message(
...
```
jyao1, line 135: How that is useful ?

The compiler may optimize it.

PR #1979, opened 2023-04-23
Fix unit_test using local variable as response

Wenxing-hou:


steven-bellock:
@jyao1 You can merge as you see fit.

Issue #1825, opened 2023-03-21
if response message size exceeds `MaxSPDMmsgSize` of requester then responder shall respond with `ErrorCode == ResponseTooLarge` or silently discard the request.


Zhiqiang520:
If the size of a response message exceeds the size of the `MaxSPDMmsgSize` of the requesting SPDM endpoint, the responding
SPDM endpoint shall respond with `ErrorCode == ResponseTooLarge` or silently discard the request.
1. Refer to paragraph 239  in [DSP0274_1.2.1](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf).

```
If the size of a response message exceeds the size of the `MaxSPDMmsgSize` of the requesting SPDM endpoint, the responding
SPDM endpoint shall respond with `ErrorCode == ResponseTooLarge` or silently discard the request.
```

2. Refer to paragraph 450 Table 48 — Error code and error data in [DSP0274_1.2.1](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf).

![image](https://user-images.githubusercontent.com/90754803/226577803-6fb2bc75-04dd-4359-b246-4979737d5583.png)

3. I think:

- a. the code at following line 622 :

`if (my_response_size < large_buffer_size) {` 
 may should be like:
`if (my_response_size <= large_buffer_size) {`.

- b. the code at following line 665~666: 

`status = libspdm_generate_error_response(context, `
`                                          SPDM_ERROR_CODE_LARGE_RESPONSE, `
may should be like:
`status = libspdm_generate_extended_error_response(context, `
`                                                  SPDM_ERROR_CODE_RESPONSE_TOO_LARGE,`
https://github.com/DMTF/libspdm/blob/0af7c1197cab8e99615fd9f062653aaf32552c03/library/spdm_responder_lib/libspdm_rsp_receive_send.c#L622-L669

steven-bellock:
`large_buffer_size` is the size of the Responder's scratch buffer, no? In that case this is an internal error to the Responder and it should be `Unspecified`. 

I think `libspdm_build_response` needs a major refactoring. It's difficult to figure out what's going on in it.

jyao1:
According to our definition:

```
#if LIBSPDM_ENABLE_CAPABILITY_CHUNK_CAP
...
[...]

PR #1984, opened 2023-04-24
Handle response too large and add unit_test

Wenxing-hou:
Fix: #1825

commit1:  Add check for `SPDM_ERROR_CODE_RESPONSE_TOO_LARGE`;
commit2:  Fix unit_test when add check for `SPDM_ERROR_CODE_RESPONSE_TOO_LARGE`;
commit3:  Add unit_test case;

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_receive_send.c
+++ b/library/spdm_responder_lib/libspdm_rsp_receive_send.c
@@ -434,6 +434,7 @@ libspdm_return_t libspdm_build_response(void *spdm_context, const uint32_t *sess
     libspdm_chunk_info_t* get_info;
     spdm_chunk_response_response_t *chunk_rsp;
     uint8_t *chunk_ptr;
+    uint32_t actual_size;
     #endif /* LIBSPDM_ENABLE_CAPABILITY_CHUNK_CAP */
 
     context = spdm_context;
@@ -627,7 +628,15 @@ libspdm_return_t libspdm_build_response(void *spdm_context, const uint32_t *sess
                        libspdm_get_scratch_buffer_large_message_offset(spdm_context);
         large_buffer_size = libspdm_get_scratch_buffer_large_message_capacity(spdm_context);
 
-        if (my_response_size < large_buffer_size) {
+        if (my_response_size > context->connection_info.capability.max_spdm_msg_size) {
+            actual_size = (uint32_t)my_response_size;
+            status = libspdm_generate_extended_error_response(context,
+                                                              SPDM_ERROR_CODE_RESPONSE_TOO_LARGE,
+                                                              0,
+                                                              sizeof(uint32_t),
+                                                              (uint8_t *)&actual_size,
+                                                              &my_response_size, my_response);
+        } else {
             get_info->chunk_in_use = true;
             /* Increment chunk_handle here as opposed to end of chunk_get handler
              * in case requester never issues chunk_get. */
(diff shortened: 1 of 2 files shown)


library/spdm_responder_lib/libspdm_rsp_receive_send.c
```diff
@@ -627,7 +627,12 @@ libspdm_return_t libspdm_build_response(void *spdm_context, const uint32_t *sess
                        libspdm_get_scratch_buffer_large_message_offset(spdm_context);
         large_buffer_size = libspdm_get_scratch_buffer_large_message_capacity(spdm_context);
 
-        if (my_response_size < large_buffer_size) {
+        if (my_response_size > context->connection_info.capability.max_spdm_msg_size) {
+            status = libspdm_generate_error_response(context,
+                                                     SPDM_ERROR_CODE_RESPONSE_TOO_LARGE,
```
Zhiqiang520, line 632: `ResponseTooLarge` ERROR response message has its `ExtendedErrorData`, so I think here `libspdm_generate_error_response()` function should be changed to `libspdm_generate_extended_error_response()` function.

library/spdm_responder_lib/libspdm_rsp_receive_send.c
```diff
@@ -664,17 +669,15 @@ libspdm_return_t libspdm_build_response(void *spdm_context, const uint32_t *sess
...
+                       "Warning: Could not save chunk. Scratch buffer too small.\n"));
 
-            status = libspdm_generate_error_response(context,
-                                                     SPDM_ERROR_CODE_LARGE_RESPONSE,
-                                                     spdm_request->request_response_code,
-                                                     &my_response_size, my_response);
-        }
+        status = libspdm_generate_error_response(context,
+                                                 SPDM_ERROR_CODE_LARGE_RESPONSE,
+                                                 spdm_request->request_response_code,
```
Zhiqiang520, line 678: `ResponseLarge` ERROR response message has its `ExtendedErrorData`, so I think here `libspdm_generate_error_response()` function should be changed to `libspdm_generate_extended_error_response()` function.

library/spdm_responder_lib/libspdm_rsp_receive_send.c
Wenxing-hou: Thanks. I have changed `libspdm_generate_error_response()` function should be changed to `libspdm_generate_extended_error_response()`.

library/spdm_responder_lib/libspdm_rsp_receive_send.c
Wenxing-hou: Thanks. I have changed `libspdm_generate_error_response()` function should be changed to `libspdm_generate_extended_error_response()`.

library/spdm_responder_lib/libspdm_rsp_receive_send.c
```diff
@@ -627,7 +627,14 @@ libspdm_return_t libspdm_build_response(void *spdm_context, const uint32_t *sess
                        libspdm_get_scratch_buffer_large_message_offset(spdm_context);
         large_buffer_size = libspdm_get_scratch_buffer_large_message_capacity(spdm_context);
 
-        if (my_response_size < large_buffer_size) {
+        if (my_response_size > context->connection_info.capability.max_spdm_msg_size) {
+            status = libspdm_generate_extended_error_response(context,
+                                                              SPDM_ERROR_CODE_RESPONSE_TOO_LARGE,
+                                                              0,
+                                                              sizeof(my_response_size),
```
Zhiqiang520, line 634: I think maybe `sizeof(my_response_size)` at line 634 should be changed to`sizeof(uint32_t)`.
Because `ExtendedErrorData` **length** for `ResponseTooLarge` is explicitly 4 in `Table 52 — ExtendedErrorData format for ResponseTooLarge` in DSP0274_1.2.1.pdf

jyao1: Agree with Zhiqiang.

Wenxing-hou: Yes. I have added the `actual_size` to ensure that the code is right whenever using in big endian or little endian.

PR #1995, opened 2023-04-28
Allow Integrator to retrieve opaque data

### LINKED ISSUE #1694: Requester functions should return opaque data if desired
Most Requester functions only print out the opaque data returned by the Responder. For example : https://github.com/steven-bellock/libspdm/blob/6e0d7afe5c73a6a4b3f8f55b9910f7b588dcc00c/library/spdm_requester_lib/libspdm_req_challenge.c#L280-L285 The Integrator should have the option to store this information in a buffer.

steven-bellock:
Fix #1694.

### DIFF
--- a/include/library/spdm_requester_lib.h
+++ b/include/library/spdm_requester_lib.h
@@ -306,36 +306,40 @@ libspdm_return_t libspdm_start_session(void *spdm_context, bool use_psk,
  * If encapsulated mutual authentication is requested from the responder,
  * this function also perform the encapsulated mutual authentication.
  *
- * @param  spdm_context              A pointer to the SPDM context.
- * @param  use_psk                   False means to use KEY_EXCHANGE/FINISH to start a session.
- *                                   True means to use PSK_EXCHANGE/PSK_FINISH to start a session.
- * @param  psk_hint                  The psk_hint in PSK_EXCHANGE. It is ignored if use_psk is false.
- * @param  psk_hint_size             The size in bytes of psk_hint. It is ignored if use_psk is false.
- * @param  measurement_hash_type     The type of the measurement hash.
- * @param  slot_id                   The number of slot for the certificate chain.
- * @param  session_policy            The policy for the session.
- * @param  session_id                The session ID of the session.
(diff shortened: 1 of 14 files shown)


include/library/spdm_requester_lib.h
```diff
@@ -347,7 +351,9 @@ libspdm_return_t libspdm_start_session_ex(void *spdm_context, bool use_psk,
                                           void *requester_random,
                                           size_t *requester_random_size,
                                           void *responder_random,
-                                          size_t *responder_random_size);
+                                          size_t *responder_random_size,
+                                          void *opaque_data,
```
jyao1, line 355: Different from CHALLENGE and GET_MEASUREMENT, KEY_EXCHANGE and PSK_EXCHANGE have requester_opaque_data.

Do we want to allow integrator input requester_opaque_data ? or return requester_opaque_data from libspdm ?

steven-bellock: Probably should, but will need to integrate the Integrator's opaque data with DSP0277 opaque data. If `OpaqueDataFmt1` is selected then libspdm needs to know the number of opaque elements.

jyao1: I think we can just let integrator fill **everything**.

I think the logic could be:
```
if (requester_opaque_data == null) {
    opaque_key_exchange_req_size =
        libspdm_get_opaque_data_supported_version_data_size(spdm_context);
    libspdm_write_uint16(ptr, (uint16_t)opaque_key_exchange_req_size);
    ptr += sizeof(uint16_t);
    libspdm_build_opaque_data_supported_version_data(
        spdm_context, &opaque_key_exchange_req_size, ptr);
    ptr += opaque_key_exchange_req_size;
} else {
    opaque_key_exchange_req_size  = requester_opaque_data_size;
    libspdm_write_uint16(ptr, (uint16_t)opaque_key_exchange_req_size);
    ptr += sizeof(uint16_t);
    libspdm_copy_mem (ptr, requester_opaque_data_size, requester_opaque_data, requester_opaque_data_size)
}
```

We already did that in GET_CSR API.

jyao1 (approved):

Issue #1987, opened 2023-04-25
Sending GET_VERSION inside a session (which is prohibited by the spec) resets the connection


PavelPu123:
From DSP0274_1.2.1 (page 37, table 6, last table line) it seems that GET_VERSION is not allowed inside the session. On my understanding there should be an error message issued in this case and connection reset should be prevented. In reality we see that the reset is happening in this case.

steven-bellock:
Filed #1988 for version checks.

PR #2000, opened 2023-04-29
add check for GET_VERSION in session.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/1987


steven-bellock:
This also applies to `(V)CA`.

jyao1:
I think CA cannot success because of the connection_state check.

jyao1:
```

steven-bellock:
> I think CA cannot success because of the connection_state check.

Ok.

steven-bellock (approved):

Issue #2005, opened 2023-05-01
SPDM mutual authentication bypass


jyao1:
A vulnerability has been identified in SPDM session establishment.
If a device supports both DHE session and PSK session with mutual
authentication, the attacker may be able to establish the session with
KEY_EXCHANGE and PSK_FINISH to bypass the mutual authentication.  This
is most likely to happen when the Requester begins a session using 
one method (DHE, for example) and then uses the other method's finish
(PSK_FINISH in this example) to establish the session.  The session 
hashes would be expected to fail in this case, but the condition was
not detected.

Impacted Function:

This issue only impacts the SPDM responder, which supports KEY_EX_CAP=1 and
PSK_CAP=10b at same time with mutual authentication requirement.
The SPDM requester is not impacted.
The SPDM responder is not impacted if KEY_EX_CAP=0 or PSK_CAP=0 or PSK_CAP=01b.
The SPDM responder is not impacted if mutual authentication is not required.

Severity scoring:

CVSS: https://www.first.org/cvss/calculator/3.1#CVSS:3.1/AV:A/AC:L/PR:L/UI:N/S:C/C:H/I:H/A:H - 9.0
Severity: Critical

Acknowledgement:

This issue was reported on 20 March 2023 by Cas Cremers <cremers@cispa.de>,
together with Naska, Aurora <aurora.naska@cispa.de>;
Dax, Alexander <alexander.dax@cispa.de>.

PR #2007, opened 2023-05-01
Add handshake mode switch check in FINISH and PSK_FINISH.

jyao1:
Reference: DMTF-2023-0001

Fix: https://github.com/DMTF/libspdm/issues/2005


steven-bellock (approved):
You'll want to edit `VERSION.md` before releasing 2.3.2.

PR #2012, opened 2023-05-03
libspdm_doe_pcidoe: Add DOE discovery encode/decode helpers

alistair23:
Add helper functions for encoding/decoding the DOE discovery protocol. All DOE requesters and responders must implement this protocol.

There isn't a straight forwrad way to integrate this into libspdm so let's instead implement helper functions that can be used directly by the custom send and receive functions.

### DIFF
--- a/include/library/spdm_transport_pcidoe_lib.h
+++ b/include/library/spdm_transport_pcidoe_lib.h
@@ -65,6 +65,21 @@ libspdm_return_t libspdm_transport_pci_doe_encode_message(
     size_t *transport_message_size, void **transport_message);
 
 /**
+ * Encode a DOE discovery message.
+ *
+ * @param  message_size                         Size in bytes of the message data buffer.
+ * @param  message                              A pointer to a source buffer to store the message.
+ * @param  transport_message_size               Size in bytes of the transport message data buffer.
+ * @param  transport_message                    A pointer to a destination buffer to store the transport message.
+ *
+ * @retval LIBSPDM_STATUS_SUCCESS               The message is encoded successfully.
+ * @retval LIBSPDM_STATUS_INVALID_PARAMETER     The message is NULL or the message_size is zero.
+ **/
+libspdm_return_t libspdm_pci_doe_encode_discovery(size_t message_size, void *message,
+                                                  size_t *transport_message_size,
+                                                  void **transport_message);
+
+/**
  * Decode an SPDM or APP message from a transport layer message.
  *
  * For normal SPDM message, it removes the transport layer wrapper,
--- a/include/library/spdm_transport_pcidoe_lib.h
+++ b/include/library/spdm_transport_pcidoe_lib.h
@@ -101,6 +101,51 @@ libspdm_return_t libspdm_transport_pci_doe_decode_message(
     size_t *message_size, void **message);
 
 /**
+ * Decode a DOE discovery request message.
+ *
+ * @param  transport_message_size               Size in bytes of the transport message data buffer.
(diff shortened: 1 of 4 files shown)


include/library/spdm_transport_pcidoe_lib.h
```diff
@@ -62,6 +62,21 @@ libspdm_return_t libspdm_transport_pci_doe_encode_message(
...
 
+/**
+ * Encode a DOE discovery message.
+ *
+ * @param  message_size                   size in bytes of the message data buffer.
+ * @param  message                       A pointer to a source buffer to store the message.
+ * @param  transport_message_size         size in bytes of the transport message data buffer.
+ * @param  transport_message             A pointer to a destination buffer to store the transport message.
+ *
+ * @retval RETURN_SUCCESS               The message is encoded successfully.
```
steven-bellock, line 73: Use the `LIBSPDM_STATUS_` macros. We are in the process of converting the old `RETURN_` macros.

library/spdm_transport_pcidoe_lib/libspdm_doe_pcidoe.c
```diff
@@ -128,6 +128,61 @@ libspdm_return_t libspdm_pci_doe_encode_message(const uint32_t *session_id,
     return LIBSPDM_STATUS_SUCCESS;
 }
 
+/**
```
steven-bellock, line 131: Delete the doxygen comments from non-static functions in .c files that already have the doxygen comments in the function declaration.

jyao1:
May I know which component should be the consumer for this new API?

alistair23:
The idea is that the send/receive implementations of `libspdm_register_device_io_func()` can use this to check for DOE discovery messages and handle them

jyao1:
do you mean the [spdm-emu](https://github.com/DMTF/spdm-emu) should use this API?
If so, would you please propose a patch in spdm-emu? I am trying to understand how that works.

jyao1:
SPDM-EMU is using libspdm and it did send and receive DOE discovery.

Would you please propose a change to help us understand how to use this library API?

alistair23:
Done! There is an example of what this would look like here: https://github.com/DMTF/spdm-emu/pull/268

alistair23:
Any comments on this?

steven-bellock:
> Any comments on this?

This is just straight non-SPDM PCIE discovery? Aren't there more generic PCIE libraries that do this? This seems too far removed from SPDM to be in libspdm.

alistair23:
It is generic PCIe discovery. My thinking is that as DOE discovery is pretty simple and it's critical to SPDM (over DOE), these helper functions should be included in libspdm.

I think all DOE over PCIe users will require these or similar functions, [spdm-emu](https://github.com/DMTF/spdm-emu/pull/268/files) implemented something itself for example.

These could be in a different generic PCIe library, but then anyone using libspdm needs to include another library for something we can do in a few hundred lines on code. On top of that this API matches the style of libspdm, which makes it easier to integrate.

By having one copy of the helper functions here, we can avoid bugs and issues in individual implementations. For example we can already remove code from [spdm-emu](https://github.com/DMTF/spdm-emu/pull/268/files) we can do a similar thing for all libspdm users.

I don't feel it's too far removed from SPDM to be included

alistair23:
Any more thoughts on this?

jyao1:
https://github.com/DMTF/spdm-emu/pull/268 is still a draft.

Are you still working on it?

alistair23:
https://github.com/DMTF/spdm-emu/pull/268 is ready to go, it just needs this merged first

jyao1 (approved):

PR #2014, opened 2023-05-03
Enforce DSP0277 capabilities

### LINKED ISSUE #1986: Setting ENCRYPT_CAP to 1 and MAC_CAP to 0 in GET_CAPABILITIES message leads to LIBSPDM_ASSERT
While setting the mentioned configuration in GET_CAPABILITIES message we're getting to:

https://github.com/DMTF/libspdm/blob/6684f43117784b12e25e665df65d1d122540f11f/library/spdm_common_lib/libspdm_com_context_data_session.c#L62

As I understand there should be a verification of this combination somewhere in:
https://github.com/DMTF/libspdm/blob/6684f43117784b12e25e665df65d1d122540f11f/library/spdm_responder_lib/libspdm_rsp_capabilities.c#L70-L85
and there should be an error message since this combination violates the spec.

Thanks.

steven-bellock:
Fix #1986.

Note that this code path is already covered in a unit test that sets `ENC_CAP=0` and `MAC_CAP=0`.

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_get_capabilities.c
+++ b/library/spdm_requester_lib/libspdm_req_get_capabilities.c
@@ -74,7 +74,10 @@ static bool validate_responder_capability(uint32_t capabilities_flag, uint8_t ve
 
         /* Checks that originate from key exchange capabilities. */
         if ((key_ex_cap == 1) || (psk_cap != 0)) {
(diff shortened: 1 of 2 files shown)


steven-bellock:
@PavelPu123 as well.

jyao1 (approved):

Issue #2003, opened 2023-04-30
`GET_CSR` check enhancements


steven-bellock:
1. https://github.com/DMTF/libspdm/blob/043f81877e435c7a84d3134128f566edfdddefbb/library/spdm_requester_lib/libspdm_req_get_csr.c#L154 `csr_length` is unsigned so it cannot be less than `0`. Change `<=` to `==`. 
2. Provide a check that if `ResetRequired` is returned by the Responder then its `CERT_INSTALL_RESET_CAP` is `1`.

jyao1:
Using "UINT value <= 0" is a bug, because "< 0" is impossible.

PR #2018, opened 2023-05-04
Make csr_len check right

Wenxing-hou:
Fix: #2003

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_get_csr.c
+++ b/library/spdm_requester_lib/libspdm_req_get_csr.c
@@ -151,7 +151,7 @@ static libspdm_return_t libspdm_try_get_csr(libspdm_context_t *spdm_context,
         goto receive_done;
     }
 
-    if (spdm_response->csr_length <= 0) {
+    if (spdm_response->csr_length == 0) {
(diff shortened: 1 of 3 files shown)


steven-bellock:
Before #2003 is closed there also needs to be a pull request for. 
>Provide a check that if ResetRequired is returned by the Responder then its CERT_INSTALL_RESET_CAP is 1.

It's probably easier to do it in this one.

Wenxing-hou:
> Before #2003 is closed there also needs to be a pull request for.
> 
> > Provide a check that if ResetRequired is returned by the Responder then its CERT_INSTALL_RESET_CAP is 1.
> 
> It's probably easier to do it in this one.


Thanks. I have added a commit to provide the check:
```
    if ((last_spdm_request->header.request_response_code == SPDM_SET_CERTIFICATE ||
         last_spdm_request->header.request_response_code == SPDM_GET_CSR) &&
        (libspdm_is_capabilities_flag_supported(
             spdm_context, true, 0,
             SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_CERT_INSTALL_RESET_CAP))) {

```

steven-bellock (approved):

Issue #1832, opened 2023-03-23
If request message size exceeds `MaxSPDMmsgSize` of responder then responder shall respond with `ErrorCode=RequestTooLarge` or silently discard the request.


Zhiqiang520:
If the size of a request message exceeds the size of the `MaxSPDMmsgSize` of the responding SPDM endpoint, the
responding SPDM endpoint shall respond with `ErrorCode=RequestTooLarge` or silently discard the request.

1. Refer to paragraph 239 in [DSP0274_1.2.1](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf).
```
if the size of a request message exceeds the size of the MaxSPDMmsgSize of the responding SPDM endpoint, the
responding SPDM endpoint shall respond with ErrorCode=RequestTooLarge or silently discard the request.
```
2. Refer to paragraph 450 Table 48 — Error code and error data in [DSP0274_1.2.1](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf).
![image](https://user-images.githubusercontent.com/90754803/227152730-7de0f127-31cd-4ef7-8257-bbbd19bd0d30.png)
3. I think:
Maybe we should first to check `if (request_size <= spdm_context->connection_info.capability.max_spdm_msg_size)`, then call `libspdm_handle_large_request()` function at the following line 574 to handle the large SPDM message.
https://github.com/DMTF/libspdm/blob/cbf9e65c6dfe086acc2909e5b1ca697ea5c2c648/library/spdm_requester_lib/libspdm_req_send_receive.c#L553-L579

steven-bellock:
This one makes my head hurt, but I believe `RequestTooLarge` is used when the Requester sends `GET_ENCAPSULATED_REQUEST` and the `ENCAPSULATED_REQUEST` is larger than the Requester's `MaxSPDMmsgSize`. In such a case the Responder would reply with error `RequestTooLarge`.

But it might be good to ask the SPDM working group if this is the correct use of it.

steven-bellock:
After discussing this with the SPDM working group there's an even simpler case. The Requester attempts to send a request message that is larger than the Responder's `MaxSPDMmsgSize`. In such a case the Responder can reply with error `RequestTooLarge`. Note that a non-malicious Requester would never send the request in the first place.

jyao1:
Current libspdm cannot return `RequestTooLarge`, because transport layer does not tell such info.

[...]

PR #2027, opened 2023-05-05
Handle requester message too large

Wenxing-hou:
Fix: #1832

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_send_receive.c
+++ b/library/spdm_requester_lib/libspdm_req_send_receive.c
@@ -588,8 +588,12 @@ libspdm_return_t libspdm_send_spdm_request(libspdm_context_t *spdm_context,
             spdm_context->last_spdm_request_size = request_size;
         }
 
-        status = libspdm_handle_large_request(
-            spdm_context, session_id, request_size, request);
+        if (request_size > spdm_context->connection_info.capability.max_spdm_msg_size) {
+            return LIBSPDM_STATUS_PEER_BUFFER_TOO_SMALL;
+        } else {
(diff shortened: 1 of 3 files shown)


library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -588,8 +588,12 @@ libspdm_return_t libspdm_send_spdm_request(libspdm_context_t *spdm_context,
             spdm_context->last_spdm_request_size = request_size;
         }
 
-        status = libspdm_handle_large_request(
-            spdm_context, session_id, request_size, request);
+        if (request_size > spdm_context->connection_info.capability.max_spdm_msg_size) {
+            return LIBSPDM_STATUS_SEND_FAIL;
```
jyao1, line 592: @steven-bellock , do you think if we need a special error code?

steven-bellock: Yes, `LIBSPDM_STATUS_SEND_FAIL` is for transport issues, but this is a buffer-too-small issue. So let's make 
>/* Message is too large to send to peer. */
#define LIBSPDM_STATUS_PEER_BUFFER_TOO_SMALL \

steven-bellock (changes requested):

library/spdm_requester_lib/libspdm_req_send_receive.c
Wenxing-hou: Thanks. I have added the `LIBSPDM_STATUS_PEER_BUFFER_TOO_SMALL` MACRO.

steven-bellock (approved):

Issue #1712, opened 2023-02-20
OpaqueData field shall conform to the selected opaque data format in OtherParamsSelection.


Zhiqiang520:
Currently there is **no format check** for `OpaqueData` Field in CHALLENGE_AUTH response message, MEASUREMENTS response message and GET_CSR request message.

1. Refer to paragraph 353 Table 36 — Successful CHALLENGE_AUTH response message format in
https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf
**OpaqueData** Field:
The Responder can include Responder-specific information and/or
information that its transport defines. If present, this field **shall
conform** to the selected opaque data format in
`OtherParamsSelection` .

2. Refer to paragraph 806 in
https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf
If the `OpaqueDataFmt1` bit is selected in `OtherParamsSelection` of `ALGORITHMS` , then all opaque data fields in SPDM messages shall use the format that Table 92 — General opaque data format defines.

steven-bellock:
What can be checked?

Zhiqiang520:
> What can be checked?

For example,
Refer to paragraph 808 Table 93 — Opaque element defines the format for each element in OpaqueList.
1. `ID` Field: **Shall be one of the values** in the ID column of Table 50 — Registry or standards body ID. 

Need add check if `ID` Field is one of the values  in the ID column of Table 50 — Registry or standards body ID.

2. `AlignPadding` Field: If 4 + VendorIDLen + OpaqueElementDataLen **does not fall on a 4-byte boundary**, this field shall be present and of the correct length to ensure 4 + VendorIDLen +OpaqueElementDataLen + AlignPaddingSize is a multiple
of 4. The value of this field shall be all zeros and the size of this field shall be 0, 1, 2 or 3.

Need add check if 4 + VendorIDLen + OpaqueElementDataLen fall on a 4-byte boundary.

PR #2040, opened 2023-05-11
OpaqueData field shall conform to the selected opaque data format in OtherParamsSelection

Xiaohanjlll:
Fix: #1712

### DIFF
--- a/library/spdm_common_lib/libspdm_com_opaque_data.c
+++ b/library/spdm_common_lib/libspdm_com_opaque_data.c
@@ -230,3 +230,105 @@ bool libspdm_get_element_from_opaque_data(libspdm_context_t *spdm_context,
 
     return result;
 }
+
+/**
+ *  Process general opaque data check
+ *
+ * @param  data_in_size                  size in bytes of the data_in.
+ * @param  data_in                       A pointer to the buffer to store the opaque data version selection.
+ *
+ * @retval true                           check opaque data successfully
+ * @retval false                          check opaque data failed
+ **/
+bool libspdm_process_general_opaque_data_check(libspdm_context_t *spdm_context,
+                                               size_t data_in_size,
+                                               void *data_in)
+{
+    const spdm_general_opaque_data_table_header_t
+    *spdm_general_opaque_data_table_header;
+    const opaque_element_table_header_t
+    *opaque_element_table_header;
+    uint8_t element_num;
+    uint8_t element_index;
+    uint16_t opaque_element_data_len;
+    size_t data_element_size;
+    size_t current_element_len;
+    size_t total_element_len;
+    uint8_t zero_padding[4] = {0};
+
+    total_element_len = 0;
+
+    LIBSPDM_ASSERT(data_in_size <= SPDM_MAX_OPAQUE_DATA_SIZE);
+
+    if (libspdm_get_connection_version(spdm_context) >= SPDM_MESSAGE_VERSION_12) {
+        if (spdm_context->connection_info.algorithm.other_params_support ==
+            SPDM_ALGORITHMS_OPAQUE_DATA_FORMAT_1) {
+            /* Check byte alignment */
+            if ((data_in_size & 3) != 0) {
+                return false;
+            }
+
+            spdm_general_opaque_data_table_header = data_in;
+            if (data_in_size < sizeof(spdm_general_opaque_data_table_header_t)) {
+                return false;
+            }
+            if (spdm_general_opaque_data_table_header->total_elements < 1) {
+                return false;
+            }
+            opaque_element_table_header = (const void *)(spdm_general_opaque_data_table_header + 1);
+
+            element_num = spdm_general_opaque_data_table_header->total_elements;
+
+            data_element_size = data_in_size - sizeof(spdm_general_opaque_data_table_header_t);
+
+            for (element_index = 0; element_index < element_num; element_index++) {
+                /*ensure the opaque_element_table_header is valid*/
+                if (total_element_len + sizeof(opaque_element_table_header_t) +
+                    sizeof(opaque_element_data_len) >
+                    data_element_size) {
+                    return false;
+                }
+
+                /*check element header id*/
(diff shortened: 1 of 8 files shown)


library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -230,3 +230,96 @@ bool libspdm_get_element_from_opaque_data(libspdm_context_t *spdm_context,
...
+    uint8_t element_num;
+    uint8_t element_index;
+    uint16_t opaque_element_data_len;
+    size_t data_element_size;
+    size_t current_element_len;
+    size_t total_element_len;
+
+    total_element_len = 0;
+
+    if(data_in_size > SPDM_MAX_OPAQUE_DATA_SIZE)
```
steven-bellock, line 260: Have this check be a `LIBSPDM_ASSERT`.

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -230,3 +230,96 @@ bool libspdm_get_element_from_opaque_data(libspdm_context_t *spdm_context,
...
+    {
+        return false;
+    }
+
+    if(libspdm_get_connection_version (spdm_context) >= SPDM_MESSAGE_VERSION_12) {
```
steven-bellock, line 265: `if(` -> `if (` in multiple places.
I'm surprised uncrustify doesn't catch this. I'll see if it can be enabled.

steven-bellock (changes requested):

library/spdm_common_lib/libspdm_com_opaque_data.c
Xiaohanjlll: Thanks,updated to AA `LIBSPDM_ASSERT`

library/spdm_common_lib/libspdm_com_opaque_data.c
Xiaohanjlll: Thanks ,I have changed ` if( ` to  `if  (`

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -230,3 +230,93 @@ bool libspdm_get_element_from_opaque_data(libspdm_context_t *spdm_context,
...
+                }
+
+                opaque_element_data_len = *(uint16_t *)((size_t)(opaque_element_table_header + 1)) +
+                                          opaque_element_table_header->vendor_len;
+
+                current_element_len = sizeof(opaque_element_table_header_t) +
+                                      opaque_element_table_header->vendor_len +
+                                      sizeof(opaque_element_data_len) +
+                                      opaque_element_data_len;
+                /* Add Padding*/
```
steven-bellock, line 303: Can also check that if there is padding then the padding bytes are zeroes.

steven-bellock (changes requested):
`libspdm_process_general_opaque_data_check` can be used in other functions, correct? `KEY_EXCHANGE` / `PSK_EXCHANGE`?

library/spdm_common_lib/libspdm_com_opaque_data.c
```diff
@@ -230,3 +230,93 @@ bool libspdm_get_element_from_opaque_data(libspdm_context_t *spdm_context,
...
+            for (element_index = 0; element_index < element_num; element_index++) {
+                /*ensure the opaque_element_table_header is valid*/
+                if (total_element_len + sizeof(opaque_element_table_header_t) +
+                    sizeof(opaque_element_data_len) >
+                    data_element_size) {
+                    return false;
+                }
+
+                /*check element header id*/
+                if (opaque_element_table_header->id > SPDM_REGISTRY_ID_JEDEC) {
```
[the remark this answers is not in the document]
steven-bellock: https://github.com/DMTF/libspdm/issues/2064

Zhiqiang520:
> `libspdm_process_general_opaque_data_check` can be used in other functions, correct? `KEY_EXCHANGE` / `PSK_EXCHANGE`?

Yes, I think `libspdm_process_general_opaque_data_check` can be used in `KEY_EXCHANGE` / `PSK_EXCHANGE`.
But due to `KEY_EXCHANGE` / `PSK_EXCHANGE` are related with the `Secured Messages opaque data format` in [DSP0277](https://www.dmtf.org/dsp/DSP0277), maybe need check the compatibility, so I suggest that we can submit **another PR** to use `libspdm_process_general_opaque_data_check` in `KEY_EXCHANGE` / `PSK_EXCHANGE`.

library/spdm_common_lib/libspdm_com_opaque_data.c
Xiaohanjlll: Thanks. Padding byte check updated

steven-bellock:
0277 1.0 also states 
> In many SPDM requests and response, an opaque data field exists to accommodate transport, standard organizations, or vendor specific use cases. The [Secured Message general opaque data](https://github.com/DMTF/SPDM-WG/blob/master/docs/DSP0277/DSP0277.md#gen-opaque-table) table defines the general format for all opaque data fields. If the selected SPDM protocol version is 1.1, then all opaque data fields in SPDM messages shall utilize the format defined by [Secured Message general opaque data](https://github.com/DMTF/SPDM-WG/blob/master/docs/DSP0277/DSP0277.md#gen-opaque-table). This format allows an SPDM message to contain multiple vendor or standard bodies' opaque data without collision.

Which means that if an endpoint supports 0277, `(KEY_EX_CAP || PSK_EX_CAP)`, then all of its opaque data needs to be formatted to the Secured Message general opaque data format, not just the opaque data in key exchange. But that can be a separate issue as well.

steven-bellock (approved):

jyao1:
I think it is good enough to check only SPDM 1.2 and above.
For SPDM 1.0/1.1, there is not rule for opaque data for MEASUREMENTS/CHALLENG_AUTH. We shall not do any check.

Issue #2047, opened 2023-05-17
The format of `*.csr` file in `test_csr` folder shall be DER format


Zhiqiang520:
The format of `*.csr` file in `test_csr` folder shall be DER format.
1. It looks like the format of `*.csr` file in `test_csr` folder is PEM format, not DER format.
https://github.com/DMTF/libspdm/blob/36e5a0e4c4c3570b522169eefdab9f8c8d9c815b/unit_test/sample_key/test_csr/ecp256.csr#L1-L8
2. But `CSRdata` Field shall be `DER-encoded` in `Table 80 — CSR response message format` in [DSP0274_1.2.1.pdf](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf).
![image](https://github.com/DMTF/libspdm/assets/90754803/dd91fd1e-2c95-4b10-b82b-40aa94981456)

PR #2050, opened 2023-05-18
Change CSR format from pem to DER

Wenxing-hou:
Fix: #2047


steven-bellock:
@jyao1 Feel free to merge.

Issue #2039, opened 2023-05-11
The OpaqueData offset in `libspdm_req_get_csr.c` and `libspdm_rsp_csr.c` not match with the `Byte offset` of `OpaqueData` Field in `GET_CSR` request message format


Zhiqiang520:
The **OpaqueData offset** in `libspdm_req_get_csr.c` and `libspdm_rsp_csr.c` does not match with the `Byte offset` of `OpaqueData` Field in `Table 79 — GET_CSR request message format` in [DSP0274_1.2.1.pdf](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf).

1. The `OpaqueData offset` in `libspdm_req_get_csr.c` should be 8 + `RequesterInfoLength`, following the spec.
GET_CSR request message format
![image](https://github.com/DMTF/libspdm/assets/90754803/962a5aff-737b-405b-a32f-f7d38cf8352a)

2. But now the `OpaqueData offset` in `libspdm_req_get_csr.c` is `spdm_request + 1`(=8) at the following line 97, it is actually `RequesterInfo` Field, not `OpaqueData` Field.
https://github.com/DMTF/libspdm/blob/e545b07b7b6f3176f6568087c93983a4a7fda464/library/spdm_requester_lib/libspdm_req_get_csr.c#L96-L106

3. The same issue is also in `libspdm_rsp_csr.c` at the following line 116.
https://github.com/DMTF/libspdm/blob/e545b07b7b6f3176f6568087c93983a4a7fda464/library/spdm_responder_lib/libspdm_rsp_csr.c#L116

steven-bellock:
Looks like that's been there since the very beginning : https://github.com/DMTF/libspdm/commit/0ab27873ba24d392bb6c58fd271bcec8c22c4740 Nobody has complained probably because
1. Nobody has a use for those fields.
2. Nobody is using `GET_CSR`.
But this highlights a deficiency in the unit tests, as they aren't checking the values of those fields.

jyao1:
I think probably we also need to fix in 2.3 branch.

Xiaohanjlll:
I can take this issue

steven-bellock:
> I think probably we also need to fix in 2.3 branch.
> 
> Also need to check unit test.

We can bundle it with the other fix.

PR #2057, opened 2023-05-23
Fix byte offset mismatch in OpaqueData field in GET_CSR request message format

Xiaohanjlll:
Fix: #2039


jyao1 (approved):

steven-bellock:
We should add 2.3 to CI/CD.

steven-bellock (approved):

steven-bellock:
I ran CI/CD manually in https://github.com/DMTF/libspdm/actions/runs/5082881037.

Issue #2051, opened 2023-05-18
`requester_info` can't represent the whole `GET_CSR` request message.


Zhiqiang520:
`requester_info` is just **part** of the whole `GET_CSR` request message, it can't represent the whole `GET_CSR` request message.

1. After the Responder has processed the reset, current code decide a new `GET_CSR` request is same or different with the `GET_CSR` request that was sent before the reset, by calling `libspdm_consttime_is_mem_equal(cached_req_info, requester_info` at the following line 1142 to compare the `cached_req_info` and the new `requester_info`, but `requester_info` is just **part** of the whole `GET_CSR` request message, it can't represent the whole `GET_CSR` request message.
https://github.com/DMTF/libspdm/blob/36e5a0e4c4c3570b522169eefdab9f8c8d9c815b/os_stub/spdm_device_secret_lib_sample/lib.c#L1134-L1144

2. Refer to the paragraph 666 in [DSP0274_1.2.1.pdf](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf).
![image](https://github.com/DMTF/libspdm/assets/90754803/6d0081a7-6f53-4a31-b48a-3994ae8ef564)

3. There are many other `Field` in `Table 79 — GET_CSR request message format`.
![image](https://github.com/DMTF/libspdm/assets/90754803/b934f8bf-7896-4ab1-9a7c-d7c9edf9badc)

steven-bellock:
Yeah I was wondering how to handle that in https://github.com/DMTF/libspdm/issues/2002. Maybe libspdm should pass a pointer to the request message through `libspdm_gen_csr` and the Integrator can either store the message or a hash of it that persists between resets.

Wenxing-hou:
> Yeah I was wondering how to handle that in #2002. Maybe libspdm should pass a pointer to the request message through `libspdm_gen_csr` and the Integrator can either store the message or a hash of it that persists between resets.

I will fix this issue after the PR: #2043 and #2045 are merged.

And the total request needs to be taken as `libspdm_gen_csr` input. The `requester_info` and `opaque_data` are obtained from the total request.

    result = libspdm_gen_csr(spdm_context->connection_info.algorithm.base_hash_algo,

PR #2058, opened 2023-05-23
Check the whole last SPDM CSR request data

Wenxing-hou:
Fix: #2051

**Commit1:**
Cache the last last SPDM CSR request data;
Add `request` and `request_size` as `libspdm_gen_csr` input;

**The old parameters `requester_info `and `opaque_data` are also saved as input. Because the `spdm_get_csr_request` structure 
should not be known in `os_stub`.  The `requester_info `and `opaque_data` are only got in `library`.**

**Commit2:**
clear last SPDM CSR request data in unit_test;

### DIFF
--- a/os_stub/spdm_device_secret_lib_sample/lib.c
+++ b/os_stub/spdm_device_secret_lib_sample/lib.c
@@ -992,23 +992,23 @@ bool libspdm_read_requester_public_key(uint16_t req_base_asym_alg,
 }
 
 #if LIBSPDM_ENABLE_CAPABILITY_GET_CSR_CAP
-bool libspdm_read_cached_requester_info(uint8_t **req_info, size_t *req_info_length)
+bool libspdm_read_cached_last_csr_request(uint8_t **last_csr_request, size_t *last_csr_request_len)
 {
     bool res;
     char *file;
 
-    file = "cached_req_info";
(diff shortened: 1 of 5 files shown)


os_stub/spdm_device_secret_lib_sample/lib.c
```diff
@@ -1063,28 +1064,29 @@ bool libspdm_gen_csr(uint32_t base_hash_algo, uint32_t base_asym_algo, bool *nee
...
-    size_t cached_req_info_length;
+    uint8_t *cached_last_csr_request;
+    size_t cached_last_request_len;
     uint8_t *cached_csr;
     size_t csr_buffer_size;
 
     csr_buffer_size = *csr_len;
 
     /*device gen csr need reset*/
     if (*need_reset) {
```
steven-bellock, line 1075: `*need_reset` is an output from `libspdm_gen_csr`. Why is it evaluating it as an input?

jyao1: agree. It should be output only.

jyao1: OK. Then we can merge this one.

Wenxing-hou: Yes. There is a unit_test to cover the check.

Issue #2065, opened 2023-05-28
Add `SPDM_REGISTRY_ID_VESA`


steven-bellock:
VESA was added as a registry body in the 1.3 specification, and will be backported to earlier specifications.

Issue #2064, opened 2023-05-26
Define `LIBSPDM_REGISTRY_ID_MAX`


steven-bellock:
`SPDM_REGISTRY_ID_JEDEC` is used to detect illegal IDs. For example

https://github.com/DMTF/libspdm/blob/772338edcf727d46ebf56451c6927187d97b5d52/library/spdm_common_lib/libspdm_com_opaque_data.c#L138

Instead there should be a `LIBSPDM_REGISTRY_ID_MAX` value that can be used for this purpose, so that when a new ID is added to the specification we don't have to update the macro in multiple places.

PR #2067, opened 2023-05-29
Add more SPDM_REGISTRY_ID_

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2064
Fix https://github.com/DMTF/libspdm/issues/2065


include/industry_standard/spdm.h
```diff
@@ -365,6 +365,9 @@ typedef struct {
 #define SPDM_REGISTRY_ID_MIPI 6
 #define SPDM_REGISTRY_ID_CXL 7
 #define SPDM_REGISTRY_ID_JEDEC 8
+#define SPDM_REGISTRY_ID_VESA 9
+#define SPDM_REGISTRY_ID_IANA_CBOR 0xa
+#define SPDM_REGISTRY_ID_MAX  0xa
```
steven-bellock, line 370: This should be defined in `libspdm_common_lib.h` as `LIBSPDM_REGISTRY_ID_MAX`.

steven-bellock (changes requested):

include/industry_standard/spdm.h
```diff
@@ -365,6 +365,9 @@ typedef struct {
 #define SPDM_REGISTRY_ID_MIPI 6
 #define SPDM_REGISTRY_ID_CXL 7
 #define SPDM_REGISTRY_ID_JEDEC 8
+#define SPDM_REGISTRY_ID_VESA 9
+#define SPDM_REGISTRY_ID_IANA_CBOR 0xa
```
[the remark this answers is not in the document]
jyao1: this is to align with SPDM spec.

include/industry_standard/spdm.h
jyao1: Why? I think it should be in SPDM definition. Irrelevant to SPDM implementation.

steven-bellock: `spdm.h` should only reflect what's in the specification.

jyao1: Right. That is why we have below:
```
#define SPDM_MAX_VERSION_COUNT 3
#define SPDM_MAX_SLOT_COUNT 8
#define SPDM_MAX_OPAQUE_DATA_SIZE 1024
```
There is nothing wrong to add `SPDM_REGISTRY_ID_MAX` IMHO, because that is what the spec has defined.

steven-bellock: Ok.

steven-bellock (approved):

PR #2070, opened 2023-05-29
Prevent undefined behavior

### LINKED ISSUE #2068: Responder can Invoke Undefined Behavior in libspdm Requester
Following a successful CAPABILITIES response a libspdm Requester stores the Responder's CTExponent into its context without validation. If the Requester sends a request message that requires a cryptography operation by the Responder, such as CHALLENGE, libspdm will calculate the timeout value using the Responder's unvalidated CTExponent.

```
((uint64_t)2 << context->connection_info.capability.ct_exponent)
```

If the value of CTExponent is greater than or equal to 64 then C language undefined behavior is invoked.
Note that libspdm provides a method to query the value of the CTExponent after VCA.

This also happens with RDTExponent. Note also that shifting the value `2` is incorrect. It should be `1`.

steven-bellock:
Fix #2068 in 2.3 branch.

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -14,6 +14,8 @@
 #include "library/spdm_device_secret_lib.h"
 
 #define INVALID_SESSION_ID 0
+#define LIBSPDM_MAX_CT_EXPONENT 31
+#define LIBSPDM_MAX_RDT_EXPONENT 31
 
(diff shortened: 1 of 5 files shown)


steven-bellock:
CI/CD is in https://github.com/steven-bellock/libspdm/actions/runs/5114484444. For some reason it's not running automatically.

Issue #2073, opened 2023-05-30
Bug: the leaf_cert_public_key in bundle_requester.certchain.der and bundle_requester.certchain1.der are ddifferent


Wenxing-hou:
When in commit https://github.com/DMTF/libspdm/commit/c4bd8d17fc6dd0956cfe7297a8eb4e18e6f427bf, the end_key in ecp384 and ecp521 are changed.
Then the `leaf_cert_public_key` will be different for `bundle_requester.certchain.der` and `bundle_requester.certchain1.der`, which will cause the spdm-emu verify failed.

Wenxing-hou:
I will check the issue.

PR #1956, opened 2023-04-19
X.509 certificate spdm defined oids check

xiangfeima:
Fix: #1594
Fix: #1949

1. Conditionally check SPDM defined extended key usage in X.509 certificate according to SPDM 1.2 specification section 10.8.2.1.
2. Fix SPDM defined extension check memory comparison issue.
3. Correct SPDM defined extension check function name.
4. Enable SPDM defined extension check unit test case.

Signed-off-by: Xiangfei Ma <xiangfei.ma@samsung.com>

PR #2074, opened 2023-05-30
Make slot_1 certchain has the same public key with slot_0 

Wenxing-hou:
Fix: #2073


jyao1:
@xiangfeima would you please take a look?

xiangfeima:
Sure, I'm looking into it now.

xiangfeima:
In the original ECP384 and ECP521 certificates for testing, the subject name all wrongly set to "ECP256", such as "DMTF libspdm ECP256 CA" or " DMTF libspdm ECP256 intermediate cert", "DMTF libspdm ECP256 requseter cert".
In https://github.com/DMTF/libspdm/commit/c4bd8d17fc6dd0956cfe7297a8eb4e18e6f427bf those certificatse were updated to proper subject name according to its algorithm and also new keys were generated.
However `bundle_requester.certchain.der` for each algorithm was updated but `bundle_requester.certchain1.der` was left without updating.
I'm very sorry for the missing.
Anyway I think we just need to update `bundle_requester.certchain1.der` instead of a whole restore.

Reference: https://github.com/DMTF/libspdm/pull/1956#discussion_r1172115513

Wenxing-hou:
> In the original ECP384 and ECP521 certificates for testing, the subject name all wrongly set to "ECP256", such as "DMTF libspdm ECP256 CA", " DMTF libspdm ECP256 intermediate cert", "DMTF libspdm ECP256 requseter cert" (ECP384 or ECP521 it should be).
> In [c4bd8d1](https://github.com/DMTF/libspdm/commit/c4bd8d17fc6dd0956cfe7297a8eb4e18e6f427bf) those certificatse were updated to proper subject name according to its algorithm and also new keys were generated.
> However `bundle_requester.certchain.der` for each algorithm was updated but `bundle_requester.certchain1.der` was left without updating.
> I'm very sorry for the missing.
> Anyway I think we just need to update `bundle_requester.certchain1.der` instead of a whole restore.

Thanks. I will update the `bundle_requester.certchain1.der` for ecp384 and ecp521.

xiangfeima:
@Wenxing-hou Thanks a lot. Would you help to take a look all the test certificates/keys for slot 1? I'm not sure if `bundle_requester.certchain1.der` is the only one need to be updated.

Issue #2072, opened 2023-05-30
Bug:  input should not include signature size when measurement generate signature


Wenxing-hou:
`libspdm_append_message_m` should not include the signature size.
https://github.com/DMTF/libspdm/blob/60e1a4ab19e5f1060f0caf2e55d98c2311ab8e77/library/spdm_responder_lib/libspdm_rsp_measurements.c#L369-L381

`status = libspdm_append_message_m(spdm_context, session_info, spdm_response, *response_size); ` should be 
`status = libspdm_append_message_m(spdm_context, session_info, spdm_response, *response_size - signature_size); `

Wenxing-hou:
I will fix the issue.

jyao1:
It seems regression caused by https://github.com/DMTF/libspdm/commit/e25b49cf1acd0ee255c88eed9b0aea33ba8e088a.

steven-bellock:
At least in https://github.com/DMTF/libspdm/blob/e9a62040ba556f64ea2041f506f1e49766b48d14/unit_test/test_spdm_responder/measurements.c#L403-L407 it's not checking the value of the signature.

PR #2079, opened 2023-06-01
Fix measurement signature generate input

Wenxing-hou:
Fix: #2072


Wenxing-hou:
The PR need @steven-bellock  review. Thanks.

steven-bellock (approved):

PR #2095, opened 2023-06-06
Fix logic error when use libspdm_consttime_is_mem_equal

Wenxing-hou:
The logic is wrong when use `libspdm_consttime_is_mem_equal`.


steven-bellock (approved):
How did the test pass before this change?

Wenxing-hou:
> How did the test pass before this change?

The old test logic is wrong. When the old test call the `libspdm_find_buffer`, the memory is different at first. The `libspdm_find_buffer` return true directly instead of looking for the same value.

```
    for (index = 0; index < src_len - dst_len; index++) {
        if ((*(src + index) == *dst) &&
            (libspdm_consttime_is_mem_equal(src + index, dst, dst_len) == 0)) {
            return true;
        }
    }
```

Issue #2077, opened 2023-05-31
the Requester should allow issue `SET_CERTIFICATE` outside of a secure session for slots 1-7


Zhiqiang520:
the Requester should allow issue SET_CERTIFICATE outside of a secure session for slots 1-7.
1. The current code behavior is The Requester **shall** issue SET_CERTIFICATE **inside** a secure session for slot 1-7 provisioning.
https://github.com/DMTF/libspdm/blob/aa51e293809194e601af7db7c96f5e855d4d78cb/library/spdm_requester_lib/libspdm_req_set_certificate.c#L58-L62
2. Refer to the paragraph 675 in [DSP0274_1.2.1.pdf](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf).
![image](https://github.com/DMTF/libspdm/assets/90754803/5119751d-1c52-4f1b-ad29-02953fd5dbfe)
![image](https://github.com/DMTF/libspdm/assets/90754803/6270b96c-2a31-4d2a-b7fb-b4eec45c97ca)
3. Base on the above specification, `"should" ("recommended")`, I think:
For slots 1-7, if the provisioning happens in a trusted environment, the Requester should **allow** issue `SET_CERTIFICATE` **outside** of a secure session.

steven-bellock:
From 5-Jun-2023 meeting: we could add a bit to the `spdm_context` or require a function (in secretlib) that says whether the Responder is in a trusted environment or not.

PR #2102, opened 2023-06-07
Fix Set cert slot number check.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2077

### DIFF
--- a/include/hal/library/responder_secretlib.h
+++ b/include/hal/library/responder_secretlib.h
@@ -274,4 +274,14 @@ extern bool libspdm_gen_csr(uint32_t base_hash_algo, uint32_t base_asym_algo, bo
                             bool is_device_cert_model);
 #endif /* LIBSPDM_ENABLE_CAPABILITY_GET_CSR_CAP */
 
+#if LIBSPDM_ENABLE_CAPABILITY_SET_CERT_CAP
+/**
+ * return if current code is running in a trusted environment.
+ *
+ * @retval  true   It is in a trusted environment.
+ * @retval  false  It is not in a trusted environment.
+ **/
+extern bool libspdm_is_in_trusted_environment();
+#endif /* LIBSPDM_ENABLE_CAPABILITY_SET_CERT_CAP */
+
 #endif /* RESPONDER_SECRETLIB_H */
--- a/include/hal/library/responder_secretlib.h
+++ b/include/hal/library/responder_secretlib.h
@@ -282,6 +282,21 @@ extern bool libspdm_gen_csr(uint32_t base_hash_algo, uint32_t base_asym_algo, bo
  * @retval  false  It is not in a trusted environment.
  **/
 extern bool libspdm_is_in_trusted_environment();
+
+/**
+ * Stores a certificate chain in non-volatile memory.
+ *
+ *
+ * @param[in]  slot_id          The number of slot for the certificate chain.
+ * @param[in]  cert_chain       The pointer for the certificate chain to set.
+ * @param[in]  cert_chain_size  The size of the certificate chain to set.
+ *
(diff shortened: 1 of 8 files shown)


include/hal/library/responder_secretlib.h
```diff
@@ -270,4 +270,29 @@ extern bool libspdm_gen_csr(uint32_t base_hash_algo, uint32_t base_asym_algo, bo
...
+ *
+ *
+ * @param[in]  slot_id          The number of slot for the certificate chain.
+ * @param[in]  cert_chain       The pointer for the certificate chain to set.
+ * @param[in]  cert_chain_size  The size of the certificate chain to set.
+ *
+ * @retval true   The certificate chain was successfully written to non-volatile memory.
+ * @retval false  Unable to write certificate chain to non-volatile memory.
+ **/
+extern bool libspdm_write_certificate_to_nvm(uint8_t slot_id, const void * cert_chain,
```
steven-bellock, line 293: Why move this to secretlib? Certificates aren't secret.

jyao1: All certificate are managed by secretlib.

platformlib deal with watchdog / timer, etc.

jyao1: Writing flash requires integrity guarantee. Generic platformlib is not a good place.

At least it is handled by [spdm_device_secret_lib_sample](https://github.com/DMTF/libspdm/tree/main/os_stub/spdm_device_secret_lib_sample) today, because all certificate management is done by secretlib.

steven-bellock: @xiaoyuruan can you be the tie-breaker? Should the code that writes the certificate chain to FLASH live in the secure enclave or in the platform library with timers and watchdogs?

xiaoyuruan: According to the current definition of spdm_device_secret_lib

[spdm_device_secret_lib](https://github.com/DMTF/libspdm/blob/main/include/library/spdm_device_secret_lib.h)

This library handles the private key signing, PSK HMAC operation, and measurement collection.

I agree that "certificate write to FLASH" does not fall under this definition.

os_stub/spdm_device_secret_lib_sample/lib.c
```diff
@@ -2015,6 +2015,11 @@ bool libspdm_psk_master_secret_hkdf_expand(
 #endif /* LIBSPDM_ENABLE_CAPABILITY_PSK_EX_CAP */
 
 #if LIBSPDM_ENABLE_CAPABILITY_SET_CERT_CAP
+bool libspdm_is_in_trusted_environment()
+{
+    return false;
```
steven-bellock, line 2020: Should make this settable from `unit_test` so that the `true` path can be tested as well.

include/hal/library/responder_secretlib.h
jyao1: [platform_lib](https://github.com/DMTF/libspdm/blob/main/include/hal/library/platform_lib.h) provides sleep function and watchdog function, 

I don't think it belongs to platform_lib.

We can have below options:
1) redefine secret lib
2) redefine platform lib
3) define a new lib

steven-bellock: We can have a separate issue to sort out the libraries and headers for the 3.0 release.

jyao1: agree. I filed https://github.com/DMTF/libspdm/issues/2124

os_stub/spdm_device_secret_lib_sample/lib.c
jyao1: Agree. Code is updated and unit tests are added.

PR #2105, opened 2023-06-08
Supply slot_id to libspdm_*_data_sign()

alistair23:
Supply the slot_id to libspdm_responder_data_sign() and libspdm_requester_data_sign(). These functions use certificates to sign an SPDM message so they need to know what slot is being used to sign the data.

For CHALLENGE_AUTH for example the slot_id is sent as part of Param1 and section 365 of spec version 1.2.1 specifies that that slot leaf certificate should be used for signing. Currently there is no way to know which leaf certificate should be used when generating signatures, this patch ensures that we know what slot certificate should be used.

### DIFF
--- a/include/hal/library/requester_secretlib.h
+++ b/include/hal/library/requester_secretlib.h
@@ -16,6 +16,7 @@
  *
  * @param  req_base_asym_alg Indicates the signing algorithm.
  * @param  base_hash_algo    Indicates the hash algorithm.
+ * @param  slot_id           The number of slot for the certificate chain.
  * @param  is_data_hash      Indicates the message type.
  *                           If true, raw message before hash.
  *                           If false, message hash.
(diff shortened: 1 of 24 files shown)


jyao1:
NOTE: in SPDM 1.0/1.1/1.2, all slots shares the same private key.
But for SPDM 1.3, we will support multiple keys. That feature may help for SPDM 1.3+.

steven-bellock:
Yeah this isn't needed for SPDM <= 1.2. When we get to SPDM 1.3 multi-key, I imagine libspdm will have a slot ID to `KeyPairID` mapping and `KeyPairID` will be used to specify keys instead of slot IDs.

alistair23:
I'm not sure I follow. The 1.2.1 version of the spec says:

```
PrivKey shall be the private key associated with the leaf certificate of the Responder in
slot=Param1 of the CHALLENGE request message. If the public key of the Responder was
provisioned to the Requester, then PrivKey shall be the associated private key.
```

Where each slot contains `one complete certificate chain`. It seems like each slot can have a different private key, I don't see anything saying they all have the same private key

jyao1:
See: SPDM 1.2.1 - 10.6.1 Certificates and certificate chains:
```
The SPDM endpoint shall contain a single public-private key pair per supported algorithm for its leaf certificate,
regardless of how many certificate chains are stored on the device. The Responder selects a single asymmetric key
signature algorithm per Requester.
```

alistair23:
Ah! Your right, thanks for that

PR #2120, opened 2023-06-13
CMakeLists: Allow disabling tests

### LINKED ISSUE #1560: Testing should be opt in
Now that there appears to be a class of Integrators who will be using libspdm as a real library and who don't care about running its unit and fuzz testing (for example https://github.com/altera-opensource/verifier), we should make testing opt in when CMake is invoked; ie `-DTEST=On`.

alistair23:
Fix #1560.

Add support for disabling the tests. If building libspdm as a library (for example in buildroot) we often don't want to build the tests, so add the option to disable the tests.


jyao1:
Would you please also update readme to show how to use this feature?

alistair23:
Done

doc/build.md
```diff
@@ -248,3 +248,11 @@ a) [NIOS2 GNU](https://www.intel.com/content/www/us/en/docs/programmable/683689/
    cmake -DARCH=riscv32 -DTOOLCHAIN=RISCV_XPACK -DTARGET=Release -DCRYPTO=mbedtls ..
    ```
    Note: `make -j` can be used to accelerate the build.
+
+### Disabling the unit tests
+
+If you don't want to build or run the unit tests they can be disabled by adding the following to the cmake command
```
steven-bellock, line 254: "Unit tests can be disable by adding `-DDISABLE_TESTS=1` to CMake."

steven-bellock (changes requested):
This needs to be added to either https://github.com/DMTF/libspdm/blob/main/.github/workflows/build.yml or maybe create another workflow that just builds libspdm without tests.

alistair23:
I have added tests to the GitHub actions

jyao1:
@steven-bellock , we should not do "Merge branch"

steven-bellock:
> @steven-bellock , we should not do "Merge branch"

Yeah that was Github. I'll revert.

alistair23:
I have rebased the PR

steven-bellock (approved):

Issue #2119, opened 2023-06-12
Misleading error code


steven-bellock:
For example
```
    status = libspdm_append_message_k(spdm_context, session_info, true, spdm_request,
                                      spdm_request_size);
    if (LIBSPDM_STATUS_IS_ERROR(status)) {
        libspdm_free_session_id(spdm_context, *session_id);
        libspdm_secured_message_dhe_free(
            spdm_context->connection_info.algorithm.dhe_named_group, dhe_context);
        status = LIBSPDM_STATUS_BUFFER_FULL;
        goto receive_done;
    }
```
When `LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT==1` then `LIBSPDM_STATUS_BUFFER_FULL` is the correct code. But when `LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT==0` then `libspdm_append_message_*` fails due to a cryptography error and should be `LIBSPDM_STATUS_CRYPTO_ERROR`.

jyao1:
It can just propagate the error code.

PR #2125, opened 2023-06-13
propagate error code from libspdm_append_message_x.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2119


steven-bellock (approved):
Since we don't have crypto error injection the unit tests don't need to be updated.

Issue #2114, opened 2023-06-10
An error response with any error code other than `ResponseNotReady` shall re-initialize L1/L2 to null.


Zhiqiang520:
An error response with **any** error code **other than** `ResponseNotReady` shall re-initialize L1/L2 to null.

1. Refer to the paragraph 432 in [DSP0274_1.2.1.pdf](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf).
![image](https://github.com/DMTF/libspdm/assets/90754803/e432e178-abb7-41c9-9f6c-67b20d9790f9)
2. So I think in `libspdm_get_response_measurements` function, should call `libspdm_reset_message_m(spdm_context, session_info);` before return **any** `libspdm_generate_error_response` **other than** `ResponseNotReady` error.
3. the Requester issues **a number of requests** without requiring signatures followed by a final request requiring a signature over the entire set of request-response messages exchanged.
> GET_MEASUREMENTS_REQUEST1
MEASUREMENTS_RESPONSE1
GET_MEASUREMENTS_REQUEST2
MEASUREMENTS_RESPONSE2
... ...
GET_MEASUREMENTS_REQUESTn
MEASUREMENTS_RESPONSEn

4. In the above scenario, if `MEASUREMENTS_RESPONSE1` successful, but `MEASUREMENTS_RESPONSE2` **error** at the following line 364 or 372, then should call `libspdm_reset_message_m(spdm_context, session_info);`  before return  `libspdm_generate_error_response`.
https://github.com/DMTF/libspdm/blob/83cbb2fd450e67a60e7e549069b595905099e637/library/spdm_responder_lib/libspdm_rsp_measurements.c#L362-L375
5. there are many same issues is in `libspdm_rsp_measurements.c`, for example,
https://github.com/DMTF/libspdm/blob/83cbb2fd450e67a60e7e549069b595905099e637/library/spdm_responder_lib/libspdm_rsp_measurements.c#L95-L107
6. there are also many similar issues in `libspdm_req_get_measurements.c`, for example, the line 414~415,
https://github.com/DMTF/libspdm/blob/83cbb2fd450e67a60e7e549069b595905099e637/library/spdm_requester_lib/libspdm_req_get_measurements.c#L411-L424

jyao1:
Please help check if M1/M2 has similar issue.

Zhiqiang520:
> Please help check if M1/M2 has similar issue.

[...]

PR #2126, opened 2023-06-13
reset L1/L2 in case of error

jyao1:
fix: https://github.com/DMTF/libspdm/issues/2114

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_get_measurements.c
+++ b/library/spdm_requester_lib/libspdm_req_get_measurements.c
@@ -307,7 +307,6 @@ static libspdm_return_t libspdm_try_get_measurement(libspdm_context_t *spdm_cont
             goto receive_done;
         }
     } else if (spdm_response->header.request_response_code != SPDM_MEASUREMENTS) {
-        libspdm_reset_message_m(spdm_context, session_info);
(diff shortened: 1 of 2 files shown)


library/spdm_requester_lib/libspdm_req_get_measurements.c
```diff
@@ -609,6 +601,11 @@ static libspdm_return_t libspdm_try_get_measurement(libspdm_context_t *spdm_cont
     #endif /* LIBSPDM_ENABLE_MSG_LOG */
 
 receive_done:
+    if ((status != LIBSPDM_STATUS_SUCCESS) &&
+        (status != LIBSPDM_STATUS_NOT_READY_PEER) &&
+        (status != LIBSPDM_STATUS_BUSY_PEER)) {
```
steven-bellock, line 606: `LIBSPDM_STATUS_BUSY_PEER` needs to be clarified by the SPDM Working Group. Right now the specification says that only `LIBSPDM_STATUS_NOT_READY_PEER` is applicable here. See https://github.com/DMTF/SPDM-WG/issues/2917.

jyao1: OK. I dropped BUSY.

Issue #2128, opened 2023-06-13
`GET_VERSION` issued to set M1/M2 to `null` also for basic mutual authentication.


Zhiqiang520:
I think `GET_VERSION` issued to set M1/M2 to `null` also for basic mutual authentication.
1. Refer to the paragraph 382 `Table 38 — Request ordering and message transcript computation rules for M1/M2` in [DSP0274_1.2.1.pdf](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf).
<img width="545" alt="image" src="https://github.com/DMTF/libspdm/assets/90754803/96708e89-5a15-41f2-8306-5ca8482f086d">

2.  So I think maybe the following two `libspdm_reset_message_mut_*` functions should be added after the line 58.

> libspdm_reset_message_mut_b(spdm_context);
libspdm_reset_message_mut_c(spdm_context);

https://github.com/DMTF/libspdm/blob/04baaa92bc59e03aa528988edd46bfc15ce01e2b/library/spdm_requester_lib/libspdm_req_get_version.c#L56-L59
3.  So I think maybe the following two `libspdm_reset_message_mut_*` functions should be added after the line 68.

> libspdm_reset_message_mut_b(spdm_context);
libspdm_reset_message_mut_c(spdm_context);

https://github.com/DMTF/libspdm/blob/04baaa92bc59e03aa528988edd46bfc15ce01e2b/library/spdm_responder_lib/libspdm_rsp_version.c#L66-L69

PR #2129, opened 2023-06-13
GET_VERSION issued to set M1/M2 to null also for basic mutual authent…

Zhiqiang520:
…ication.

Fix: #2128


jyao1:
See my comment in the issue.

Zhiqiang520:
Close it due to #2128 not a real issue.

Issue #2101, opened 2023-06-07
consider adding policy to maximum sequence number


jyao1:
Current code uses hardcoded maximum sequence number: 2^64.
```
    if (sequence_number == (uint64_t)-1) {
        return LIBSPDM_STATUS_SEQUENCE_NUMBER_OVERFLOW;
    }
```

However, there are multiple papers discussing smaller boundary. For example, to resist 2^−60 Attack Success Probability, we can only send 2^24.5 messages, with AES-GCM-256.

We may consider letting integrator to determine and configure the maximum sequence number, such as 2^32 or 2^24, instead of hardcoded to 2^64.

Reference:
[1] https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-aead-limits
[2] https://www.isg.rhul.ac.uk/~kp/TLS-AEbounds.pdf

PR #2130, opened 2023-06-13
Add LIBSPDM_MAX_SPDM_SESSION_SEQUENCE_NUMBER

jyao1:
Fix: https://github.com/DMTF/libspdm/issues/2101

Tested in spdm-emu. Setting to small number will cause emu stop communicating.

### DIFF
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -127,6 +127,19 @@ typedef enum {
     LIBSPDM_DATA_MAX_DHE_SESSION_COUNT,
     LIBSPDM_DATA_MAX_PSK_SESSION_COUNT,
 
+    /* DSP0277 defines 64bit sequence number.
+     * The default value is max number 0xFFFFFFFFFFFFFFFFull (64bit).
+     * 0 means the default value.
+     *
+     * https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-aead-limits describes
+     * how to limit the use of keys in order to bound the advantage given to an attacker.
+     *
+     * The integrator can override the default value, such as 0xFFFFFFFF (32bit) or 0xFFFFFF (24bit).
+     * If KEY_UPDATE is not sent before the max sequence number is reached,
+     * the SPDM session will be terminated.
+     */
+    LIBSPDM_DATA_MAX_SPDM_SESSION_SEQUENCE_NUMBER,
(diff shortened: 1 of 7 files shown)


include/library/spdm_lib_config.h
```diff
@@ -12,6 +12,18 @@
...
 
+/* DSP0277 defines 64bit sequence number. The max number is 0xFFFFFFFFFFFFFFFFull.
+ * https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-aead-limits describes
+ * how to limit the use of keys in order to bound the advantage given to an attacker.
+ *
+ * The integritor can override the default value, such as 0xFFFFFFFF (32bit) or 0xFFFFFF (24bit).
+ * If KEY_UPDATE is not sent before the LIBSPDM_MAX_SPDM_SESSION_SEQUENCE_NUMBER is reached,
+ * the SPDM session will be terminated.
+ */
+#ifndef LIBSPDM_MAX_SPDM_SESSION_SEQUENCE_NUMBER
```
steven-bellock, line 23: This should be settable at runtime. The default value should be `UINT64_MAX` but Integrator can override it at runtime.

jyao1: agree. updated.

include/library/spdm_common_lib.h
```diff
@@ -127,6 +127,19 @@ typedef enum {
...
     LIBSPDM_DATA_MAX_PSK_SESSION_COUNT,
 
+    /* DSP0277 defines 64bit sequence number.
+     * The default value is max number 0xFFFFFFFFFFFFFFFFull (64bit).
+     * 0 means the default value.
+     *
+     * https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-aead-limits describes
+     * how to limit the use of keys in order to bound the advantage given to an attacker.
+     *
+     * The integritor can override the default value, such as 0xFFFFFFFF (32bit) or 0xFFFFFF (24bit).
```
steven-bellock, line 137: `integritor` -> "Integrator`

steven-bellock:
@jyao1 feel free to commit after the typo is fixed.

include/library/spdm_common_lib.h
jyao1: thanks. fixed.

Issue #2082, opened 2023-06-02
should we keep sequence number (not increase) in case of decryption error


jyao1:
In current code, we do
* increase sequence number in local storage
* decrypt message
* if fail, return error

It means that the sequence number is still increased, in case of failure. That follows the DSP0277.
```
For each record, both SPDM endpoints shall follow these steps as prescribed:
1. Zero extend the sequence number to iv_length according to the selected AEAD cipher suite in
ALGORITHMS messages.
2. Perform a bitwise XOR of the zero-extended sequence number with the appropriate IV derived in the
SPDM key schedule. The sequence number shall be padded with zeros to the left (most significant) to the
size of the IV.
◦ The output of this step is called the per-message nonce.
3. Increment the sequence number by a value of one for the next Secured Message
```

Is that expected way?
Or should we keep sequence number as original in case of failure ?

steven-bellock:
From 5-Jun-2023 meeting: should probably terminate the session. If someone needs more flexibility to retry DSP0277 messages then we can revisit it.

Issue #2081, opened 2023-06-02
need remove session in case of LIBSPDM_STATUS_SEQUENCE_NUMBER_OVERFLOW


jyao1:
Currently, there is code to detect session overflow.

In `libspdm_decode_secured_message`, it sets last error.
```
    if (sequence_number == (uint64_t)-1) {
        libspdm_secured_message_set_last_spdm_error_struct(
            spdm_secured_message_context, &spdm_error);
        return LIBSPDM_STATUS_SEQUENCE_NUMBER_OVERFLOW;
    }
```
In `libspdm_encode_secured_message`, it does not.
```
    if (sequence_number == (uint64_t)-1) {
        return LIBSPDM_STATUS_SEQUENCE_NUMBER_OVERFLOW;
    }
```
I think the behavior should be consistent.

In requester - `libspdm_receive_data`, it will end session.
```
    if (spdm_response->header.request_response_code == SPDM_ERROR) {
        if ((spdm_response->header.param1 == SPDM_ERROR_CODE_DECRYPT_ERROR) &&
            (session_id != NULL)) {
            libspdm_free_session_id(context, *session_id);
            libspdm_release_receiver_buffer (context);
            return LIBSPDM_STATUS_SESSION_MSG_ERROR;
        }
    }
```
In responder - `libspdm_build_response`, it does not.
```
        switch (context->last_spdm_error.error_code) {
        case SPDM_ERROR_CODE_DECRYPT_ERROR:
            /* session ID is valid. Use it to encrypt the error message.*/
            if((context->handle_error_return_policy &
                LIBSPDM_DATA_HANDLE_ERROR_RETURN_POLICY_DROP_ON_DECRYPT_ERROR) == 0) {
                status = libspdm_generate_error_response(
                    context, SPDM_ERROR_CODE_DECRYPT_ERROR, 0,
                    &my_response_size, my_response);
```
I think the behavior should be consistent.

steven-bellock:
Since the counter is 64 bits this will never actually happen.

jyao1:
if we add such check, that means we assume it might happen. As such we need do the clean up.

PR #2131, opened 2023-06-13
Free session_id case of error in session error

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2081
Fix https://github.com/DMTF/libspdm/issues/2082

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_send_receive.c
+++ b/library/spdm_requester_lib/libspdm_req_send_receive.c
@@ -104,7 +104,6 @@ libspdm_return_t libspdm_send_request(void *spdm_context, const uint32_t *sessio
                        status));
         if ((session_id != NULL) &&
             ((status == LIBSPDM_STATUS_SEQUENCE_NUMBER_OVERFLOW) ||
-             (status == LIBSPDM_STATUS_LOW_ENTROPY) ||
              (status == LIBSPDM_STATUS_CRYPTO_ERROR))) {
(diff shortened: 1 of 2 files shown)


library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -102,6 +102,12 @@ libspdm_return_t libspdm_send_request(void *spdm_context, const uint32_t *sessio
     if (LIBSPDM_STATUS_IS_ERROR(status)) {
         LIBSPDM_DEBUG((LIBSPDM_DEBUG_INFO, "transport_encode_message status - %p\n",
                        status));
+        if ((session_id != NULL) &&
+            ((status == LIBSPDM_STATUS_SEQUENCE_NUMBER_OVERFLOW) ||
+             (status == LIBSPDM_STATUS_LOW_ENTROPY) ||
```
steven-bellock, line 107: `LIBSPDM_STATUS_LOW_ENTROPY` and `LIBSPDM_STATUS_CRYPTO_ERROR` aren't fatal to the session. The Requester could try again, maybe after more entropy has been collected by the system.

jyao1: I dropped LIBSPDM_STATUS_LOW_ENTROPY check.

For crypto error, we have discussed in https://github.com/DMTF/libspdm/issues/2082. We should terminate the session.

steven-bellock (approved):

Issue #2124, opened 2023-06-13
need split secretlib to multiple smaller lib


jyao1:
See the discussion at https://github.com/DMTF/libspdm/pull/2102#discussion_r1225021952

PR #2134, opened 2023-06-14
reorg the secretlib.

jyao1:
fix: https://github.com/DMTF/libspdm/issues/2124

change secretlib to asymsignlib, csrlib, measlib, psklib, setcertlib.
change platformlib to timerlib, watchdoglib.

Some open for discussion:
1) Do we need `requester_` or `responder_` prefix?
2) Do we need to distinguish requester_psk function from responder_psk function? Currently they are same.


steven-bellock:
I'll have to think about this one. Let's discuss it in the next meeting.

steven-bellock:
>Do we need requester_ or responder_ prefix?

We can have 
>include/hal/library/requester/
include/hal/library/responder/

and get rid of the prefix. Then the Integrator only needs to look at the appropriate directory.

include/hal/library/requester/reqasymsignlib.h
```diff
@@ -4,8 +4,8 @@
  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
  **/
 
-#ifndef REQUESTER_SECRETLIB_H
-#define REQUESTER_SECRETLIB_H
+#ifndef REQUESTER_REQASYMSIGNLIB_H
```
steven-bellock, line 7: Filename can be renamed to `asymsignlib.h`.

jyao1: This is requester. The SPDM defines the algo to be req_asym.
We had better align with SPDM spec.

steven-bellock: Ok.

Issue #2135, opened 2023-06-14
The timer for the Heartbeat period shall start at transmission of `PSK_EXCHANGE_RSP` when `ResponderContext` is absent 


Zhiqiang520:
The timer for the Heartbeat period shall start at the transmission of `PSK_EXCHANGE_RSP` response messages, when `ResponderContext` is **absent** in `PSK_EXCHANGE_RSP` response message.

1. Refer to paragraph 564 in [DSP0274_1.2.1](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf).
![image](https://github.com/DMTF/libspdm/assets/90754803/3f122a2c-a3d0-4ed1-8237-67cc23b9c2ba)
2. But for the special **case** `ResponderContext` is **absent** in `PSK_EXCHANGE_RSP` response message, the Requester **shall not** send `PSK_FINISH`.
I think: 
The timer for the Heartbeat period shall start at the transmission of `PSK_EXCHANGE_RSP` response messages.
Refer to paragraph 549 in [DSP0274_1.2.1](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf).
![image](https://github.com/DMTF/libspdm/assets/90754803/7518bc9e-4310-4999-87b9-4d4c915a108a)
3. I think `libspdm_start_watchdog` function should be called in the code block between the following line 409 ~ 432 in `libspdm_rsp_psk_exchange.c`, just like the line 230 ~ 243 in `libspdm_rsp_psk_finish.c`.
https://github.com/DMTF/libspdm/blob/2b8bfc06c775308a7c0fe06bc552496191d167da/library/spdm_responder_lib/libspdm_rsp_psk_exchange.c#L407-L432
https://github.com/DMTF/libspdm/blob/2b8bfc06c775308a7c0fe06bc552496191d167da/library/spdm_responder_lib/libspdm_rsp_psk_finish.c#L230-L243
4. It is better to add the description for this special case in the paragraph 564 in [DSP0274_1.2.1](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf).

steven-bellock:
Oh good catch. I'll file an issue against the specification.

PR #2136, opened 2023-06-14
start watchdog in case of PSK without context.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2135

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_psk_exchange.c
+++ b/library/spdm_responder_lib/libspdm_rsp_psk_exchange.c
@@ -428,6 +428,22 @@ libspdm_return_t libspdm_get_response_psk_exchange(libspdm_context_t *spdm_conte
                 0, response_size, response);
         }
 
+        #if LIBSPDM_ENABLE_CAPABILITY_HBEAT_CAP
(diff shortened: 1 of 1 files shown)


library/spdm_responder_lib/libspdm_rsp_psk_exchange.c
```diff
@@ -428,6 +428,22 @@ libspdm_return_t libspdm_get_response_psk_exchange(libspdm_context_t *spdm_conte
                 0, response_size, response);
         }
 
+        #if LIBSPDM_ENABLE_CAPABILITY_HBEAT_CAP
+        if (libspdm_is_capabilities_flag_supported(
```
steven-bellock, line 432: This is only happens if Responder's `PSK_CAP==01`.

steven-bellock (changes requested):

library/spdm_responder_lib/libspdm_rsp_psk_exchange.c
steven-bellock: Ah ok.

steven-bellock (approved):

Issue #2062, opened 2023-05-25
The description for `require a reset to complete the SET_CERTIFICATE request` in the specification is not clear


Zhiqiang520:
The description for `require a reset to complete the SET_CERTIFICATE request` in the specification is not clear.

It doesn't mention --- **After** the Responder has processed the reset:
if the Requester sends the **same** `SET_CERTIFICATE` that was sent before the reset, what to do then?
if the Requester sends the **different** `SET_CERTIFICATE` request than the request that was sent before the reset, what to do then?

1. Refer to paragraph 675 in [DSP0274_1.2.1](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf).
`The device might require a reset to complete the SET_CERTIFICATE request, potentially so that the
device can generate AliasCert certificates using lower firmware layers. If the device requires a reset to complete
the SET_CERTIFICATE request, then the device shall respond with an ErrorCode=ResetRequired response.`

2. I think more description for `require a reset to complete the SET_CERTIFICATE request` should be added in the specification  [DSP0274_1.2.1](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf).
May be just like the paragraph 666 in [DSP0274_1.2.1](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf) for `requires a reset to complete the GET_CSR request`.
`If the device requires a reset to complete the GET_CSR request, then the device shall respond with an
ErrorCode=ResetRequired response. After the Responder has processed the reset, the Requester sends the same
GET_CSR request that was sent before the reset, which signals to the Responder to send the CSR response. If the
Requester sends a different GET_CSR request than the request that was sent before the reset, then the Responder
shall treat it as a new request or reject the different GET_CSR request with an ErrorCode=UnexpectedRequest
response. On receipt of a new GET_CSR request, the Responder can discard any existing CSR responses.`

steven-bellock:
Unlike `GET_CSR`, the Requester doesn't have to do anything if a Responder requires a reset to complete `SET_CERTIFICATE`. If, after a reset, the Requester sends another `SET_CERTIFICATE` the Responder processes it, regardless if it's the same certificate or a different certificate.

jyao1:
> If, after a reset, the Requester sends another SET_CERTIFICATE the Responder processes it, regardless if it's the same certificate or a different certificate.

Is that in SPDM spec? I am surprised on that.

[...]

PR #2144, opened 2023-06-16
Add unit_test for set_cert to check overwrite same slot_id

Wenxing-hou:
Fix: #2062

### DIFF
--- a/os_stub/spdm_device_secret_lib_sample/lib.c
+++ b/os_stub/spdm_device_secret_lib_sample/lib.c
@@ -9,7 +9,7 @@
  * It follows the SPDM Specification.
  **/
 #include <base.h>
-#if defined(_MSC_VER) || (defined(__clang__) && (defined (LIBSPDM_CPU_AARCH64) || \
+#if defined(_WIN32) || (defined(__clang__) && (defined (LIBSPDM_CPU_AARCH64) || \
     defined(LIBSPDM_CPU_ARM)))
 #else
     #include <fcntl.h>
@@ -1590,18 +1590,18 @@ bool libspdm_is_in_trusted_environment()
 bool libspdm_write_certificate_to_nvm(uint8_t slot_id, const void * cert_chain,
                                       size_t cert_chain_size)
 {
-#if defined(_MSC_VER) || (defined(__clang__) && (defined (LIBSPDM_CPU_AARCH64) || \
+#if defined(_WIN32) || (defined(__clang__) && (defined (LIBSPDM_CPU_AARCH64) || \
     defined(LIBSPDM_CPU_ARM)))
     FILE *fp_out;
 #else
     uint64_t fp_out;
 #endif
 
-    char file_name[] = {'s','l','o','t','_','i','d','_','0','\0'};
+    char file_name[] = "slot_id_0_cert_chain.der";
+    /*change the file name, for example: slot_id_1_cert_chain.der*/
+    file_name[8] = (char)(slot_id+'0');
 
-    file_name[strlen(file_name) - 1] = (char)(slot_id+'0');
-
-#if defined(_MSC_VER) || (defined(__clang__) && (defined (LIBSPDM_CPU_AARCH64) || \
+#if defined(_WIN32) || (defined(__clang__) && (defined (LIBSPDM_CPU_AARCH64) || \
     defined(LIBSPDM_CPU_ARM)))
     if ((fp_out = fopen(file_name, "w+b")) == NULL) {
         printf("Unable to open file %s\n", file_name);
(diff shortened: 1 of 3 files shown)


os_stub/spdm_device_secret_lib_sample/lib.c
```diff
@@ -1597,7 +1597,7 @@ bool libspdm_write_certificate_to_nvm(uint8_t slot_id, const void * cert_chain,
     uint64_t fp_out;
 #endif
 
-    char file_name[] = {'s','l','o','t','_','i','d','_','0','\0'};
+    char file_name[] = "slot_id_0";
```
jyao1, line 1600: please change to "slot_id_0_cert_chain.der".

Wenxing-hou: Thanks. I have changed the file name.

unit_test/test_spdm_responder/set_certificate_rsp.c
```diff
@@ -27,6 +27,14 @@ void libspdm_test_responder_set_cetificate_rsp_case1(void **state)
     size_t cert_chain_size;
     spdm_set_certificate_request_t *m_libspdm_set_certificate_request;
 
+#if defined(_MSC_VER) || (defined(__clang__) && (defined (LIBSPDM_CPU_AARCH64) || \
```
jyao1, line 30: why we need this?

Wenxing-hou: Because of the PR: https://github.com/DMTF/libspdm/commit/00885640d5dec19fcaa0051a3e6add56ccf363ee, there are two ways(fwrite or write) to write NVM. 
And when use `libspdm_read_responder_public_certificate_chain_per_slot` get the cert, it use fread to get cert. 
So add the MACRO to check for the `fwrite` way.

```
bool libspdm_write_certificate_to_nvm(uint8_t slot_id, const void * cert_chain,
                                      size_t cert_chain_size)
{
#if defined(_MSC_VER) || (defined(__clang__) && (defined (LIBSPDM_CPU_AARCH64) || \
    defined(LIBSPDM_CPU_ARM)))
    FILE *fp_out;
#else
    uint64_t fp_out;
#endif

    char file_name[] = {'s','l','o','t','_','i','d','_','0','\0'};

    file_name[strlen(file_name) - 1] = (char)(slot_id+'0');

#if defined(_MSC_VER) || (defined(__clang__) && (defined (LIBSPDM_CPU_AARCH64) || \
    defined(LIBSPDM_CPU_ARM)))
    if ((fp_out = fopen(file_name, "w+b")) == NULL) {
        printf("Unable to open file %s\n", file_name);
        return false;
    }

    if ((fwrite(cert_chain, 1, cert_chain_size, fp_out)) != cert_chain_size) {
        printf("Write output file error %s\n", file_name);
        fclose(fp_out);
        return false;
    }

    fclose(fp_out);
#else
    if ((fp_out = open(file_name, O_WRONLY | O_CREAT, S_IRWXU)) == -1) {
        printf("Unable to open file %s\n", file_name);
        return false;
    }

    if ((write(fp_out, cert_chain, cert_chain_size)) != cert_chain_size)
    {
        printf("Write output file error %s\n", file_name);
        close(fp_out);
        return false;
    }

    close(fp_out);
#endif

    return true;
}
```

jyao1: I dont see `else` branch.

How does that work in GCC?

unit_test/test_spdm_responder/set_certificate_rsp.c
```diff
@@ -77,6 +302,71 @@ void libspdm_test_responder_set_cetificate_rsp_case1(void **state)
...
+    /*test overwirte same slot_id cert*/
+#if defined(_MSC_VER) || (defined(__clang__) && (defined (LIBSPDM_CPU_AARCH64) || \
+    defined(LIBSPDM_CPU_ARM)))
+    /*read a different cert_chain*/
+    libspdm_read_responder_public_certificate_chain_per_slot(1, m_libspdm_use_hash_algo,
+                                                             m_libspdm_use_asym_algo,
+                                                             &cert_chain_slot_1,
+                                                             &cert_chain_size, NULL, NULL);
+#else
+    libspdm_test_read_responder_public_certificate_chain_per_slot(1, m_libspdm_use_hash_algo,
```
jyao1, line 315: this is not acceptable. Please use one function.

unit_test/test_spdm_responder/set_certificate_rsp.c
Wenxing-hou: Thanks. I have changed the code. And the CI has passed.

Issue #2107, opened 2023-06-08
Build failure when `LIBSPDM_CERT_PARSE_SUPPORT` is `0`


steven-bellock:
```
[ 19%] Building C object library/spdm_common_lib/CMakeFiles/spdm_common_lib.dir/libspdm_com_context_data.c.obj
libspdm_com_context_data.c
C:\repo\steven-bellock\libspdm\library\spdm_common_lib\libspdm_com_context_data.c(455): error C2220: the following warning is treated as an error
C:\repo\steven-bellock\libspdm\library\spdm_common_lib\libspdm_com_context_data.c(455): warning C4013: 'libspdm_x509_get_cert_from_cert_chain' undefined; assuming extern returning int
```

steven-bellock:
@jyao1 What's the expectation when `LIBSPDM_CERT_PARSE_SUPPORT` is `0`? In particular nothing really works since both certificate chains and raw public keys need parsing. I guess PSK should work fine.

jyao1:
I don't think you can set this to 0, but leave other open. e.g. CERT_CAP or KEY_EX_CAP.

```
/* This can be set to 0 for the device which does not need X509 parser.*/
#ifndef LIBSPDM_CERT_PARSE_SUPPORT
#define LIBSPDM_CERT_PARSE_SUPPORT 1
#endif
```

In my case, the responder (without SET_CERT_CAP or CSR_CAP) sets this flag, because responder does not need parse X509 cert. 
PSK should work, I agree.

Wenxing-hou:
I have submitted a PR https://github.com/Wenxing-hou/libspdm/pull/10 in my branch to test the disabling `LIBSPDM_CERT_PARSE_SUPPORT` case.

steven-bellock:
Thanks. It was a bit too much for me.

steven-bellock:
We also need to add `LIBSPDM_CERT_PARSE_SUPPORT==0` to CI/CD.

steven-bellock:
This issue is still present in the code base. 
> implicit declaration of function 'libspdm_get_leaf_cert_public_key_from_cert_chain' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
    result = libspdm_get_leaf_cert_public_key_from_cert_chain(
    
@jyao1 even if `LIBSPDM_CERT_PARSE_SUPPORT` is `0`, the local endpoint still needs to be able to get the public key of the peer to do much of anything.

PR #2156, opened 2023-06-20
Fix CERT_PARSER build error.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2107


jyao1:

PR #2157, opened 2023-06-20
Fix build error on RESPOND_IF_READY_SUPPORT=0

jyao1:


jyao1:

Issue #2145, opened 2023-06-16
Handling of `CHUNK_SEND` is incomplete


Zhiqiang520:
Handling of `CHUNK_SEND` is incomplete.
1.  I think `spdm_context->connection_info.capability.data_transfer_size` should be change to `LIBSPDM_MIN(spdm_context->connection_info.capability.data_transfer_size, context->local_context.capability.sender_data_transfer_size)`
at the following line 384 and 388 in the current latest libspdm.
https://github.com/DMTF/libspdm/blob/6674aa87c5a3b6991e25e22fcd1bf44e2a111331/library/spdm_requester_lib/libspdm_req_send_receive.c#L384-L390
2.  I think `(uint32_t)request_size != spdm_context->local_context.capability.data_transfer_size` should be **removed**
at the following line 100 and 147 in the current latest libspdm, due to `request_size` may be equal to the requester's `sender_data_transfer_size` and not equal to `spdm_context->local_context.capability.data_transfer_size`.
https://github.com/DMTF/libspdm/blob/6674aa87c5a3b6991e25e22fcd1bf44e2a111331/library/spdm_responder_lib/libspdm_rsp_chunk_send_ack.c#L98-L105
https://github.com/DMTF/libspdm/blob/6674aa87c5a3b6991e25e22fcd1bf44e2a111331/library/spdm_responder_lib/libspdm_rsp_chunk_send_ack.c#L144-L150

Issue #2141, opened 2023-06-15
Responder's handling of `CHUNK_GET` is incomplete


steven-bellock:
A Responder may return a `LargeResponse` error if the size of the response is larger than the `DataTransferSize` of the Requester or the transmit buffer size of the Responder. However libspdm ignores the case when the transmit buffer of the Responder is small.

https://github.com/DMTF/libspdm/blob/6674aa87c5a3b6991e25e22fcd1bf44e2a111331/library/spdm_responder_lib/libspdm_rsp_receive_send.c#L614-L620 checks if the response is larger than at least one of the two buffers. If it is then `LargeResponse` is eventually sent to the Requester. When the Responder receives `CHUNK_GET` it ignores its own transmit buffer. For example
https://github.com/DMTF/libspdm/blob/6674aa87c5a3b6991e25e22fcd1bf44e2a111331/library/spdm_responder_lib/libspdm_rsp_chunk_get.c#L116-L119 This leads to hitting assertions like https://github.com/DMTF/libspdm/blob/main/library/spdm_responder_lib/libspdm_rsp_chunk_get.c#L107 and https://github.com/DMTF/libspdm/blob/6674aa87c5a3b6991e25e22fcd1bf44e2a111331/library/spdm_responder_lib/libspdm_rsp_chunk_get.c#L122

This should be fixed in 2.3 as well. Note that I fixed this issue in the specification for `CHUNK_SEND` but not `CHUNK_GET` : https://github.com/DMTF/SPDM-WG/pull/2715.

steven-bellock:
@jyao1 @Wenxing-hou I don't understand how `spdm-emu` passes the "test" where Responder and Requester have different buffer sizes.

jyao1:
@Zhiqiang520 , please take a look.

Zhiqiang520:
@jyao1 @steven-bellock @Wenxing-hou  
I found the root cause why the **current** [`spdm-emu`](https://github.com/DMTF/spdm-emu/commit/94668a9b76d39401b5c043552b7045ca505c87bc) passes the "test" where Responder and Requester have different buffer sizes.

[...]

PR #2158, opened 2023-06-20
Fix Chunk issue

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2141
Fix https://github.com/DMTF/libspdm/issues/2145


jyao1:

Issue #1344, opened 2022-11-02
Capability flags (and many other fields) should be masked before being stored in SPDM context


steven-bellock:
Currently libspdm blindly copies all bits from a request / response into its context. For example : https://github.com/DMTF/libspdm/blob/7002d5202b28f0f89bba3beb90c9b3f54eea54af/library/spdm_requester_lib/libspdm_req_get_capabilities.c#L274 In the case of `CAPABILITIES`, a malicious actor could fill the reserved portion of the `Flags` field that may be used in a ROP-type attack. Instead libspdm should mask the reserved or unused fields which hopefully reduces the chance of exploitation. In the case of a 1.0 `CAPABILITIES` response that would entail setting the upper bytes of the `Flags` field to 0. Ie
```
spdm_context->connection_info.capability.flags = spdm_response->flags & 0x0000004f;
```

hiption:
in SPDM 1.0, the used bit size is 1byte.
Why is the mask 0x0000004f, not 0x000000ff?

steven-bellock:
If you're talking about the `PSS_CAP` field that's a bug in the specification. It shouldn't be there. 1.0 only has 6 bits for capabilities.

hiption:
I wondered why PSS_CAP is described in SPDM 1.0.
Thank you for letting me know the bug

jyao1:
11/7/2022 - agree. Good catch.

PR #2160, opened 2023-06-20
Add capability mask.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/1344


mcsmonk (approved):

steven-bellock (approved):
Hopefully this does not break `spdm-emu`.

PR #2163, opened 2023-06-21
Fix the responder chunk checking

jyao1:
Currently the responder does not check requester's chunk cap.

This patch adds requester's chunk cap.


library/spdm_responder_lib/libspdm_rsp_receive_send.c
```diff
@@ -619,7 +619,7 @@ libspdm_return_t libspdm_build_response(void *spdm_context, const uint32_t *sess
          ((context->local_context.capability.sender_data_transfer_size != 0) &&
           (my_response_size > context->local_context.capability.sender_data_transfer_size))) &&
         libspdm_is_capabilities_flag_supported(
-            context, false, 0,
+            context, false, SPDM_GET_CAPABILITIES_REQUEST_FLAGS_CHUNK_CAP,
```
steven-bellock, line 622: This logic will probably change as the other chunk issues get resolved. But if you're going issue-by-issue it's fine to change, so long as at the end the logic makes sense.

steven-bellock (approved):

PR #2169, opened 2023-06-23
Add new macros and fix old ones

### LINKED ISSUE #1682: Capability / support macros
Currently the use of the `_CAPABILITY_` macros are not in line with the actual SPDM capabilities of the endpoint. For example the Requester has to enable `LIBSPDM_ENABLE_CAPABILITY_CERT_CAP` to send a `GET_CERTIFICATE` request to the Responder, whereas the Requester's `CERT_CAP` capabilty specifies whether the Requester can provide a `CERTIFICATE` response to a Responder's `GET_CERTIFICATE` request.

To fix this I think there should be two sets of macros. The `_CAPABILITY_` macros we have now would be aligned to the capabilities reported in `GET_CAPABILITIES / CAPABILITIES`. Then a new set of macros, `_SUPPORT_`, would be used to enable messages that aren't communicated through `GET_CAPABILITIES / CAPABILITIES`. 

So, for example, if a Requester is never going to send a `GET_CERTIFICATE` request, it would set `LIBSPDM_ENABLE_CERT_SUPPORT` to `0`, and it would still be free to set `LIBSPDM_ENABLE_CAPABILITY_CERT_CAP` to 1.

### LINKED ISSUE #2146: Evaluation of `ENCAP_CAP` and `MUT_AUTH_CAP` macros are incorrect
The SPDM 1.2.0 specification erroneously deprecated `ENCAP_CAP` and lumped its functionality in with `MUT_AUTH_CAP`. Now that `ENCAP_CAP` has been undeprecated libspdm needs to update its evaluation of `ENCAP_CAP` and `MUT_AUTH_CAP` macros. In particular there are many instances of `#if (LIBSPDM_ENABLE_CAPABILITY_MUT_AUTH_CAP) || (LIBSPDM_ENABLE_CAPABILITY_ENCAP_CAP)` that are incorrect, as an endpoint may support mutual authentication through raw public keys and not set `ENCAP_CAP`. Most of these instances need to be changed to `#if (LIBSPDM_ENABLE_CAPABILITY_MUT_AUTH_CAP) && (LIBSPDM_ENABLE_CAPABILITY_ENCAP_CAP)` where appropriate.

Note that SPDM 1.1 still says 
>`ENCAP_CAP` - If mutual authentication is supported, this field shall be set.

which is not right. Not sure what should be done there.

steven-bellock:
Fix #1682.
Fix #2146.


steven-bellock:
Need to do some testing first then I'll open it up for review.

steven-bellock:
@jyao1 The moral of this pull request is that `MUT_AUTH_CAP` does not have to equal `ENCAP_CAP` although SPDM 1.1 still says that.

Issue #873, opened 2022-04-29
[SET_CERTIFICATE] Save certificate into SPDM context


rw8896:
https://github.com/DMTF/libspdm/blob/a27fe8164d8edcbe3143e5b9e65f0c378c2df77a/library/spdm_responder_lib/libspdm_rsp_set_certificate.c#L138-L147

If reset is not required, should the certificate chain be added into spdm_context?
Note: in case of AliasCert model, it might need to extract Alias cert from slot 0 and append the Alias cert to compose a complete certificate chain.

jyao1:
DeviceCert mode + reset needed: Not save to context
AliasCert mode + reset needed: Not save to context

DeviceCert mode + reset not needed: Can save to context <==
AliasCert mode + reset not needed: Not sure if it is valid? 

Question: libspdm save the context? or Integrator save the context ?
Conclusion: It is easy that libspdm save the context.
We can let integrator save to the context by using SetData(), because the integrator can know the slot id and full cert chain data. But libspdm_write_certificate_to_nvm() does not have libspdm_context. It seems problematic. Maybe define a callback function?

rw8896:
Integrator saves Certificates into context seems a better solution.
- libspdm_write_certificate_to_nvm() didn't have libspdm_context: could we pass spdm context to the function?
Anyways, the integrator should have that context. Otherwise the integrator won't be able to set Certificates during libspdm initialization.
- current libspdm_write_certificate_to_nvm() returns bool, could we change it to return a status?
If integrator can't update the full Certificates into the spdm context, it should return SPDM_ERROR_CODE_RESET_REQUIRED or a specific error code. spdm_context->need_reset_to_set_cert is not necessary then.

jyao1:
Another special corner case, that a device may setup 2 spdm_context, such as PCI_DOE and MCTP.

If a certificate chain is setup from one context, do we need provision to the other context ?

jyao1:
> Integrator saves Certificates into context seems a better solution.
> 
> * libspdm_write_certificate_to_nvm() didn't have libspdm_context: could we pass spdm context to the function?
>   Anyways, the integrator should have that context. Otherwise the integrator won't be able to set Certificates during libspdm initialization.
> * current libspdm_write_certificate_to_nvm() returns bool, could we change it to return a status?
[...]

PR #2174, opened 2023-06-27
library: Pass spdm_context to libspdm_write_certificate_to_nvm()

alistair23:
If the implementor wants to update the certificate without rebooting the device they will need to know the spdm_context in order to modify the stored certificates. Let's pass that information to the libspdm_write_certificate_to_nvm() function so that it can use it.

### DIFF
--- a/include/hal/library/responder_secretlib.h
+++ b/include/hal/library/responder_secretlib.h
@@ -287,6 +287,7 @@ extern bool libspdm_is_in_trusted_environment();
  * Stores a certificate chain in non-volatile memory.
  *
  *
+ * @param[in]  spdm_context                  A pointer to the SPDM context.
  * @param[in]  slot_id          The number of slot for the certificate chain.
  * @param[in]  cert_chain       The pointer for the certificate chain to set.
  * @param[in]  cert_chain_size  The size of the certificate chain to set.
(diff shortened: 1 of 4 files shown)


jyao1:
I am not sure why we need spdm_context.

alistair23:
It's the only way to update the certificate for a running responder. Unless I'm missing something

jyao1:
you need update all SPDM context, right? not only this one.

The SPDM integrator should maintain all SPDM context and perform the update.

steven-bellock:
Corresponding issue : https://github.com/DMTF/libspdm/issues/873.

alistair23:
Thanks for pointing out the issue.

If there are multiple SPDM contexts then yes, they will all need to be updated. I think in that case the implementation (`responder_secretlib`) will need to know what to do here.

In general though I don't think we should require `responder_secretlib` to globally store the context information. Having a global to access the information that instead could be passed directly isn't ideal. If libspdm was used with Rust code for example that would required `unsafe` code, which could all be avoided by passing a `void *spdm_context` instead.

Reading through https://github.com/DMTF/libspdm/issues/873 it seems that the goal was to keep `responder_secretlib` independent of context to allow implementing it entirely in hardware. I'm not sure I follow, but couldn't the hardware implementation just ignore the context and use the current reset required method?

alistair23:
Closing as this doesn't seem to be the right approach

steven-bellock:
We need to pass the `spdm_context` in the case where multiple Responders / contexts are using the same libspdm binary. See #2494. So in that sense it's the right approach.

PR #2176, opened 2023-06-27
Assert LargeResponse and ResponseTooLarge can only be sent when spdm …

### LINKED ISSUE #2153: Ensure `LargeResponse` and `ResponseTooLarge` are only sent when negotiated version is >= 1.2
Original issue : https://github.com/DMTF/libspdm/issues/2141#issuecomment-1594855129.

`LargeResponse` and `ResponseTooLarge` can only be sent if the negotiated SPDM version is >= 1.2 There is no version check in `libspdm_build_response`. To catch this sooner, maybe `libspdm_generate_error_response` should have a bunch of assertions that check that the error code is appropriate for the negotiated version.

jyao1:
…version >= 1.2

Fix #2153


steven-bellock (approved):
This is a start, but it can obviously expanded to all the other error codes.

Issue #2165, opened 2023-06-22
Build fails in absence of algorithms


steven-bellock:
For example https://github.com/DMTF/libspdm/blob/c70fc237601628c271e576bca823aec481a35b8a/include/library/spdm_crypt_lib.h#L54-L58
If Integrator does not support AEAD then build will fail. 

Cryptography algorithms should be added to CI/CD.

DavidPu:
There is another build failure in this header file if some asymmetric algorithms are not enabled and the C preprocessor reaches to: https://github.com/DMTF/libspdm/blob/c70fc237601628c271e576bca823aec481a35b8a/include/library/spdm_crypt_lib.h#L41..L43

> spdm/lib/spdm//libspdm/include/library/spdm_crypt_lib.h:41:73: error: operator '||' has no right operand
 #elif ((LIBSPDM_ECDSA_P256_SUPPORT) || (LIBSPDM_SM2_DSA_P256_SUPPORT) ||
                                                                         ^

jyao1:
@DavidPu , that is also fixed in #2179.

PR #2179, opened 2023-06-27
fix build error when all DHE/AEAD disabled

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2165

NOTE: This patch does not enable CI, because lots of unit test (negotiate_algorithm and algorithms) depends on the presence of algorithms.

We need enable CI later.


jyao1:
@steven-bellock , please review if you have any comments.

steven-bellock (approved):

PR #2185, opened 2023-06-28
os_stub: spdm_device_secret_lib_sample: Fixup SVN index

alistair23:
The security version number (SVN) index should be 0x07 not 0x10 to match the DMTFSpecMeasurementValueType defined in table 45 in the 1.2.1 SPDM spec.


alistair23:
Whoops! Misread the table

Issue #2188, opened 2023-06-28
unaligned access shall be avoided


DavidPu:
There are some potential unaligned memory access which may trigger CPU exception if unaligned access is not supported.

PR #2190, opened 2023-06-28
library: fix unaligned type casts

DavidPu:
Unaligned access is not supported on some ARCHs. It will trigger CPU load/store fault. This change fixes unaligned access reported by -Wcast-align

Fix https://github.com/DMTF/libspdm/issues/2188

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -187,7 +187,7 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
 
     if(TOOLCHAIN STREQUAL "GCC")
         SET(CMAKE_C_COMPILER gcc)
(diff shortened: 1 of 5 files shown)


steven-bellock:
@DavidPu also add `-Wcast-align` to https://github.com/DMTF/libspdm/blob/main/CMakeLists.txt so that this is prevented in the future.

jyao1:
@DavidPu , could you please add one more commit to update CMakeLists.txt?

Then we can merge this PR.

DavidPu:
> @DavidPu , could you please add one more commit to update CMakeLists.txt?
> 
> Then we can merge this PR.

sorry for the delay! pushed new patchset to add -Wcast-align into CMakeLists.txt

steven-bellock (approved):

PR #2194, opened 2023-07-01
Clean up unit test

l1k:
~~Two~~ One assorted cleanup for debug code which has accumulated on my development branch:

- Elimination of superfluous code in a unit test
- ~~Addition of ASCII output to hex dumps~~

Thanks!


jyao1:
@l1k , thanks for the PR.
I think they are totally 2 different cleanup. Orthogonal.
In our rule, each PR should address only one issue.
A such, would you please separate them to 2 different PR?

l1k:
@jyao1 Fair enough, I've moved the hex+ascii commit to separate pull #2196.

PR #2195, opened 2023-07-02
fix the large_response_size check.

jyao1:


steven-bellock:
What's wrong with using `large_response_capacity`? `libspdm_get_scratch_buffer_large_message_capacity` returns `spdm_context->local_context.capability.max_spdm_msg_size`

jyao1:
large_response_capacity may be larger, fully depending on the libspdm implementation on scratch buffer.
max_spdm_msg_size only depends on SPDM spec, which is the one we should use.

This also fixed the return status code.

steven-bellock (approved):

PR #2196, opened 2023-07-02
Augment debug hex dumps with ASCII representation

l1k:
When emitting a hex dump, emit an ASCII representation as well.

This eases debugging if the data to be dumped contains human-readable strings.


steven-bellock:
@l1k has this helped you? Most of what gets printed is "....".

PR #2197, opened 2023-07-02
Make sequence number big-endian

### LINKED ISSUE #2166: TLS1.3 compliance issue w.r.t sequence number handling (AEAD encryption and decryption)
Hi,

Libspdm does not handle sequence number in correct byte order, as per the spec it's encoded in network byte order so it should be handled in the network byte order.

As per TLS1.3 spec section 5.3 (below)

https://datatracker.ietf.org/doc/html/rfc8446#section-5.3

"
The per-record nonce for the AEAD
   construction is formed as follows:

   1.  The 64-bit record sequence number is encoded in network byte
       order and padded to the left with zeros to iv_length.

   2.  The padded sequence number is XORed with either the static
       client_write_iv or server_write_iv (depending on the role).

   The resulting quantity (of length iv_length) is used as the
   per-record nonce.
"

Libspdm reference code -  (file: libspdm_secmes_encode_decode.c)
=============================================

```
libspdm_return_t libspdm_decode_secured_message(
--
--
--
--
    if (sequence_number > 0) {
        data64 = libspdm_read_uint64((const uint8_t *)salt) ^
                 (sequence_number - 1) ^ sequence_number;
    }
--
--
}
```

As a result the salt/IV produced results to incorrect values which causes failures from device (being as responder)

Thanks,
Manish

steven-bellock:
Fix #2166.

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -1588,6 +1588,14 @@ uint64_t libspdm_read_uint64(const uint8_t *buffer);
 void libspdm_write_uint64(uint8_t *buffer, uint64_t value);
 
 /**
+ * Converts a 64-bit little endian number to big endian.
+ *
+ * @param buffer Buffer to hold the converted value.
+ * @param value  The 64-bit little endian number.
+ */
+void libspdm_ltob_uint64(uint8_t *buffer, uint64_t value);
+
+/**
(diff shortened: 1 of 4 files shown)


steven-bellock:
@jyao1 tests need to be fixed but you can start reviewing the change.

jyao1:
We need wait for feedback from SPDM WG.

PR #2201, opened 2023-07-03
Create 3.0 change log

steven-bellock:


steven-bellock:
@jyao1 feel free to edit / add to this as you see fit.

Issue #2207, opened 2023-07-04
`spdm_responder_emu.exe` will `LIBSPDM_ASSERT` when `-DLIBSPDM_SENDER_BUFFER_SIZE=170 -DLIBSPDM_RECEIVER_BUFFER_SIZE=170 -DLIBSPDM_MAX_SPDM_MSG_SIZE=170`.


Zhiqiang520:
`spdm_responder_emu.exe` will `LIBSPDM_ASSERT` when `-DLIBSPDM_SENDER_BUFFER_SIZE=170 -DLIBSPDM_RECEIVER_BUFFER_SIZE=170 -DLIBSPDM_MAX_SPDM_MSG_SIZE=170`.

Verified on the current latest commit https://github.com/DMTF/spdm-emu/commit/5dd34068a580e96e3dc64d10ab35289ad4e35828.
The detail step is as following:

**Step-1**:

> mkdir build_x64_Debug_mbedtls_size_170
cd build_x64_Debug_mbedtls_size_170
cmake -E env CFLAGS="-DLIBSPDM_SENDER_BUFFER_SIZE=170 -DLIBSPDM_RECEIVER_BUFFER_SIZE=170 -DLIBSPDM_MAX_SPDM_MSG_SIZE=170" cmake -G"NMake Makefiles" -DARCH=x64 -DTOOLCHAIN=VS2019 -DTARGET=Debug -DCRYPTO=mbedtls ..
nmake copy_sample_key
nmake

**Step-2**:

> cd build_x64_Debug_mbedtls_size_170\bin
spdm_responder_emu.exe

**Step-3**:

> cd build_x64_Debug_mbedtls_size_170\bin
spdm_requester_emu.exe

**Step-4**:
at the end,  `spdm_responder_emu.exe` `LIBSPDM_ASSERT`.
![image](https://github.com/DMTF/spdm-emu/assets/90754803/22f27c21-8820-4855-a8fc-9dfee3dbc570)

jyao1:
Discussed in SPDM WG. It seems we dont want to support the case that cert chain size larger than MaxSPDMmsgSize.
As such, we need to return ResponseTooLarge in this case.

steven-bellock:
@Zhiqiang520 You have misconfigured the Responder. Presumably the certificates are larger than `170 B` and so the Responder is unable to form a response. An assertion is a correct action to a misconfigured endpoint.

Zhiqiang520:
@steven-bellock agree.

PR #2209, opened 2023-07-05
Rsp check max_spdm_msg_size size to avoid assert

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2207

### DIFF
--- a/library/spdm_common_lib/libspdm_com_context_data.c
+++ b/library/spdm_common_lib/libspdm_com_context_data.c
@@ -977,6 +977,7 @@ libspdm_return_t libspdm_get_data(void *spdm_context, libspdm_data_type_t data_t
 bool libspdm_check_context (void *spdm_context)
 {
     libspdm_context_t *context;
+    size_t index;
 
     context = spdm_context;
 
@@ -1016,6 +1017,25 @@ bool libspdm_check_context (void *spdm_context)
                        context->local_context.capability.sender_data_transfer_size));
         return false;
     }
+
+    if (((context->local_context.capability.flags &
+          SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_CHUNK_CAP) != 0) &&
+        (context->local_context.capability.max_spdm_msg_size != 0)) {
+        for (index = 0; index < SPDM_MAX_SLOT_COUNT; index++) {
+            if ((context->local_context.local_cert_chain_provision_size[index] != 0) &&
+                (context->local_context.local_cert_chain_provision_size[index] +
+                 sizeof(spdm_certificate_response_t) >
+                 context->local_context.capability.max_spdm_msg_size)) {
+                LIBSPDM_DEBUG((LIBSPDM_DEBUG_ERROR,
+                               "max_spdm_msg_size (%d) must be greater than or "
+                               "equal to local_cert_chain_provision_size[%d] (%d).\n",
+                               context->local_context.capability.max_spdm_msg_size, index,
+                               context->local_context.local_cert_chain_provision_size[index]));
+                return false;
+            }
+        }
+    }
+
     return true;
 }
 #endif /* LIBSPDM_CHECK_CONTEXT */
--- a/library/spdm_responder_lib/libspdm_rsp_receive_send.c
+++ b/library/spdm_responder_lib/libspdm_rsp_receive_send.c
@@ -612,10 +612,12 @@ libspdm_return_t libspdm_build_response(void *spdm_context, const uint32_t *sess
     }
 
     if (status == LIBSPDM_STATUS_SUCCESS) {
+        LIBSPDM_ASSERT (my_response_size <= context->local_context.capability.max_spdm_msg_size);
         /* large SPDM message is the SPDM message whose size is greater than the DataTransferSize of the receiving
          * SPDM endpoint or greater than the transmit buffer size of the sending SPDM endpoint */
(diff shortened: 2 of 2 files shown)


library/spdm_responder_lib/libspdm_rsp_certificate.c
```diff
@@ -111,6 +113,36 @@ libspdm_return_t libspdm_get_response_certificate(libspdm_context_t *spdm_contex
         length = (uint16_t)(spdm_context->local_context.local_cert_chain_provision_size[slot_id] -
                             offset);
     }
+
+    /* In chunk case, ensure the length request can be served, otherwise return error. */
+    if (libspdm_is_capabilities_flag_supported(spdm_context, false,
```
steven-bellock, line 118: Why is this needed? Wouldn't `libspdm_build_response` handle chunking like for all the other messages?

jyao1: The length must be calculated and checked here properly.

Otherwise, it will cause ASSERT at line 152 - `LIBSPDM_ASSERT(*response_size >= sizeof(spdm_certificate_response_t) + length);`

steven-bellock: Ah so this is a case where the Integrator has provisioned a certificate chain that is larger than its `MaxSPDMmsgSize`. In that case this is true even if chunking is not enabled and the
```
LIBSPDM_ASSERT (length + sizeof(spdm_certificate_response_t) <=
                        spdm_context->local_context.capability.max_spdm_msg_size);
```
is unconditional, assuming that `max_spdm_msg_size != 0`.

steven-bellock: That check can also be added to `libspdm_check_context` to help them catch that sooner.

library/spdm_responder_lib/libspdm_rsp_certificate.c
```diff
@@ -111,6 +113,36 @@ libspdm_return_t libspdm_get_response_certificate(libspdm_context_t *spdm_contex
...
+            LIBSPDM_DEBUG((LIBSPDM_DEBUG_ERROR, "cert size > req max_spdm_msg_size\n"));
+            actual_size = length + sizeof(spdm_certificate_response_t);
+            return libspdm_generate_extended_error_response(
+                spdm_context, SPDM_ERROR_CODE_RESPONSE_TOO_LARGE, 0,
+                sizeof(uint32_t), (uint8_t *)&actual_size, response_size, response);
+        }
+    }
+
+    /* Adjust the length according to max_spdm_msg_size to ensure it can be fit into response buffer. */
+    if (spdm_context->connection_info.capability.max_spdm_msg_size != 0) {
```
steven-bellock, line 135: The specification is explicit that
>If the size of a response message exceeds the size of the MaxSPDMmsgSize of the requesting SPDM endpoint, the responding SPDM endpoint shall respond with an ERROR message of ErrorCode=ResponseTooLarge.

so the Responder shouldn't truncate the response based on the Requester's `MaxSPDMmsgSize`.

library/spdm_responder_lib/libspdm_rsp_certificate.c
jyao1: For local_context, I dont believe we need 0 check.

jyao1: Uh, I am not sure if `libspdm_check_context` can help here, because the certificate chain may be provisioned later.
What we can do is only "if cert_chain != NULL, then check".

But we really dont know when `libspdm_check_context` will be invoked by the integrator. Maybe never.
Then we still need ASSERT here, even we add to `libspdm_check_context`

steven-bellock: The assertion would remain here but it can also be checked in `libspdm_check_context`.

library/spdm_responder_lib/libspdm_rsp_certificate.c
jyao1: done. This check is removed.

library/spdm_responder_lib/libspdm_rsp_certificate.c
jyao1: agree. Done.

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -1016,6 +1017,24 @@ bool libspdm_check_context (void *spdm_context)
...
+        for (index = 0; index < SPDM_MAX_SLOT_COUNT; index++) {
+            if ((context->local_context.local_cert_chain_provision_size[index] != 0) &&
+                (context->local_context.local_cert_chain_provision_size[index] +
+                 sizeof(spdm_certificate_response_t) >
+                 context->local_context.capability.max_spdm_msg_size)) {
+                LIBSPDM_DEBUG((LIBSPDM_DEBUG_ERROR,
+                               "max_spdm_msg_size (%d) must be greater than or "
+                               "equal to local_cert_chain_provision_size(%d) (%d).\n",
+                               context->local_context.capability.max_spdm_msg_size, index,
+                               context->local_context.local_cert_chain_provision_size[index]));
```
steven-bellock, line 1033: Needs to return `false`.

jyao1: fixed.

Issue #2206, opened 2023-07-04
`spdm_requester_emu.exe` will `LIBSPDM_ASSERT` if enable chunk and set `MaxSPDMmsgSize` equal to `MinDataTransferSize` +1.


Zhiqiang520:
`spdm_requester_emu.exe` will `LIBSPDM_ASSERT` if enable chunk and set `MaxSPDMmsgSize` equal to `MinDataTransferSize` +1.

Verified on the current latest commit https://github.com/DMTF/spdm-emu/commit/5dd34068a580e96e3dc64d10ab35289ad4e35828.
The detail step is as following:

**Step-1**:

> mkdir build_x64_Debug_mbedtls_default
cd build_x64_Debug_mbedtls_default
cmake -G"NMake Makefiles" -DARCH=x64 -DTOOLCHAIN=VS2019 -DTARGET=Debug -DCRYPTO=mbedtls ..
nmake copy_sample_key
nmake

**Step-2**:

> mkdir build_x64_Debug_mbedtls_size_170
cd build_x64_Debug_mbedtls_size_170
cmake -E env CFLAGS="-DLIBSPDM_SENDER_BUFFER_SIZE=170 -DLIBSPDM_RECEIVER_BUFFER_SIZE=170 -DLIBSPDM_MAX_SPDM_MSG_SIZE=43" cmake -G"NMake Makefiles" -DARCH=x64 -DTOOLCHAIN=VS2019 -DTARGET=Debug -DCRYPTO=mbedtls ..
nmake copy_sample_key
nmake

**Step-3**:

> cd build_x64_Debug_mbedtls_default\bin
spdm_responder_emu.exe

**Step-4**:

> cd build_x64_Debug_mbedtls_size_170\bin
spdm_requester_emu.exe

Then `LIBSPDM_ASSERT`:
![image](https://github.com/DMTF/spdm-emu/assets/90754803/b5bebe7d-a65b-48e7-bf19-233878f5ea8e)

**Step-5**:
After debugging, I found it assert in `libspdm_copy_mem` at the following line 629.
https://github.com/DMTF/libspdm/blob/1c34043d8142ed285ead67eb07d4aa09c8d0a8d8/library/spdm_requester_lib/libspdm_req_send_receive.c#L619-L634
![image](https://github.com/DMTF/spdm-emu/assets/90754803/027cfd97-bfb7-46c4-a0c1-76a5dca5aef9)

PR #2210, opened 2023-07-05
req check max_spdm_msg_size to avoid assert.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2206

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_send_receive.c
+++ b/library/spdm_requester_lib/libspdm_req_send_receive.c
@@ -601,6 +601,13 @@ libspdm_return_t libspdm_send_spdm_request(libspdm_context_t *spdm_context,
         }
     }
 
+    if ((spdm_context->connection_info.capability.max_spdm_msg_size != 0) &&
+        (request_size > spdm_context->connection_info.capability.max_spdm_msg_size)) {
+        LIBSPDM_DEBUG((LIBSPDM_DEBUG_ERROR, "request_size > rsp max_spdm_msg_size\n"));
+        return LIBSPDM_STATUS_PEER_BUFFER_TOO_SMALL;
+    }
+    LIBSPDM_ASSERT (request_size <= spdm_context->local_context.capability.max_spdm_msg_size);
+
     #if LIBSPDM_ENABLE_MSG_LOG
     /* First save the size of the message log buffer. If there is an error it will be reverted. */
     msg_log_size = libspdm_get_msg_log_size(spdm_context);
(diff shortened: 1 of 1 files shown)


library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -601,6 +601,18 @@ libspdm_return_t libspdm_send_spdm_request(libspdm_context_t *spdm_context,
         }
     }
 
+    if ((spdm_context->connection_info.capability.max_spdm_msg_size != 0) &&
+        (request_size > spdm_context->connection_info.capability.max_spdm_msg_size)) {
+        LIBSPDM_DEBUG((LIBSPDM_DEBUG_ERROR, "request_size > rsp max_spdm_msg_size\n"));
+        return LIBSPDM_STATUS_PEER_BUFFER_TOO_SMALL;
+    }
+    if (request_size > spdm_context->local_context.capability.max_spdm_msg_size) {
```
steven-bellock, line 609: `local_context.capability.max_spdm_msg_size` is the same size / buffer for both sending and receiving message, correct?

jyao1: correct.
We only distinguish device hw buffer for sender/receiver. 
max_spdm_msg_size is sw buffer and same for sender/receiver.

steven-bellock: Ok. I would replace all of this with
```
LIBSPDM_ASSERT(request_size <= spdm_context->local_context.capability.max_spdm_msg_size);
```
since the request message has already been formed and by definition it can't be larger than the local `max_spdm_msg_size`.

jyao1: done.

library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -601,6 +601,13 @@ libspdm_return_t libspdm_send_spdm_request(libspdm_context_t *spdm_context,
         }
     }
 
+    if ((spdm_context->connection_info.capability.max_spdm_msg_size != 0) &&
+        (request_size > spdm_context->connection_info.capability.max_spdm_msg_size)) {
+        LIBSPDM_DEBUG((LIBSPDM_DEBUG_ERROR, "request_size > rsp max_spdm_msg_size\n"));
+        return LIBSPDM_STATUS_PEER_BUFFER_TOO_SMALL;
+    }
+    LIBSPDM_ASSERT (request_size <= spdm_context->local_context.capability.max_spdm_msg_size);
```
steven-bellock, line 609: Sorry, this should probably also include `max_spdm_msg_size != 0` in the assertion.

jyao1: This is local_context.
I dont think we need add zero check.

steven-bellock (approved):

PR #2235, opened 2023-07-11
os_stub: Allow disabling time checks

alistair23:
Some implementations might not have accurate clocks, so allow disabling time checks.

### DIFF
--- a/doc/user_guide.md
+++ b/doc/user_guide.md
@@ -301,6 +301,8 @@ Refer to spdm_server_init() in [spdm_responder.c](https://github.com/DMTF/spdm-e
 
    0.5, implement required SPDM device IO functions - `libspdm_device_send_message_func` and `libspdm_device_receive_message_func` according to [spdm_common_lib](https://github.com/DMTF/libspdm/blob/main/include/library/spdm_common_lib.h).
 
+   0.6, if the device does not have access to a real-time clock and if the device uses OpenSSL or Mbed TLS then undefine `OPENSSL_CHECK_TIME` or `MBEDTLS_HAVE_TIME_DATE`.
+
 0. Implement a proper spdm_device_secret_lib.
 
 1. Initialize SPDM context (similar to SPDM Requester)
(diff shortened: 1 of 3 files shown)


os_stub/cryptlib_openssl/pk/x509.c
```diff
@@ -1887,6 +1887,7 @@ bool libspdm_x509_verify_cert(const uint8_t *cert, size_t cert_size,
      * still trusted intermediate certificate. Also disable time checks.*/
 
     X509_STORE_set_flags(cert_store, X509_V_FLAG_PARTIAL_CHAIN);
+    X509_STORE_set_flags(cert_store, X509_V_FLAG_NO_CHECK_TIME);
```
jyao1, line 1890: Can we just fix comment?

I forget why it is comment in this way.

alistair23: A lot of SPDM devices don't have reliable clocks, so I thought skipping the time check made a lot of sense

jyao1: Then, this is a functional change.
Please file an issue and let's discuss it in SPDM TF meeting.

alistair23: It's not really a functional change, the comment already says we do this. I have created an issue here: https://github.com/DMTF/libspdm/issues/2237

jyao1: If code does not have the check, then it is a functional change.

You can treat it as a comment bug.

doc/user_guide.md
```diff
@@ -301,6 +301,8 @@ Refer to spdm_server_init() in [spdm_responder.c](https://github.com/DMTF/spdm-e
 
    0.5, implement required SPDM device IO functions - `libspdm_device_send_message_func` and `libspdm_device_receive_message_func` according to [spdm_common_lib](https://github.com/DMTF/libspdm/blob/main/include/library/spdm_common_lib.h).
 
+   0.6, if the device doesn't have an accurate time undefine OPENSSL_CHECK_TIME or MBEDTLS_HAVE_TIME_DATE
```
steven-bellock, line 304: if the device does not have access to a real-time clock and if the device uses OpenSSL or Mbed TLS then undefine  `OPENSSL_CHECK_TIME` or `MBEDTLS_HAVE_TIME_DATE`.

PR #2236, opened 2023-07-11
Add AEAD limit doc

jyao1:


steven-bellock:
I'll add this to https://github.com/DMTF/libspdm/blob/main/doc/api/common_api.md
Probably don't need a separate document.

jyao1:
I merged #2239, but I don't think there is any conflict.

I prefer to create it and remind people that they have such option for security.
Also, the AEAD limit is a new feature I would like to mention in release note.

PR #2252, opened 2023-07-18
Check `LargeResponse` error

### LINKED ISSUE #1207: Requester validation of LargeResponse
Currently the Requester checks if the Responder is chunking responses that it shouldn't be, like `VERSION`, after it has already received the chunked response at the end of `libspdm_receive_spdm_response`. Instead the check should occur immediately after the `LargeResponse` error from the Responder and be based on the request message. So if the Requester sends `GET_VERSION` and the Responder replies with `LargeResponse` then the Requester should immediately error out, rather then receiving the chunked `VERSION` response and then erroring out.

steven-bellock:
Fix #1207.

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_send_receive.c
+++ b/library/spdm_requester_lib/libspdm_req_send_receive.c
@@ -734,6 +734,27 @@ libspdm_return_t libspdm_receive_spdm_response(libspdm_context_t *spdm_context,
 
     if (spdm_response->request_response_code == SPDM_ERROR
         && spdm_response->param1 == SPDM_ERROR_CODE_LARGE_RESPONSE) {
+        spdm_message_header_t *request_message;
+
+        request_message = (spdm_message_header_t *)spdm_context->last_spdm_request;
+
+        switch (request_message->request_response_code)
+        {
+        case SPDM_GET_VERSION:
+        case SPDM_GET_CAPABILITIES:
+        case SPDM_PSK_FINISH:
+        case SPDM_HEARTBEAT:
+        case SPDM_KEY_UPDATE:
+        case SPDM_END_SESSION:
+        case SPDM_SET_CERTIFICATE:
+            /* The responses to these requests cannot be chunked as they are smaller than the
+             * MinDataTransferSize. */
+            status = LIBSPDM_STATUS_ERROR_PEER;
+            goto receive_done;
+        default:
+            break;
+        }
+
         status = libspdm_handle_error_large_response(
             spdm_context, session_id,
             response_size, (void*) spdm_response, response_capacity);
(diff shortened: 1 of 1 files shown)


steven-bellock:
@Zhiqiang520 too.

library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -734,6 +734,27 @@ libspdm_return_t libspdm_receive_spdm_response(libspdm_context_t *spdm_context,
...
+
+        switch (request_message->request_response_code)
+        {
+        case SPDM_GET_VERSION:
+        case SPDM_GET_CAPABILITIES:
+        case SPDM_PSK_FINISH:
+        case SPDM_HEARTBEAT:
+        case SPDM_KEY_UPDATE:
+        case SPDM_END_SESSION:
+        case SPDM_SET_CERTIFICATE:
```
jyao1, line 749: SPDM spec only ensures `SPDM_GET_VERSION` and `SPDM_GET_CAPABILITIES` cannot be chunked.

Why we put other here? Those messages may change in the future.

steven-bellock: If they change in the future then we can put version logic in there. But for now they cannot be chunked.

jyao1: I don't think it is useful to add such check.
1) The list is unnecessary, because the list not defined in the spec.
2) The list is insufficient, because any command/response, whose length is smaller than `DataTransferSize`, should be forbidden and add to the list.

Adding more here just creates confusing to me.

I think we should choose one of below:
1) Just check SPDM_GET_VERSION and SPDM_GET_CAPABILITIES, which is defined in SPDM code.
2) Add a generic check for the response length and `DataTransferSize`.

steven-bellock: >The list is unnecessary, because the list not defined in the spec.

The specification says
> For SPDM messages that are less than or equal to both the DataTransferSize of the receiving SPDM endpoint and the transmit buffer size of the sending SPDM endpoint, the sending SPDM endpoint shall not utilize this transfer mechanism.

Once the specification has been clarified that the transmit buffer size must be larger than `MinDataTransferSize` then the combination of the above clause plus the fact that the specification defines the size of responses (ie, `KEY_UPDATE_ACK` is defined to be four bytes) then the list is defined in the specification.

>The list is insufficient, because any command/response, whose length is smaller than DataTransferSize, should be forbidden and add to the list.

The Responder's transmit buffer size can be smaller than the Requester's `DataTransferSize` in which case the Responder may reply with `LargeResponse` to a message that is less than the Requester's `DataTransferSize`.

steven-bellock: I will put this logic in a generic function.

steven-bellock:
`FINISH_RSP` is a funny one because whether or not it is smaller than `MinSPDMMesgSize` depends on
- The size of the negotiated hashing algorithm AND
- `HANDSHAKE_IN_THE_CLEAR_CAP`

so I may use the negotiated parameters to determine if that can be chunked or not.

PR #2262, opened 2023-07-22
Add more doc to readme.

jyao1:


README.md
```diff
@@ -57,6 +67,8 @@
 
    Support to be linked by other language. For example, [JAVA verifier](https://github.com/altera-opensource/verifier).
 
+   Support interoperability testing with other SPDM implementation. For example, [intel-server-prot-spdm](https://github.com/intel/intel-server-prot-spdm) and [rust-spdm](https://github.com/intel/rust-spdm).
```
steven-bellock, line 70: `implementation` -> `implementations`

steven-bellock (changes requested):

README.md
jyao1: > `implementation` -> `implementations`

Done

jyao1: > Is intel-server-prot-spdm a separate implementation? Doesn't it use libspdm?

It is just a set of test. The use case is simple - just do authentication and measurement, no secure session. See https://github.com/intel/intel-server-prot-spdm/blob/main/intelprot/spdm.py.

The spdm-emu is used for interoperability test - https://github.com/intel/intel-server-prot-spdm/blob/main/intelprot/testprot.py.

steven-bellock (approved):

PR #2264, opened 2023-07-22
Allow no selected OpaqueDataFmt

### LINKED ISSUE #1389: No selected OpaqueDataFmt should be legal
libspdm requires `OpaqueDataFmt1` to be selected during `NEGOTIATE_ALGORITHMS`. See : https://github.com/DMTF/libspdm/blob/21729b5456e7075a23ca8918d0b12b3a4a0eac43/library/spdm_requester_lib/libspdm_req_negotiate_algorithms.c#L476-L481

However it should be legal to have no `OpaqueData`. From the specification
>If the Requester or the Responder does not set a bit, then all OpaqueData fields in this specification shall be absent by setting the respective OpaqueDataLen field to a value of zero.

steven-bellock:
Fix #1389.


steven-bellock:
To summarize this pull request.
- `OpaqueDataFmt1` is required for key exchange.
- If no `OpaqueDataFmt` is selected then endpoints will check that `OpaqueDataLength` is `0`.
- For `GET_CSR`, if no `OpaqueDataFmt` is selected and the Integrator provides opaque data then libspdm will override the parameter and return a warning to the Integrator.

library/spdm_requester_lib/libspdm_req_negotiate_algorithms.c
```diff
@@ -195,6 +195,13 @@ static libspdm_return_t libspdm_try_negotiate_algorithms(libspdm_context_t *spdm
         status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
         goto receive_done;
     }
+    if (spdm_request->header.spdm_version >= SPDM_MESSAGE_VERSION_12) {
+        if (!libspdm_onehot0(spdm_response->other_params_selection &
+                             SPDM_ALGORITHMS_OPAQUE_DATA_FORMAT_MASK)) {
```
jyao1, line 200: This check is not enough, because we dont want to support `SPDM_ALGORITHMS_OPAQUE_DATA_FORMAT_0`.

I think it should be "((selection & MASK) != FORMAT_NONE) && ((selection & MASK) != FORMAT_1)"

library/spdm_responder_lib/libspdm_rsp_algorithms.c
```diff
@@ -252,6 +252,8 @@ libspdm_return_t libspdm_get_response_algorithms(libspdm_context_t *spdm_context
 
     uint32_t other_params_support_priority_table[] = {
         SPDM_ALGORITHMS_OPAQUE_DATA_FORMAT_1,
+        SPDM_ALGORITHMS_OPAQUE_DATA_FORMAT_0,
```
jyao1, line 255: How do we support "FORMAT_0" ?

library/spdm_requester_lib/libspdm_req_negotiate_algorithms.c
steven-bellock: `SPDM_ALGORITHMS_OPAQUE_DATA_FORMAT_0` is fine for libspdm. It just means that the Integrator needs to provide / parse all of the opaque data and they can't do key exchange.

library/spdm_responder_lib/libspdm_rsp_algorithms.c
steven-bellock: `SPDM_ALGORITHMS_OPAQUE_DATA_FORMAT_0` is fine for libspdm. It just means that the Integrator needs to provide / parse all of the opaque data and they can't do key exchange.

steven-bellock:
@jyao1 the change to `NEGOTIATE_ALGORITHMS` has been reverted. It will fail if key exchange is enabled and `OpaqueFmt1` isn't selected.

library/spdm_responder_lib/libspdm_rsp_algorithms.c
```diff
@@ -683,16 +699,6 @@ libspdm_return_t libspdm_get_response_algorithms(libspdm_context_t *spdm_context
...
-            if (spdm_response->header.spdm_version >= SPDM_MESSAGE_VERSION_12) {
-                if ((spdm_context->connection_info.algorithm.other_params_support &
-                     SPDM_ALGORITHMS_OPAQUE_DATA_FORMAT_MASK) !=
-                    SPDM_ALGORITHMS_OPAQUE_DATA_FORMAT_1) {
-                    return libspdm_generate_error_response(
-                        spdm_context,
-                        SPDM_ERROR_CODE_INVALID_REQUEST, 0,
-                        response_size, response);
-                }
-            }
```
jyao1, line 695: Should we revert this as well?

steven-bellock: Sure. Long-term I don't think the Responder should return the error.

Issue #2270, opened 2023-07-26
No need to set Length=0xFFFF in chunk enabled case for GET_CERTIFICATE


jyao1:
See https://github.com/DMTF/libspdm/blob/main/library/spdm_requester_lib/libspdm_req_get_certificate.c#L127-L131

The spec just says:
```
If the Requester sets Offset to 0 and Length to 0xFFFF in the GET_CERTIFICATE request, the Responder shall set PortionLength equal to the size of the complete certificate chain stored in the requested slot,
```

There is no requirement to set length=0xFFFF.

steven-bellock:
A libspdm Requester always gets the entire certificate chain. So while it is not a specification requirement, it is a libspdm requirement.

Wenxing-hou:
In S`PDM spec 1.2.1 317 line`, there are description:
```
Endpoints that support the large SPDM message transfer mechanism message set, shall use the large SPDM
message transfer mechanism messages to manage the transfer of the requested certificate chain. Specifically:
• The Requester shall set Offset to 0 and Length to 0xFFFF in the GET_CERTIFICATE request.
```

But in the `SPDM spec 1.3 389 line`, there are description:
```
Endpoints that support the large SPDM message transfer mechanism message set shall use the large SPDM
message transfer mechanism messages to manage the transfer of the requested certificate chain when the
CERTIFICATE response is larger than either the DataTransferSize of the Requester or the transmit buffer of the
Responder. Specifically:
```
```
If the Requester sets Offset to 0 and Length to 0xFFFF in the GET_CERTIFICATE request, the Responder
shall set PortionLength equal to the size of the complete certificate chain stored in the requested slot, shall set
RemainderLength to 0,
```
And I have submitted the PR to sync it. Thanks.

PR #2304, opened 2023-08-07
Make get_cert sync the SPDM spec 1.3

Wenxing-hou:
Fix: #2270

According to spec 1.3, get_cert requester should not assume that it always receives the entire Cert or sets 0xFFFF when enabling the chunk.
 
 And the test pass the emu test.

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_get_certificate.c
+++ b/library/spdm_requester_lib/libspdm_req_get_certificate.c
@@ -124,12 +124,6 @@ static libspdm_return_t libspdm_try_get_certificate(libspdm_context_t *spdm_cont
                                                SPDM_GET_CAPABILITIES_REQUEST_FLAGS_CHUNK_CAP,
                                                SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_CHUNK_CAP);
 
-    if (chunk_enabled) {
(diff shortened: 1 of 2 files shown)


library/spdm_requester_lib/libspdm_req_get_certificate.c
```diff
@@ -261,11 +249,6 @@ static libspdm_return_t libspdm_try_get_certificate(libspdm_context_t *spdm_cont
             status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
             goto done;
         }
-        if (chunk_enabled && (spdm_response->remainder_length != 0)) {
```
jyao1, line 264: We should use (chunk_enable && (length == 0xffff) && (spdm_response->remainder_length != 0))

Wenxing-hou: Thanks. I have changed the code.

jyao1 (approved):

jyao1:
There is no need to use chunking for CERTIFICATE, because it already includes the chunk mechanism.
At least, I dont see the benefit, and SPDM spec does not mandate it.

However, if we expose the interface, the integrator may request full cert with all 0xFFFF as cert length.

PR #2310, opened 2023-08-11
cryptlib_openssl: Remove internal OpenSSL crypto include

alistair23:

Using the internal APIs makes it difficult to use libspdm as a library
instead use the public API.

We can also assume that the internal APIs don't provide backwards
compatible guarantees


alistair23:
Urgh, there is a `ossl_ec_group_do_inverse_ord()` that needs to be replaced as well.

I'd still appricate a review on what I currently have. I'll fix the `LIBSPDM_FIPS_MODE` next week

alistair23:
Closing for separate PRs to help with debugging Windows issues

PR #2313, opened 2023-08-14
cryptlib_openssl: x509: Remove internal OpenSSL crypto include

alistair23:

Using the internal APIs makes it difficult to use libspdm as a library
instead use the public API.

We can also assume that the internal APIs don't provide backwards
compatible guarantees

### DIFF
--- a/os_stub/cryptlib_openssl/pk/x509.c
+++ b/os_stub/cryptlib_openssl/pk/x509.c
@@ -17,7 +17,6 @@
 #include <openssl/bn.h>
 #include <openssl/pem.h>
 #include <openssl/bio.h>
-#include <crypto/x509.h>
 
 #if LIBSPDM_CERT_PARSE_SUPPORT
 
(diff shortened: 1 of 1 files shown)


jyao1 (dismissed):

jyao1:
@alistair23, Do you want to remove `#include <crypto/x509.h>` ?

alistair23:
Whoops, yep. Removed now

alistair23:
Urgh. Ok, I don't think it's possible to get the subject from a `X509_REQ_INFO` in a supported way

os_stub/cryptlib_openssl/pk/x509.c
```diff
@@ -2268,6 +2267,20 @@ bool libspdm_set_subject_name(X509_NAME *x509_name, char *subject_name)
...
+ *
+ * TODO: Stop accessing private parts of OpenSSL
+ */
+typedef struct X509_req_info_st {
+    ASN1_ENCODING enc;          /* cached encoding of signed part */
+    ASN1_INTEGER *version;      /* version, defaults to v1(0) so can be NULL */
+    X509_NAME *subject;         /* certificate request DN */
+    X509_PUBKEY *pubkey;        /* public key of request */
+    /* ... */
+} libspdm_X509_req_info_st;
```
jyao1, line 2282: I dont think it is a good idea to copy the data structure from openssl.

What if the internal structure is changed in new version?

jyao1: In general, if there is error, I hope it can be detected at build time (compilation error), but not runtime (crash).

IMHO, copying data structure brings bigger risk due to runtime error.

alistair23:
This is the last user of the internal APIs. I have pushed a change to avoid copying the OpenSSL struct

jyao1 (approved):

PR #2314, opened 2023-08-14
cryptlib_openssl: ecd: Remove internal OpenSSL crypto include

alistair23:

Using the internal APIS makes it difficult to use libspdm as a library
instead use the public API.

The current ECD code uses internal APIs, making it unsuitable for use in
production code or libraries.

The supported way to do this is via OSSL params, either with
EVP_PKEY_fromdata() [1] or using EVP_PKEY_set_octet_string_param().

Unfortunately this isn't supported in OpenSSL and ed25519_set_params()
and ed448_set_params() will always return 1, indicating no support.

As there doesn't appear to be a supported method in OpenSSL to set the
public and private keys, let's instead allow users to disable this
support so the library can be used with the regular OpenSSL libraries.

1: https://www.openssl.org/docs/man3.0/man3/EVP_PKEY_fromdata.html

### DIFF
--- a/os_stub/cryptlib_openssl/pk/ecd.c
+++ b/os_stub/cryptlib_openssl/pk/ecd.c
@@ -12,6 +12,9 @@
  **/
 
 #include "internal_crypt_lib.h"
+
+#if (LIBSPDM_EDDSA_ED25519_SUPPORT) || (LIBSPDM_EDDSA_ED448_SUPPORT)
+
 #include <openssl/evp.h>
 #include <crypto/evp.h>
 
@@ -471,3 +474,4 @@ bool libspdm_eddsa_verify(const void *ecd_context, size_t hash_nid,
     EVP_MD_CTX_free(ctx);
     return true;
 }
+#endif /* (LIBSPDM_EDDSA_ED25519_SUPPORT) || (LIBSPDM_EDDSA_ED448_SUPPORT) */


library/spdm_crypt_lib/fips/libspdm_selftest_eddsa.c
```diff
@@ -122,14 +105,6 @@ bool libspdm_fips_selftest_eddsa(void *fips_selftest_context)
         goto update;
     }
 
-    if (!libspdm_consttime_is_mem_equal(signature_25519, expected_signature_25519,
```
jyao1, line 125: why we cannot compare it?

alistair23: The signature isn't constant

alistair23: Good question! I have no idea. I was hoping whoever originally put the `evp_pkey_copy_downgraded()` call there would know.

For some reason not calling `evp_pkey_copy_downgraded()` means the signature is changing and I have no idea why. The [original commit](https://github.com/DMTF/libspdm/commit/882602c9487d1b15cbd892f2402b10640349b80f) added the call to  `evp_pkey_copy_downgraded()`, but I don't see a justification for the internal library call in the commit message

jyao1: @Wenxing-hou , please provide your comment.

Wenxing-hou: In the `libspdm_ecd_set_pri_key` and `libspdm_ecd_set_pub_key`, the purpose is setting the private key or public key to the `ecd_context`.

Using `libspdm_ecd_set_pub_key` as an example,  **the code in your PR is:**

```
    evp_key = (EVP_PKEY *)ecd_context;    (in line 102)

    evp_key = EVP_PKEY_new_raw_public_key(EVP_PKEY_id(evp_key), NULL,
                                          public_key, public_key_size);

    if (evp_key == NULL) {
        return false;
    }
```
This code will change the `evp_key` and will not achieve the goal(setting the private key or public key to the `ecd_context`).

**The origin code is:**
```
    evp_key = (EVP_PKEY *)ecd_context;  (in line 104)

    new_evp_key = EVP_PKEY_new_raw_public_key(EVP_PKEY_id(evp_key), NULL,
                                              public_key, public_key_size);

    if (new_evp_key == NULL) {
        return false;
    }

    if (evp_pkey_copy_downgraded(&evp_key, new_evp_key) != 1) {
        EVP_PKEY_free(new_evp_key);
        return false;
    }

    EVP_PKEY_free(new_evp_key);
```

Wenxing-hou: Then I find a API `EVP_PKEY_copy_parameters` in the `include\openssl\evp.h`.
**I changed the origin code:**
```
    new_evp_key = EVP_PKEY_new_raw_public_key(EVP_PKEY_id(evp_key), NULL,
                                              public_key, public_key_size);

    if (new_evp_key == NULL) {
        return false;
    }

    // if (evp_pkey_copy_downgraded(&evp_key, new_evp_key) != 1) {
    //     EVP_PKEY_free(new_evp_key);
    //     return false;
    // }

    if (EVP_PKEY_copy_parameters(evp_key, new_evp_key) != 1) {
        EVP_PKEY_free(new_evp_key);
        return false;
    }

    EVP_PKEY_free(new_evp_key);
    return true;
```

But the fips test will fail at:
```
    if (!libspdm_consttime_is_mem_equal(get_pub_key_25519, public_key_25519,
                                        sizeof(public_key_25519))) {
        LIBSPDM_DEBUG((LIBSPDM_DEBUG_INFO, "EDDSA 25519 get key content compare failed \n"));
        libspdm_ecd_free(ecd_25519);
        result = false;
        goto update;
    }
```

Then I  searched for various materials. It seems that the `EVP_PKEY_copy_parameters` only copy part of the EVP_PKEY parameters.

alistair23: Yeah, `EVP_PKEY_copy_parameters()` isn't what you want.

It really seems like this should use `EC_KEY` instead of `EVP_PKEY`. The `ec.c` seems like a much better approach to do this

alistair23:
What about adding an option to disable the ECD support at build time? At least until the implementation is updated to use public OpenSSL APIs?

library/spdm_crypt_lib/fips/libspdm_selftest_eddsa.c
alistair23: Following: https://www.openssl.org/docs/manmaster/man7/Ed448.html and the OpenSSL code, `EVP_PKEY_set_octet_string_param()` should work, but it doesn't

alistair23:
```c
static int ed25519_set_params(void *key, const OSSL_PARAM params[])
{
    return 1;
}
```

So OpenSSL doesn't support setting params for `ed25519`. `ed448` also has the same issue

alistair23:
I have updated the PR. It now jut allows user to disable the ECD support. Looking at the [OpenSSL source](https://github.com/openssl/openssl/blob/c2a8226cba2757b251729620aedffeed23d73623/providers/implementations/keymgmt/ecx_kmgmt.c#L436) it seems like ECD isn't well supported.

[This patch](https://github.com/alistair23/libspdm/commit/67a60ec0026669b6bfd3eef153e8495e1128f5ce) should fix things up to allow ECD support, but OpenSSL doesn't support it, at least not yet.

In the long run, the solution should be to either add support to OpenSSL or convert the ECD implementation to use `EC_KEY`. I think someone who understands ECD more then me will have to do the conversion though

jyao1:
I am OK to add macro around it.

jyao1 (approved):

Issue #2318, opened 2023-08-19
libspdm_secured_message_export_session_keys() does not export salt correctly.


jyao1:
In libspdm_secured_message_export_session_keys(), the salt is returned directly. See https://github.com/DMTF/libspdm/blob/main/library/spdm_secured_message_lib/libspdm_secmes_context_data.c#L245
```
    libspdm_copy_mem(secured_message_context->application_secret.response_data_salt,
                     sizeof(secured_message_context->application_secret.response_data_salt),
                     ptr, secured_message_context->aead_iv_size);
```

But salt is XORed with sequence_number. See https://github.com/DMTF/libspdm/blob/main/library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c#L129C1-L133C6
```
    if (sequence_number > 0) {
        data64 = libspdm_read_uint64((const uint8_t *)salt) ^
                 (sequence_number - 1) ^ sequence_number;
        libspdm_write_uint64(salt, data64);
    }
```

PR #2319, opened 2023-08-19
fix the salt value in export/import function.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2318


steven-bellock:
Why not keep the salt and sequence number in `libspdm_secure_session_keys_struct_t` completely separate and only combine them, locally, when encrypting / decrypting?

jyao1:
Looking at the history, I think we want to prevent the data leakage in stack...

Agree this could be a simpler way.

steven-bellock (approved):

Issue #2151, opened 2023-06-19
question on RSA/ECDSA signature endianness in SPDM 1.0/1.1


jyao1:
Current signature endianness is clarified in SPDM 1.2.1.

But SPDM 1.0/1.1 does not include the clarification.

SPDM issue: https://github.com/DMTF/SPDM-WG/issues/2929

steven-bellock:
Ideally there should be no byte-swapping, as that was the intent starting with 1.0.

xiaoyuruan:
According to the SPDM WG, SPDM 1.0/1.1 allows big or little endian, chosen by implementation.  As such, I propose that:

1. Provide compile-time option for libspdm Requester's signature generation, if negotiated VERSION == 1.0 or 1.1: choose one from: {big endian; little endian}.
2. Provide compile-time option for libspdm Responder's signature verification, if negotiated VERSION == 1.0 or 1.1: choose one from: {big endian; little endian; flexible - try both and accept signature if one passes}.

Understood this will not intercept libspdm3.0.0, but later.

Please assign Rich Kong to it. @richkong88

jyao1:
I think we need define a set of MACRO to control the different behavior:
```
LIBSPDM_SPDM_10_RSA_LITTLE_ENDIAN
LIBSPDM_SPDM_10_ECDSA_LITTLE_ENDIAN
LIBSPDM_SPDM_10_RSA_DUAL_VERIFICATION
LIBSPDM_SPDM_10_ECDSA_DUAL_VERIFICATION
LIBSPDM_SPDM_11_RSA_LITTLE_ENDIAN
LIBSPDM_SPDM_11_ECDSA_LITTLE_ENDIAN
LIBSPDM_SPDM_11_RSA_DUAL_VERIFICATION
LIBSPDM_SPDM_11_ECDSA_DUAL_VERIFICATION
```

richkong88:
> I think we need define a set of MACRO to control the different behavior:
> 
> ```
> LIBSPDM_SPDM_10_RSA_LITTLE_ENDIAN
> LIBSPDM_SPDM_10_ECDSA_LITTLE_ENDIAN
> LIBSPDM_SPDM_10_RSA_DUAL_VERIFICATION
> LIBSPDM_SPDM_10_ECDSA_DUAL_VERIFICATION
> LIBSPDM_SPDM_11_RSA_LITTLE_ENDIAN
[...]

PR #2324, opened 2023-08-24
Enable little-endian signing and dual-endian verification in SPDM 1.0…

richkong88:
…\1.1

Fix: https://github.com/DMTF/libspdm/issues/2151


jyao1:
@richkong88 , would you please squash the commit? There is no need to add patch for coding style fix.

include/library/spdm_lib_config.h
```diff
@@ -383,4 +383,68 @@
 #define LIBSPDM_CHECK_SPDM_CONTEXT 1
 #endif
 
+/* Enable Endianness for signatures on SPDM1.0 and SPDM1.1.
+* Big-endian is default, but little-endian may need to be supported as well.
+* See issue: https://github.com/DMTF/libspdm/issues/2151 */
```
jyao1, line 388: I dont think we need this line.

richkong88:
> @richkong88 , would you please squash the commit? There is no need to add patch for coding style fix.

Yes, I'll put everything into a single CL, once all feedback is incorporated.

jyao1:
I think we need to update req_asym version as well, such as `libspdm_req_asym_sign`, `libspdm_req_asym_verify`.

library/spdm_crypt_lib/libspdm_crypt_asym.c
```diff
@@ -833,28 +940,97 @@ bool libspdm_asym_verify_hash(
...
     }
 }
 
+void libspdm_asym_signature_swap_endian_if_necessary(
+    spdm_version_number_t spdm_version, uint32_t base_asym_algo,
+    uint8_t *signature, size_t sig_size)
+{
+#if LIBSPDM_SPDM_10_11_SIGN_LITTLE_ENDIAN_RSA_ECDSA
+    uint32_t swap_endian_rsa = 0;
+    uint32_t swap_endian_ecdsa = 0;
```
jyao1, line 977: I think swap_endian_rsa and swap_endian_ecdsa should be a `bool` type.

library/spdm_crypt_lib/libspdm_crypt_asym.c
```diff
@@ -589,12 +589,49 @@ bool libspdm_sm2_dsa_verify_wrap (void *context, size_t hash_nid,
 }
 #endif
 
+static void libspdm_copy_signature_swap_endian(
+    uint8_t *endian_swapped_signature_buffer,
+    size_t endian_swapped_signature_buffer_size,
+    bool is_dual_buffer,
```
jyao1, line 595: `is_dual_buffer` brings confusion to me, and it takes me sometime to understand what it really means.


Maybe just define two functions: libspdm_copy_signature_swap_endian_for_ecdsa() and libspdm_copy_signature_swap_endian_for_rsa() ?

Then those functions can be reused by libspdm_asym_signature_swap_endian_if_necessary().

library/spdm_crypt_lib/libspdm_crypt_asym.c
richkong88: Agree.

richkong88:
> I think we need to update req_asym version as well, such as `libspdm_req_asym_sign`, `libspdm_req_asym_verify`.
> 
> Also, can we have some unit test or CI to ensure there will be no regression?

Will update req_asym versions and add unit tests.

library/spdm_crypt_lib/libspdm_crypt_asym.c
richkong88: Agree. I split into rsa and ecdsa versions and tried to make the code clearer.

include/library/spdm_lib_config.h
richkong88: Removed.

steven-bellock:
In general the macros should only be used if they impart a significant reduction in size. In this case they don't so I would like to see this set at runtime through `libspdm_set_data`. Also I don't think we need to differentiate between 1.0 and 1.1. Just lump them together. I can't imagine a Responder that would use LE for 1.0 and BE for 1.1.

richkong88:
> In general the macros should only be used if they impart a significant reduction in size. In this case they don't so I would like to see this set at runtime through `libspdm_set_data`. 

While I'm not generally opposed to a runtime decision, it seems that the libspdm_crypt_asym routines do not receive an "spdm_context" to extract the libspdm_set_data field? I think any "context" parameters are crypto contexts.  So making it runtime would involve changing a lot of interfaces (to take another parameter to indicate the runtime choic) which I was trying to minimize. Please correct me if I'm misunderstanding this.

> Also I don't think we need to differentiate between 1.0 and 1.1. Just lump them together. I can't imagine a Responder that would use LE for 1.0 and BE for 1.1.

I'm fine lumping 1.0 and 1.1 together. @jyao1 Jiewen, do you think lumping them together is ok?

jyao1:
If we use runtime mechanism, we can combine 1.0 and 1.1 because the integrator can make final decision based upon version.

But if we want to keep MACRO, then I think we should separate them to allow the flexibility, because the text in 1.0 and 1.1 are different.

richkong88:
Going to try a runtime mechanism. Closing this pull request.

PR #2330, opened 2023-08-30
CMakeLists.txt: Allow disabling EDDSA support from command line

alistair23:


alistair23:
The Windows failures are unrelated to this PR

alistair23:
That can be difficult to do when using build environments or packaging systems. Having a nicely defined argument makes it easier to disable these

steven-bellock:
> That can be difficult to do when using build environments or packaging systems. Having a nicely defined argument makes it easier to disable these

Why single out EDDSA though? Presumably this type of fix would be needed for all macros.

alistair23:
Because the EDDSA support is currently [not very good](https://github.com/DMTF/libspdm/commit/e87687d72688e980b929920b7d77dca26fff169e)

twilfredo (approved):

jyao1:
I don't have strong opinion. I am OK either way.

steven-bellock (approved):

PR #2332, opened 2023-08-30
Add runtime support for signing as little endian and dual endian verification.

richkong88:
@jyao1 @steven-bellock 

Looking for feedback on this runtime mechanism for supporting endian changes in spdm versions 1.0/11.  Not really excited about this, but I pass a "spdm_10_11_asym_algos_sign_little_endian" and "spdm_10_11_asym_algos_verify_dual_endian" fields around (retrieved from spdm_context) which are bitfields to indicate the asymmetric algorithms we should apply the endian changes too.  

I could have made it more generic (removing the spdm_10_11 from the name and allowing it to be applied for all versions). It might look less specific in that case and give a cleaner look and more flexibility to the integrator, but may be against specs if it applied to non 1.0/1.1 algorithms. 

Looking for your guys thoughts.

I'll add tests after we bottom out on a design.

### DIFF
--- a/library/spdm_crypt_lib/libspdm_crypt_asym.c
+++ b/library/spdm_crypt_lib/libspdm_crypt_asym.c
@@ -250,11 +250,117 @@ uint32_t libspdm_get_asym_signature_size(uint32_t base_asym_algo)
     }
 }
 
+static void libspdm_copy_signature_swap_endian_rsa(
+    uint8_t *dst,
+    size_t dst_size,
+    const uint8_t *src,
+    size_t src_size)
+{
+    /* RSA signature is a single buffer to be swapped */
+    size_t i;
+
+    if (src == dst) {
+        LIBSPDM_ASSERT(dst_size == src_size);
+
+        /* src and dst are same buffer. Swap in place. */
+
+        uint8_t byte;
+        for (i = 0; i < dst_size / 2; i++) {
+            byte = dst[i];
+            dst[i] = dst[dst_size - i - 1];
+            dst[dst_size - i - 1] = byte;
+        }
+    } else {
+        /* src and dst are different non-overlapping buffers */
+        LIBSPDM_ASSERT(dst_size >= src_size);
+        LIBSPDM_ASSERT((src < dst && src + src_size <= dst) ||
+                       (dst < src && dst + dst_size <= src));
+
+        for (i = 0; i < src_size; i++) {
+            dst[i] = src[src_size - i - 1];
+        }
+    }
+}
+
+static void libspdm_copy_signature_swap_endian_ecdsa(
+    uint8_t *dst,
+    size_t dst_size,
+    const uint8_t *src,
+    size_t src_size)
+{
+    /* ECDSA signature is actually 2 buffers (x & y) and each must be swapped individually */
+
+    size_t i;
+
+    if (src == dst) {
+        LIBSPDM_ASSERT(dst_size == src_size);
+
+        /* src and dst are same buffer. Swap ecdsa 2 internal buffers in place. */
+
+        size_t x_size;
+        size_t y_size;
+        uint8_t* x;
+        uint8_t* y;
+        uint8_t byte;
+
+        x_size = dst_size / 2;
+        y_size = x_size;
+
+        x = dst;
+        y = x + x_size;
+
+        for (i = 0; i < x_size / 2; i++) {
+            byte = x[i];
+            x[i] = x[x_size - i - 1];
+            x[x_size - i - 1] = byte;
+        }
+
+        for (i = 0; i < y_size / 2; i++) {
+            byte = y[i];
+            y[i] = y[y_size - i - 1];
+            y[y_size - i - 1] = byte;
+        }
+    } else {
+        /* src and dst are different non-overlapping buffers */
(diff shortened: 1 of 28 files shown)


jyao1:
This change updated the API, which breaks our rule - no API change in minor version.

@steven-bellock, do we must enable runtime support?

steven-bellock:
So now that I think about it more...

Presumably the signing implementation is only going to support one endianness per SPDM version. This is up to the implementer of `*asymsignlib` to decide and they already have `spdm_version_number_t spdm_version` as a parameter. For `os_stub` we can include a macro in https://github.com/DMTF/libspdm/blob/main/os_stub/mbedtlslib/include/mbedtls/config.h and whatever is there for OpenSSL to set the signing endianness.

For verification the runtime options should be (for SPDM version 1.0/1.1)
- Verify with BE only.
- Verify with LE only.
- Verify with both (try/fail).

The verification functions also know the SPDM version. I would assume the default would be BE-only, as that is consistent with 1.2/1.3.

richkong88:
So based on the comments so far, I'm envisioning these control macros in spdm_lib_config.h:
```
LIBSPDM_SPDM_10_RSA_SIGN_LITTLE_ENDIAN
LIBSPDM_SPDM_10_RSA_SIGN_BIG_ENDIAN
LIBSPDM_SPDM_10_RSA_VERIFY_LITTLE_ENDIAN
LIBSPDM_SPDM_10_RSA_VERIFY_BIG_ENDIAN
LIBSPDM_SPDM_10_RSA_VERIFY_DUAL_ENDIAN

LIBSPDM_SPDM_10_ECDSA_SIGN_LITTLE_ENDIAN
LIBSPDM_SPDM_10_ECDSA_SIGN_BIG_ENDIAN
LIBSPDM_SPDM_10_ECDSA_VERIFY_LITTLE_ENDIAN
LIBSPDM_SPDM_10_ECDSA_VERIFY_BIG_ENDIAN
LIBSPDM_SPDM_10_ECDSA_VERIFY_DUAL_ENDIAN

LIBSPDM_SPDM_11_RSA_SIGN_LITTLE_ENDIAN
LIBSPDM_SPDM_11_RSA_SIGN_BIG_ENDIAN
LIBSPDM_SPDM_11_RSA_VERIFY_LITTLE_ENDIAN
LIBSPDM_SPDM_11_RSA_VERIFY_BIG_ENDIAN
LIBSPDM_SPDM_11_RSA_VERIFY_DUAL_ENDIAN

LIBSPDM_SPDM_11_ECDSA_SIGN_LITTLE_ENDIAN
LIBSPDM_SPDM_11_ECDSA_SIGN_BIG_ENDIAN
LIBSPDM_SPDM_11_ECDSA_VERIFY_LITTLE_ENDIAN
LIBSPDM_SPDM_11_ECDSA_VERIFY_BIG_ENDIAN
LIBSPDM_SPDM_11_ECDSA_VERIFY_DUAL_ENDIAN
```
Seems like a lot of complexity, but ok.

I'm also wondering where the signing/verification given this information should be? 

Given that libspdm_crypt_asm.c calls the cryptlib_mbedtls/cryptlib_openssl to do the actual signing/verification, I see two options:

1. libspdm_crypt_asym.c assumes the cryptlib_*** works with big endian only and libspdm_crypt_asym.c must convert to little endian if defined by the control macros.  This leaves the cryptlib_*** untouched.

2. libspdm_crypt_asym.c passes to the cryptlib_*** what the endianness should be (based on the control switches) and the cryptlib_*** performs the action in that endianess within the cryptlib.  This would require an additional parameter to from library\spdm_crypt_lib\libspdm_crypt_asm.c to os\cryptlib_*** for the signing and verification functions.  Is it ok to do this?

I feel option 2 is clear/explicit and option 1 makes an assumption about the cryptlib endianess.

steven-bellock:
>So based on the comments so far, I'm envisioning these control macros in spdm_lib_config.h

No, no macros in `spdm_lib_config.h`. If there are macros put them in (something like) https://github.com/DMTF/libspdm/blob/main/os_stub/mbedtlslib/include/mbedtls/config.h. Also I don't think we need the 1.0/1.1 granularity. I can't imagine anyone implementing 1.0 BE signatures and 1.1 LE signatures on the same endpoint.

richkong88:
I created another Draft PR with just the framework for determining endianness to use (and not where to handle it).
https://github.com/DMTF/libspdm/pull/2340

Regarding"macro" in spdm_lib_config.h, I should have said #define. Just a switch. On or Off.

library/spdm_crypt_lib/libspdm_crypt_asym.c
```diff
@@ -250,11 +250,117 @@ uint32_t libspdm_get_asym_signature_size(uint32_t base_asym_algo)
...
+        LIBSPDM_ASSERT((src < dst && src + src_size <= dst) ||
+                       (dst < src && dst + dst_size <= src));
+
+        for (i = 0; i < src_size; i++) {
+            dst[i] = src[src_size - i - 1];
+        }
+    }
+}
+
+static void libspdm_copy_signature_swap_endian_ecdsa(
```
PrithviAPai, line 285: For ECDSA swap Endianness why is reverse of array NOT performed ?

richkong88:
Closing after receiving comments.

PR #2341, opened 2023-09-06
libspdm_rsp_algorithms: fixup spec conformance

twilfredo:
`libspdm_get_response_algorithms()` currently hardcodes the number of `RespAlgStruct` returned in the response regardless of what the requester has sent, and returns all 4 structs.

However, according to `SPDM 1.3.0 Table 22` it seems to imply that the response should include only those that are selected by the requester.

The following quotes from `SPDM 1.3.0 Table 22` seems to support the above interpretation:

-"Shall be the bit mask listing Responder-supported fixed algorithm requested by the Requester."

-"Number of bytes required to describe Requester-supported SPDM-enumerated fixed algorithms"

-"Number of *Requester-supported*, Responder-selected, extended algorithms"

-"Shall be the bit mask for indicating a Requester-supported, Responder-selected, SPDM-enumerated algorithm."

-"If present: shall be a Requester-supported, Responder-selected, extended algorithm. Responder shall select at most one extended algorithm."

This was discovered while testing against the kernel implementaion of the spdm requester, see [1] for more context.

This change ensures we return only the `RespAlgStructs` as supported by the requester.

[1] https://github.com/l1k/linux/issues/3


jyao1:
@twilfredo, thanks for the fix. I am ok with that.

Would you please fix CI failure? That is code format issue.

twilfredo:
> @twilfredo, thanks for the fix. I am ok with that.
> 
> Would you please fix CI failure? That is code format issue.

Fixed! thanks :)

jyao1 (approved):

steven-bellock:
The Requester needs similar treatment due to
> The algorithm structure table shall be present only if the Requester supports that AlgType.

Feel free to fix that in this pull request or a subsequent one, or don't fix that and file an issue.

There are issues in the specification with respect to the Responder. Let me get those sorted out before I comment on what the Responder's behavior can / should be.

steven-bellock:
We discussed this at the SPDM Working Group meeting. The outcome was that

>An algorithm structure table needs to be present only if the Responder supports that AlgType.

is incorrect in two different ways. It needs to be normative and it's only present if the Requester supports that `AlgType`. So it would become
>An algorithm structure table shall be present only if the Requester supports that AlgType.

So the Responder needs to check that, if a Requester sends an `AlgType` that at least one algorithm in `AlgSupported` or `AlgExternal` is selected. The Requester needs to check that `NEGOTIATE_ALGORITHMS.Param1 == ALGORITHMS.Param1`, and that the `AlgTypes` returned by the Responder are the `AlgTypes` sent by the Requester.

twilfredo:
> We discussed this at the SPDM Working Group meeting. The outcome was that
> 
> > An algorithm structure table needs to be present only if the Responder supports that AlgType.
> 
> is incorrect in two different ways. It needs to be normative and it's only present if the Requester supports that `AlgType`. 

Shouldn't it be if the Requester **and** the Responder supports that `AlgType` (in the response)? 

>So it would become
> 
> > An algorithm structure table shall be present only if the Requester supports that AlgType.
> 
> So the Responder needs to check that, if a Requester sends an `AlgType` that at least one algorithm in `AlgSupported` or `AlgExternal` is selected. 

What about a case for a given `AlgType` none of  `AlgSupported` or `AlgExternal` is supported by the responder? It would just mean that `AlgStructure` isn't returned back to the requester? So `ALGORITHMS.Param1` is now less than `NEGOTIATE_ALGORITHMS.Param1`

>The Requester needs to check that `NEGOTIATE_ALGORITHMS.Param1 == ALGORITHMS.Param1`, and that the `AlgTypes` returned by the Responder are the `AlgTypes` sent by the Requester.

With the above, wouldn't the requester check/assert `ALGORITHMS.Param1 <= NEGOTIATE_ALGORITHMS.Param1`. My thinking is that Responder may not support all that was requested.

twilfredo:
Updates: Fixed up the `/* -=[Process Request Phase]=- */` to align with the initial changes.

jyao1:
> > The Requester needs to check that `NEGOTIATE_ALGORITHMS.Param1 == ALGORITHMS.Param1`, and that the `AlgTypes` returned by the Responder are the `AlgTypes` sent by the Requester.
> 
> With the above, wouldn't the requester check/assert `ALGORITHMS.Param1 <= NEGOTIATE_ALGORITHMS.Param1`. My thinking is that Responder may not support all that was requested.

Do we must check those two? The spec is NOT super clear on that area.
Restrict speaking, they SHOULD be equal.
But for maximum compatibility, I recommend we don't check them. Smaller, bigger, and equal had better all be supported.
At least, I dont see it harmful.

twilfredo:
> 



> > > The Requester needs to check that `NEGOTIATE_ALGORITHMS.Param1 == ALGORITHMS.Param1`, and that the `AlgTypes` returned by the Responder are the `AlgTypes` sent by the Requester.
> > 
> > 
> > With the above, wouldn't the requester check/assert `ALGORITHMS.Param1 <= NEGOTIATE_ALGORITHMS.Param1`. My thinking is that Responder may not support all that was requested.
> 
> Do we must check those two? The spec is NOT super clear on that area. Restrict speaking, they SHOULD be equal. But for maximum compatibility, I recommend we don't check them. Smaller (responder remove the one from requester supported, but not responder supported), bigger (responder report some responder can support, but requester cannot support), and equal had better all be supported. **NOTE**: Even equal may still be insufficient, e.g. requester support type 2,3,4, but responder support type 2,4,5. As long as the finally result is same, I dont see it harmful.

Does the bigger case make sense though? In a case where "responder report some responder can support, but requester cannot support", this seems like wasted bandwidth as the requester has no use for this information as far as I understand. 

So it seems like the responder should report less (responder supports **some** requested types, but not all) or equal (if responder supports all requested types) to the requester supported, responder selected types. Which is this assertion from the requester `ALGORITHMS.Param1 <= NEGOTIATE_ALGORITHMS.Param1`.

As for this
> NOTE: Even equal may still be insufficient, e.g. requester support type 2,3,4, but responder support type 2,4,5.

In such a case the responder should only respond with 2 and 4, is my understanding from the spec. Thoughts?

jyao1:
The sentence in current SPDM spec is: "An algorithm structure table needs to be present only if **the Responder supports that AlgType** ."
That means, a responder may report what it can support, beyond what a requester requests.
I am not saying, that is a right way. (it is useless, I agree.)
I am just saying, that is a possible way, if people interprets the sentence in this way.

jyao1:
>> NOTE: Even equal may still be insufficient, e.g. requester support type 2,3,4, but responder support type 2,4,5.

> In such a case the responder should only respond with 2 and 4, is my understanding from the spec. Thoughts?

NO. 
We discussed and decision is to let responder report what **requester** has requested honestly.
That means even a responder support type 2,4,5, it should still report type 2,3,4. Of course the selected algo for 3 is zero.

twilfredo:
> The sentence in current SPDM spec is: "An algorithm structure table needs to be present only if **the Responder supports that AlgType** ." That means, a responder may report what it can support, beyond what a requester requests. I am not saying, that is a right way. (it is useless, I agree.) I am just saying, that is a possible way, if people interprets the sentence in this way.

Okay I follow you. In regards to what @steven-bellock mentioned earlier this sentence was **incorrect** and I'm assuming (?) the spec is being amended to instead say (as was mentioned in his prior response):

> An algorithm structure table shall be present only if the Requester supports that AlgType.

So do we wait for the spec to change before implementing the necessary changes for this?

twilfredo:
> > > NOTE: Even equal may still be insufficient, e.g. requester support type 2,3,4, but responder support type 2,4,5.
> 
> > In such a case the responder should only respond with 2 and 4, is my understanding from the spec. Thoughts?
> 
> NO. We discussed and decision is to let responder report what **requester** has requested honestly. That means even a responder support type 2,4,5, it should still report type 2,3,4. Of course the selected algo for 3 is zero.
> 
> That is why I said: it is not super clear. and we had better loosen the check to keep the compatibility. And there is no harm and no difference to see a type 3 with value 0 to indicate "unsupported" v.s. type 3 is absent. Also, if responder return type 5, it will be 0, too. No harm.

Got it, thanks for the clarification!

twilfredo:
@jyao1 Do you see any more necessary changes that need to be added to this PR in this case?

jyao1:
I am OK with that.

l1k:
@jyao1

<blockquote>
But for maximum compatibility, I recommend we don't check them. Smaller (responder remove the one from requester supported, but not responder supported), bigger (responder report some responder can support, but requester cannot support), and equal had better all be supported.
</blockquote>

It seems like a very bad idea to allow the responder to include extraneous RespAlgStructs which the requester didn't request: It forces the requester to allocate a buffer for the response which is sufficiently large to accommodate all possible RespAlgStructs that the responder could ever include. Right now (SPDM 1.2 and 1.3) that could amount to 4 RespAlgStructs. What if a future SPDM version defines further RespAlgStructs which requesters that comply only with an older SPDM version do not support? Do you want to allow responders to still include those extra RespAlgStructs even though the requester didn't allocate a sufficiently large buffer for them because it doesn't know about these newer RespAlgStruct types? If the buffer isn't large enough, the requester cannot compute the hash over the VCA transcript!

By all means don't do that. The responder should never ever be allowed to include RespAlgStructs which the requester didn't ask for. The spec is very clear about that ("Shall be the bit mask listing Responder-supported fixed algorithm <b>requested by the Requester</b>. -- SPDM 1.3.0 Table 22.)

Don't suddenly relax the spec by allowing additional RespAlgStructs to be included!

l1k:
You've already made my life as maintainer of the kernel SPDM implementation harder by allowing the responder to return more bytes than requested in a GET_CERTIFICATE response, thus forcing me to allocate a huge 64k buffer for the cert chain:
https://github.com/l1k/linux/blob/75dbba70720aafcb57783525b52c4c448f5cbe91/lib/spdm_requester.c#L1026

Please don't repeat that mistake!

If you make extraneous RespAlgStructs legal, I'll also have to allocate a larger than necessary buffer for the ALGORITHMS response. And I don't even know if it'll ever be sufficiently large because some future SPDM version might define further RespAlgStructs.

steven-bellock:
>Shouldn't it be if the Requester and the Responder supports that AlgType (in the response)?

>What about a case for a given AlgType none of AlgSupported or AlgExternal is supported by the responder? It would just mean that AlgStructure isn't returned back to the requester? So ALGORITHMS.Param1 is now less than NEGOTIATE_ALGORITHMS.Param1

`AlgCount` says (emphasis mine)
>Bit [7:4]. Number of Bytes required to describe **Requester supported** SPDM-enumerated
fixed algorithms (= FixedAlgCount).

and does not mention any Responder support, so the proposed normative clarification is consistent with that. 

>Do we must check those two? The spec is NOT super clear on that area.
Restrict speaking, they SHOULD be equal.
But for maximum compatibility, I recommend we don't check them. Smaller (responder remove the one from requester supported, but not responder supported), bigger (responder report some responder can support, but requester cannot support), and equal had better all be supported.

Yes, in terms of the Requester checking the Responder I don't think the specification is clear enough, so some leniency is fine. In fact I think this normative clarification should only go into the 1.4 specification, and Requesters that support < 1.4 would need to be flexible.

>By all means don't do that. The responder should never ever be allowed to include RespAlgStructs which the requester didn't ask for. The spec is very clear about that ("Shall be the bit mask listing Responder-supported fixed algorithm requested by the Requester. -- SPDM 1.3.0 Table 22.)

The problem is that in the older specifications that sentence was non-normative. 

>Bit mask listing Responder supported fixed algorithm requested by the Requester.

So there's not much normative language to restrict the Responder such that the Requester can make strict checks. Even in 1.3 `AlgCount` is conditional on the `Response AlgStructure field format` even existing, and the existence of those fields is currently governed by
>An algorithm structure table needs to be present only if the Responder supports that AlgType .

which is both non-normative and wrong (Responder should be Requester).

steven-bellock (approved):
I filed #2344 for `NEGOTIATE_ALGORITHMS`.

l1k:
@steve-bellock
If in SPDM <= 1.3, the responder is allowed to *omit* a RespAlgStruct despite the requester having sent a corresponding ReqAlgStruct, that's relatively harmless.

What I object to is allowing the responder to send an *extraneous* RespAlgStruct even though the responder sent no corresponding ReqAlgStruct. If that's allowed, the requester has to allocate a buffer for the response which can accommodate all RespAlgStructs that the responder could ever send. I don't see language in the spec that would allow such behavior. SPDM 1.3.0 Table 22 very clearly says that the RespAlgStructs must be "Requester supported".

jyao1:
2023/09/11 - discussed in SPDM Code TF meeting.
This update is for the responder. The responder will return what is requested from the requester. Nothing wrong here.
Approve to merge it.

PR #2345, opened 2023-09-08
Support big/little/both endians on signature verification on SPDM1.0/1.1

richkong88:
Changes to support runtime decision on endianness (BE\LE\Both) for signature verification on SPDM1.0/1.1.


jyao1:
@richkong88, thanks for the update.

I suggest to do the crypto handling (swap, double verification) in spdm_crypto_lib, to make the solution more scalable.
We will have more command requiring signing verification in the future 1.4.
I dont want to bring the burden to SPDM command writer. I suggest we HIDE the logic.

My comment for current API in spdm_crypto_lib, is to add a new API, dont change old API.
The old API implementation can invoke the new API, with default endianness.

richkong88:
@jyao1 Ok. I can do it the way you suggest. I prefer that too. I just did not want to change any APIs so I did it this way.

include/internal/libspdm_common_lib.h
```diff
@@ -565,6 +565,10 @@ typedef struct {
 #if LIBSPDM_FIPS_MODE
     libspdm_fips_selftest_context fips_selftest_context;
 #endif /* LIBSPDM_FIPS_MODE */
+
+    /* Endianness (BE/LE/Both) to use for signature verification on SPDM 1.0 and 1.1 */
```
jyao1, line 569: Maybe add comment say: This field is ignored on SPDM 1.2 and above.

richkong88:
Moving to different PR.

PR #2349, opened 2023-09-08
Enable different endians for signature verification on SPDM 1.0/1.1

richkong88:


include/library/spdm_crypt_lib.h
```diff
@@ -421,9 +447,17 @@ void libspdm_asym_free(uint32_t base_asym_algo, void *context);
...
     spdm_version_number_t spdm_version, uint8_t op_code,
     uint32_t base_asym_algo, uint32_t base_hash_algo,
-    void *context, const uint8_t *message,
-    size_t message_size, const uint8_t *signature,
-    size_t sig_size);
+    void *context,
+    const uint8_t *message, size_t message_size,
+    const uint8_t *signature, size_t sig_size);
+
+bool libspdm_asym_verify_ex(
```
richkong88, line 454: Introducing "_ex" functions which have an additional "flags" parameter, which you can use to pass the BE Only\LE Only\Both.

library/spdm_crypt_lib/libspdm_crypt_asym.c
```diff
@@ -757,24 +911,55 @@ bool libspdm_asym_verify(
         if (!result) {
             return false;
         }
-        return libspdm_asym_verify_wrap(context, hash_nid, base_asym_algo,
-                                        param, param_size,
-                                        message_hash, hash_size,
-                                        signature, sig_size);
+        result = false;
```
richkong88, line 914: I put the 
```
if (try_big_endian) {....}
if (result == false && try_little_endian) {...}
```
endian handling pattern here at this level since the spdm_version check is here is available here. 
Looking at this again, we could put this endian handling inside the `libspdm_asym_verify_wrap` function instead, to save some code and potential for missing it, but that would not have a spdm_version check at the same layer. I'm open to which ever what you prefer.

jyao1 (approved):

library/spdm_crypt_lib/libspdm_crypt_asym.c
```diff
@@ -526,6 +526,144 @@ static bool libspdm_asym_func_need_hash(uint32_t base_asym_algo)
...
+
+        uint8_t byte;
+        for (i = 0; i < dst_size / 2; i++) {
+            byte = dst[i];
+            dst[i] = dst[dst_size - i - 1];
+            dst[dst_size - i - 1] = byte;
+        }
+    }
+    else {
+        /* src and dst are different non-overlapping buffers */
```
steven-bellock, line 552: `!(src == dst)` doesn't mean that they are non-overlapping.

library/spdm_crypt_lib/libspdm_crypt_asym.c
```diff
@@ -526,6 +526,144 @@ static bool libspdm_asym_func_need_hash(uint32_t base_asym_algo)
...
+        }
+
+        for (i = 0; i < y_size / 2; i++) {
+            byte = y[i];
+            y[i] = y[y_size - i - 1];
+            y[y_size - i - 1] = byte;
+        }
+    }
+    else {
+        /* src and dst are different non-overlapping buffers */
```
steven-bellock, line 603: `!(src == dst)` doesn't mean that they are non-overlapping.

library/spdm_crypt_lib/libspdm_crypt_asym.c
```diff
@@ -526,6 +526,144 @@ static bool libspdm_asym_func_need_hash(uint32_t base_asym_algo)
...
+}
+
+static const uint32_t spdm_10_11_rsa_algos =
+    SPDM_ALGORITHMS_BASE_ASYM_ALGO_TPM_ALG_RSASSA_2048 |
+    SPDM_ALGORITHMS_BASE_ASYM_ALGO_TPM_ALG_RSAPSS_2048 |
+    SPDM_ALGORITHMS_BASE_ASYM_ALGO_TPM_ALG_RSASSA_3072 |
+    SPDM_ALGORITHMS_BASE_ASYM_ALGO_TPM_ALG_RSAPSS_3072 |
+    SPDM_ALGORITHMS_BASE_ASYM_ALGO_TPM_ALG_RSASSA_4096 |
+    SPDM_ALGORITHMS_BASE_ASYM_ALGO_TPM_ALG_RSAPSS_4096;
+
```
steven-bellock, line 643: These can go into `libspdm_copy_signature_swap_endian`.

library/spdm_crypt_lib/libspdm_crypt_asym.c
```diff
@@ -757,24 +911,55 @@ bool libspdm_asym_verify(
...
-                                        message_hash, hash_size,
-                                        signature, sig_size);
+        result = false;
+        if (try_big_endian) {
+            result = libspdm_asym_verify_wrap(context, hash_nid, base_asym_algo,
+                                              param, param_size,
+                                              message_hash, hash_size,
+                                              signature, sig_size);
+        }
+        if (result == false && try_little_endian) {
```
steven-bellock, line 921: `if (result == false && try_little_endian)`
->
`if (!result && try_little_endian)`

and in subsequent instances.

include/library/spdm_crypt_lib.h
```diff
@@ -404,6 +412,24 @@ bool libspdm_asym_get_public_key_from_der(uint32_t base_asym_algo,
  **/
 void libspdm_asym_free(uint32_t base_asym_algo, void *context);
 
+
```
steven-bellock, line 415: Everything in this header, besides the `SIG_SIZE` macros, can go into `internal/libspdm_crypt_lib.h`.

steven-bellock (changes requested):

library/spdm_crypt_lib/libspdm_crypt_asym.c
jyao1: can we just swap in place?

library/spdm_crypt_lib/libspdm_crypt_asym.c
jyao1: can we just swap in place?

library/spdm_crypt_lib/libspdm_crypt_asym.c
richkong88: @steven-bellock  Right. Perhaps I should have said this function only supports in-place same buffer swap, or completely different buffers (no overlap). That is what the asserts were supposed to do on line 544-555.

@jyao1 I can do in place, but I need to cast away const-ness, or change the parameter type in a higher up function to remove the const.  Plus it leaves the original buffer modified/swapped unless I swap it back. Feels like a bad side-effect to me, but I can see the benefit of saving the additional stack space, which can be considerable for some cases (512 bytes RSA 4072). I'm open to whatever you prefer.

library/spdm_crypt_lib/libspdm_crypt_asym.c
richkong88: @steven-bellock Sure.

library/spdm_crypt_lib/libspdm_crypt_asym.c
richkong88: Agree.

library/spdm_crypt_lib/libspdm_crypt_asym.c
richkong88: See same comment earlier.

include/library/spdm_crypt_lib.h
richkong88: @steven-bellock I think I only added the *SIG_SIZE macros for the PR.  Perhaps other general cleanup changes should go into a different PR.

library/spdm_crypt_lib/libspdm_crypt_asym.c
```diff
@@ -757,24 +911,55 @@ bool libspdm_asym_verify(
         if (!result) {
             return false;
         }
-        return libspdm_asym_verify_wrap(context, hash_nid, base_asym_algo,
-                                        param, param_size,
-                                        message_hash, hash_size,
-                                        signature, sig_size);
+        result = false;
+        if (try_big_endian) {
```
steven-bellock, line 915: It would be good to perform the try/fail only once. Once the endianness is determined then that can be stored in the context and would be used exclusively afterwards. Otherwise there is currently a performance hit for little endian, as it is always trying big endian first.

Zhiqiang520:
image

the function code of `libspdm_req_key_exchange.c` also changed, but there is no corresponding unit test to cover it.

richkong88:
> image
> 
> the function code of `libspdm_req_key_exchange.c` also changed, but there is no corresponding unit test to cover it.

I would argue that only change was going from libspdm_asm_verify(_hash) to the libspdm_asm_verif(_hash)_ex version which is already tested by other ULTs that I added.

library/spdm_crypt_lib/libspdm_crypt_asym.c
richkong88: @steven-bellock  This is an interesting optimization, In this case, should it try little endian then big endian? So the code would be something like...
```
result = false;
if (!context->try_little_endian_first && try_big_endian) {
    result = verify_in_big_endian(); 
     try_big_endian = false; 
}
if (!result && try_little_endian) {
    result = verify_in_little_endian(); 
    if (result) {
        context->try_little_endian_first = true
    }
}
if (!result && try_big_endian) {
    result = verify_in_big_endian(); 
    if (result) {
        context->try_little_endian_first = false;
    }
}
```

library/spdm_crypt_lib/libspdm_crypt_asym.c
richkong88: @jyao1 Waiting for your comments on this.

Zhiqiang520:
> > image
> > the function code of `libspdm_req_key_exchange.c` also changed, but there is no corresponding unit test to cover it.
> 
> I would argue that only change was going from libspdm_asm_verify(_hash) to the libspdm_asm_verif(_hash)_ex version which is already tested by other ULTs that I added.

ok, agree. Thanks.

library/spdm_crypt_lib/libspdm_crypt_asym.c
jyao1: I think it is OK to have it, to support additional buffer.

library/spdm_crypt_lib/libspdm_crypt_asym.c
jyao1: Once the endianness is determined after one success verification, then we should stick to it, in this SPDM connection.

No need to detect again.

library/spdm_crypt_lib/libspdm_crypt_asym.c
richkong88: @steven-bellock Updated the comments in my latest push to try to be clear. Overlap is not a supported usage model, except when they are the exact same buffer.

library/spdm_common_lib/libspdm_com_crypto_service.c
```diff
@@ -1047,13 +1047,43 @@ bool libspdm_verify_challenge_auth_signature(libspdm_context_t *spdm_context,
...
+         */
+        LIBSPDM_ASSERT(sizeof(uint32_t) ==
+                       sizeof(spdm_context->spdm_10_11_verify_signature_endian));
+        LIBSPDM_ASSERT(LIBSPDM_SPDM_10_11_ASYM_VERIFY_FLAG_LITTLE_ENDIAN_ONLY ==
+                       LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_LITTLE_ENDIAN_ONLY);
+        LIBSPDM_ASSERT(LIBSPDM_SPDM_10_11_ASYM_VERIFY_FLAG_LITTLE_BIG_ONLY ==
+                       LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_LITTLE_BIG_ONLY);
+        LIBSPDM_ASSERT(LIBSPDM_SPDM_10_11_ASYM_VERIFY_FLAG_BIG_OR_LITTLE_ENDIAN ==
+                       LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_BIG_OR_LITTLE_ENDIAN);
+
```
richkong88, line 1079: Note there are two different types here for specifying "endianness". One is in the context and is of type libspdm_spdm_10_11_verify_signature_endian_t. The other is within the spdm_crypt_lib and is a uint32_t with macros for each endian type. The values overlap for each type so you can use one in place of the other. I don't really like this (using 2 types for the same meaning) but I did it this way because the spdm_crypt_lib seems to be agnostic about types defined for the context.  Another option would be to put a single type spdm_10_11_endian types in spdm.h, but I wasn't sure if this ok this feels like a workaround for endianness missing in the spdm 1.0/1.1 spec. If you prefer adding a single type to spdm.h, I can do that.

include/library/spdm_crypt_lib.h
```diff
@@ -428,7 +428,6 @@ void libspdm_copy_signature_swap_endian(
 #define LIBSDPM_SPDM_10_11_ASYM_VERIFY_FLAG_BIG_ENDIAN_ONLY        (0x0)
 #define LIBSPDM_SPDM_10_11_ASYM_VERIFY_FLAG_LITTLE_ENDIAN_ONLY     (0x1)
 #define LIBSDPM_SPDM_10_11_ASYM_VERIFY_FLAG_BIG_OR_LITTLE_ENDIAN   (0x2)
-#define LIBSDPM_SPDM_10_11_ASYM_VERIFY_FLAG_ENDIAN_FLAGS           (0x3)
```
richkong88, line 431: These are defines specifying endianness in the spdm_crypt_lib. These are different than the context but have the same values so they can be used interchangeably with the context type. See my comment about this later in the review in libspdm_com_crypto_service.c.

include/library/spdm_crypt_lib.h
```diff
@@ -457,7 +458,7 @@ bool libspdm_asym_verify_ex(
     void* context,
     const uint8_t* message, size_t message_size,
     const uint8_t* signature, size_t sig_size,
-    uint32_t flags);
+    uint32_t *endian);
```
richkong88, line 461: The spdm_context is not passed into the libspdm_asm_verify functions that do endian swap and verification, so need an in\out parameter for endianness.

library/spdm_common_lib/libspdm_com_crypto_service.c
steven-bellock: What was wrong with defining the flags in `spdm_crypt_lib.h`?

steven-bellock (changes requested):

library/spdm_common_lib/libspdm_com_crypto_service.c
richkong88: Good point. I thought I would have to pull in another header file into spdm_context, but looks like I don't. I can make defines in spdm_crypt_lib.h and use them in the context files.

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -688,6 +688,19 @@ libspdm_return_t libspdm_set_data(void *spdm_context, libspdm_data_type_t data_t
             context->max_spdm_session_sequence_number = LIBSPDM_MAX_SPDM_SESSION_SEQUENCE_NUMBER;
         }
         break;
+    case LIBSPDM_DATA_SPDM_VERSION_10_11_VERIFY_SIGNATURE_ENDIAN:
+        if (data_size != sizeof(libspdm_spdm_10_11_verify_signature_endian_t)) {
+            return LIBSPDM_STATUS_INVALID_PARAMETER;
+        }
+        if (*(uint32_t*)data != LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_BIG_ENDIAN_ONLY &&
```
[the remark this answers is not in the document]
richkong88: Noted. But isn't there some compiler flag to change enum sizes to smaller than int (byte size)? I thought I had encountered that in the past before.

richkong88:
Closing this PR and putting everything so far into a new single set of changes.

PR #2357, opened 2023-09-14
Specify sequence number endianness

### LINKED ISSUE #2166: TLS1.3 compliance issue w.r.t sequence number handling (AEAD encryption and decryption)

steven-bellock:
Fix #2166.

### DIFF
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -134,6 +134,8 @@ typedef enum {
     /* For SPDM 1.0 and 1.1, allow signature verification in big, little, or both endians. */
     LIBSPDM_DATA_SPDM_VERSION_10_11_VERIFY_SIGNATURE_ENDIAN,
 
+    LIBSPDM_DATA_SESSION_SEQUENCE_NUMBER_ENDIAN,
+
     /* MAX */
     LIBSPDM_DATA_MAX
 } libspdm_data_type_t;
@@ -212,12 +214,21 @@ typedef enum {
     LIBSPDM_RESPONSE_STATE_MAX
 } libspdm_response_state_t;
 
-/* The Endian values apply only if the spdm version is 1.0/1.1.
+/* These macros apply only if the negotiated SPDM version is 1.0 or 1.1.
  * The default verification mode is big endian only. */
-#define LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_ENDIAN_BIG_ONLY        (0)
-#define LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_ENDIAN_LITTLE_ONLY     (1)
-#define LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_ENDIAN_BIG_OR_LITTLE   (2)
+#define LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_ENDIAN_BIG_ONLY 0
+#define LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_ENDIAN_LITTLE_ONLY 1
+#define LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_ENDIAN_BIG_OR_LITTLE 2
 
+/* Set the endianness of the AEAD sequence number. These macros apply only if the negotiated
+ * SPDM Secured Message version is 1.0 or 1.1.
+ * *_LITTLE and *_BIG immediately return an error on decryption failure.
+ * *_BOTH tries the opposite endianness on decryption failure.
+ * The default endianness is little endian for both encryption and decryption. */
+#define LIBSPDM_DATA_SESSION_SEQ_NUM_ENC_LITTLE_DEC_LITTLE 0
+#define LIBSPDM_DATA_SESSION_SEQ_NUM_ENC_BIG_DEC_BIG 1
+#define LIBSPDM_DATA_SESSION_SEQ_NUM_ENC_LITTLE_DEC_BOTH 2
+#define LIBSPDM_DATA_SESSION_SEQ_NUM_ENC_BIG_DEC_BOTH 3
 
 /*
  * +--------------------------+------------------------------------------+---------+
--- a/library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
+++ b/library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
@@ -6,6 +6,62 @@
 
 #include "internal/libspdm_secured_message_lib.h"
 
+static void generate_iv(uint64_t sequence_number, uint8_t *iv, const uint8_t *salt,
+                        size_t aead_iv_size, uint8_t endian)
+{
+    uint8_t iv_temp[LIBSPDM_MAX_AEAD_IV_SIZE];
+    size_t index;
+
+    /* Form the AEAD IV from the salt and the sequence number. */
+    libspdm_copy_mem(iv, LIBSPDM_MAX_AEAD_IV_SIZE, salt, aead_iv_size);
+
+    switch (endian) {
+    case LIBSPDM_DATA_SESSION_SEQ_NUM_ENC_LITTLE_DEC_LITTLE:
+    case LIBSPDM_DATA_SESSION_SEQ_NUM_ENC_LITTLE_DEC_BOTH:
+        /* If little-endian then the sequence number is zero-extended to the higher indices.
+         * The sequence number begins at the lowest index (0). */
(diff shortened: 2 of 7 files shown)


steven-bellock:
@jyao1 since re-try involves re-establishing sessions I've left it up to the Requester's Integrator to perform the retry. So for now they would
1. Attempt to establish session.
2. Session fails due to endianness mismatch.
3. Integrator changes endianness.
4. Attempt to re-establish session.

In particular the Responder is not required to send `DecryptError` back to the Requester, so this might be tough to automate entirely in libspdm.

jyao1:
Althougth auto detection is hard for the requester, the smart responder may choose to decrypt (little endian) then try decrypt (big endian), and cache the right endianness, right?

library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
```diff
@@ -397,8 +404,15 @@ libspdm_return_t libspdm_decode_secured_message(
...
+    /* Form the AEAD IV from the salt and the sequence number. */
     libspdm_copy_mem (iv, LIBSPDM_MAX_AEAD_IV_SIZE, salt, aead_iv_size);
-    data64 = libspdm_read_uint64((const uint8_t *)iv) ^ sequence_number;
+
+    if (secured_message_context->sequence_number_endian ==
+        LIBSPDM_DATA_SESSION_SEQUENCE_NUMBER_ENDIAN_BIG) {
+        data64 = libspdm_read_uint64((const uint8_t *)iv) ^ libspdm_le_to_be_64(sequence_number);
+    } else {
+        data64 = libspdm_read_uint64((const uint8_t *)iv) ^ sequence_number;
+    }
```
jyao1, line 415: I think we need support dual_decode, like how we support dual_verify.

For encoder, choose one of little and big, like how we support asym_sign (no change is required here)

steven-bellock:
> Althougth auto detection is hard for the requester, the smart responder may choose to decrypt (little endian) then try decrypt (big endian), and cache the right endianness, right?

Yeah it is a cheap operation. I'll add that.

library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
```diff
@@ -127,8 +127,15 @@ libspdm_return_t libspdm_encode_secured_message(
         return LIBSPDM_STATUS_SEQUENCE_NUMBER_OVERFLOW;
     }
 
+    /* Form the AEAD IV from the salt and the sequence number. */
     libspdm_copy_mem (iv, LIBSPDM_MAX_AEAD_IV_SIZE, salt, aead_iv_size);
-    data64 = libspdm_read_uint64((const uint8_t *)iv) ^ sequence_number;
+
+    if (secured_message_context->sequence_number_endian ==
+        LIBSPDM_DATA_SESSION_SEQUENCE_NUMBER_ENDIAN_BIG) {
+        data64 = libspdm_read_uint64((const uint8_t *)iv) ^ libspdm_le_to_be_64(sequence_number);
```
jyao1, line 135: If the intention is to simulate what TLS does.

Then I think we should XOR IV's most significant bit, not least significant bit.

See the picture at https://github.com/DMTF/libspdm/issues/2166#issuecomment-1608838324

jyao1:
@steven-bellock , reminder to resolve conflict.

library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
steven-bellock: Yes, that needs to be fixed.

steven-bellock:
Still need unit tests for big-endian and try-fail. May do that in a separate pull request.

include/library/spdm_common_lib.h
```diff
@@ -218,6 +220,12 @@ typedef enum {
 #define LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_ENDIAN_LITTLE_ONLY     (1)
 #define LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_ENDIAN_BIG_OR_LITTLE   (2)
 
+typedef enum {
+    LIBSPDM_DATA_SESSION_SEQ_NUM_ENC_LITTLE_DEC_LITTLE,
+    LIBSPDM_DATA_SESSION_SEQ_NUM_ENC_BIG_DEC_BIG,
+    LIBSPDM_DATA_SESSION_SEQ_NUM_ENC_LITTLE_DEC_BOTH,
+    LIBSPDM_DATA_SESSION_SEQ_NUM_ENC_BIG_DEC_BOTH
+} libspdm_sequence_number_endian_t;
```
jyao1, line 228: Can we use consistent way to handle session endianness and signature endianness?

I am ok with either way - enum or uint8_t.

steven-bellock: Sure, I can change it to `uint8_t` and use macros.

library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
```diff
@@ -6,6 +6,62 @@
 
 #include "internal/libspdm_secured_message_lib.h"
 
+static void generate_iv(uint64_t sequence_number, uint8_t *iv, const uint8_t *salt,
```
steven-bellock, line 9: @jyao1 please check that this makes sense.

jyao1: looks good to me.

@manishc88, would you please double check?

include/library/spdm_common_lib.h
```diff
@@ -214,10 +216,14 @@ typedef enum {
...
 /* The Endian values apply only if the spdm version is 1.0/1.1.
  * The default verification mode is big endian only. */
-#define LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_ENDIAN_BIG_ONLY        (0)
-#define LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_ENDIAN_LITTLE_ONLY     (1)
-#define LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_ENDIAN_BIG_OR_LITTLE   (2)
-
+#define LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_ENDIAN_BIG_ONLY 0
+#define LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_ENDIAN_LITTLE_ONLY 1
+#define LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_ENDIAN_BIG_OR_LITTLE 2
+
```
jyao1, line 222: /* The Endian values apply only if the secured spdm version is 1.0/1.1.
 * The default sequence number mode is enc little and dec little endian only. */

library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
steven-bellock: Right now the default behavior is, for both 277 1.0 and 1.1, that little endian is used exclusively. If the Integrator wants something different then they'll need to set it in the SPDM context.

include/library/spdm_common_lib.h
```diff
@@ -212,12 +214,20 @@ typedef enum {
...
+/* Set the endianness of the AEAD sequence number.
```
jyao1, line 223: please clarify this is for 1.0 and 1.1 only.

jyao1 (approved):

jyao1:
@steven-bellock , when the test case will be ready?

steven-bellock:
> @steven-bellock , when the test case will be ready?

Hopefully by the end of the week. Feel free to merge this and I'll obviously fix bugs if the tests find any.

PR #2358, opened 2023-09-14
Remove `SPDM_ERROR_CODE_INVALID_SESSION`

### LINKED ISSUE #2315: Remove `SPDM_ERROR_CODE_INVALID_SESSION`
That error code was introduced in SPDM 1.1 but it had a bug. It was ultimately decided to remove the error code from the specification in 1.1.1.

steven-bellock:
Fix #2315.


library/spdm_transport_mctp_lib/libspdm_mctp_common.c
```diff
@@ -250,7 +250,7 @@ libspdm_return_t libspdm_transport_mctp_decode_message(
             libspdm_get_secured_message_context_via_session_id(
                 spdm_context, *secured_message_session_id);
         if (secured_message_context == NULL) {
-            spdm_error.error_code = SPDM_ERROR_CODE_INVALID_SESSION;
+            spdm_error.error_code = SPDM_ERROR_CODE_DECRYPT_ERROR;
```
jyao1, line 253: You changed the behavior.

`SPDM_ERROR_CODE_INVALID_SESSION` means to ignore, not response.

`SPDM_ERROR_CODE_DECRYPT_ERROR` may return error message based upon `LIBSPDM_DATA_HANDLE_ERROR_RETURN_POLICY_DROP_ON_DECRYPT_ERROR`.

I suggest we may keep `SPDM_ERROR_CODE_INVALID_SESSION` to internal at first, because we still need a flag to tell other module, that it gets an invalid session.

steven-bellock: It seems like `SPDM_ERROR_CODE_INVALID_SESSION` needs to remain since it's a breaking change. I can move it to `library/spdm_secured_message_lib.h` though so that it's out of `spdm.h`.

PR #2369, opened 2023-09-22
[RFC] crypt_lib: allow partial certchain verification

twilfredo:
The new addition to the changed api allows us to verify a partial certificate chain, that is, one that does not contain a leaf certificate.

This maybe useful, for example, when the requester is doing a `SET_CERTIFICATE` and is loading in the immutable certificates. If using the AliasCert model, then this chain contains only the rootCA->...->deviceCA (the `immutable` certificates). Composed of only CAs (no leaf). In such a case, we can still use this function to validate most of the chain, but skip the leaf check.

### DIFF
--- a/include/library/spdm_crypt_lib.h
+++ b/include/library/spdm_crypt_lib.h
@@ -974,20 +974,23 @@ bool libspdm_get_dmtf_subject_alt_name(const uint8_t *cert, const size_t cert_si
 /**
  * This function verifies the integrity of certificate chain data without spdm_cert_chain_t header.
  *
- * @param  cert_chain_data       The certificate chain data without spdm_cert_chain_t header.
- * @param  cert_chain_data_size  Size in bytes of the certificate chain data.
- * @param  base_asym_algo        SPDM base_asym_algo
- * @param  base_hash_algo        SPDM base_hash_algo
- * @param  is_requester_cert     Is the function verifying requester or responder cert.
- * @param  is_device_cert_model  If true, the cert chain is DeviceCert model.
- *                               If false, the cert chain is AliasCert model.
+ * @param  cert_chain_data          The certificate chain data without spdm_cert_chain_t header.
+ * @param  cert_chain_data_size      size in bytes of the certificate chain data.
+ * @param  base_asym_algo            SPDM base_asym_algo
+ * @param  base_hash_algo            SPDM base_hash_algo
+ * @param  is_requester_cert         Is the function verifying requester or responder cert.
+ * @param  is_device_cert_model      If true, the cert chain is DeviceCert model;
+ *                                   If false, the cert chain is AliasCert model;
+ * @param  is_partial_chain          If true, the cert chain does not contain a leaf certificate;
+ *                                   If false, the cert chain is complete;
  *
- * @retval true  Certificate chain data integrity verification pass.
- * @retval false Certificate chain data integrity verification fail.
+ * @retval true  certificate chain data integrity verification pass.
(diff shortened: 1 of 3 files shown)


jyao1:
@twilfredo , I am not sure if this is needed.

SET_CERTIFICATE should only input Root->DeviceCert even in Alias mode, per my understanding.

twilfredo:
> @twilfredo , I am not sure if this is needed.
> 
> SET_CERTIFICATE should only input Root->DeviceCert even in Alias mode, per my understanding.

Yeah, so from a requester perspective when inputting this chain inclusive of only Root->DeviceCert, these are all CA's (and does not contain an end entity/leaf). So with that chain if you invoke `libspdm_verify_cert_chain_data()` it will fail at the leaf check. Because that chain doesn't contain a leaf certificate. 

With this change, you could pass the argument to skip only the leaf check and still validate rest of the chain. The use case I've come across is, a libspdm based requester loading the immutable certificate chain during a `SET_CERTIFICATE` setup, calling  `libspdm_verify_cert_chain_data()` before it sends the `SET_CERTIFICATE` request. 

What do you guys think?

steven-bellock:
> What do you guys think?

I think you should [file an issue](https://github.com/DMTF/libspdm/issues/new) and we can discuss it there.

twilfredo:
> > What do you guys think?
> 
> I think you should [file an issue](https://github.com/DMTF/libspdm/issues/new) and we can discuss it there.

No worries, I've created an issue https://github.com/DMTF/libspdm/issues/2372

jyao1:
It seems you only want to skip leaf cert check, right?

Another way is to use zero base_asym_algo and base_hash_algo to indicate leaf cert skipping.

twilfredo:
> It seems you only want to skip leaf cert check, right?
> 
> Another way is to use zero base_asym_algo and base_hash_algo to indicate leaf cert skipping.

hmm that seems hacky, also we will hit https://github.com/DMTF/libspdm/blob/137b6f7056be9df0295165bf1a30517814dfc5ad/library/spdm_crypt_lib/libspdm_crypt_cert.c#L398 if `base_asym_algo` is 0.

The chain of calls is:` libspdm_verify_cert_chain_data() ->libspdm_x509_certificate_check()-> libspdm_x509_common_certificate_check()-> libspdm_verify_cert_subject_public_key_info() ->libspdm_get_public_key_algo_OID()->libspdm_get_public_key_algo_OID_len()`

jyao1:
NOTE: We cannot change public API in minor release. As such, changing API in spdm_crypto_lib.h is not allowed.

Another way is to add a new API for cert chain verification only - cover the first half of `libspdm_verify_cert_chain_data()`.
But I am not sure if that will be useful, because you can call `libspdm_x509_get_cert_from_cert_chain()`, then `libspdm_x509_verify_cert_chain()` directly.

twilfredo:
> NOTE: We cannot change public API in minor release. As such, changing API in spdm_crypto_lib.h is not allowed.
> 

that's fair enough.

> Another way is to add a new API for cert chain verification only - cover the first half of `libspdm_verify_cert_chain_data()`. But I am not sure if that will be useful, because you can call `libspdm_x509_get_cert_from_cert_chain()`, then `libspdm_x509_verify_cert_chain()` directly.

That's a good point, from the perspective of someone implementing a requester. I think it would be easier/more readable to just have a new API in libspdm (less coupling for the requester side too). If you are open to this, I'm happy try and implement it.

twilfredo:
closing as we can continue this in the new PR here: https://github.com/DMTF/libspdm/pull/2373

Issue #2372, opened 2023-09-24
crypt_lib: allow partial certchain verification


twilfredo:
This is a continuation of #2369 

# Issue

Currently when a certificate chain is validated, it is required that the certificate chain is a complete certificate chain. That is, it contains an end-entify/leaf certificate at the end. The certificate validation takes place in `libspdm_verify_cert_chain_datalibspdm_verify_cert_chain_data()`. Although this makes sense for most use cases, we can't use this functions to validate a chain that is valid but is only partial (no leaf).

An example of such a case maybe, when a libspdm based requester is loading an immutable certificate (RootCA->...->DeviceCA **only**, no leaf) chain prior to a `SET_CERTIFICATE` request (`AliasCertificate Model`). Before sending the chain over to the responder, if the requester tries to validate the certificate chain locally using `libspdm_verify_cert_chain_datalibspdm_verify_cert_chain_data()`, it will fail (rightfully) during the leaf check. Which means in the case above, we can't use `libspdm_verify_cert_chain_datalibspdm_verify_cert_chain_data()` in it's current state.

# Proposed Change

A small modification could be made to `libspdm_verify_cert_chain_datalibspdm_verify_cert_chain_data()` skip the leaf check for such a case. See https://github.com/DMTF/libspdm/pull/2369 for proposed changes.

# Misc

Is there a better way to address the case above?

steven-bellock:
Validation of the `SET_CERTIFICATE` certificate chain currently bypasses `spdm_crypt_lib` and goes straight to `crypt_lib` : https://github.com/DMTF/libspdm/blob/137b6f7056be9df0295165bf1a30517814dfc5ad/library/spdm_responder_lib/libspdm_rsp_set_certificate.c#L13 If you'd like to bring that to `spdm_crypt_lib` it should probably be a new `libspdm_verify_cert_chain_data_ex` function.

twilfredo:
I was referring the validation of the certificate chain on the requester  (locally) before even the `libspdm_set_certificate()`. So we don't even bother sending something that isn't valid to the responder. 

Starting at  reading the cert chain file (on a requester device), then calling `libspdm_verify_cert_chain_data()` to verify things locally first (currently fail here), if OK, only then invoke `libspdm_set_certificate()`. 

Could that be in the form of a new function (in `libspdm_crypt_cert.c`) as mentioned here: https://github.com/DMTF/libspdm/pull/2369#issuecomment-1733766041 ?

[...]

PR #2373, opened 2023-09-26
spdm_crypt_lib: add new api for partial certificate chain verification

twilfredo:
## Overview:

This addresses the issues discussed in https://github.com/DMTF/libspdm/pull/2369 and https://github.com/DMTF/libspdm/issues/2372

## Testing:

Tested by running responder/requester unit tests locally. Tested the new function by connecting it to a libspdm requester, parsing the certificate chain with the new function, then attempting to do a SET_CERTIFICATE.

### DIFF
--- a/doc/api/requester_api.md
+++ b/doc/api/requester_api.md
@@ -392,7 +392,9 @@ TBD
 ---
 
 ### Description
-Sends `SET_CERTIFICATE` to deposit a certificate chain into a Responder.
+Sends `SET_CERTIFICATE` to deposit a certificate chain into a Responder. It is recommended that the certificate chain be 
+validated by `libspdm_verify_cert_chain_data_ex` prior to invoking `libspdm_set_certificate`. This ensures that the
+certificate chain is validated locally first.
 
 ### Parameters
 
--- a/include/library/spdm_crypt_lib.h
+++ b/include/library/spdm_crypt_lib.h
@@ -990,6 +990,32 @@ bool libspdm_verify_cert_chain_data(uint8_t *cert_chain_data, size_t cert_chain_
                                     bool is_requester_cert, bool is_device_cert_model);
 
 /**
+ * This function verifies the integrity of certificate chain data without spdm_cert_chain_t header for a
+ * SET_CERTIFICATE
+ *
+ * This function also allows for not-validating the leaf certificate in the chain. This maybe useful for a
+ * requester to validate the immutable certificates ahead of a `SET_CERTIFICATE` request
+ * (i.e using the AliasCertificate model).
+ *
+ * @param  cert_chain_data          The certificate chain data without spdm_cert_chain_t header.
+ * @param  cert_chain_data_size      size in bytes of the certificate chain data.
+ * @param  base_asym_algo            SPDM base_asym_algo
+ * @param  base_hash_algo            SPDM base_hash_algo
+ * @param  is_requester_cert         Is the function verifying requester or responder cert.
+ * @param  is_device_cert_model      If true, the cert chain is DeviceCert model;
+ *                                   If false, the cert chain is AliasCert model;
+ * @param  check_leaf_cert           If true, verify the entire certificate chain;
+ *                                   If false, skip leaf certificate validation;
+ *
+ * @retval true  certificate chain data integrity verification pass.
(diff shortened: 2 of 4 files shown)


doc/api/requester_api.md
```diff
@@ -392,7 +392,10 @@ TBD
 ---
 
 ### Description
-Sends `SET_CERTIFICATE` to deposit a certificate chain into a Responder.
+Sends `SET_CERTIFICATE` to deposit a certificate chain into a Responder. It's recommended
```
steven-bellock, line 395: >It's recommended for the `cert_chain` to be validated with `libspdm_verify_cert_chain_data_ex` by the requester, prior to invoking `libspdm_set_certificate`.

->
>It is recommended that the certificate chain be validated by `libspdm_verify_cert_chain_data_ex` prior to invoking `libspdm_set_certificate`.

include/library/spdm_crypt_lib.h
```diff
@@ -971,6 +971,32 @@ bool libspdm_get_dmtf_subject_alt_name(const uint8_t *cert, const size_t cert_si
                                        size_t *name_buffer_size,
                                        uint8_t *oid, size_t *oid_size);
 
+/**
```
steven-bellock, line 974: This should go immediately after `libspdm_verify_cert_chain_data`.

include/library/spdm_crypt_lib.h
```diff
@@ -971,6 +971,32 @@ bool libspdm_get_dmtf_subject_alt_name(const uint8_t *cert, const size_t cert_si
...
+ * This function also allows for not-validating the leaf certificate in the chain. This maybe useful for a
+ * requester to validate the immutable certificates ahead of a `SET_CERTIFICATE` request
+ * (i.e using the AliasCertificate model).
+ *
+ * @param  cert_chain_data          The certificate chain data without spdm_cert_chain_t header.
+ * @param  cert_chain_data_size      size in bytes of the certificate chain data.
+ * @param  base_asym_algo            SPDM base_asym_algo
+ * @param  base_hash_algo            SPDM base_hash_algo
+ * @param  is_requester_cert         Is the function verifying requester or responder cert.
+ * @param  is_device_cert_model      If true, the cert chain is DeviceCert model;
```
steven-bellock, line 987: SPDM 1.3 introduces a third certificate model, generic certificates. In anticipation of that this parameter should be something like `uint8_t cert_model`. @jyao1 do you agree?

steven-bellock (changes requested):

include/library/spdm_crypt_lib.h
jyao1: If we need consider SPDM 1.3, yes, change to uint8_t.

Also, we need to add spdm_version, right?

twilfredo: do you mean as another argument to this function? if so why?

twilfredo: > SPDM 1.3 introduces a third certificate model, generic certificates. In anticipation of that this parameter should be something like `uint8_t cert_model`. @jyao1 do you agree?

I feel like this should be a separate PR? There's a few other function that need to be updated, and probably the addition of new macros for each type also.

steven-bellock: >Also, we need to add spdm_version, right?

In this particular case no, but it might not hurt in the long-run.

>do you mean as another argument to this function?

No, it would replace `is_device_cert_model`.

>if so, why?

So we won't need `libspdm_verify_cert_chain_data_ex2` when we start to implement 1.3.

twilfredo: I can add this change, but we will need to change the api for libspdm_x509_set_cert_certificate_check() and all subsequent calls to accommodate the `uint8_t cert_model`. Is that still appropriate for this PR? or is it more appropriate to just translate the u8 into a bool for now within `_data_ex()`?

jyao1:
Cert model is a common issue. I filed https://github.com/DMTF/libspdm/issues/2505.

jyao1:
https://github.com/DMTF/libspdm/issues/2505 is fixed.
What do we do for this PR?

twilfredo:
We can close this if it's no longer relevant.

PR #2378, opened 2023-09-28
Remove 1.2.1 capability checks from Requester

### LINKED ISSUE #2375: Lack of backward compatibility for version 1.2.0
According to [SPDM specification 1.2.1](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf) new capabilities where added between 1.2.0 and 1.2.1:
`Added SET_CERT_CAP , CSR_CAP and CERT_INSTALL_RESET_CAP capabilities bits.`
This means, that any SPDM Responder implemented to 1.2.0 specification (there are already such implementors in the field) does not know about these capabilities and send `0` n CAPABILITIES response.

However, libspdm library always performs a check if this flag is set e.g. during [requester set certificate](https://github.com/DMTF/libspdm/blob/137b6f7056be9df0295165bf1a30517814dfc5ad/library/spdm_requester_lib/libspdm_req_set_certificate.c#L46):
```
if (!libspdm_is_capabilities_flag_supported(
        spdm_context, true, 0,
        SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_SET_CERT_CAP)) {
}
```

Thus, SPDM command SET_CERTIFICATE will always be blocked by libspdm when communicating with SPDM Responder with version 1.2.0 even if the responder supports SET_CERTIFICATE.

My proposal is to add additional check for 0x121 version and check the flag only if SPDM Responder version >= 0x121.
Proposal for SET_CERTIFICATE is [on the Pull Request from fork](https://github.com/DMTF/libspdm/pull/2374) - it will be probably necessary to also implement this for CSR_CAP and CERT_INSTALL_RESET_CAP where applicable.

steven-bellock:
Fix #2375.

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_get_csr.c
+++ b/library/spdm_requester_lib/libspdm_req_get_csr.c
@@ -52,11 +52,8 @@ static libspdm_return_t libspdm_try_get_csr(libspdm_context_t *spdm_context,
         return LIBSPDM_STATUS_UNSUPPORTED_CAP;
     }
 
-    if (!libspdm_is_capabilities_flag_supported(
-            spdm_context, true, 0,
-            SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_CSR_CAP)) {
-        return LIBSPDM_STATUS_UNSUPPORTED_CAP;
-    }
+    /* Do not check the Responder's CSR_CAP capability as it may be a 1.2.0 Responder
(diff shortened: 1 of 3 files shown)


steven-bellock:
Apparently none of these conditions were unit tested.

jyao1 (approved):

PR #2385, opened 2023-10-03
Add SESSION_SEQUENCE_NUMBER_ENDIAN

steven-bellock:
And revert the default value back to little-endian only.


jyao1 (approved):

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2166

PR #2396, opened 2023-10-09
Skip AEAD endianness detection

### LINKED ISSUE #2393: AEAD sequence number endianness detection does not work
If the endianness of the sequence number has not been determined, then on decryption libspdm will attempt to determine the endianness through https://github.com/DMTF/libspdm/blob/28571c8653f657ff26cdd6dfcd36414eff7dab86/library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c#L534-L568 This does not work in practice as the first sequence number is `0` which has the same byte layout in both big and little endian. Endianness can only be determined on the subsequent decryption, when the sequence number is `1`.

steven-bellock:
Fix #2393.


steven-bellock:
Well this is annoying. This only fails on Ubuntu / GCC / Release for some reason. Maybe undefined behavior or excessive compiler optimization.

jyao1:
Could you please squash the commit? I dont see the reason to split.

steven-bellock:
> Could you please squash the commit? I dont see the reason to split.

It is done.

PR #2397, opened 2023-10-09
Fix for wrong endian determination when signature is same for both

richkong88:
endians.

### DIFF
--- a/library/spdm_crypt_lib/libspdm_crypt_asym.c
+++ b/library/spdm_crypt_lib/libspdm_crypt_asym.c
@@ -689,6 +689,64 @@ void libspdm_copy_signature_swap_endian(
     }
 }
 
+/**
+ * libspdm_is_palindrome
+ * Checks to see if a buffer is a palindrome.
+ * If the buffer is a palindrone, it is the same for both endians,
+ * and therefore endianness cannot be determined.
+ **/
+bool libspdm_is_palindrome(const uint8_t* buf, size_t buf_size)
+{
+    size_t head;
+    size_t tail;
+
+    head = 0;
+    tail = buf_size - 1;
+
+    while (head < tail) {
+        if (buf[head] != buf[tail]) {
+            return false;
+        }
+        head++;
+        tail--;
+    }
+
+    return true;
+}
+
+bool libspdm_is_signature_buffer_palindrome(
+    uint32_t base_asym_algo, const uint8_t *buf, size_t buf_size)
+{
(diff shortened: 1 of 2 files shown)


library/spdm_crypt_lib/libspdm_crypt_asym.c
```diff
@@ -979,10 +1004,12 @@ bool libspdm_asym_verify_ex(
         }
     }
     if (try_big_endian && try_little_endian && result) {
-        if (little_endian_succeeded) {
-            *endian = LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_ENDIAN_LITTLE_ONLY;
-        } else {
-            *endian = LIBSPDM_SPDM_10_11_VERIFY_SIGNATURE_ENDIAN_BIG_ONLY;
+        if (!libspdm_is_palindrome(signature, sig_size)) {
```
jyao1, line 1007: This is for whole signature, right?

RSA is OK.

But ECDSA includes 2 parts - R and S.
Should we test R and S separately?

richkong88: You are correct. I forgot.

richkong88: @jyao1 I updated the changes for your review.

library/spdm_crypt_lib/libspdm_crypt_asym.c
```diff
@@ -689,6 +689,64 @@ void libspdm_copy_signature_swap_endian(
...
+
+    const uint32_t spdm_10_11_ecdsa_algos =
+        SPDM_ALGORITHMS_BASE_ASYM_ALGO_TPM_ALG_ECDSA_ECC_NIST_P256 |
+        SPDM_ALGORITHMS_BASE_ASYM_ALGO_TPM_ALG_ECDSA_ECC_NIST_P384 |
+        SPDM_ALGORITHMS_BASE_ASYM_ALGO_TPM_ALG_ECDSA_ECC_NIST_P521;
+
+    if (base_asym_algo & spdm_10_11_rsa_algos) {
+        return libspdm_is_palindrome(buf, buf_size);
+    } else if (base_asym_algo & spdm_10_11_ecdsa_algos) {
+        size_t x_size = buf_size / 2;
```
steven-bellock, line 736: `x_size` and `y_size` can be `const` as well.

jyao1 (approved):

steven-bellock (approved):

PR #2399, opened 2023-10-10
os_stub: cryptlib: CSR: Allow copying attributes from an existing cert

alistair23:
Section 754 of SPDM 1.3 states that:

```
    The CSR shall be populated with a combination of attributes
    provided by the Requester via the RequesterInfo field and other
    attributes contributed by the Responder itself.
```

libspdm currently supports populating the CSR based on the RequesterInfo, but we don't allow the Responder to add custom attributes.

This patch updates libspdm_gen_x509_csr() to pass in a base certificate that is used to provide CSR attributes. libspdm_gen_x509_csr() will copy the extensions from `base_cert` to the CSR.

This way devices can choose to apply the same extensions from a leaf certificate (or other certificate already on the device) or create a new cert just to pass the extensions in.

By copying the extensions from an existing cert the API should be generic between OpenSSL and MBedTLS implementations.

### DIFF
--- a/os_stub/cryptlib_mbedtls/pk/x509.c
+++ b/os_stub/cryptlib_mbedtls/pk/x509.c
@@ -1931,6 +1931,8 @@ bool libspdm_set_attribute_for_req(mbedtls_x509write_csr *req,
  * @param[in, out]      csr_pointer           For input, csr_pointer is buffer address to store CSR.
  *                                            For output, csr_pointer is address for stored CSR.
  *                                            The csr_pointer address will be changed.
+ * @param[in]           base_cert             An optional leaf certificate whose
+ *                                            extensions should be copied to the CSR
  *
  * @retval  true   Success.
  * @retval  false  Failed to gen CSR.
@@ -1939,7 +1941,8 @@ bool libspdm_gen_x509_csr(size_t hash_nid, size_t asym_nid,
                           uint8_t *requester_info, size_t requester_info_length,
                           bool is_ca,
                           void *context, char *subject_name,
-                          size_t *csr_len, uint8_t *csr_pointer)
+                          size_t *csr_len, uint8_t *csr_pointer,
+                          void *base_cert)
 {
     int ret;
     bool result;
@@ -1947,11 +1950,16 @@ bool libspdm_gen_x509_csr(size_t hash_nid, size_t asym_nid,
 
     mbedtls_x509write_csr req;
     mbedtls_md_type_t md_alg;
+    mbedtls_asn1_sequence extns;
+    mbedtls_asn1_sequence *next;
+    mbedtls_x509_buf buf;
+    mbedtls_x509_crt *cert;
     mbedtls_pk_context key;
 
     uint8_t pubkey_buffer[LIBSPDM_MAX_PUBKEY_DER_BUFFER_SIZE];
     uint8_t *pubkey_der_data;
     size_t pubkey_der_len;
+    size_t tag_len;
 
     /*basic_constraints: CA: false */
     #define BASIC_CONSTRAINTS_STRING_FALSE {0x30, 0x00}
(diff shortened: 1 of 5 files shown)


steven-bellock:
> This patch updates libspdm_gen_x509_csr() to pass in a base certificate that is used to provide CSR attributes.

Why not provide a sequence of attributes directly? Are there other fields in the certificate that are needed?

alistair23:
> Why not provide a sequence of attributes directly? Are there other fields in the certificate that are needed?

I couldn't think of a nice way to pass the attributes in directly. I can use openssl functions to create attributes and then pass an array of them in, but that then wouldn't work for mbedtls. So it's pretty clunky.

We could add libspdm helper functions to get around this. But it still ends up being clunky to allocate memory for the array of attributes and pass that all around. I found it cleaner to just say "here is a certificate, use the ones from this". I suspect in reality most Responders will use some existing certificate as a base anyway, so it should be pretty easy to implement.

os_stub/cryptlib_mbedtls/pk/x509.c
```diff
@@ -1961,6 +1964,11 @@ bool libspdm_gen_x509_csr(size_t hash_nid, size_t asym_nid,
     #define BASIC_CONSTRAINTS_STRING_TRUE {0x30, 0x03, 0x01, 0x01, 0xFF}
     uint8_t basic_constraints_true[] = BASIC_CONSTRAINTS_STRING_TRUE;
 
+    if (base_cert) {
```
steven-bellock, line 1967: `base_cert != NULL`

os_stub/cryptlib_openssl/pk/x509.c
```diff
@@ -2618,17 +2623,26 @@ bool libspdm_gen_x509_csr(size_t hash_nid, size_t asym_nid,
...
 
-    basic_constraints_ext = X509V3_EXT_conf_nid(
-        NULL, NULL, NID_basic_constraints,
-        is_ca ? "CA:TRUE" : "CA:FALSE");
-    if (!basic_constraints_ext) {
-        sk_X509_EXTENSION_free(exts);
-        ret = 0;
-        goto free_all;
+    /*set basicConstraints*/
+    if (base_cert) {
```
steven-bellock, line 2627: `base_cert != NULL`

steven-bellock (changes requested):

steven-bellock (approved):

os_stub/cryptlib_openssl/pk/x509.c
```diff
@@ -2618,17 +2623,26 @@ bool libspdm_gen_x509_csr(size_t hash_nid, size_t asym_nid,
...
-        ret = 0;
-        goto free_all;
+    /*set basicConstraints*/
+    if (base_cert != NULL) {
+        num_exts = X509_get_ext_count(base_cert);
+
+        for (int i = 0; i < num_exts; i++) {
+            sk_X509_EXTENSION_push(exts, X509_get_ext(base_cert, i));
+        }
+    } else {
```
Wenxing-hou, line 2633: I have a question: is it consistent for the `basic_constraints` between `base_cert` and `is_ca`?

alistair23: Are you worried about not appending a `CA:TRUE`/`CA:FALSE`?

I think it would be up to the caller to make sure the constraints supplied are correct

jyao1:
@alistair23, what test you have run to ensure the code is correct?

Could you please add some unit test?

alistair23:
I have been using SPDM-Utils for testing. You can see the commit adding support to SPDM-Utils here: https://github.com/westerndigitalcorporation/spdm-utils/commit/86e9add4d3c67ec9c059c1d160f7f6511be5cbf0

I can add some unit tests as well

alistair23:
The `Style Compliance` failure is a false positive. Let me know what I should do here. Should I change it to be incorrect, but pass the tests or leave it correct with a failing test case

alistair23:
Unit tests added, it's implemented and passing for OpenSSL and mbedtls now as well!

jyao1:
> The Style Compliance failure is a false positive. Let me know what I should do here. Should I change it to be incorrect, but pass the tests or leave it correct with a failing test case

Please just follow the suggestion to make it pass, so far.

alistair23:
Ok! All tests including the formatting tests are now passing

steven-bellock (approved):

PR #2402, opened 2023-10-10
Added shared library and pkg-config support for libspdm

uvinrg:
Changes:
- added pkg-config libspdm.pc.in file for cmake to use as template

CMakeLists.txt changes:
- added BUILD_SHARED_LIB cmake parameter (ON/OFF) to enable/disable building of libspdm.so in addition to the static libraries
- added two build paths for libspdm.so that generate distinct pkg-config pc files, for runtime linking with either mbedtls or openssl

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,6 @@
 cmake_minimum_required(VERSION 2.8.12)
 
 project("libspdm" C)
-SET(LIB_NAME "spdm")
-SET(CMAKE_PROJECT_NAME "LibSPDM")
-SET(CMAKE_PROJECT_DESCRIPTION "LibSPDM Shared Library")
 FILE(STRINGS ${CMAKE_CURRENT_SOURCE_DIR}/VERSION.md CMAKE_INPUT_PROJECT_VERSION)
 string(REPLACE " " "_" CMAKE_PROJECT_VERSION ${CMAKE_INPUT_PROJECT_VERSION})
 
@@ -21,7 +18,7 @@ SET(CMAKE_BUILD_TYPE ${TARGET} CACHE STRING "Choose the target of build: Debug R
 SET(CRYPTO ${CRYPTO} CACHE STRING "Choose the crypto of build: mbedtls openssl" FORCE)
 SET(GCOV ${GCOV} CACHE STRING "Choose the target of Gcov: ON  OFF, and default is OFF" FORCE)
 SET(STACK_USAGE ${STACK_USAGE} CACHE STRING "Choose the target of STACK_USAGE: ON  OFF, and default is OFF" FORCE)
-SET(BUILD_SHARED_LIB ${BUILD_SHARED_LIB} CACHE STRING "Choose if shared library libspdm.co should be built: ON OFF, and default is OFF" FORCE)
+SET(BUILD_LINUX_SHARED_LIB ${BUILD_LINUX_SHARED_LIB} CACHE STRING "Choose if libspdm shared library should be built for linux: ON OFF, and default is OFF" FORCE)
 
 if(NOT GCOV)
     SET(GCOV "OFF")
@@ -31,8 +28,8 @@ if(NOT STACK_USAGE)
     SET(STACK_USAGE "OFF")
 endif()
 
-if(NOT BUILD_SHARED_LIB)
-    SET(BUILD_SHARED_LIB "OFF")
+if(NOT BUILD_LINUX_SHARED_LIB)
+    SET(BUILD_LINUX_SHARED_LIB "OFF")
 endif()
 
 SET(LIBSPDM_DIR ${PROJECT_SOURCE_DIR})
@@ -985,7 +982,11 @@ else()
         endif()
     endif()
 
-    if (BUILD_SHARED_LIB STREQUAL "ON")
+    if ((BUILD_LINUX_SHARED_LIB STREQUAL "ON") AND (CMAKE_SYSTEM_NAME MATCHES "Linux"))
+        SET(LIB_NAME "spdm")
+        SET(CMAKE_PROJECT_NAME "libspdm")
+        SET(CMAKE_PROJECT_DESCRIPTION "libspdm shared library")
+
         if(CRYPTO STREQUAL "mbedtls")
             SET(CRYPTO_DEPS "-lmbedtls -lmbedx509 -lmbedcrypto")
             add_library(${LIB_NAME} SHARED 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -987,51 +987,59 @@ else()
         SET(CMAKE_PROJECT_NAME "libspdm")
         SET(CMAKE_PROJECT_DESCRIPTION "libspdm shared library")
 
+        SET(CMAKE_SKIP_BUILD_RPATH FALSE)
+        SET(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
+        SET(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
+        SET(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+
+        add_library(${LIB_NAME} SHARED 
+            $<TARGET_OBJECTS:spdm_common_lib>
+            $<TARGET_OBJECTS:spdm_crypt_lib>
+            $<TARGET_OBJECTS:spdm_requester_lib>
+            $<TARGET_OBJECTS:spdm_responder_lib>
+            $<TARGET_OBJECTS:spdm_secured_message_lib>
+            $<TARGET_OBJECTS:spdm_transport_mctp_lib>
+            $<TARGET_OBJECTS:spdm_transport_pcidoe_lib>
+        )
+
+        add_library(${LIB_NAME}_platform SHARED 
+            $<TARGET_OBJECTS:debuglib>
+            $<TARGET_OBJECTS:malloclib>
+            $<TARGET_OBJECTS:platform_lib>
+            $<TARGET_OBJECTS:rnglib>
+            $<TARGET_OBJECTS:memlib>
+        )
+
+        add_library(${LIB_NAME}_crypto SHARED 
+            $<TARGET_OBJECTS:spdm_crypt_ext_lib>
+            $<TARGET_OBJECTS:cryptlib_${CRYPTO}>
+        )
+
         if(CRYPTO STREQUAL "mbedtls")
             SET(CRYPTO_DEPS "-lmbedtls -lmbedx509 -lmbedcrypto")
-            add_library(${LIB_NAME} SHARED 
-                                            $<TARGET_OBJECTS:cryptlib_${CRYPTO}>
-                                            $<TARGET_OBJECTS:debuglib>
-                                            $<TARGET_OBJECTS:malloclib>
-                                            $<TARGET_OBJECTS:platform_lib>
-                                            $<TARGET_OBJECTS:rnglib>
-                                            $<TARGET_OBJECTS:spdm_common_lib>
-                                            $<TARGET_OBJECTS:spdm_crypt_ext_lib>
-                                            $<TARGET_OBJECTS:spdm_crypt_lib>
-                                            $<TARGET_OBJECTS:spdm_device_secret_lib_sample>
-                                            $<TARGET_OBJECTS:spdm_requester_lib>
-                                            $<TARGET_OBJECTS:spdm_responder_lib>
-                                            $<TARGET_OBJECTS:spdm_secured_message_lib>
-                                            $<TARGET_OBJECTS:spdm_transport_mctp_lib>
-                                            $<TARGET_OBJECTS:spdm_transport_pcidoe_lib>
-                                            $<TARGET_OBJECTS:memlib>
-                                            )
-            
+            TARGET_LINK_LIBRARIES(${LIB_NAME}_crypto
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -974,4 +984,53 @@ else()
             endif()
         endif()
     endif()
+
+    if (BUILD_SHARED_LIB STREQUAL "ON")
+        if(CRYPTO STREQUAL "mbedtls")
+            SET(CRYPTO_DEPS "-lmbedtls -lmbedx509 -lmbedcrypto")
+            add_library(${LIB_NAME} SHARED 
+                                            $<TARGET_OBJECTS:cryptlib_${CRYPTO}>
```
steven-bellock, line 992: This is open for debate, but as a shared library `libspdm.so` should only include the files in `libspdm/library/`
- spdm_common_lib
- spdm_crypt_lib
- spdm_requester_lib
- spdm_responder_lib
- spdm_secured_message_lib
- spdm_transport_mctp_lib
- spdm_transport_pcidoe_lib

Functionality beyond that would go in a separate shared library or be statically linked into the executable binary.

os_stub/spdm_device_secret_lib_sample/lib.c
```diff
@@ -1495,6 +1495,15 @@ uint8_t m_libspdm_bin_str0[0x11] = {
     0x64, 0x65, 0x72, 0x69, 0x76, 0x65, 0x64,
 };
 
+void libspdm_dump_hex_str_sample(const uint8_t *buffer, size_t buffer_size)
```
steven-bellock, line 1498: Please make a separate pull request for this change.

steven-bellock (changes requested):

CMakeLists.txt
```diff
@@ -16,6 +21,7 @@ SET(CMAKE_BUILD_TYPE ${TARGET} CACHE STRING "Choose the target of build: Debug R
 SET(CRYPTO ${CRYPTO} CACHE STRING "Choose the crypto of build: mbedtls openssl" FORCE)
 SET(GCOV ${GCOV} CACHE STRING "Choose the target of Gcov: ON  OFF, and default is OFF" FORCE)
 SET(STACK_USAGE ${STACK_USAGE} CACHE STRING "Choose the target of STACK_USAGE: ON  OFF, and default is OFF" FORCE)
+SET(BUILD_SHARED_LIB ${BUILD_SHARED_LIB} CACHE STRING "Choose if shared library libspdm.co should be built: ON OFF, and default is OFF" FORCE)
```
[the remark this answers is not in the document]
jyao1: Good point. Windows shared lib is DLL.

I suggest we remove the `libspdm.so` term to avoid confusing. Just saying `shared library` is enough.

BTW: Does this PR support shared lib in windows?

os_stub/spdm_device_secret_lib_sample/lib.c
jyao1: Agree with Steven. Rule is: each PR should only address one problem.
I am not sure why this change is relevant to shared library.

Would you please clarify: Why we cannot use old function?

CMakeLists.txt
```diff
@@ -1,6 +1,11 @@
 cmake_minimum_required(VERSION 2.8.12)
 
 project("libspdm" C)
+SET(LIB_NAME "spdm")
+SET(CMAKE_PROJECT_NAME "LibSPDM")
```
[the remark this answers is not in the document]
uvinrg: Will do, thank you

CMakeLists.txt
```diff
@@ -1,6 +1,11 @@
...
+SET(CMAKE_PROJECT_DESCRIPTION "LibSPDM Shared Library")
```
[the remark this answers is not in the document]
uvinrg: Will change it, thanks

CMakeLists.txt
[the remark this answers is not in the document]
uvinrg: Hi, the change was built and tested on linux only. Pkg-config support is only available on linux also, as far as I know. I do not have the know-how to do the integration on Windows and I also don't have access to a windows machine and windows integration code to verify it works.
Are you guys ok with making this PR to enable only Linux shared library and pkg-config support? I can rename BUILD_SHARED_LIB to BUILD_LINUX_SHARED_LIB.

CMakeLists.txt
uvinrg: I agree with your comments, but in this scenario we are trying to build a functional shared library that can be included and used directly in any libspdm integration. So I think debugging code, memory allocation, cryptographic library hooks, rng generator and time management should be part of the built shared library in my opinion. 
But you guys are the owners, so I can also create two shared libraries if you think that works better, like libspdm_main.so with what you mentioned and libspdm_platform.so that includes all the required primitives to be able to instantiate libspdm in an integration.
What is your opinion?

os_stub/spdm_device_secret_lib_sample/lib.c
uvinrg: This change is required only for the shared library PR, that is why I included it here.
The regular libspdm_dump_hex_str() method is defined in multiple places, but only inside the "unit_test" folder of libspdm, so it is not available for code outside the scope of unit tests, like spdm_secret_lib_sample() is.
Regular integrations of libspdm with static libraries usually include at least one of the unit test libraries that contain an implementation of the method, so it is not an issue there.
Considering this, is it ok to keep this change as part of this PR?

uvinrg:
Hi, thank you for reviewing my change and for the feedback. I added some comments above and fixed some of the issues raised.

CMakeLists.txt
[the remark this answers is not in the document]
jyao1: I am OK to limit the support to Linux.

We just need better documentation to tell people.

os_stub/spdm_device_secret_lib_sample/lib.c
jyao1: Please help me understand why os_stub/spdm_device_secret_lib_sample/lib.c is  impacted.

This is just a **sample**. Should not be used by other people, except unit test.

Do you want to create a shared lib for spdm_device_secret_lib_sample ?

CMakeLists.txt
jyao1: Why not use 1:1 mapping for static lib and dynamic lib? Why we have to create 1 .so file including everything?

steven-bellock:
@uvinrg what is the use case for this? Will this primarily be used to construct SPDM Requesters, Responders, or both?

CMakeLists.txt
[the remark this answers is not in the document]
uvinrg: Hi, I will update the documentation as well in the next commit.

CMakeLists.txt
uvinrg: Hi, we want to package libspdm as a regular shared library on linux distributions, for future work around SPDM attestation happening in a regular linux distribution (libspdm can act as either responder or requester, depending on target platform and usecase).
To that end we want to update the libspdm project to have an option to build a shared library like many other libraries in a linux operating system.
I understand your concern regarding need for customization of different components of a libspdm target, such as custom crypto library or HAL.
To that end I will create a new commit with a 1:1 mapping for static lib to shared lib, with an overarching libspdm.so shared library that depends on all the other libspdm libraries to run (using TARGET_LINK_LIBRARIES in CMakeLists.txt). 
This way, from the user point of view, a simple linking with libspdm.so is enough to compile the program, as long as all the other libraries are built and available in the system to use at runtime, as libspdm.so itself will internally depend on the other libraries.
Does that sound ok?

os_stub/spdm_device_secret_lib_sample/lib.c
uvinrg: You are right, of course. SPDM_device_secret_lib_sample should not be part of the shared library, and it can be implemented in the application using the library. So there is no need for the change and I will remove it from shared library target in the next commit.

uvinrg:
Hi Steven, the intent is to package libspdm as a regular shared library on linux distributions for future work around SPDM attestation. Depending on target platform and usecase, the libspdm shared library can act as either requester or responder.
Thank you for all the feedback, I responded to some of the comments and concerns raised and I will make the changes in the next commit.

CMakeLists.txt
jyao1: Thanks @uvinrg . Sounds good to me to have 1:1 mapping.

BTW: For Linux use case, do you anticipate requester only? or responder as well?
I have confidence for the requester, such as attestation use case. That is standard CMA flow.
But I am not sure about the responder, which has more strict requirement on security (e.g. private key management) and platform (e.g. measurement reporting, watchdog for heartbeat). It would be good to know what your thought is.

uvinrg: Hi, while we initially target requester usecase, responder as a shared library will be added later as well, as it is useful even if just for testing a system. Regarding security, there is only so much a library, shared or statically linked can do to protect its private key. Security of the running environment of the responder should be provided by the hardware platform and the operating system, so we don't see the shared library linking as an issue for the responder.
For the 1:1 mapping I have an issue as CMake does not allow both a shared and static taget with the same name to be generated. As such, if shared libraries are enabled, I have to disable building of the static libraries, which will also disable the building of the unit_tests, as they rely on the static libraries. Or change the name of the shared libraries, for example libspdm_requester_lib.a and libspdm_requester_lib_sh.so
What is your opinion?

steven-bellock: > Or change the name of the shared libraries, for example libspdm_requester_lib.a and libspdm_requester_lib_sh.so.

Yes, appending the `_sh` suffix should be fine.

steven-bellock: Yeah I'm fine with (at least) two libraries. The platform library would include
- debuglib
- malloclib
- memlib
- rnglib
- platformlib

The cryptography portion could either be part of the platform library or its own library.

Note that some of these are considered sample code, and since the intent is that they be used as production code they should be cleaned up and hardened in subsequent pull requests.

uvinrg:
Sorry for the delay, I committed the new change with three shared libraries, one for libspdm, one for the platform code and one for the crypto libraries interfaces. Thank you.

doc/build.md
```diff
@@ -263,6 +263,27 @@ cmake -DARCH=<arch> -DTOOLCHAIN=NONE -DTARGET=<Debug|Release> -DCRYPTO=<mbedtls|
 make
 ```
 
+### Linux Shared Library Builds
+
+Supports shared libraries building and pkg-config '.pc' file generation and installation.
+Will generate:
+ - libspdm.so - main library code, all subprojects from "library" folder
+ - libspdm_platform.so - subprojects in the "os_stub" folder related to platform code, like memory allocation, random number generator, etc.
+  - libspdm_crypto.so - cryptography related code for libspdm linking to either mbedtls or openssl libraries
```
steven-bellock, line 272: > cryptography related code for libspdm to dynamically link to either Mbed TLS or OpenSSL shared libraries.

doc/build.md
```diff
@@ -263,6 +263,27 @@ cmake -DARCH=<arch> -DTOOLCHAIN=NONE -DTARGET=<Debug|Release> -DCRYPTO=<mbedtls|
...
 ```
 
+### Linux Shared Library Builds
+
+Supports shared libraries building and pkg-config '.pc' file generation and installation.
+Will generate:
+ - libspdm.so - main library code, all subprojects from "library" folder
+ - libspdm_platform.so - subprojects in the "os_stub" folder related to platform code, like memory allocation, random number generator, etc.
+  - libspdm_crypto.so - cryptography related code for libspdm linking to either mbedtls or openssl libraries
+All three libraries are required for an application that uses libspdm, but the developer is free to implement their own versions of libspdm_platform or libspdm_crypto libraries and link with their implementations.
```
steven-bellock, line 273: `developer` -> `Integrator`

CMakeLists.txt
```diff
@@ -974,4 +981,65 @@ else()
...
+            $<TARGET_OBJECTS:spdm_responder_lib>
+            $<TARGET_OBJECTS:spdm_secured_message_lib>
+            $<TARGET_OBJECTS:spdm_transport_mctp_lib>
+            $<TARGET_OBJECTS:spdm_transport_pcidoe_lib>
+        )
+
+        add_library(${LIB_NAME}_platform SHARED 
+            $<TARGET_OBJECTS:debuglib>
+            $<TARGET_OBJECTS:malloclib>
+            $<TARGET_OBJECTS:platform_lib>
```
steven-bellock, line 1008: FYI : the [watchdog timers](https://github.com/DMTF/libspdm/blob/main/os_stub/platform_lib/watchdog.c) currently don't do anything, so you should file an issue to have these implemented on Linux.

os_stub/spdm_device_secret_lib_sample/lib.c
steven-bellock, line 1498: This file can be removed from the pull request.

steven-bellock (changes requested):

CMakeLists.txt
uvinrg: I created an issue for this, thank you.

steven-bellock (approved):

Kendidi:
Is it possible to add support to build shared libraries for Windows platform?  Thanks a lot in advance.

uvinrg:
> Is it possible to add support to build shared libraries for Windows platform? Thanks a lot in advance.

Unfortunately, I only used libspdm under linux, so I don't have the know-how. Feel free to add the Windows support if you need it.

Issue #2282, opened 2023-08-02
[SPDM 1.3] CHALLENGE: Add context field


jyao1:

l1k:
The <a href="https://lore.kernel.org/all/cover.1695921656.git.lukas@wunner.de/">Linux kernel's SPDM requester</a> supports SPDM 1.3 and current master of libspdm claims that it supports SPDM 1.3 as well, but the missing `RequesterContext` field in CHALLENGE_AUTH responses means that the kernel will reject the response:

```
[   10.475246] nvme 0000:00:03.0: SPDM: Truncated challenge_auth response
```

So libspdm's current master is broken when used against the Linux kernel. libspdm should not select version 1.3 if it doesn't actually support it.

This issue was reported by @tsarvela. Heads up @twilfredo @alistair23 !

twilfredo:
> The [Linux kernel's SPDM requester](https://lore.kernel.org/all/cover.1695921656.git.lukas@wunner.de/) supports SPDM 1.3 and current master of libspdm claims that it supports SPDM 1.3 as well, but the missing `RequesterContext` field in CHALLENGE_AUTH responses means that the kernel will reject the response:
> 
> ```
> [   10.475246] nvme 0000:00:03.0: SPDM: Truncated challenge_auth response
> ```
> 
> So libspdm's current master is broken when used against the Linux kernel. libspdm should not select version 1.3 if it doesn't actually support it.
> 
> This issue was reported by @tsarvela. Heads up @twilfredo @alistair23 !

I had a quick go at trying to reproduce this today. But it seemed to work... I was testing on QEMU with libspdm@2565e47321 based responder and kernel on linux@b6c714ff335e (latest on doe branch). 

With the usual process of adding the certificates to the key-ring then authenticating.

```
nvme 0000:00:03.0: CMA: authenticated with certificate slot 0
```
Was this addressed or did I miss something? Looks like the issue is still open too...

l1k:
@twilfredo Hm, perhaps you've specified `--ver 1.2` on the `spdm_responder_emu` command line? I can still reproduce the issue with f029916052f1. Reverting ea5c9246ed4e avoids the issue. But I see @jyao1 is on it. Thanks for testing!

twilfredo:
[...]

PR #2420, opened 2023-10-31
Add 1.3 support for challenge

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2282

Validated with spdm-emu.
The requester_context can be filled as expected.

### DIFF
--- a/include/industry_standard/spdm.h
+++ b/include/industry_standard/spdm.h
@@ -21,6 +21,7 @@
 
 #define SPDM_NONCE_SIZE 32
 #define SPDM_RANDOM_DATA_SIZE 32
+#define SPDM_REQ_CONTEXT_SIZE 8
 
 /* SPDM response code (1.0) */
 #define SPDM_DIGESTS 0x01
@@ -480,6 +481,7 @@ typedef struct {
     /* param1 == slot_id
      * param2 == HashType*/
     uint8_t nonce[32];
+    /*uint8_t requester_context[SPDM_REQ_CONTEXT_SIZE]; */
 } spdm_challenge_request_t;
 
 /* SPDM CHALLENGE response */
@@ -492,6 +494,7 @@ typedef struct {
      * uint8_t                measurement_summary_hash[digest_size];
      * uint16_t               opaque_length;
      * uint8_t                opaque_data[opaque_length];
+     * uint8_t                requester_context[SPDM_REQ_CONTEXT_SIZE];
      * uint8_t                signature[key_size];*/
 } spdm_challenge_auth_response_t;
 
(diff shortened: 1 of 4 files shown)


include/industry_standard/spdm.h
```diff
@@ -21,6 +21,7 @@
 
 #define SPDM_NONCE_SIZE 32
 #define SPDM_RANDOM_DATA_SIZE 32
+#define SPDM_REQ_CONTEXT_SIZE 8
```
steven-bellock, line 24: This is specific to `CHALLENGE` / `CHALLENGE_AUTH`.

include/library/spdm_requester_lib.h
```diff
@@ -197,6 +197,41 @@ libspdm_return_t libspdm_challenge_ex(void *spdm_context, void *reserved,
...
+ * this function also performs the basic mutual authentication.
+ *
+ * @param  spdm_context           A pointer to the SPDM context.
+ * @param  reserved               Reserved for session_id and is ignored.
+ * @param  slot_id                The number of slot for the challenge.
+ * @param  measurement_hash_type  The type of the measurement hash.
+ * @param  measurement_hash       A pointer to a destination buffer to store the measurement hash.
+ * @param  slot_mask              A pointer to a destination to store the slot mask.
+ * @param  requester_nonce_in     A buffer to hold the requester nonce (32 bytes) as input, if not NULL.
+ * @param  requester_context      A buffer to hold the requester context (8 bytes) as input, if not NULL.
```
steven-bellock, line 215: If `requester_context` is non-NULL and the negotiated version is <= 1.2 then it should return `LIBSPDM_STATUS_OVERRIDDEN_PARAMETER`. Or maybe make a new warning `LIBSPDM_STATUS_IGNORED_PARAMETER`.

steven-bellock (changes requested):

include/industry_standard/spdm.h
jyao1: NO. MEASUREMENT also need this.

include/library/spdm_requester_lib.h
jyao1: I am not sure if that is a good idea, because it adds burden to the integrator to understand the version.

SPDM 1.1/1.2 already added new extension, and we don't return such overridden or ignored parameter. Why we need to add in SPDM 1.3?

steven-bellock: The header documentation should then state that this parameter is ignored if the negotiated version is <= 1.2.

include/industry_standard/spdm.h
steven-bellock: Ok.

include/library/spdm_requester_lib.h
jyao1: OK I will update the comment.

steven-bellock (approved):

PR #2428, opened 2023-11-07
CMakeLists: Add and support a RISCV_NONE toolchain

alistair23:
Add a RISCV_NONE toolchain to target an embedded RISC-V environment.

This allows building libspdm libraries with mbedtls in an embedded environment. We can then use newlibc to create final binaries with libspdm

Tested with:

```
cmake -DARCH=riscv32 -DTOOLCHAIN=RISCV_NONE -DTARGET=Debug -DCRYPTO=mbedtls -DDISABLE_TESTS=1 ..
```


steven-bellock (changes requested):
https://github.com/DMTF/libspdm/blob/main/doc/build.md also needs to be updated.

alistair23:
Done!

doc/build.md
```diff
@@ -291,3 +291,15 @@ Unit tests can be disable by adding -DDISABLE_TESTS=1 to CMake.
 ```shell
 -DDISABLE_TESTS=1
 ```
+
+### Embedded builds for RISC-V
```
jyao1, line 295: Thanks to put this section. 

1) We have a section 7 - `Supported Architecture and Cross-Compiler based on X64 platform` in https://github.com/DMTF/libspdm/blob/main/README.md#features.

You may find: [RISCV_GNU](https://github.com/riscv/riscv-gnu-toolchain), [RISCV64_GCC](https://packages.ubuntu.com/bionic/gcc-riscv64-linux-gnu),
[RISCV_XPACK](https://github.com/xpack-dev-tools/riscv-none-elf-gcc-xpack)

Could you please add a similar item there? RISCV_NONE.

2) In https://github.com/DMTF/libspdm/blob/main/doc/build.md#compiler-for-riscv32riscv64-choose-one, we describe how to get the compiler.

Could you please add a section on how to get this `riscv64-elf-gcc` ?

3) For the name, why you call `riscv64-elf-gcc` as `RISCV-none` ? But not `RISCV64-ELF-GCC` ?

alistair23: I added 1 and 2.

For the name, it matches the other names. You have `RISCV_GNU` for using the GNU tools (like glibc), `RISCV64_GCC` for using GCC and `RISCV_XPACK` for using xpack. I thought `RISCV_NONE` describes that we aren't using a libc to create the binaries.

I'm happy to change it. I could go with `RISCV_NOSTDLIB` if that's better?

doc/build.md
```diff
@@ -97,6 +97,9 @@ b) [RISCV GNU](https://github.com/riscv-collab/riscv-gnu-toolchain)
 c) [RISCV64 GCC](https://packages.ubuntu.com/bionic/gcc-riscv64-linux-gnu) for RISCV64 only
   - `sudo apt-get install gcc-riscv64-linux-gnu`
 
+c) [RISCV NONE](https://github.com/riscv/riscv-gnu-toolchain)
+  - `sudo pacman -Syu riscv32-elf-binutils riscv32-elf-newlib riscv64-elf-binutils riscv64-elf-gcc riscv64-elf-newlib`
```
jyao1, line 101: I think the index should be d). c) is already used.

I think the URL should be https://archlinux.org/packages/extra/x86_64/riscv64-elf-gcc/, right?

README.md
```diff
@@ -92,6 +92,7 @@
 | [RISCV_GNU](https://github.com/riscv/riscv-gnu-toolchain) |  -   |  -  |  -  |    -    | riscv32-unknown-linux-gnu-gcc | riscv64-unknown-linux-gnu-gcc |
 | [RISCV64_GCC](https://packages.ubuntu.com/bionic/gcc-riscv64-linux-gnu) |  -   |  -  |  -  |    -    |    -    | riscv64-linux-gnu-gcc |
 | [RISCV_XPACK](https://github.com/xpack-dev-tools/riscv-none-elf-gcc-xpack) |  -   |  -  |  -  |    -    | riscv-none-elf-gcc | riscv-none-elf-gcc |
+| [RISCV_NONE](https://github.com/riscv/riscv-gnu-toolchain) |  -   |  -  |  -  |    -    | riscv64-elf-gcc | riscv64-elf-gcc |
```
jyao1, line 95: I think the URL should be https://archlinux.org/packages/extra/x86_64/riscv64-elf-gcc/, right?

doc/build.md
jyao1: I checked riscv64-elf-gcc in ArchLinux - "The GNU Compiler Collection - cross compiler for RISCV64 (bare-metal) target" at  https://archlinux.org/packages/extra/x86_64/riscv64-elf-gcc/

I suggest `RISCV64_NONE`.

alistair23: I have changed the URL to `https://gcc.gnu.org/` to indicate that it's just upstream GCC. This isn't Arch specific so I don't want to point to the Arch package.

I think `RISCV64_NONE` is confusing as it supports both 32 and 64 bit RISC-V. In fact it probably will be used exclusively for 32-bit as that's what embedded systems commonly are

doc/build.md
alistair23: Changed to `d` and the URL now points to upstream GCC as that is where the source comes form

README.md
alistair23: I think it should be https://gcc.gnu.org/ as that is where the source is from

doc/build.md
jyao1: > I have changed the URL to `https://gcc.gnu.org/` to indicate that it's just upstream GCC. This isn't Arch specific so I don't want to point to the Arch package.

That is confusing. In build document, you have below:
```
d) [RISCV NONE](https://gcc.gnu.org/)
  - `sudo pacman -Syu riscv32-elf-binutils riscv32-elf-newlib riscv64-elf-binutils riscv64-elf-gcc riscv64-elf-newlib`
```

NOTE: `pacman` is only for Arch linux.
If it is generic, then we need a better way to describe it.
For example, I am still using ubuntu, does it support?

README.md
jyao1: Are you saying, we need build the compiler from the source?

If so, we need add those steps in build document.

All in all, I think we need a clear step-by-step document to tell people, how to get the compiler binary. (Either download/install it from somewhere, or build from source)

So far, I still don't understand.

doc/build.md
alistair23: Ok, I have updated the docs to describe what is required

It should work on Ubuntu, you will just need to install the required toolchain

README.md
alistair23: > Are you saying, we need build the compiler from the source?

No, I'm not saying that

> 
> If so, we need add those steps in build document.
> 
> All in all, I think we need a clear step-by-step document to tell people, how to get the compiler binary. (Either download/install it from somewhere, or build from source)

A user will need to install a GCC based RISC-V toolchain. By pointing to https://gcc.gnu.org/ we are saying you can use standard GCC instead of a fork (like https://github.com/riscv/riscv-gnu-toolchain which is a fork).

Most distros should expose a RISC-V toolchain, it should just involved installing it.

> 
> So far, I still don't understand.

doc/build.md
```diff
@@ -97,6 +97,14 @@ b) [RISCV GNU](https://github.com/riscv-collab/riscv-gnu-toolchain)
 c) [RISCV64 GCC](https://packages.ubuntu.com/bionic/gcc-riscv64-linux-gnu) for RISCV64 only
   - `sudo apt-get install gcc-riscv64-linux-gnu`
 
+d) [RISCV NONE](https://gcc.gnu.org/)
+  - Use a GCC compiler configured for building baremetal (not Linux) binaries
+
+  - For example on Arch you can install it with
+      ```
+      sudo pacman -Syu riscv32-elf-binutils riscv32-elf-newlib riscv64-elf-binutils riscv64-elf-gcc riscv64-elf-newlib
```
jyao1, line 105: what is command for ubuntu? I got failure below:

```
jyao1@jyao1-desk:~$ sudo apt-get install riscv32-elf-binutils riscv32-elf-newlib riscv64-elf-binutils riscv64-elf-gcc riscv64-elf-newlib
[sudo] password for jyao1:
Reading package lists... Done
Building dependency tree
Reading state information... Done
E: Unable to locate package riscv32-elf-binutils
E: Unable to locate package riscv32-elf-newlib
E: Unable to locate package riscv64-elf-binutils
E: Unable to locate package riscv64-elf-gcc
E: Unable to locate package riscv64-elf-newlib
```

doc/build.md
jyao1: I got failure on ubuntu. Please help me understand how to.

```
jyao1@jyao1-desk:~$ sudo apt-get install riscv32-elf-binutils riscv32-elf-newlib riscv64-elf-binutils riscv64-elf-gcc riscv64-elf-newlib
[sudo] password for jyao1:
Reading package lists... Done
Building dependency tree
Reading state information... Done
E: Unable to locate package riscv32-elf-binutils
E: Unable to locate package riscv32-elf-newlib
E: Unable to locate package riscv64-elf-binutils
E: Unable to locate package riscv64-elf-gcc
E: Unable to locate package riscv64-elf-newlib
```

doc/build.md
alistair23: I don't use Ubuntu, so I'm not sure.

I have written the instructions to be clear for the distro that I use, in the same way the rest of the instructions are only clear for Ubunutu users and not other users. I hopefully have provided enough information to explain what is required so that people can figure it out for their distro

jyao1: It is OK if you only validated Arch Linux.

My request is that: **you should say: it is only validated in Arch Linux.**

If so, the right URL should be  https://archlinux.org/packages/extra/x86_64/riscv64-elf-gcc/

doc/build.md
jyao1: > I think RISCV64_NONE is confusing as it supports both 32 and 64 bit RISC-V. 

My original goal is to let the compiler name match the compiler binary. Since the compiler name has `riscv64-`, I dont think anything wrong to use `RISCV64_` as prefix.

> In fact it probably will be used exclusively for 32-bit as that's what embedded systems commonly are

I am not sure what you mean - `exclusively for 32-bit`. Would you please elaborate a little bit?

alistair23: > My original goal is to let the compiler name match the compiler binary. Since the compiler name has `riscv64-`, I dont think anything wrong to use `RISCV64_` as prefix.

Hmmm... I see what you mean. I just find it really confusing to call it `RISCV64_NONE` when it supports 32/64-bit RISC-V

> I am not sure what you mean - `exclusively for 32-bit`. Would you please elaborate a little bit?

This is targeting an embedded system not running Linux. That is pretty much only going to be 32-bit systems

doc/build.md
alistair23: I have updated the instructions to more explicitly say I have only tested this on Arch

doc/build.md
```diff
@@ -291,3 +301,15 @@ Unit tests can be disable by adding -DDISABLE_TESTS=1 to CMake.
 ```shell
 -DDISABLE_TESTS=1
 ```
+
+### Embedded builds for RISC-V
+
+The libspdm libraries can be build along with mbedtls to target an embedded
```
steven-bellock, line 307: >The libspdm libraries can be built along with Mbed TLS to target an embedded
environment. The Integrator must provide a C library and runtime, such as Newlib.

doc/build.md
```diff
@@ -291,3 +301,15 @@ Unit tests can be disable by adding -DDISABLE_TESTS=1 to CMake.
...
 -DDISABLE_TESTS=1
 ```
+
+### Embedded builds for RISC-V
+
+The libspdm libraries can be build along with mbedtls to target an embedded
+environment. You can then use newlibc to create final binaries linking with
+libspdm and mbedtls
+
+To build libspdm with mbedtls for RISC-V 32-bit run the following
```
steven-bellock, line 311: `mbedtls` -> `Mbed TLS`.

steven-bellock (changes requested):

README.md
```diff
@@ -92,6 +92,7 @@
 | [RISCV_GNU](https://github.com/riscv/riscv-gnu-toolchain) |  -   |  -  |  -  |    -    | riscv32-unknown-linux-gnu-gcc | riscv64-unknown-linux-gnu-gcc |
 | [RISCV64_GCC](https://packages.ubuntu.com/bionic/gcc-riscv64-linux-gnu) |  -   |  -  |  -  |    -    |    -    | riscv64-linux-gnu-gcc |
 | [RISCV_XPACK](https://github.com/xpack-dev-tools/riscv-none-elf-gcc-xpack) |  -   |  -  |  -  |    -    | riscv-none-elf-gcc | riscv-none-elf-gcc |
+| [RISCV_NONE](https://gcc.gnu.org/) |  -   |  -  |  -  |    -    | riscv64-elf-gcc | riscv64-elf-gcc |
```
jyao1, line 95: The URL should be https://archlinux.org/packages/extra/x86_64/riscv64-elf-gcc/

doc/build.md
jyao1: Thank you!

jyao1 (approved):

Issue #2280, opened 2023-08-02
[SPDM 1.3] CAPABILITIES/ALGORITHMS: Add Multiple Key Support (MULTI_KEY_CAP)


jyao1:
1. Update CAPABILITIES - add `MULTI_KEY_CAP`
2. Update ALGORITHMS - add `RequesterMultiKeyConnSel` and `ResponderMultiKeyConn`

PR #2431, opened 2023-11-13
1.3 CAPABILITIES/ALGORITHMS: Add more capability

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2280


library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -85,6 +91,11 @@ static bool validate_responder_capability(uint32_t capabilities_flag, uint8_t ve
                 (hbeat_cap == 1) || (key_upd_cap == 1)) {
                 return false;
             }
+            if (version == SPDM_MESSAGE_VERSION_13) {
```
steven-bellock, line 94: I'd move this down to line 156 with the rest of the 1.3 checks.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -107,6 +118,11 @@ static bool validate_responder_capability(uint32_t capabilities_flag, uint8_t ve
             if ((chal_cap == 1) || (key_ex_cap == 1) || (meas_cap == 2) || (mut_auth_cap == 1)) {
                 return false;
             }
+            if (version == SPDM_MESSAGE_VERSION_13) {
```
steven-bellock, line 121: I'd move this down to line 156 with the rest of the 1.3 checks.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -137,6 +153,22 @@ static bool validate_responder_capability(uint32_t capabilities_flag, uint8_t ve
...
+        /* Illegal to return reserved values. */
+        if ((ep_info_cap == 3) || (multi_key_cap == 3)) {
+            return false;
+        }
+        /* check multi-key */
+        if ((multi_key_cap != 0) && (get_key_pair_info_cap == 0)) {
+            return false;
+        }
+        if ((multi_key_cap == 0) &&
+            ((get_key_pair_info_cap == 1) || (set_key_pair_info_cap == 1))) {
```
steven-bellock, line 167: Filed https://github.com/DMTF/SPDM-WG/issues/3270 for this one.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -137,6 +153,22 @@ static bool validate_responder_capability(uint32_t capabilities_flag, uint8_t ve
         }
     }
 
+    /* Checks specific to 1.3. */
```
steven-bellock, line 156: Public key and multi-key are mutually exclusive. If one is set then the other must be cleared.

include/industry_standard/spdm.h
```diff
@@ -190,6 +190,20 @@ typedef struct {
         SPDM_GET_CAPABILITIES_REQUEST_FLAGS_11_MASK | \
         SPDM_GET_CAPABILITIES_REQUEST_FLAGS_CHUNK_CAP)
 
+/* SPDM GET_CAPABILITIES request flags (1.3) */
+#define SPDM_GET_CAPABILITIES_REQUEST_FLAGS_EP_INFO_CAP (0x00400000 | 0x00800000)
```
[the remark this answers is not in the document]
jyao1: change title.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
jyao1: I add it here purposely.

event_cap relies on other session cap. Most logic can be merged.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
jyao1: I add it here purposely.

event_cap relies on other session cap. Most logic can be merged.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
jyao1: Really? Would you please point to me the text in spec?

library/spdm_requester_lib/libspdm_req_get_capabilities.c
jyao1: Will remove this check.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
steven-bellock: I believe that was the intent of the specification. If `PUB_KEY_ID_CAP == 1 && MULTI_KEY_CAP != 0` is legal then there needs to be a lot of clarifications and restrictions.

steven-bellock: Filed issue as https://github.com/DMTF/SPDM-WG/issues/3277.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -124,8 +140,8 @@ static bool validate_responder_capability(uint32_t capabilities_flag, uint8_t ve
         }
     }
 
-    /* Checks specific to 1.2. */
-    if (version == SPDM_MESSAGE_VERSION_12) {
+    /* Checks specific to 1.2 and higher. */
```
steven-bellock, line 143: `specific` -> `common`

library/spdm_requester_lib/libspdm_req_get_capabilities.c
steven-bellock: This breaks the flow of the checking with respect to SPDM versions. Here, 1.3 capabilities are introduced before 1.2 capabilities.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
jyao1: I think we can add the check, **after** the spec clarification.

I prefer to keep this patch as is here.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
jyao1: I think we need to use the most efficient way to do the check.

Otherwise, if we use check 1.0, check 1.1, check 1.2, check 1.3 style, there will be lots of duplication.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
jyao1: fixed

steven-bellock (approved):

PR #2439, opened 2023-11-21
Add unit test for SPDM 1.3 MEASUREMENTS: context field.

Xiaohanjlll:


Zhiqiang520:
It looks like good to me.

steven-bellock (approved):

PR #2440, opened 2023-11-22
os_stub/cryptlib_*: x509: Don't copy some NIDs

alistair23:
Avoid copying the basic_constraints NID as we already set it ourselves
and also avoid copying the authority_key_identifier as it won't be
correct on the CSR.

This PR also fixes the mbedtls NID copying

### DIFF
--- a/os_stub/cryptlib_mbedtls/pk/x509.c
+++ b/os_stub/cryptlib_mbedtls/pk/x509.c
@@ -2110,6 +2110,17 @@ bool libspdm_gen_x509_csr(size_t hash_nid, size_t asym_nid,
             goto free_all;
         }
 
+        if (MBEDTLS_OID_CMP_RAW(MBEDTLS_OID_BASIC_CONSTRAINTS, next_oid->buf.p, oid_tag_len) == 0) {
+            next_oid = next_oid->next;
+            continue;
+        }
+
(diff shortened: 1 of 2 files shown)


os_stub/cryptlib_mbedtls/pk/x509.c
```diff
@@ -2073,55 +2073,66 @@ bool libspdm_gen_x509_csr(size_t hash_nid, size_t asym_nid,
...
+    /*set basicConstraints*/
+    if (mbedtls_x509write_csr_set_extension(&req, MBEDTLS_OID_BASIC_CONSTRAINTS,
+                                            MBEDTLS_OID_SIZE(MBEDTLS_OID_BASIC_CONSTRAINTS),
+                                            is_ca ? basic_constraints_true : basic_constraints_false,
+                                            is_ca ?
+                                            sizeof(basic_constraints_true) :
+                                            sizeof(basic_constraints_false)
+                                            ) != 0) {
+        ret = 1;
+        LIBSPDM_DEBUG((LIBSPDM_DEBUG_INFO,
```
steven-bellock, line 2085: Separate issue but most of these `LIBSPDM_DEBUG_INFO`s should be `LIBSPDM_DEBUG_ERROR`s.

PR #2450, opened 2023-11-30
Add 1.3 CSR and SET_CERT format

jyao1:


include/industry_standard/spdm.h
```diff
@@ -1001,29 +1001,42 @@ typedef struct {
...
-     * param2 == RSVD
-     * void * cert_chain*/
+    /* param1 == BIT[0:3]=slot_id, Request Attribute in 1.3
+     * param2 == key_pair_id in 1.3
+     * uint8_t cert_chain[]; */
 } spdm_set_certificate_request_t;
 
+#define SPDM_SET_CERTIFICATE_REQUEST_SLOT_ID_MASK 0xF
+
+/* SPDM SET_CERTIFICATE request Attributes */
```
steven-bellock, line 1011: Can also define fields for `CSRCertModel` and `SetCertModel`.

jyao1: I don't define it because the spec said the value and format is same.

`CSRCertModel` - `This field indicates the desired certificate model of the CSR. The value and format of this field shall be the same as CertModel in Certificate info`
`SetCertModel` - `This field indicates the certificate model of the certificate chain. The value and format of this field shall be the same as CertModel in Certificate info.`

Issue #2474, opened 2023-12-12
[SPDM 1.3] The Responder should report error when matching the invalid condition of MULTI_KEY_CONN_RSP value calculation


Zhiqiang520:
The Responder should report error when matching the **invalid** condition of `MULTI_KEY_CONN_RSP value calculation`.

1. Refer to the paragraph 333 `Table 32 — MULTI_KEY_CONN_RSP value calculation` In [DSP0274_1.3.0.pdf](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.3.0.pdf).
![image](https://github.com/DMTF/libspdm/assets/90754803/f90df1c7-e986-4bed-9c36-c850ebe08fa5)

2. But the code will not report error when matching the invalid condition of MULTI_KEY_CONN_RSP value calculation.
    The root cause is as following:
2.1 `ResponderMultiKeyConn in NEGOTIATE_ALGORITHMS` was saved into `spdm_context->connection_info.algorithm.other_params_support` at the line 556.
https://github.com/DMTF/libspdm/blob/3bbcb24dbae81a25cb04dd185b0fe948e6ac6b25/library/spdm_responder_lib/libspdm_rsp_algorithms.c#L555-L562 
2.2 But `spdm_context->connection_info.algorithm.other_params_support` was overwritten by `spdm_response->other_params_selection` at the line 735, then the original `ResponderMultiKeyConn in NEGOTIATE_ALGORITHMS` value saved in `spdm_context->connection_info.algorithm.other_params_support` was missed.
https://github.com/DMTF/libspdm/blob/3bbcb24dbae81a25cb04dd185b0fe948e6ac6b25/library/spdm_responder_lib/libspdm_rsp_algorithms.c#L734-L736
2.3 Here `spdm_context->connection_info.algorithm.other_params_support` was sitll taken as `ResponderMultiKeyConn in NEGOTIATE_ALGORITHMS` at line 818.
https://github.com/DMTF/libspdm/blob/3bbcb24dbae81a25cb04dd185b0fe948e6ac6b25/library/spdm_responder_lib/libspdm_rsp_algorithms.c#L817-L839
2.4 We have created test cases to catch this issue.
https://github.com/DMTF/libspdm/pull/2472

PR #2475, opened 2023-12-12
Mask the multi-key bit in connection_info.

jyao1:
The reason is the bit in req msg is different from the bit in rsp msg.

Fix https://github.com/DMTF/libspdm/issues/2474

Pass test at https://github.com/DMTF/libspdm/pull/2472


steven-bellock (approved):
`other_params_support_priority_table` should be renamed but I'll file a separate issue for that.

PR #2486, opened 2023-12-21
rnglib: add support for a nostd target

twilfredo:
## Overview

For nostd/baremetal targets, rnglib cannot currently be used. This change allows the integrator to provide a cryptographically safe implementation for `libspdm_get_random_number_64()`. Such that it is abstracted away from libspdm.

It is the integrators responsibility to provide an implementation that is crypto-graphically safe.

### DIFF
--- a/os_stub/rnglib/CMakeLists.txt
+++ b/os_stub/rnglib/CMakeLists.txt
@@ -5,7 +5,11 @@ INCLUDE_DIRECTORIES(${LIBSPDM_DIR}/include
                     ${LIBSPDM_DIR}/os_stub/rnglib
 )
 
-if(CMAKE_SYSTEM_NAME MATCHES "Linux" AND NOT TOOLCHAIN STREQUAL "ARM_DS2022")
+set (src_rnglib "")
+
+if(TOOLCHAIN STREQUAL "ARM_GNU_BARE_METAL" OR TOOLCHAIN STREQUAL "RISCV_NONE")
+    # The target must provide the required implementation at link time.
+elseif(CMAKE_SYSTEM_NAME MATCHES "Linux" AND NOT TOOLCHAIN STREQUAL "ARM_DS2022")
 SET(src_rnglib
     rng_linux.c
 )
@@ -19,4 +23,8 @@ SET(src_rnglib
 )
 endif()
 
-ADD_LIBRARY(rnglib STATIC ${src_rnglib})
+if (NOT src_rnglib STREQUAL "")
+    ADD_LIBRARY(rnglib STATIC ${src_rnglib})
+else ()
+    message(WARNING "An rnglib implementation must be provided at link time")
+endif()
--- a/os_stub/spdm_crypt_ext_lib/cryptlib_ext.h
+++ b/os_stub/spdm_crypt_ext_lib/cryptlib_ext.h
@@ -11,6 +11,20 @@
 #include "hal/library/cryptlib.h"
 
 /**
+ * Generates a 64-bit random number.
+ *
+ * An implementation must be provided by the target environment to provide
+ * a cryptographically safe 64-bit random number.
+ *
+ * @param[out] rand_data     buffer pointer to store the 64-bit random value.
+ *
+ * @retval true         Random number generated successfully.
+ * @retval false        Failed to generate the random number.
+ *
+ **/
+extern bool libspdm_get_random_number_64(uint64_t *rand_data);
+
+/**
  * Retrieve the common name (CN) string from one X.509 certificate.
  *
  * @param[in]      cert              Pointer to the DER-encoded X509 certificate.


jyao1:
My original thought is that: the integrator can implement a rnglib and just link their own in non-std. (Not use the one in libspdm project)

twilfredo:
> My original thought is that: the integrator can implement a rnglib and just link their own in non-std. (Not use the one in libspdm project)

If I am understanding you correctly, that would imply making changes to the libspdm code base. That is the integrator will need to call it's rng functions replacing the libspdm ones. Is this what you mean? If so, I don't think it is ideal. 

My thinking is that, since, we already have many external stubs for an integrator to implement required by libspdm. Having the integrator implement an rnglib stub(s) required by libspdm, seems like the most idiomatic way.

jyao1:
> If I am understanding you correctly, that would imply making changes to the libspdm code base.

NO. I dont think we need libspdm code base. The integrator can choose a different lib.

Please take a look at https://github.com/DMTF/spdm-emu/tree/main/spdm-device-sample/spdm_device_sample.
It defines its own library override at https://github.com/DMTF/spdm-emu/tree/main/spdm-device-sample/spdm_device_sample/library, and its own CMake at https://github.com/DMTF/spdm-emu/blob/main/spdm-device-sample/spdm_device_sample/CMakeLists.txt.

twilfredo:
> > If I am understanding you correctly, that would imply making changes to the libspdm code base.
> 
> NO. I dont think we need libspdm code base. The integrator can choose a different lib.
> 
> Please take a look at https://github.com/DMTF/spdm-emu/tree/main/spdm-device-sample/spdm_device_sample. It defines its own library override at https://github.com/DMTF/spdm-emu/tree/main/spdm-device-sample/spdm_device_sample/library, and its own CMake at https://github.com/DMTF/spdm-emu/blob/main/spdm-device-sample/spdm_device_sample/CMakeLists.txt.

It seems like this creates a forward dependency for libspdm for  an rnglib by the integrator. Meaning that in the above context libspmd cannot be built into a statically linkable library as it's **own entity**. With the changes in this PR an integrator can do do **both**,

That is, integrator may choose to:

- (Simple) Link to a standalone build of libspdm and **only** have to worry about implementing the rnglib function as required.
- (Not so simple) Provide it's own library and the surrounding buildsystem to support building libspmd and other targets.

jyao1:
If you look at all libraries we need to override at https://github.com/DMTF/spdm-emu/tree/main/spdm-device-sample/spdm_device_sample/library, I believe it is NOT realistic to "built into a statically linkable library as it's own entity."

E.g. how you plan to resolve asymsignlib, psklib?

twilfredo:
> I believe it is NOT realistic to "built into a statically linkable library as it's own entity."

The functions prototypes in https://github.com/DMTF/libspdm/tree/main/include/hal/library/responder of-course need to be defined somewhere. But not at compile time. So you can create a static lib and link with an integrator that implements such functions. 

In this way, we have been able to get a  libspdm responder running on an embedded bare-metal target. That is building a static lib of libsdpm then linking it to an integrator which implements all of the required libspdm stubs.  If these are not defined anywhere, then ofcourse the linker will complain.

To allow `libspdm_get_random_number_64()` in rnglib to be implemented in the same way, allows for us to be consistent with this approach.

jyao1:
@twilfredo, how you plan to resolve asymsignlib, psklib?

twilfredo:
> @twilfredo, how you plan to resolve asymsignlib, psklib?

Just so we are talking about the same thing, you mean these right? [psklib](https://github.com/DMTF/libspdm/blob/main/include/hal/library/responder/psklib.h) and [asymsignlib](https://github.com/DMTF/libspdm/blob/main/include/hal/library/responder/asymsignlib.h)

In that case, as I said before it is up-to the integrator:

- Override the library like how the [`spdm-device-sample` does](https://github.com/DMTF/spdm-emu/blob/aa72b0420b9888d6026d25156ce73ed28726a56b/spdm-device-sample/spdm_device_sample/CMakeLists.txt#L808).
- Implement the stub functions to be resolved when linking.

jyao1:
> > @twilfredo, how you plan to resolve asymsignlib, psklib?
> 
> Just so we are talking about the same thing, you mean these right? [psklib](https://github.com/DMTF/libspdm/blob/main/include/hal/library/responder/psklib.h) and [asymsignlib](https://github.com/DMTF/libspdm/blob/main/include/hal/library/responder/asymsignlib.h)
> 
> In that case, as I said before it is up-to the integrator. You could do either:
> 
> * Override the library like how the [`spdm-device-sample` does](https://github.com/DMTF/spdm-emu/blob/aa72b0420b9888d6026d25156ce73ed28726a56b/spdm-device-sample/spdm_device_sample/CMakeLists.txt#L808).
> * Implement the stub functions to be resolved when linking (Like [this asymsignlib](https://github.com/westerndigitalcorporation/spdm-utils/blob/63ff071b26f9592666a85561fcb483565357ff68/src/libspdm/spdm.rs#L729) implementation of `libspdm_responder_data_sign()`.


That is something I don't understand. Why you cannot treat rnglib in the same way as asymsignlib and psklib?

jyao1:
> Implement the stub functions to be resolved when linking (Like [this asymsignlib](https://github.com/westerndigitalcorporation/spdm-utils/blob/63ff071b26f9592666a85561fcb483565357ff68/src/libspdm/spdm.rs#L729) implementation of libspdm_responder_data_sign().

Actually, I dont understand this sentence very well.
What do you mean stub function?
To me, you implemented the required function in asymsignlib. It is same as your first option "Override the library like how the [spdm-device-sample does](https://github.com/DMTF/spdm-emu/blob/aa72b0420b9888d6026d25156ce73ed28726a56b/spdm-device-sample/spdm_device_sample/CMakeLists.txt#L808)."

twilfredo:
> > Implement the stub functions to be resolved when linking (Like [this asymsignlib](https://github.com/westerndigitalcorporation/spdm-utils/blob/63ff071b26f9592666a85561fcb483565357ff68/src/libspdm/spdm.rs#L729) implementation of libspdm_responder_data_sign().
> 
> Actually, I dont understand this sentence very well. What do you mean stub function? To me, you implemented the required function in asymsignlib. It is same as your first option "Override the library like how the [spdm-device-sample does](https://github.com/DMTF/spdm-emu/blob/aa72b0420b9888d6026d25156ce73ed28726a56b/spdm-device-sample/spdm_device_sample/CMakeLists.txt#L808)." Please enlighten me on this part.

Yeah,  you are right! they are both the same. My point is that we cannot do this for rnglib _currently_ because `libspdm_get_random_number_64()` is not exposed as an external function like the functions is `asymsignlib` and `psklib`. Unless I am missing something 😁.

For example, when you build the [`spdm-device-sample`](https://github.com/DMTF/spdm-emu/blob/aa72b0420b9888d6026d25156ce73ed28726a56b/spdm-device-sample/spdm_device_sample/CMakeLists.txt#L808) we use an integrator implemented `rnglib` thus providing `libspdm_get_random_number_64()` at compile time.  But this can't work if an integrator wants to statically link libspdm to itself, and provide `libspdm_get_random_number_64()` at link time (as in the case you saw above, in the Rust code). So exposing it as an `extern bool libspdm_get_random_number_64(uint64_t *rand_data);` for unsupported targets, allows us to do the same thing we already do with `asymsignlib` and `psklib` for instance.

steven-bellock:
@twilfredo you need to provide an implementation of https://github.com/DMTF/libspdm/blob/main/include/hal/library/cryptlib/cryptlib_rng.h#L28.

twilfredo:
> @twilfredo you need to provide an implementation of https://github.com/DMTF/libspdm/blob/main/include/hal/library/cryptlib/cryptlib_rng.h#L28.

We are using `mbedtls` so libspdm defaults to using `libspdm_random_bytes()` [from here](https://github.com/DMTF/libspdm/blob/159ecd11d70f09c16fe0278821c80513dc99b4bd/os_stub/cryptlib_mbedtls/rand/rand.c#L26). In which case, we only need to provide `libspdm_get_random_number_64()`

If the changes in this PR aren't suitable, perhaps we can add an `extern bool libspdm_get_random_number_64(uint64_t *rand_data);` to this file https://github.com/DMTF/libspdm/blob/main/include/hal/library/cryptlib/cryptlib_rng.h#L28? Which I _think_ should accomplish the same goal.

jyao1:
@twilfredo , I understand what you mean now.

I think we should add `libspdm_get_random_number_64()` to https://github.com/DMTF/libspdm/blob/main/include/hal/library/cryptlib/cryptlib_rng.h.

steven-bellock:
If that is needed it would go in https://github.com/DMTF/libspdm/blob/main/os_stub/spdm_crypt_ext_lib/cryptlib_ext.h or some other header in `os_stub/`.

twilfredo:
> If that is needed it would go in https://github.com/DMTF/libspdm/blob/main/os_stub/spdm_crypt_ext_lib/cryptlib_ext.h or some other header in `os_stub/`.

Okay great! I've added this change.

jyao1 (approved):

include/hal/library/cryptlib/cryptlib_rng.h
```diff
@@ -27,4 +27,18 @@
...
+ * An implementation must be provided by the target environment to provide
+ * a cryptographically safe 64-bit random number.
+ *
+ * @param[out] rand_data     buffer pointer to store the 64-bit random value.
+ *
+ * @retval true         Random number generated successfully.
+ * @retval false        Failed to generate the random number.
+ *
+ **/
+extern bool libspdm_get_random_number_64(uint64_t *rand_data);
```
steven-bellock, line 42: This should be in https://github.com/DMTF/libspdm/blob/main/os_stub/spdm_crypt_ext_lib/cryptlib_ext.h.

steven-bellock (changes requested):

include/hal/library/cryptlib/cryptlib_rng.h
twilfredo: sorry about that! fixed.

steven-bellock (approved):

PR #2487, opened 2023-12-21
Add parameter for the SPDM context to `libspdm_requester_data_sign()` and `libspdm_responder_data_sign()`

colinhect:
We have a use case where there are multiple SPDM contexts and the code path taken in the signing functions is different based on which SPDM context it is signing for.

This change adds the SPDM context as a parameter to the signing functions.


jyao1:
@colinhect  API change is NOT allowed in minor version upgrade.

Would you please help me understand why it is needed?

steven-bellock:
> Would you please help me understand why it is needed?

I would assume the case of two Responders using the same libspdm binary. They have different private/public keys and the function needs to know which one to use.

colinhect:
Within a single binary we instantiate multiple SPDM requesters which potentially use different private/public keys depending on what they are connected to. The signing functions currently have no context passed in which can discriminate which SPDM context it is signing for.

I realize that this is an API breaking change and I'm not expecting this to get merged any time soon. But I figured it was worth trying to get this incorporated properly in the library since this use case seems reasonable.

jyao1:
OK. @colinhect , just a reminder, the CI failed. I think we need your "Signed-off-by" tag.

steven-bellock:
There are multiple functions that would need this. I filed #2494.

jyao1:
resolved in https://github.com/DMTF/libspdm/pull/2592

Issue #2491, opened 2023-12-25
need to improve the OID checking logic


jyao1:
Following code is scanning the OID in memory 1 by 1.

https://github.com/DMTF/libspdm/blob/159ecd11d70f09c16fe0278821c80513dc99b4bd/library/spdm_crypt_lib/libspdm_crypt_cert.c#L888C42-L902

https://github.com/DMTF/libspdm/blob/159ecd11d70f09c16fe0278821c80513dc99b4bd/library/spdm_crypt_lib/libspdm_crypt_cert.c#L963C1-L971

The right behavior is to check the tag and skip the length.
In case of any passing error, the code should stop and return error.

Issue #2325, opened 2023-08-28
SEQUENCE OF and SEQUENCE for "id-DMTF-hardware-identity"


rioone2023:
In the sample certificates for unit_test, such as end_responder1.cert.der, where "id-DMTF-hardware-identity" is implemented as follows
[https://github.com/DMTF/libspdm/blob/main/unit_test/sample_key/ecp384/end_responder1.cert.der](url)

<pre>
$ openssl asn1parse -in ./end_responder1.cert.der -dump -inform DER -strparse 436
    0:d=0  hl=2 l=  26 cons: SEQUENCE
    2:d=1  hl=2 l=  10 prim: OBJECT            :1.3.6.1.4.1.412.274.6
   14:d=1  hl=2 l=  12 prim: OCTET STRING
      0000 - 06 0a 2b 06 01 04 01 83-1c 82 12 02               ..+.........
$ openssl asn1parse -in ./end_responder1.cert.der -dump -inform DER -strparse 450
    0:d=0  hl=2 l=  10 prim: OBJECT            :1.3.6.1.4.1.412.274.2
</pre>

Doesn't it require SEQUENCE OF and SEQUENCE as defined in SPDM v1.2?

<pre>
id-DMTF-spdm-extension Extension ::=
{
	extnID { id-DMTF-spdm 6 }
	critical FALSE
	extnValue id-spdm-cert-oids
}
id-spdm-cert-oids ::= SEQUENCE SIZE (1..MAX) OF id-spdm-cert-oid
id-spdm-cert-oid ::= SEQUENCE
{
	spdmOID OBJECT IDENTIFIER
	spdmOIDdefinition OCTET STRING OPTIONAL
}

id-DMTF-hardware-identity id-spdm-cert-oid :: = {
	spdmOID { id-DMTF-spdm 2 }
	spdmOIDdefinition ABSENT
}
</pre>

Wenxing-hou:
The generated cert does have the format error.  I will fix it. Thanks!

Wenxing-hou:
Hi, @rioone2023, I have regenerated the cert.

**The before id-DMTF-hardware-identity is:**
Parse from https://lapo.it/asn1js/:
![image](https://github.com/DMTF/libspdm/assets/87060980/93081407-3f5f-4250-8347-e8267afa1bbf)
Parse from Openssl command:
![image](https://github.com/DMTF/libspdm/assets/87060980/228a63e6-81fc-4eb0-97cd-c181b76c868e)



[...]

PR #2496, opened 2023-12-27
Improve the OID check logic in cert extension

Wenxing-hou:
Fix: #2491 
The patch has passed the existing `test_spdm_crypt` unit_test.

### DIFF
--- a/library/spdm_crypt_lib/libspdm_crypt_cert.c
+++ b/library/spdm_crypt_lib/libspdm_crypt_cert.c
@@ -865,9 +865,10 @@ static bool libspdm_verify_leaf_cert_spdm_eku(const uint8_t *cert, size_t cert_s
     bool status;
     uint8_t eku[256];
     size_t eku_size;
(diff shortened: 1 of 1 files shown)


jyao1:
question: What is wrong with original cert?

Wenxing-hou:
> question: What is wrong with original cert?

The original long_chians cert was generated a long time ago, which have the wrong `id-DMTF-hardware-identity`.
The related issue is https://github.com/DMTF/libspdm/issues/2325

jyao1:
How you regenerate the new long_chians cert? I think we need script. Not only update binary.

Wenxing-hou:
> How you regenerate the new long_chians cert? I think we need script. Not only update binary.
> 
> Also, I think regenerating cert should be separated to another PR.

Thanks. I have spilt this PR and submitted the new PR: https://github.com/DMTF/libspdm/pull/2504

steven-bellock (approved):

PR #2354, opened 2023-09-12
Regenerate cert with right id-DMTF-hardware-identity format

Wenxing-hou:
Fix: #2325

Regenerate all related cert with right id-DMTF-hardware-identity format.

PR #2504, opened 2024-01-03
Update long_chains leaf cert DMTF SPDM Extension Format

Wenxing-hou:


jyao1:
missing part of https://github.com/DMTF/libspdm/pull/2354

PR #2509, opened 2024-01-05
Make unit test send/receive buffer tracking more granular

### LINKED ISSUE #2508: Make unit test send/receive buffer tracking more granular
The unit tests currently use a single boolean when tracking the state of the send/receive buffer.
https://github.com/DMTF/libspdm/blob/743f1db79a7c3fcf7f1784546e8fe0f20e0cc95a/unit_test/spdm_unit_test_common/common.c#L11
https://github.com/DMTF/libspdm/blob/743f1db79a7c3fcf7f1784546e8fe0f20e0cc95a/unit_test/spdm_unit_test_common/common.c#L37-L39

However this does not detect the case where libspdm acquires the sender buffer and then releases the receiver buffer. There needs to be two booleans to track the state of each buffer.

steven-bellock:
Fix #2508

### DIFF
--- a/unit_test/spdm_unit_test_common/common.c
+++ b/unit_test/spdm_unit_test_common/common.c
@@ -8,8 +8,10 @@
 
 static libspdm_test_context_t *m_spdm_test_context;
 
-static bool m_send_receive_buffer_acquired = false;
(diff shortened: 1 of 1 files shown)


steven-bellock:
@Xiaohanjlll @Zhiqiang520 too.

Issue #2495, opened 2023-12-27
[SPDM 1.3] The `SlotID` fields shall not specify this certificate slot when the corresponding `Key Usage` is not set.


Zhiqiang520:
The `SlotID` fields **shall not** specify this certificate slot when the corresponding `MULTI_KEY_CONN_REQ` or `MULTI_KEY_CONN_RSP` is true and the corresponding `Key Usage` is not set, for CHALLENGE, MEASUREMENTS, KEY_EXCHANGE and so on.

1. Refer to the paragraph 373 `Table 37 — Key usage bit mask` In [DSP0274_1.3.0.pdf](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.3.0.pdf).
![image](https://github.com/DMTF/libspdm/assets/90754803/84b606d4-d368-4887-a474-7da61c7a0d49)

2. For example, `slot_id` received from CHALLENGE request message at line 94 was directly set to `auth_attribute` without checking its corresponding `Key Usage` at line 141.
https://github.com/DMTF/libspdm/blob/159ecd11d70f09c16fe0278821c80513dc99b4bd/library/spdm_responder_lib/libspdm_rsp_challenge_auth.c#L94
https://github.com/DMTF/libspdm/blob/159ecd11d70f09c16fe0278821c80513dc99b4bd/library/spdm_responder_lib/libspdm_rsp_challenge_auth.c#L139-L141

jyao1:
@Zhiqiang520 please create a test for that.

Zhiqiang520:
@jyao1 We have created test cases to catch this issue.
https://github.com/DMTF/libspdm/pull/2518

PR #2520, opened 2024-01-12
Add check for KeyUsage bit 

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2495

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_key_exchange.c
+++ b/library/spdm_requester_lib/libspdm_req_key_exchange.c
@@ -536,6 +536,17 @@ static libspdm_return_t libspdm_try_send_receive_key_exchange(
                 status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
                 goto receive_done;
             }
+
+            if ((spdm_request->header.spdm_version >= SPDM_MESSAGE_VERSION_13) &&
+                spdm_context->connection_info.multi_key_conn_req) {
+                if ((spdm_context->local_context.local_key_usage_bit_mask[*req_slot_id_param] &
+                     SPDM_KEY_USAGE_BIT_MASK_KEY_EX_USE) == 0) {
+                    libspdm_secured_message_dhe_free(
+                        spdm_context->connection_info.algorithm.dhe_named_group, dhe_context);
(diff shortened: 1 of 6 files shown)


steven-bellock:
Requester needs to check `KEY_EXCHANGE_RSP.SlotIDParam`, if it exists.

jyao1:
That is good point. I think we need SPDM spec may need to clarify that.
The `SlotID` usage for `FINISH` should be covered by `KeyExUse`.

library/spdm_requester_lib/libspdm_req_key_exchange.c
```diff
@@ -536,6 +536,19 @@ static libspdm_return_t libspdm_try_send_receive_key_exchange(
                 status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
                 goto receive_done;
             }
+
+            /* relax the key_usage_bit check for slot 0, because the responder might no know the key usage yet */
```
steven-bellock, line 540: This is a bug in the specification.

steven-bellock, line 540: If the Responder has chosen `SPDM_KEY_EXCHANGE_RESPONSE_MUT_AUTH_REQUESTED` then it should already know the Requester's multi-key restrictions, else the Responder would issue `GET_DIGESTS`. So the `(*req_slot_id_param != 0x0)` is not needed.

jyao1: This field is mandatory.

Without the check, how do you know if this field should be used or not?

steven-bellock: `SPDM_KEY_EXCHANGE_RESPONSE_MUT_AUTH_REQUESTED` means the Responder wants to skip the encapsulated flow and go straight to `FINISH`. If Responder specifies slot 0 then it means slot 0 and the Requester should check if its slot 0 key can do key exchange.

jyao1: Ah, I see. Agree.

steven-bellock (approved):

PR #2545, opened 2024-01-26
Fix SPDM 1.3 MEL algorithm negotiate responder

Xiaohanjlll:
Fix： #2546


jyao1 (approved):

steven-bellock (approved):
The same logic should be performed to `MeasurementSpecificationSel` to keep them consistent.

PR #2572, opened 2024-02-15
Initialize variable to quiet compiler finding

bluerise:
The compiler believes that the variable secured_context might be used in the switch-case LIBSPDM_DATA_SESSION_SEQUENCE_NUMBER_REQ_DIR without being initialized.  This though does not happen because in that case need_session_info_for_data() will return true and secured_context will be set.


steven-bellock:
Please specify which compiler(s) complain in the commit description.

jyao1:
please fix CI failure

Issue #2554, opened 2024-02-04
A Requester may set more than one bit in the table to indicate each supported format.


Zhiqiang520:
A **Requester** may set **more than one bit** in the table to indicate each supported format.

1. Refer to the paragraph 310 In [DSP0274_1.3.0.pdf](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.3.0.pdf).
<img width="545" alt="image" src="https://github.com/DMTF/libspdm/assets/90754803/b2866815-0fe7-47e3-aa39-66495350fe5a">

2. but in the received `NEGOTIATE_ALGORITHMS request message`, if spdm_request->other_params_support = SPDM_ALGORITHMS_OPAQUE_DATA_FORMAT_0 | SPDM_ALGORITHMS_OPAQUE_DATA_FORMAT_1, then the libspdm(`spdm_responder_lib`) will generate_error_response.
https://github.com/DMTF/libspdm/blob/eded450e4f2e569aecb41b378572bd7ce8b25549/library/spdm_responder_lib/libspdm_rsp_algorithms.c#L400-L412

steven-bellock:
Note that while libspdm allows `SPDM_ALGORITHMS_OPAQUE_DATA_FORMAT_0`, `SPDM_ALGORITHMS_OPAQUE_DATA_FORMAT_1` is required for key exchange.

PR #2577, opened 2024-02-18
Remove request msg check for OPAQUE_DATA_FORMAT, since reserved bit i…

jyao1:
…s legal.

Fix: https://github.com/DMTF/libspdm/issues/2554


steven-bellock (approved):
This pull request is fine, but we need other changes as well for `OpaqueDataFmt`. libspdm should only be looking at the first two bits but `SPDM_ALGORITHMS_OPAQUE_DATA_FORMAT_MASK` is four bits. I'll file a separate issue for that.

PR #2464, opened 2023-12-06
Fix up for build error in build environment with ARM architecture

gixttl:
There are couple of problems when I try to build libspdm in a Yocto project with ARM architecture.
This pull request fixes the problems which I've met.
I used following cmake command.
```
EXTRA_OECMAKE = "-DARCH=${TUNE_ARCH} -DTOOLCHAIN=NONE -DTARGET=Release -DCRYPTO=openssl -DBUILD_LINUX_SHARED_LIB=ON"
```

### Fixed problems:
1. openssllib and cryptolib_openssl prints "Unknown ARCH" and stop building, because 'os_stub/openssllib/CMakeLists.txt' and 'os_stub/cryptlib_openssl/CMakeLists.txt' has 2 if statement for 'ARCH' cmake parameter and first one prints "Unknown ARCH" with FATAL_ERROR.

2. Warnings and errors from sub-modules related with compile flags.
For ex, following error needs '-include base.h'.
```
| .../libspdm/os_stub/openssllib/include/openssl/configuration.h:55:9: error: unknown type name 'size_t'
|    55 | typedef size_t UINTN;
|       |         ^~~~~~
```

3. Cannot find -lssl and -lcrypto on linking lib/libspdm_crypto.so
```
.../libspdm/git-r0/recipe-sysroot-native/usr/bin/arm-openbmc-linux-gnueabi/../../libexec/arm-openbmc-linux-gnueabi/gcc/arm-openbmc-linux-gnueabi/11.2.0/ld: cannot find -lssl
.../libspdm/git-r0/recipe-sysroot-native/usr/bin/arm-openbmc-linux-gnueabi/../../libexec/arm-openbmc-linux-gnueabi/gcc/arm-openbmc-linux-gnueabi/11.2.0/ld: cannot find -lcrypto
collect2: error: ld returned 1 exit status
```

4. Redundant RPATH warning by Yocto QA tests
```
libspdm: /usr/lib/libspdm.so contains probably-redundant RPATH /usr/lib [useless-rpaths]
```

PR #2578, opened 2024-02-19
CMakeLists: add disabling flto option for NONE

twilfredo:
~Similar to [8071c13](https://github.com/DMTF/libspdm/commit/8071c139a89a0b90cb256e787a603c34f028380d), this adds an option to disable flto for the `NONE` toolchain. This is useful when linking against other libraries with foreign FFIs (rust etc...). @alistair23~ 

"The [point](https://github.com/DMTF/libspdm/blob/main/doc/build.md#linux-builds-inside-build-environments) of the "NONE" option though was to provide none of these compiler options and the build project should provide them instead. It seems like https://github.com/DMTF/libspdm/pull/2464 introduced a bug in setting these." As mentioned by Alistair below.

 However, the changes introduced
in https://github.com/DMTF/libspdm/commit/811f2b596def04b3a36368cf2098546d7907767f set certain compiler/linker
option that does not comply with the definition of the options as
specified in [1]. This change removes those options.

[1] https://github.com/DMTF/libspdm/blob/main/doc/build.md#linux-builds-inside-build-environments

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -618,25 +618,6 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
         SET(CMAKE_EXE_LINKER_FLAGS "")
 
         SET(CMAKE_C_LINK_EXECUTABLE "")
-
-    elseif(TOOLCHAIN STREQUAL "NONE")
-        ADD_COMPILE_OPTIONS(-fshort-wchar -fno-strict-aliasing -Wall -Werror -Wno-array-bounds -ffunction-sections -fdata-sections -fno-common -Wno-address -fpie -fno-asynchronous-unwind-tables -flto -DUSING_LTO  -Wno-maybe-uninitialized -Wno-uninitialized  -Wno-builtin-declaration-mismatch -Wno-nonnull-compare -Werror-implicit-function-declaration)
-        if(CMAKE_BUILD_TYPE STREQUAL "Debug")
-            ADD_COMPILE_OPTIONS(-g)
-        endif()
-        if(GCOV STREQUAL "ON")
-        ADD_COMPILE_OPTIONS(--coverage -fprofile-arcs -ftest-coverage)
-        endif()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -620,7 +620,18 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
...
+            SET(FLTO_LINK_OPTS "-fno-lto")
+        else()
+            if (DEFINED USING_LTO)
+                SET(FLTO_COMPILE_OPTS "-flto" "-${USING_LTO}")
+            else()
+                SET(FLTO_COMPILE_OPTS "-flto")
+            endif()
+            SET(FLTO_LINK_OPTS "-flto")
+        endif()
+        ADD_COMPILE_OPTIONS(-fshort-wchar -fno-strict-aliasing -Wall -Werror -Wno-array-bounds -ffunction-sections -fdata-sections -fno-common -Wno-address -fpie -fno-asynchronous-unwind-tables ${FLTO_COMPILE_OPTS}  -Wno-maybe-uninitialized -Wno-uninitialized  -Wno-builtin-declaration-mismatch -Wno-nonnull-compare -Werror-implicit-function-declaration)
```
alistair23, line 634: This is a good improvement.

The [point](https://github.com/DMTF/libspdm/blob/main/doc/build.md#linux-builds-inside-build-environments) of the "NONE" option though was to provide none of these options (and the build project should provide them). It seems like https://github.com/DMTF/libspdm/pull/2464 introduced a bug setting these.

I think it's probably better to revert these changes out to fix the build issues

twilfredo:
Update: partially reverts https://github.com/DMTF/libspdm/pull/2464/commits/811f2b596def04b3a36368cf2098546d7907767f as per @alistair23 comments.

gixttl (commented):
I agree on this patch.

steven-bellock (approved):

Issue #1822, opened 2023-03-20
`spdm_secured_messsage_lib` macros


steven-bellock:
`spdm_secured_messsage_lib` shouldn't exist unless asymmetric or pre-shared key exchange is enabled at compile time.

jyao1:
2023/3/27 - make a proposal

PR #2582, opened 2024-02-21
fix default build options for generate_iv()

bluerise:
generate_iv() iterates over sizeof(uint64_t), accessing an array with LIBSPDM_MAX_AEAD_IV_SIZE number of elements.  With the default build options when no AEAD algorithms are compiled in, this array consists of a single entry.  Gcc 12.2.0 complains about this and aborts the compilation.  Bump the default define to 8 to allow build to succeed.


steven-bellock:
Related issue : #1822.

steven-bellock (approved):

Issue #2494, opened 2023-12-26
Pass `spdm_context` to HAL functions


steven-bellock:
From #2487. 

Integrators may instantiate multiple endpoints using the libraries in `libspdm/library`. However many of the Integrator-implemented functions in https://github.com/DMTF/libspdm/tree/main/include/hal/library assume a single instance. For example when `libspdm_generate_measurement_summary_hash` is called the function cannot discern between the multiple instances. To alleviate this the `spdm_context` should be passed to these functions. `LIBSPDM_DATA_APP_CONTEXT_DATA` can probably be used to identify the context. If not then #1951 can be implemented.

steven-bellock:
Closed via pull request #2592.

PR #2591, opened 2024-02-27
pass spdm context as cookie to libspdm_*_data_sign()

bluerise:
Hi there,

is there a reason that these two are not being passed the spdm context?

Right now there's no state connected to the signing backend. This means it needs to be hardcoded which credentials are being used. Passing the context allows the backend code to retrieve the app-specific context to access local credentials.

The initial diff passes the context as void *, as the header for the prototypes are included in the common lib before libspdm_context_t is defined. There's probably a way to work around this though.

There might be more crypto wrappers where this might be useful, trying to start the discussion using this PR.

Opinions?

Cheers,
Patrick

### DIFF
--- a/include/hal/library/requester/reqasymsignlib.h
+++ b/include/hal/library/requester/reqasymsignlib.h
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2023 DMTF. All rights reserved.
+ *  Copyright 2023-2024 DMTF. All rights reserved.
  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
  **/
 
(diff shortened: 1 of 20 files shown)


mfischer:
@bluerise have you considered introducing a

```
struct libspdm_context {
  void *data;
};
```

This might make future extension of the API easier?

steven-bellock:
This is #2494. If it is to go into libspdm 3.x then it will need to be behind a disabled-by-default macro so as to not break current builds, and it is needed for (pretty much) all HAL functions.

bluerise:
Ah, yep, I agree. I support #2494, closing this PR then!

bluerise:
@steven-bellock Ah, I thought you're working on this, looks you mainly created the issue. Should I come up with macros and changes to the other HAL functions?

steven-bellock:
Yeah go for it.

PR #2592, opened 2024-02-27
Pass spdm context as cookie to secret lib HAL functions

bluerise:
Right now there's no state connected to the signing backend.  This means it needs to be hardcoded which credentials are being used.  Passing the context allows the backend code to retrieve the app-specific context to access local credentials. This should resolve issue #2494.

The only remaining to convert are libspdm_psk_handshake_secret_hkdf_expand() and libspdm_psk_master_secret_hkdf_expand(), but these are more complicated because the secured messaging library only receives a secured message context, no spdm context. Should we convert those to pass the spdm context as well?  e.g. libspdm_generate_session_handshake_key(), and there are much more. Or should we keep a pointer to the spdm context in the secured message context?


bluerise:
Sorry, @steven-bellock, Github didn't let me re-open the old PR #2591 anymore because I force-pushed before I re-opened it :/

include/hal/library/responder/setcertlib.h
```diff
@@ -16,16 +16,23 @@
...
  *
+ * @param[in]  spdm_context  A pointer to the SPDM context.
+ *
  * @retval  true   It is in a trusted environment.
  * @retval  false  It is not in a trusted environment.
  **/
-extern bool libspdm_is_in_trusted_environment();
+extern bool libspdm_is_in_trusted_environment(
+#if LIBSPDM_HAL_PASS_SPDM_CONTEXT
+    void *spdm_context
```
steven-bellock, line 26: This one probably doesn't need it since it applies to all contexts. Ie, the device is in a secure manufacturing environment or not.

include/library/spdm_lib_config.h
```diff
@@ -400,4 +400,9 @@
 #define LIBSPDM_CHECK_SPDM_CONTEXT 1
 #endif
 
+/* Enable passing the SPDM context to HAL functions. */
```
steven-bellock, line 403: "This macro will be removed when libspdm 4.0 is released."

steven-bellock:
>The only remaining to convert are libspdm_psk_handshake_secret_hkdf_expand() and libspdm_psk_master_secret_hkdf_expand(), but these are more complicated because the secured messaging library only receives a secured message context, no spdm context.

Let me think about those. There might be a lot more information needed when determining which PSK to use.

include/hal/library/responder/setcertlib.h
bluerise: I understand what you're saying, in my case I'd use it so I can access callbacks that would tell me the actual state. Otherwise I'd have to make this a global. So it's less about it being context specific, but more about allowing me to access outside helpers to get the result without hardcoding global variables. But I can understand you being hesitant to this.

jyao1: I don't mind to simply add `spdm_context` to all APIs, just in case someone wants it.

jyao1:
> Or should we keep a pointer to the spdm context in the secured message context?

I think this could be a good idea.

include/hal/library/responder/setcertlib.h
steven-bellock: Yeah that's fine.

steven-bellock:
If it's a pointer stored in the secured message context then the location of the SPDM context cannot change after the session has been established. Let's bring this pull request in first and then we can discuss those two functions.

@bluerise please add `LIBSPDM_HAL_PASS_SPDM_CONTEXT` to https://github.com/DMTF/libspdm/blob/e0cda37e2752e15aa43979bd8c0e25bb7910dffd/.github/workflows/build.yml#L50-L51 and then this can be merged.

bluerise:
I'm assuming simple add it to the existing configurations on line 50-51, not create new ones? I just pushed it like that, please verify.

steven-bellock:
>I'm assuming simply to add it to the existing configurations on line 50-51, not create new ones?

Correct.

steven-bellock (approved):

Issue #2571, opened 2024-02-15
Support for `SPDM Responder` to send/receive application data


PrithviAPai:
<img width="833" alt="image" src="https://github.com/DMTF/libspdm/assets/95281374/7d085936-1af5-4996-8027-33503dab1e21">

After secure session establishment, SPDM Spec allows either Requester pr Responder to send/receive application data. 
In libspdm there is a support for Responder to register the receive function callback. However, in [SPDM Responder library](https://github.com/DMTF/libspdm/tree/main/library/spdm_responder_lib) we dont have send capability. 
Is there any plans in developing that ?

Thanks,
Prithvi A Pai

steven-bellock:
>Is there any plans in developing that ?

There can be. Presumably `libspdm_send_data` and `libspdm_receive_data` would move to `common_lib`.

PR #2598, opened 2024-02-29
Add send/receive to `spdm_common_lib`

PrithviAPai:
Once secure session is established, in application phase both SPDM Requester and SPDM Responder can send/receive messages. Added patch which will enable SPDM Responders also to send message during Appication phase.

Resolves issue: https://github.com/DMTF/libspdm/issues/2571


jyao1:
@PrithviAPai , please squash your commit.
There is no need to create a patch to fix format issue.

PrithviAPai:
> @PrithviAPai , please squash your commit. There is no need to create a patch to fix format issue.

Addressed the comment. 
Squashed the commits. Thank You

Wenxing-hou:
Thanks for your commit. And I compare the change: no code changes, only positional changes. LGTM.
Please submit commit too in spdm-emu if the PR is merged.

jyao1:
Would you please show an example on how responder uses those APIs?

PrithviAPai:
> Would you please show an example on how responder uses those APIs?

SPDM Responder(Integrator) daemon can expose APIs for applications to send encrypted data to SPDM EndPoint.
The payload received by the applications can be sent to EID using `libspdm_send_data`

Issue #2589, opened 2024-02-27
`libspdm_verify_leaf_cert_spdm_extension( )` doesn't correctly check the return value of `libspdm_x509_get_extension_data( )`.


Zhiqiang520:
`libspdm_verify_leaf_cert_spdm_extension( )` doesn't correctly check the return value of `libspdm_x509_get_extension_data( )`.

I think the following line 947~ 945 is incorrect.
How about the situation when `(len == 0) && status == false`, still `return true` ?
https://github.com/DMTF/libspdm/blob/467cd6981d89d5111bd36add09b5f5cd61316ca0/library/spdm_crypt_lib/libspdm_crypt_cert.c#L969-L978

For example:
if code run into the following block line 1368 ~ 1371, then may `status == false` and `*extension_data_size == 0`, I think this means something is wrong.
https://github.com/DMTF/libspdm/blob/467cd6981d89d5111bd36add09b5f5cd61316ca0/os_stub/cryptlib_openssl/pk/x509.c#L1339-L1371

PR #2617, opened 2024-03-11
Enhance libspdm_x509_get_extension_data returned value check

Wenxing-hou:
Fix: #2589

### DIFF
--- a/include/hal/library/cryptlib/cryptlib_cert.h
+++ b/include/hal/library/cryptlib/cryptlib_cert.h
@@ -112,9 +112,9 @@ extern bool libspdm_x509_get_issuer_name(const uint8_t *cert, size_t cert_size,
  * @param[out]     extension_data       Extension bytes.
  * @param[in, out] extension_data_size  Extension bytes size.
  *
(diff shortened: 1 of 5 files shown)


include/hal/library/cryptlib/cryptlib_cert.h
```diff
@@ -206,8 +206,11 @@ extern bool libspdm_x509_get_key_usage(const uint8_t *cert, size_t cert_size, si
  * @param[out]     usage       Key usage bytes.
  * @param[in, out] usage_size  Key usage buffer size in bytes.
  *
- * @retval  true
- * @retval  false
+ * @return true   If the returned usage_size == 0, it means that cert and oid are valid, but the Extended key usage is not found;
```
steven-bellock, line 209: `return` -> `retval`
`retval` is used for specific return values, like `true` and `false`.

Applies to all instances in this pull request.

steven-bellock (changes requested):

include/hal/library/cryptlib/cryptlib_cert.h
Wenxing-hou: Thanks. I have changed the comment.

steven-bellock (approved):

Issue #2291, opened 2023-08-02
[SPDM 1.3] CSR: add param1 and param2


jyao1:
1. param1 - KeyPairID
2. param2 - Request Attributes

jyao1:
I think we may need a truth table for that

| Overwrite | CSRTrackingTag | Pending CSR | Reset | Action |
|------------|------------------|----------------|-------|---------|
|No             | 0                       | No                 | -        | ResetRequired|
|No             | 0                       | Yes                 | -        | ResetRequired or Busy|
|No             | Non-0               | No Match      | N/A   | Unexpected |
|No             | Non-0               | Match            | Before | Busy |
|No             | Non-0               | Match            | After    | CSR |
|Yes             | 0                       | No                 | -          | Invalid (?) or ResetRequired | 
|Yes             | 0                       | Yes                 | -          | Invalid (?) or ResetRequired or Busy | 
|Yes             | Non-0               | No Match      | N/A   | Unexpected |
|Yes             | Non-0               | Match            | Before | Overwrite + ResetRequired |
|Yes             | Non-0               | Match            | After    | Overwrite + ResetRequired |

Wenxing-hou:
> I think we may need a truth table for the reset required use case.
> 
> Overwrite	CSRTrackingTag	Pending CSR	Reset	Action
> No	0	No	-	ResetRequired
> No	0	Yes	-	ResetRequired or Busy
> No	Non-0	No Match	-	Unexpected
> No	Non-0	Match	Before	Busy
> No	Non-0	Match	After	CSR
> Yes	0	No	-	ResetRequired
> Yes	0	Yes	-	ResetRequired
> Yes	Non-0	-	-	Invalid

The history is form: https://github.com/DMTF/SPDM-WG/issues/3286

PR #2618, opened 2024-03-12
Add comment for Get_csr keypairid

Wenxing-hou:
Fix: #2291


steven-bellock (commented):
This should just be an issue.

jyao1:
why we need this???

Issue #2619, opened 2024-03-12
Incorrect parameter for `UnsupportedRequest`


steven-bellock:
There are many instances of https://github.com/DMTF/libspdm/blob/c0eaae4e37db4a42c3c9c7504a7ac8def6e62b99/library/spdm_responder_lib/libspdm_rsp_heartbeat.c#L25-L29 where a value of `0` is given for the error data, rather than the `RequestResponseCode` in the request message.

PR #2630, opened 2024-03-22
Fix error code for unsupported_req error.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2619


steven-bellock (approved):
https://github.com/DMTF/libspdm/blob/d83ef43d3be34419e118fc3507fefdd0d0d2692c/library/spdm_requester_lib/libspdm_req_encap_certificate.c#L31-L35 code like this shouldn't even exist. Ie, if the negotiated version is 1.0 then `libspdm_get_encap_response_certificate` should never be called. I can file a separate issue for that.

PR #2633, opened 2024-03-26
spdm_crypt_lib: Expose the base_hash_algo

alistair23:
The `base_hash_algo` is useful to know if you are using libspdm as a
library.

Reading it from the `libspdm_context_t` struct isn't ideal as the struct
is private and also doesn't work if using Rust for the implementation.

This commit exposes the `base_hash_algo` publicly using a helper
function.

This PR also fixes the documentation in `include/library/spdm_crypt_lib.h`.

### DIFF
--- a/include/library/spdm_crypt_lib.h
+++ b/include/library/spdm_crypt_lib.h
@@ -81,6 +81,16 @@
 #endif /* LIBSPDM_MAX_AEAD_KEY_SIZE */
 
 /**
+ * This function returns the SPDM base hash algorithm.
+ *
+ * @param  context  A pointer to the SPDM context.
+ *
+ * @return SPDM base hash algorithm
+ *     (spdm_context->connection_info.algorithm.base_hash_algo).
+ **/
+uint32_t libspdm_get_base_hash_algo(void *context);
+
+/**
  * This function returns the SPDM hash algorithm size.
  *
  * @param  base_hash_algo SPDM base_hash_algo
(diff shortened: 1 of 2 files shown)


steven-bellock:
>also doesn't work if using Rust for the implementation.

Are you saying that it's difficult / impossible for Rust to get the `BaseHashAlgo` via `libspdm_get_data`?

alistair23:
Honestly, I forgot about `libspdm_get_data()`...

That does work. `libspdm_get_data()` is a little clunky as you end up with something like this

```rust
    let mut spdm_algo_base_hash = SpdmAlgoBaseHash(0);
    let hash_dptr = &mut spdm_algo_base_hash.0 as *mut _ as *mut c_void;

    let ret = libspdm_rs::libspdm_get_data(
        cntx_ptr,
        libspdm_rs::libspdm_data_type_t_LIBSPDM_DATA_BASE_HASH_ALGO,
        &parameter as *const libspdm_rs::libspdm_data_parameter_t,
        hash_dptr,
        &mut data_size,
    );

    if LibspdmReturnStatus::libspdm_status_is_error(ret) {
        return Err(ret);
    }
```

but it works. I can close this if that's the preferred approach

jyao1:
if you want to expose, then we need to expose all algorithms, including hash, asym, req_asym, dhe, aead, etc.

alistair23:
No worries. I'm going to close this then

Issue #2645, opened 2024-03-31
Failure in unit test when size of encrypted SPDM message is not 4-byte aligned


steven-bellock:
From #2642.

`libspdm_test_encode_message` increases the size of an SPDM message so that it is 4-byte aligned : https://github.com/DMTF/libspdm/blob/4c92ff5ced7862e4f2eea945dd723d2e1b1fc476/unit_test/spdm_transport_test_lib/test.c#L83-L85 In the case of a secure message, where the exact size of the SPDM message is known, this is causing test failures. In particular `SUPPORTED_EVENT_TYPES` need not be 4-byte aligned.

PR #2650, opened 2024-04-02
Fix alignment issue in transport

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2645


library/spdm_transport_mctp_lib/libspdm_mctp_common.c
```diff
@@ -22,7 +22,9 @@
  * @retval RETURN_SUCCESS               The message is encoded successfully.
  * @retval RETURN_INVALID_PARAMETER     The message is NULL or the message_size is zero.
  **/
-libspdm_return_t libspdm_mctp_encode_message(const uint32_t *session_id, size_t message_size,
+libspdm_return_t libspdm_mctp_encode_message(const uint32_t *session_id,
```
steven-bellock, line 25: Production code should (hopefully) not have to be changed to fix this. Why can't this be restricted to `unit_test/`? In particular why can't `LIBSPDM_TEST_ALIGNMENT` be `1`?

jyao1: MCTP is NOT impacted, because the LIBSPDM_MCTP_ALIGNMENT is always 1 at this moment.
MCTP is changed, just because I want to align the code with TEST_ALIGNMENT.


TEST is impacted, because I do want to see if our code can handle 4 alignment correctly.

steven-bellock: I am now slightly concerned about libspdm checking the exact size of an SPDM message. Is it possible that there exist transports such that 277's `Application Data Length` may not accurately give the exact size of the SPDM message?

steven-bellock: That might be a question for the SPDM Working Group.

jyao1: Does exact size check block this PR?

If no, I suggest we file a new issue.

steven-bellock: Ok that can be a separate issue.

steven-bellock (approved):

Issue #2651, opened 2024-04-02
Miss checking whether `spdm_response_size < sizeof(spdm_csr_response_t) + spdm_response-> csr_length` before `libspdm_copy_mem`.


Zhiqiang520:
Miss checking whether `spdm_response_size < sizeof(spdm_csr_response_t) + spdm_response-> csr_length` before `libspdm_copy_mem`.

I think we should check whether `spdm_response_size < sizeof(spdm_csr_response_t) + spdm_response-> csr_length` before `libspdm_copy_mem` at the line 212.
https://github.com/DMTF/libspdm/blob/4c92ff5ced7862e4f2eea945dd723d2e1b1fc476/library/spdm_requester_lib/libspdm_req_get_csr.c#L206-L212

Issue #2649, opened 2024-04-02
`if (spdm_response_size < sizeof(spdm_csr_response_t))`, then `spdm_response->csr_length` will overflow read.


Zhiqiang520:
`if (spdm_response_size < sizeof(spdm_csr_response_t))`, then `spdm_response->csr_length` will overflow read.

Before reading its member variables `csr_length` at line 201, we should check `if (spdm_response_size < sizeof(spdm_csr_response_t))`.
https://github.com/DMTF/libspdm/blob/4c92ff5ced7862e4f2eea945dd723d2e1b1fc476/library/spdm_requester_lib/libspdm_req_get_csr.c#L201-L204

PR #2652, opened 2024-04-02
Add more check for csr response size

Wenxing-hou:
Fix: #2649
Fix: #2651

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_get_csr.c
+++ b/library/spdm_requester_lib/libspdm_req_get_csr.c
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2021-2022 DMTF. All rights reserved.
+ *  Copyright 2021-2024 DMTF. All rights reserved.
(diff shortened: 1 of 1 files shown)


library/spdm_requester_lib/libspdm_req_get_csr.c
```diff
@@ -198,11 +198,19 @@ static libspdm_return_t libspdm_try_get_csr(libspdm_context_t *spdm_context,
         goto receive_done;
     }
 
+    if (spdm_response_size < sizeof(spdm_csr_response_t)) {
```
steven-bellock, line 201: This can replace the check at line 174.

Wenxing-hou: I think it is better to split the two check. 
Because the response may be error message,  the error message size is `sizeof(spdm_message_header_t)`.
Thanks.

steven-bellock: Ah yeah that's right. I always forget about `ERROR`.

steven-bellock (approved):

PR #2666, opened 2024-04-18
Fix vendor defined request size

m-milinkovic:
The pointer of Vendor Defined Request is moved, then the request size should be adjusted accordingly.

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_vendor_response.c
+++ b/library/spdm_responder_lib/libspdm_rsp_vendor_response.c
@@ -1,5 +1,6 @@
 /**
- *  Copyright 2023 DMTF. All rights reserved.
+ *  Copyright Notice:
+ *  Copyright 2023-2024 DMTF. All rights reserved.
  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
  **/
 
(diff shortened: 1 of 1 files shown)


jyao1:
@uvinrg, could you please take a look?

jyao1 (approved):

jyao1 (changes requested):
@m-milinkovic, I think below calculation is risky, since it includes the transport padding.
```
    req_size = (uint16_t)request_size -
               sizeof(spdm_vendor_defined_request_msg_t) -
               ((const spdm_vendor_defined_request_msg_t*)request)->len -
               sizeof(uint16_t);
```

The `req_size` should be got from 2 bytes field `ReqLength` in `spdm_request`, which is precise size. See below:
```
    req_vendor_id = ((const uint8_t *)request) +
                    sizeof(spdm_vendor_defined_request_msg_t);
    req_size = *(const uint16_t *)(req_vendor_id + spdm_request->len);
```

m-milinkovic:
Also, please note that, if you decide to merge this commit, I'll check other affected repos (e.g., SPDM-Responder-Validator) and propose a solution as well.

jyao1 (approved):

uvinrg:
Hi guys, sorry for the late response. Thank you for the fix, looks good.

Issue #2678, opened 2024-04-30
Allow Integrator to ignore unsupported critical extensions


Wenxing-hou:
By default any unhandled critical extensions in certificates or (if checked) CRLs result in a fatal error.
But the actual cert may have the customized critical extension. 
It is better to add the flag to ignore unhandled critical extensions when the user want to ignore.

https://github.com/DMTF/libspdm/blob/ab6e3436413e23cd2339e7d053a91c7b0ae423ae/os_stub/openssllib/openssl_gen/openssl/x509_vfy.h#L425-L426

steven-bellock:
From [RFC 5280:](https://www.rfc-editor.org/rfc/rfc5280#section-4.2)
>The X.509 v3 certificate format also allows communities to define private extensions to carry information unique to those communities.  Each extension in a certificate is designated as either critical or non-critical.  A certificate-using system MUST reject the certificate if it encounters a critical extension it does not recognize or a critical extension that contains information that it cannot process.  A non-critical extension MAY be ignored if it is not recognized, but MUST be processed if it is recognized.

From [OpenSSL](https://www.openssl.org/docs/man3.0/man1/openssl-verification-options.html):
>Normally if an unhandled critical extension is present that is not supported by OpenSSL the certificate is rejected (as required by RFC5280).

So for the cryptography libraries that libspdm supports, I think it's fine if this check is on by default, and that it can be disabled explicitly. Note that, while this issue is currently for OpenSSL, presumably the same disablement also applies to MbedTLS. I'll rename the issue.

https://github.com/DMTF/libspdm/blob/ab6e3436413e23cd2339e7d053a91c7b0ae423ae/os_stub/mbedtlslib/include/mbedtls/config.h#L2206-L2213

PR #2679, opened 2024-04-30
Add ignore unhandle critical extension check flag for Openssl

Wenxing-hou:
Fix: #2678

The default cert check doesn't have the flag.
The user can use the following build commad to add the flag. 
`cmake -G"NMake Makefiles" -DARCH=x64 -DTOOLCHAIN=VS2019 -DTARGET=Release -DCRYPTO=openssl -DX509_IGNORE_CRITICAL=ON ..`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,7 @@ SET(CRYPTO ${CRYPTO} CACHE STRING "Choose the crypto of build: mbedtls openssl" 
 SET(GCOV ${GCOV} CACHE STRING "Choose the target of Gcov: ON  OFF, and default is OFF" FORCE)
 SET(STACK_USAGE ${STACK_USAGE} CACHE STRING "Choose the target of STACK_USAGE: ON  OFF, and default is OFF" FORCE)
 SET(BUILD_LINUX_SHARED_LIB ${BUILD_LINUX_SHARED_LIB} CACHE STRING "Choose if libspdm shared library should be built for linux: ON OFF, and default is OFF" FORCE)
+SET(X509_IGNORE_CRITICAL ${X509_IGNORE_CRITICAL} CACHE STRING "Choose if libspdm-provided cryptography libraries (OpenSSL and MbedTLS) ignore unsupported critical extensions in certificates : ON OFF, and default is OFF" FORCE)
 
 if(NOT GCOV)
     SET(GCOV "OFF")
@@ -30,6 +31,10 @@ endif()
 
 if(NOT BUILD_LINUX_SHARED_LIB)
     SET(BUILD_LINUX_SHARED_LIB "OFF")
+endif()
+
+if(NOT X509_IGNORE_CRITICAL)
+    SET(X509_IGNORE_CRITICAL "OFF")
 endif()
 
 SET(LIBSPDM_DIR ${PROJECT_SOURCE_DIR})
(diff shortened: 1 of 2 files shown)


jyao1 (approved):

CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ SET(CRYPTO ${CRYPTO} CACHE STRING "Choose the crypto of build: mbedtls openssl"
 SET(GCOV ${GCOV} CACHE STRING "Choose the target of Gcov: ON  OFF, and default is OFF" FORCE)
 SET(STACK_USAGE ${STACK_USAGE} CACHE STRING "Choose the target of STACK_USAGE: ON  OFF, and default is OFF" FORCE)
 SET(BUILD_LINUX_SHARED_LIB ${BUILD_LINUX_SHARED_LIB} CACHE STRING "Choose if libspdm shared library should be built for linux: ON OFF, and default is OFF" FORCE)
+SET(X509_IGNORE_CRITICAL ${X509_IGNORE_CRITICAL} CACHE STRING "Choose if libspdm ignore unhandled critical cert extensions : ON OFF, and default is OFF" FORCE)
```
steven-bellock, line 22: >Choose if libspdm-provided cryptography libraries (OpenSSL and MbedTLS) ignore unsupported critical extensions in certificates : ON OFF, and default is OFF

CMakeLists.txt
```diff
@@ -164,6 +169,10 @@ else()
     MESSAGE(FATAL_ERROR "Unknown CRYPTO")
 endif()
 
+if ((X509_IGNORE_CRITICAL STREQUAL "ON") AND (CRYPTO STREQUAL "openssl"))
```
steven-bellock, line 172: Also add this option to MbedTLS.

steven-bellock (changes requested):

CMakeLists.txt
Wenxing-hou: Thanks. I have updated the message.

CMakeLists.txt
Wenxing-hou: Thanks. I have added the flag for Mbedtls.
https://github.com/DMTF/libspdm/blob/93459f4720feab6e08540aeefe29a7c761272db2/CMakeLists.txt#L172-L178

The flag is from:
https://github.com/DMTF/libspdm/blob/81791b8b958085cd9ed2f1a51336de20d0eb81bd/os_stub/mbedtlslib/include/mbedtls/config.h#L2203-L2213

CMakeLists.txt
```diff
@@ -164,6 +169,14 @@ else()
     MESSAGE(FATAL_ERROR "Unknown CRYPTO")
 endif()
 
+if (X509_IGNORE_CRITICAL STREQUAL "ON")
+    if (CRYPTO STREQUAL "openssl")
+        add_definitions(-DOPENSSL_IGNORE_CRITICAL=1)
+    else()
```
steven-bellock, line 175: Should be an `else if` for `mbedtls`.

steven-bellock (changes requested):

CMakeLists.txt
Wenxing-hou: Although the current Libspdm only have two crypto library, there may be other crypto library in the future.
https://github.com/DMTF/libspdm/blob/16bf4380cdb514e0a6d7682955e2f48cc0e1e9eb/CMakeLists.txt#L147-L165

And I have added the check.  Thanks.
https://github.com/DMTF/libspdm/blob/6473dc78def184cfc0bb4fcf72a19c8b500b73f0/CMakeLists.txt#L172-L178

Issue #2694, opened 2024-05-12
Add session ID parameter to `libspdm_vendor_response_callback_func`


steven-bellock:
From #2681. 

The vendor response callback function may need to know whether the message was sent within a secure session, and if so, which secure session. Pass the session ID as new parameter that is not present by default.

m-milinkovic:
Hi @steven-bellock,
Do you mind assigning this one to me?

steven-bellock:
This was fixed by #2702.

PR #2702, opened 2024-05-16
Fix acquiring Session ID in Vendor Defined Response

m-milinkovic:
Hi @steven-bellock,
I believe we need to fix acquiring Session ID prior to #2694.

Please check the call of spdm_context->get_response_func (i.e., line 67).
session_info is not a pointer, it's an array. Therefore, Session ID is correct only if Session Info is stored at index zero.

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_vendor_response.c
+++ b/library/spdm_responder_lib/libspdm_rsp_vendor_response.c
@@ -43,6 +43,7 @@ libspdm_return_t libspdm_get_vendor_defined_response(libspdm_context_t *spdm_con
 
     libspdm_session_info_t *session_info = NULL;
     libspdm_session_state_t session_state = 0;
+    const uint32_t *session_id = NULL;
     uint8_t *resp_data = NULL;
     const uint8_t *req_vendor_id;
     const uint8_t *req_data;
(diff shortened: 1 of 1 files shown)


m-milinkovic:
Also... Line 67 is always incorrect in case if secure session is not established - a "valid" pointer is passed to spdm_context->get_response_func instead of NULL.

library/spdm_responder_lib/libspdm_rsp_vendor_response.c
```diff
@@ -56,6 +57,27 @@ libspdm_return_t libspdm_get_vendor_defined_response(libspdm_context_t *spdm_con
...
     }
 
+    if (spdm_context->last_spdm_request_session_id_valid) {
+        session_info = libspdm_get_session_info_via_session_id(
+            spdm_context,
+            spdm_context->last_spdm_request_session_id);
+        if (session_info == NULL) {
+            return libspdm_generate_error_response(
+                spdm_context,
+                SPDM_ERROR_CODE_UNEXPECTED_REQUEST, 0,
```
steven-bellock, line 67: I think this is the wrong error code, but that is a separate issue.

steven-bellock (approved):

Issue #2656, opened 2024-04-11
[SPDM 1.3] For `GenericCert` certificate models, `Signature Algorithm`, `Issuer`, `Subject Name`, `Validity` and `Key Usage` Fields is Optional, not Mandatory.


Zhiqiang520:
1. For `GenericCert` certificate models,
`Signature Algorithm`, `Issuer`, `Subject Name`, `Validity` and `Key Usage` Fields is **Optional**, **not Mandatory**.

Refer to the paragraph 397 `Table 42 — Field requirements` In [DSP0274_1.3.0.pdf](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.3.0.pdf).
![image](https://github.com/DMTF/libspdm/assets/90754803/b6aeef01-025c-45ce-8172-5fd4dc754b0b)

2. But `libspdm_x509_common_certificate_check( )` function take `Issuer`, `Subject Name`, `Validity` and `Key Usage` Fields as **Mandatory** even for `GenericCert` certificate models.
https://github.com/DMTF/libspdm/blob/828ef62524bcaeca4e90d0c021221e714872e2b5/library/spdm_crypt_lib/libspdm_crypt_cert.c#L1198-L1224
https://github.com/DMTF/libspdm/blob/828ef62524bcaeca4e90d0c021221e714872e2b5/library/spdm_crypt_lib/libspdm_crypt_cert.c#L1041-L1111

steven-bellock:
If a generic certificate only implemented the mandatory fields and did not implement the optional fields, would it pass the OpenSSL / Mbed TLS certificate chain validation functions?

Wenxing-hou:
Hi all. 
 I have investigated the OpenSSL command cert generation.
For Signature Algorithm, Issuer, Subject Name, Validity and Key Usage Fields, I make a table.

| Optional field for GenericCert   | can be absent in when use OpenSSL command generate cert |
| :--------: | :-------: |
| Signature Algorithm | false    |
|  Issuer | false    |
| Subject Name    | false    |
|  Validity | false    |
| Key Usage    | true    |

PR #2712, opened 2024-05-29
Change cert fields to optional for GenericCert model cert check

Wenxing-hou:
Fix: #2656

### DIFF
--- a/include/hal/library/cryptlib/cryptlib_cert.h
+++ b/include/hal/library/cryptlib/cryptlib_cert.h
@@ -35,10 +35,10 @@ extern bool libspdm_asn1_get_tag(uint8_t **ptr, const uint8_t *end, size_t *leng
  * @param[in, out] subject_size  The size in bytes of the cert_subject buffer on input,
  *                               and the size of buffer returned cert_subject on output.
  *
- * @retval  true   The certificate subject retrieved successfully.
- * @retval  false  Invalid certificate, or the subject_size is too small for the result.
- *                 The subject_size will be updated with the required size.
- * @retval  false  This interface is not supported.
+ * @retval  true   If the subject_size is not equal 0. The certificate subject retrieved successfully.
+ * @retval  true   If the subject_size is equal 0. The certificate parse successful. But the cert doen't have subject.
+ * @retval  false  If the subject_size is not equal 0. The certificate subject retrieved successfully.But the subject_size is too small for the result.
+ * @retval  false  If the subject_size is equal 0. Invalid certificate.
  **/
 extern bool libspdm_x509_get_subject_name(const uint8_t *cert, size_t cert_size,
                                           uint8_t *cert_subject,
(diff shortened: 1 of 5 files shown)


library/spdm_crypt_lib/libspdm_crypt_cert.c
```diff
@@ -1089,31 +1096,35 @@ bool libspdm_x509_common_certificate_check(const uint8_t *cert, size_t cert_size
...
         goto cleanup;
     }
 
-    /* 6. validaity*/
+    /* 6. validity*/
     status = libspdm_x509_get_validity(cert, cert_size, end_cert_from,
                                        &end_cert_from_len, end_cert_to,
                                        &end_cert_to_len);
     if (!status) {
         goto cleanup;
```
Zhiqiang520, line 1117: if the `Validity` field of cert is **absent** and `cert_model == SPDM_CERTIFICATE_INFO_CERT_MODEL_GENERIC_CERT`,  what is the value of `status` here? true or false?
if the `status` is false,  then even for `GenericCert` certificate model, it will return false, I think it does not make sense.

Wenxing-hou: Thanks. I have fixed the code logic.

include/hal/library/cryptlib/cryptlib_cert.h
```diff
@@ -35,10 +35,10 @@ extern bool libspdm_asn1_get_tag(uint8_t **ptr, const uint8_t *end, size_t *leng
  * @param[in, out] subject_size  The size in bytes of the cert_subject buffer on input,
  *                               and the size of buffer returned cert_subject on output.
  *
- * @retval  true   The certificate subject retrieved successfully.
- * @retval  false  Invalid certificate, or the subject_size is too small for the result.
- *                 The subject_size will be updated with the required size.
- * @retval  false  This interface is not supported.
+ * @retval  true   If the subject_size is not euqal 0. The certificate subject retrieved successfully.
```
steven-bellock, line 38: All instances of `euqal` -> `equal`.

steven-bellock (changes requested):

include/hal/library/cryptlib/cryptlib_cert.h
Wenxing-hou: Thanks. I have fixed all the typo for `equal`.

library/spdm_crypt_lib/libspdm_crypt_cert.c
```diff
@@ -1028,6 +1028,9 @@ static bool libspdm_verify_leaf_cert_spdm_extension(const uint8_t *cert, size_t
 /**
  * Certificate common Check for SPDM leaf cert when get_cert and set_cert.
  *
+ * For GenericCert certificate models,
```
steven-bellock, line 1031: Can remove this. The original comment does not say what is or is not checked.

library/spdm_crypt_lib/libspdm_crypt_cert.c
```diff
@@ -1131,8 +1161,10 @@ bool libspdm_x509_common_certificate_check(const uint8_t *cert, size_t cert_size
     if (LIBSPDM_CRYPTO_X509_KU_DIGITAL_SIGNATURE & value) {
         status = true;
     } else {
-        status = false;
-        goto cleanup;
+        if (cert_model != SPDM_CERTIFICATE_INFO_CERT_MODEL_GENERIC_CERT) {
```
steven-bellock, line 1164: Regardless of model, if `Key Usage` is present then digital signature must be set. In this logic if
- Certificate model is generic.
- Key usage is present.
- Digital signature bit is not set.

then the function will return `true` instead of `false`.

steven-bellock (changes requested):

library/spdm_crypt_lib/libspdm_crypt_cert.c
Wenxing-hou: Thanks. I have deleted the comment.

library/spdm_crypt_lib/libspdm_crypt_cert.c
Wenxing-hou: Thanks. I have updated the logic:
```
    if (!status) {
        goto cleanup;
    } else {
        if (value == 0) {
            if (cert_model != SPDM_CERTIFICATE_INFO_CERT_MODEL_GENERIC_CERT) {
                status = false;
                goto cleanup;
            }
        } else {
            if ((LIBSPDM_CRYPTO_X509_KU_DIGITAL_SIGNATURE & value) == 0) {
                status = false;
                goto cleanup;
            }
        }
    }
```

steven-bellock (approved):

Issue #2658, opened 2024-04-15
The checking for `Signature Algorithm` Fields is missed for `DeviceCert / AliasCert` certificate models.


Zhiqiang520:
The checking for `Signature Algorithm` Fields is missed for `DeviceCert / AliasCert` certificate models.

1. For `DeviceCert / AliasCert` certificate models, `Signature Algorithm` Fields is **Mandatory**.
Refer to the paragraph 397 `Table 42 — Field requirements` In [DSP0274_1.3.0.pdf](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.3.0.pdf).
![image](https://github.com/DMTF/libspdm/assets/90754803/5679e35b-6892-429d-9993-0ade2fc5ec0c)

2. But there is no checking for `Signature Algorithm` Fields in function `libspdm_x509_common_certificate_check( )` when certificate model is `DeviceCert` or `AliasCert`. 
I think `/* 3. Verify public key algorithm. */` at the following line 1083 is not the checking for `Signature Algorithm` Fields.
https://github.com/DMTF/libspdm/blob/828ef62524bcaeca4e90d0c021221e714872e2b5/library/spdm_crypt_lib/libspdm_crypt_cert.c#L1041-L1095

3. I think maybe `libspdm_x509_common_certificate_check()` should call `libspdm_x509_get_signature_algorithm()` to get `Signature Algorithm`.

steven-bellock:
https://github.com/DMTF/libspdm/blob/ebb408ee596d8104269f82f6141a63fe7ec15482/library/spdm_crypt_lib/libspdm_crypt_cert.c#L532-L533 seems to return `false` if the `Signature Algorithm` field isn't present. So I would say that `libspdm_verify_cert_subject_public_key_info` does check it when the negotiated algorithm is not SM2.

PR #2713, opened 2024-05-29
Add signature algo check for cert

Wenxing-hou:
Fix: #2658

### DIFF
--- a/include/hal/library/cryptlib/cryptlib_cert.h
+++ b/include/hal/library/cryptlib/cryptlib_cert.h
@@ -80,6 +80,25 @@ extern bool libspdm_x509_get_serial_number(const uint8_t *cert, size_t cert_size
                                            uint8_t *serial_number,
                                            size_t *serial_number_size);
 
+#if LIBSPDM_ADDITIONAL_CHECK_CERT
+/**
+ * Retrieve the signature algorithm from one X.509 certificate.
+ *
+ * @param[in]      cert       Pointer to the DER-encoded X509 certificate.
+ * @param[in]      cert_size  Size of the X509 certificate in bytes.
+ * @param[out]     oid        Signature algorithm Object identifier buffer.
+ * @param[in,out]  oid_size   Signature algorithm Object identifier buffer size.
+ *
+ * @retval  true    if the oid_size is equal 0, the cert parse successfully, but cert doesn't have signature algo.
+ * @retval  true    if the oid_size is not equal 0, the cert parse and get signature algo successfully.
+ * @retval  false   if the oid_size is equal 0, the cert parse failed.
+ * @retval  false   if the oid_size is not equal 0, the cert parse and get signature algo successfully, but the input buffer size is small.
+ **/
+extern bool libspdm_x509_get_signature_algorithm(const uint8_t *cert,
+                                                 size_t cert_size, uint8_t *oid,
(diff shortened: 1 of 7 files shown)


Zhiqiang520:
It looks good to me.

library/spdm_crypt_lib/libspdm_crypt_cert.c
```diff
@@ -1080,6 +1081,22 @@ bool libspdm_x509_common_certificate_check(const uint8_t *cert, size_t cert_size
         goto cleanup;
     }
 
+    /* 3. Verify signature algorithm. */
```
steven-bellock, line 1084: The other numbers need to be re-numbered, or maybe just removed entirely. Not sure how numbers help here.

steven-bellock (changes requested):
The other numbers need to be re-numbered, or

library/spdm_crypt_lib/libspdm_crypt_cert.c
Wenxing-hou: Thanks for your suggestions.  
Because my other PR change the file too, there will be some conflicts. 
I just make the smallest change in this PR.
I want to fix the number order after one PR merge.

steven-bellock: Ok.

jyao1:
Why this is a breaking change?

steven-bellock:
While it doesn't break the libspdm library build, libspdm now requires an implementation of `libspdm_x509_get_signature_algorithm`. Linking of `cryptlib`s that don't have this implementation will fail.

jyao1:
Since this is just to add additional check, I dont mind to add MACRO around `/* 3. Verify signature algorithm. */`

Wenxing-hou:
> Since this is just to add additional check, I dont mind to add MACRO around `/* 3. Verify signature algorithm. */`

Thanks. I have added the MACRO: `LIBSPDM_ADDITIONAL_CHECK_CERT`.

include/library/spdm_lib_config.h
```diff
@@ -410,4 +410,9 @@
 #define LIBSPDM_HAL_PASS_SPDM_CONTEXT 0
 #endif
 
+/* Enable additional checks for cert. */
+#ifndef LIBSPDM_ADDITIONAL_CHECK_CERT
+#define LIBSPDM_ADDITIONAL_CHECK_CERT 1
```
jyao1, line 415: Default should be 0.

Wenxing-hou: Thanks. I have changed the value.

jyao1 (approved):

include/hal/library/cryptlib/cryptlib_cert.h
```diff
@@ -80,6 +80,23 @@ extern bool libspdm_x509_get_serial_number(const uint8_t *cert, size_t cert_size
                                            uint8_t *serial_number,
                                            size_t *serial_number_size);
 
+/**
```
steven-bellock, line 83: Need `#if LIBSPDM_ADDITIONAL_CHECK_CERT`.

Wenxing-hou: Thanks. I have added the API under the MACRO.

PR #2717, opened 2024-06-03
Move CMake Min Requirement version to 3.5

alexteoh-intel:
Updating CMake min. required version to 3.5 in all CMake files, due to the following warning in build:
```
cmake version 3.28.3

CMake suite maintained and supported by Kitware (kitware.com/cmake).
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```
(Reference from Build-Linux CI job: https://github.com/DMTF/libspdm/actions/runs/8258546149/job/22590972198)

Cmake v3.5 has been around since 2016. This upgrade will also allow us to adopt modern CMake pattern (e.g. use of `target_sources` introduced in 3.1) in this repo.


steven-bellock (approved):
Yes, it's probably time.

jyao1 (approved):

PR #2718, opened 2024-06-03
Add length check during verify cert chain

### LINKED ISSUE #2701: Add length check when traverse certchain in libspdm_x509_verify_cert_chain
libspdm_x509_verify_cert_chain() assumes the input certchain is "One or more ASN.1 DER-encoded X.509 certificates" but it didn't traverse the whole certchain with the code below.
https://github.com/DMTF/libspdm/blob/ddfd7a1080cdbeffbe02267e6a164a0a3149be1a/os_stub/cryptlib_mbedtls/pk/x509.c#L711-L716

Should it add more check to make sure there is no more data left unchecked? E.g. 

if (ret != 0) { 
     if (current_cert < cert_chain + cert_chain_length)
           verify_flag = false;
     break; 
 }

rw8896:
Resolve #2701

### DIFF
--- a/os_stub/cryptlib_mbedtls/pk/x509.c
+++ b/os_stub/cryptlib_mbedtls/pk/x509.c
@@ -712,10 +712,18 @@ bool libspdm_x509_verify_cert_chain(const uint8_t *root_cert, size_t root_cert_l
             &tmp_ptr, cert_chain + cert_chain_length, &asn1_len,
             MBEDTLS_ASN1_CONSTRUCTED | MBEDTLS_ASN1_SEQUENCE);
         if (ret != 0) {
+            if (current_cert < cert_chain + cert_chain_length) {
+                verify_flag = false;
+            }
             break;
         }
 
         current_cert_len = asn1_len + (tmp_ptr - current_cert);
+
+        if (current_cert + current_cert_len > cert_chain + cert_chain_length) {
+            verify_flag = false;
+            break;
+        }
 
         if (libspdm_x509_verify_cert(current_cert, current_cert_len,
                                      preceding_cert,
@@ -799,6 +807,10 @@ bool libspdm_x509_get_cert_from_cert_chain(const uint8_t *cert_chain,
         }
 
         current_cert_len = asn1_len + (tmp_ptr - current_cert);
+        if (current_cert + current_cert_len > cert_chain + cert_chain_length) {
+            return false;
+        }
+
         current_index++;
 
         if (current_index == cert_index) {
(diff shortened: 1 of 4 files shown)


library/spdm_crypt_lib/libspdm_crypt_cert.c
```diff
@@ -1061,6 +1064,21 @@ bool libspdm_x509_common_certificate_check(const uint8_t *cert, size_t cert_size
     end_cert_from_len = 64;
     end_cert_to_len = 64;
 
+    /* 0. check certificate size */
```
steven-bellock, line 1067: Isn't this redundant? `libspdm_verify_cert_chain_data` already calls `ibspdm_x509_get_cert_from_cert_chain` before calling `libspdm_x509_certificate_check` and `libspdm_x509_common_certificate_check`.

rw8896: Yes, you are right. Looks like all the callers called `libspdm_x509_get_cert_from_cert_chain`. I guess I was confused by `libspdm_x509_set_cert_certificate_check_ex` unit tests, will remove /* 0. check certificate size */ and the two new tests in `libspdm_test_crypt_spdm_x509_set_cert_certificate_check_ex` below.

BTW, in  `libspdm_x509_get_cert_from_cert_chain`, do you think if I should also add a length check?
`if (current_cert + current_cert_len > cert_chain + cert_chain_length)`

rw8896: @steven-bellock Would you help review the latest commit? Thanks!

steven-bellock:
If possible the comparison should be exact.
1. Calculate the size of the entire certificate chain via x509 parsing.
2. If the size in step 1 is not equal to the size reported by the SPDM message then validation fails, else it passes.

os_stub/cryptlib_mbedtls/pk/x509.c
```diff
@@ -799,6 +806,9 @@ bool libspdm_x509_get_cert_from_cert_chain(const uint8_t *cert_chain,
         }
 
         current_cert_len = asn1_len + (tmp_ptr - current_cert);
+        if (current_cert + current_cert_len > cert_chain + cert_chain_length)
```
jyao1, line 809: please always add brace {} aound.

os_stub/cryptlib_openssl/pk/x509.c
```diff
@@ -2074,13 +2074,19 @@ bool libspdm_x509_verify_cert_chain(const uint8_t *root_cert, size_t root_cert_l
             (int *)&asn1_tag, (int *)&obj_class,
             (long)(cert_chain_length + cert_chain - tmp_ptr));
         if (asn1_tag != V_ASN1_SEQUENCE || ret & OPENSSL_ASN1_ERROR_MASK) {
+            if (current_cert < cert_chain + cert_chain_length)
```
jyao1, line 2077: please always add brace {} around.

rw8896:
> If possible the comparison should be exact.
> 
> 1. Calculate the size of the entire certificate chain via x509 parsing.
> 2. If the size in step 1 is not equal to the size reported by the SPDM message then validation fails, else it passes.

I think this commit already did exact checking:
1. During iterating the cert chain, check the x509 parsed length of each cert and make sure no cert  will exceed the cert chain.
2. After parsing all the certificates, check if there is any remaining data left in the cert chain.

Hence the total cert size can't exceed or be less than the whole cert chain length and must be exact the same as the cert chain length.

steven-bellock (approved):

PR #2720, opened 2024-06-04
Use lower-case function name convention in Cmake files

alexteoh-intel:
Modern CMake's convention uses lower case for function names. There was a mixed use of lower case and upper case in this repo. Doing a search and replace to fixed them.


steven-bellock:
From https://cliutils.gitlab.io/modern-cmake/chapters/intro/dodonot.html : 
>Use lowercase function names: CMake functions and macros can be called lower or upper case. Always use lower case. Upper case is for variables.

steven-bellock (approved):

PR #2721, opened 2024-06-06
Alistair/set cert fixup

alistair23:
When we run checks against the certificate that the requester set we have the following function calls
 - libspdm_set_cert_verify_certchain()
  - libspdm_x509_set_cert_certificate_check() ...
   - libspdm_verify_leaf_cert_spdm_extension()

At which point libspdm_verify_leaf_cert_spdm_extension() checks to make sure the id-DMTF-hardware-identity OID is not set if it's an AliasCert model.

This ends up being incorrect though. If using an AliasCert the SET_CERTIFICATE CertChain (table 93 - section 770) will "contain a partial certificate chain from the root CA to the Device Certificate CA". This means that the leaf certificate of that chain should set the the id-DMTF-hardware-identity OID as it isn't an alias certificate.

At this point the check in libspdm_verify_leaf_cert_spdm_extension() is incorrect.

Following up on https://github.com/DMTF/libspdm/pull/2708#issuecomment-2146372858 this PR fixes the current function documentation to ensure that the use of `is_requester_cert` is clear. This also skips leaf certificate checks if running set cert on a AliasCert model as we are actually checking a Device Certificate CA.


alistair23:
The `Build and Test / build (windows-latest, openssl, x64, Debug, GCC, -DLIBSPDM_ENABLE_CAPABILITY_CERT_CAP` failure is unrelated, seems to be a package mirror issue

library/spdm_crypt_lib/libspdm_crypt_cert.c
```diff
@@ -1147,11 +1149,13 @@ bool libspdm_x509_common_certificate_check(const uint8_t *cert, size_t cert_size
         goto cleanup;
     }
 
-    /* 10. verify spdm defined extension*/
-    status = libspdm_verify_leaf_cert_spdm_extension(cert, cert_size,
-                                                     is_requester_cert, cert_model);
-    if (!status) {
-        goto cleanup;
+    if (!(set_cert && cert_model == SPDM_CERTIFICATE_INFO_CERT_MODEL_ALIAS_CERT)) {
```
jyao1, line 1152: I don't think we want to verify this for SPDM_CERTIFICATE_INFO_CERT_MODEL_GENERIC_CERT, right?

Maybe change to `cert_model != SPDM_CERTIFICATE_INFO_CERT_MODEL_DEVICE_CERT`

alistair23: Correct! I forgot about generic certs. Fixed!

jyao1 (approved):

include/library/spdm_crypt_lib.h
```diff
@@ -920,7 +920,7 @@ bool libspdm_x509_certificate_check_ex(const uint8_t *cert, size_t cert_size,
  * @param[in]  cert_size             The size of certificate data in bytes.
  * @param[in]  base_asym_algo        SPDM base_asym_algo
  * @param[in]  base_hash_algo        SPDM base_hash_algo
- * @param[in]  is_requester_cert     Is the function verifying requester or responder cert.
+ * @param[in]  is_requester_cert     Is the function verifying a cert as a requester or responder.
```
steven-bellock, line 923: The parameter name would need to change back to `is_requester`.

library/spdm_crypt_lib/libspdm_crypt_cert.c
```diff
@@ -1147,11 +1149,13 @@ bool libspdm_x509_common_certificate_check(const uint8_t *cert, size_t cert_size
         goto cleanup;
     }
 
-    /* 10. verify spdm defined extension*/
-    status = libspdm_verify_leaf_cert_spdm_extension(cert, cert_size,
-                                                     is_requester_cert, cert_model);
-    if (!status) {
-        goto cleanup;
+    if ((!set_cert) || (cert_model == SPDM_CERTIFICATE_INFO_CERT_MODEL_DEVICE_CERT)) {
```
steven-bellock, line 1152: If it's
- `SET_CERTIFICATE` AND
- Responder supports `ALIAS_CERT` AND
- Device Certificate CA contains a hardware identity

then libspdm should check that it matches the SPDM hardware identity OID in `libspdm_verify_leaf_cert_spdm_extension`.

alistair23: > * `SET_CERTIFICATE` AND
> 
> * Responder supports `ALIAS_CERT` AND
> 
> * Device Certificate CA contains a hardware identity

This is invalid, "If the AliasCert model is used, this OID shall not be present in any alias certificates in the chain".

Are you saying we should check that the hardware identity OID isn't included?

steven-bellock: image
Alias certificates start right after the Device Certificate CA. In particular
>When the AliasCert model is in use, the Device Certificate is referred to as a Device Certificate CA, indicating that the certificate both contains device hardware identity information and functions as a certificate authority to sign an additional certificate. In this model, the Device Certificate CA should contain the [Hardware identity OID](https://github.com/DMTF/SPDM-WG/blob/master/docs/DSP0274/SPDMSpecification.md#hardware-identity-oid).

But reading that again it's a `should` not a `shall`. However `libspdm_verify_leaf_cert_spdm_extension` can be used in the future for other checks besides the hardware identity. Maybe add the `set_cert` parameter to `libspdm_verify_leaf_cert_spdm_extension` and move this logic to the function. That way `libspdm_verify_leaf_cert_spdm_extension` can be called unconditionally and all the set_cert / cert_model logic can be handled in the function.

include/library/spdm_crypt_lib.h
```diff
@@ -938,7 +938,7 @@ bool libspdm_x509_set_cert_certificate_check(const uint8_t *cert, size_t cert_si
...
```
steven-bellock, line 941: This should be `is_requester`?

library/spdm_crypt_lib/libspdm_crypt_cert.c
jyao1: I dont think we need such change at this moment.

`libspdm_verify_leaf_cert_spdm_extension` is to verify "leaf_cert", but in set_cert case, the last one is not leaf_cert.

I think current code is good enough - just skip the leaf_cert verification.

steven-bellock: > but in set_cert case, the last one is not leaf_cert.

libspdm does check the Device Certificate CA's CA field via https://github.com/DMTF/libspdm/blob/0ef69e7919f6949b39355849b2116b33d8167334/library/spdm_crypt_lib/libspdm_crypt_cert.c#L1326-L1328 The documentation for that function is not correct. I'll correct it in a pull request.

>I think current code is good enough - just skip the leaf_cert verification.

Ok but long-term this should be refactored as the logic is starting to get confusing.

PR #2725, opened 2024-06-11
Retire use of global functions in cmake files

alexteoh-intel:
**Summary**
This PR retires the use of global functions like include_directories and add_compile_options in CMake files. 
I'm introducing this template for building library or executable:
```
cmake_minimum_required(VERSION 3.5)

# or add_executable() for unittests
add_library(<target_name> STATIC "")

target_include_directories(<target_name>
    PRIVATE
        <directories>
)

target_sources(<target_name>
    PRIVATE
        <sources>
)

target_compile_options(<target_name> PRIVATE <options, e.g. -Wno-incompatible-pointer-types>)
```

**Background**
Coding style here is based on [Effective Modern CMake](https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1), which was referred by https://cliutils.gitlab.io/modern-cmake/chapters/intro/dodonot.html. 

Two relevant blobs that motivated this PR are: 
> Forget the commands add_compile_options, include_directories, link_directories, link_libraries.
Those commands operate on the directory level. All targets defined on that level inherit those properties. This increases the chance of hidden dependencies. Better operate on the targets directly.

> Avoid custom variables in the arguments of project commands.
Keep things simple. Don't introduce unnecessary custom variables. Instead of add_library(a ${MY_HEADERS} ${MY_SOURCES}), do add_library(a b.h b.cpp)

**Testing**
I included this version of libspdm in another CMake project via `add_subdirectory()`. It built fine for both C and CPP based projects. The programs worked as expected too. 

Please note that this change here isn't necessary for libspdm to be used in a parent project. As it is today, I can already use libspdm as sub project in my cmake project. I previously thought more changes were needed, but it turned out there was an ordering issue with `target_link_libraries`'ing multiple static libraries (from libspdm). Linker searches from left to right to resolve symbols. After fixing the order of libspdm libraries, cmake build started to work.


steven-bellock:
@alexteoh-intel did you use a script to make these changes?

alexteoh-intel:
> @alexteoh-intel did you use a script to make these changes?

Hi Steven, I did this manually with some search & replace help in vscode. It's a big change I know. If you prefer a smaller change that touches only the library cmake files, that's fine by me too. Thanks!

steven-bellock (approved):

jyao1:
@alexteoh-intel, do you want to propose similar change for other project, such as https://github.com/DMTF/spdm-emu, https://github.com/DMTF/spdm-dump, https://github.com/DMTF/SPDM-Responder-Validator?

alexteoh-intel:
> @alexteoh-intel, do you want to propose similar change for other project, such as https://github.com/DMTF/spdm-emu, https://github.com/DMTF/spdm-dump, https://github.com/DMTF/SPDM-Responder-Validator?

Hi @jyao1, sure I can look into them when I get a chance.

Issue #2697, opened 2024-05-13
The code should not call `libspdm_release_receiver_buffer( )` before calling `libspdm_acquire_receiver_buffer( )`.


Zhiqiang520:
The code should not call `libspdm_release_receiver_buffer( )` before calling `libspdm_acquire_receiver_buffer( )`.

`goto done` at the line 60 & 126 will call `libspdm_release_receiver_buffer( )` function at line 220, but in fact `libspdm_acquire_receiver_buffer( )` has not yet run before `goto done` at the line 60 & 126.

https://github.com/DMTF/libspdm/blob/999ed70ee185c065ae09fed7b85c69a19179fcf1/library/spdm_requester_lib/libspdm_req_vendor_request.c#L50-L61
https://github.com/DMTF/libspdm/blob/999ed70ee185c065ae09fed7b85c69a19179fcf1/library/spdm_requester_lib/libspdm_req_vendor_request.c#L120-L127
https://github.com/DMTF/libspdm/blob/999ed70ee185c065ae09fed7b85c69a19179fcf1/library/spdm_requester_lib/libspdm_req_vendor_request.c#L131-L135
https://github.com/DMTF/libspdm/blob/999ed70ee185c065ae09fed7b85c69a19179fcf1/library/spdm_requester_lib/libspdm_req_vendor_request.c#L219-L221

PR #2729, opened 2024-06-17
Fix wrong release_receiver_buffer before acquire

Wenxing-hou:
Fix: #2697

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_vendor_request.c
+++ b/library/spdm_requester_lib/libspdm_req_vendor_request.c
@@ -56,8 +56,7 @@ libspdm_return_t libspdm_try_vendor_send_request_receive_response(
         resp_size == NULL ||
         (*resp_size != 0 && resp_data == NULL)
         ) {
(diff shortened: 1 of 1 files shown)


jyao1 (approved):

library/spdm_requester_lib/libspdm_req_vendor_request.c
```diff
@@ -57,7 +57,7 @@ libspdm_return_t libspdm_try_vendor_send_request_receive_response(
         (*resp_size != 0 && resp_data == NULL)
         ) {
         status = LIBSPDM_STATUS_INVALID_PARAMETER;
-        goto done;
+        return status;
```
steven-bellock, line 60: Just `return LIBSPDM_STATUS_INVALID_PARAMETER;`

steven-bellock (changes requested):

library/spdm_requester_lib/libspdm_req_vendor_request.c
Wenxing-hou: Thanks. I have changed the code.

PR #2748, opened 2024-07-07
Fix SET_CERTIFICATE issues

### LINKED ISSUE #2746: `SET_CERTIFICATE.Param1` case is not handled
`Param1` is set in https://github.com/DMTF/libspdm/blob/883c31e794ea49b8d4203012e9b1e88d9873e82d/library/spdm_requester_lib/libspdm_req_set_certificate.c#L118-L139 It does not handle the case when `Erase` is not set. In such a case `Param1` is just the slot ID, which ignores the certificate model field.

steven-bellock:
Fix #2746.


jyao1 (approved):

jyao1:
@steven-bellock , I notice spdm-emu still failed after this patch.

Issue #1211, opened 2022-09-14
Secondary macros should not be required if capabilities are disabled


steven-bellock:
If a capability is disabled, like `LIBSPDM_ENABLE_CAPABILITY_PSK_EX_CAP` then macros that depend on that capability, like `LIBSPDM_PSK_MAX_HINT_LENGTH`, should not be required to be defined, and compilation and tests should be successful in their absence.

jyao1:
9/19/2022 - steven will propose a fix.

steven-bellock:
Also `LIBSPDM_MAX_SPDM_MSG_SIZE` and `LIBSPDM_DATA_TRANSFER_SIZE` should not be required when implementing a `1.0 / 1.1` endpoint.

PR #2760, opened 2024-07-09
Don't require PSK macros if PSK is not enabled

steven-bellock:
If `LIBSPDM_ENABLE_CAPABILITY_PSK_EX_CAP` is not enabled then `LIBSPDM_PSK_CONTEXT_LENGTH` and `LIBSPDM_PSK_MAX_HINT_LENGTH` are still required for a successful build. This pull request fixes that as a subtask of #1211.

### DIFF
--- a/include/library/spdm_lib_config.h
+++ b/include/library/spdm_lib_config.h
@@ -7,6 +7,121 @@
 #ifndef SPDM_LIB_CONFIG_H
 #define SPDM_LIB_CONFIG_H
 
+/* Code space optimization for optional messages.
+ *
+ * An Integrator of libspdm may not need all of the optional SPDM messages. The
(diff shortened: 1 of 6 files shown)


include/library/spdm_lib_config.h
```diff
@@ -7,6 +7,121 @@
 #ifndef SPDM_LIB_CONFIG_H
 #define SPDM_LIB_CONFIG_H
 
+/* Code space optimization for optional messages.
```
jyao1, line 10: why move the location?

steven-bellock: For `#if LIBSPDM_ENABLE_CAPABILITY_PSK_CAP` in line 151.

PR #2761, opened 2024-07-10
Update version after 3.4 release

steven-bellock:


steven-bellock:
@jyao1 this should obviously be merged before the other pull requests.

PR #2782, opened 2024-07-30
Add DOE Discovery Version

### LINKED ISSUE #2783: DOE Discovery Version shall be supported in pci_doe_discovery
DOE Discovery Version is introduced in PCIE Spec 6.1 Section 6.30.1.1. libspdm shall support this version as well.

mxu9:
Fix #2783 

DOE Discovery Version is first introduced in PCIE Spec 6.1 Section 6.30.1.1.

libspdm_pci_doe_decode_discovery_request is unchanged. New API libspdm_pci_doe_decode_discovery_request_version is added to get the version from doe_discovery_request. This is to prevent the building broken for the exsting codes.

### DIFF
--- a/library/spdm_transport_pcidoe_lib/libspdm_doe_pcidoe.c
+++ b/library/spdm_transport_pcidoe_lib/libspdm_doe_pcidoe.c
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2021-2022 DMTF. All rights reserved.
+ *  Copyright 2021-2024 DMTF. All rights reserved.
  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
  **/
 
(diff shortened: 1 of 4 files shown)


library/spdm_transport_pcidoe_lib/libspdm_doe_pcidoe.c
```diff
@@ -301,6 +301,72 @@ libspdm_return_t libspdm_pci_doe_decode_discovery_request(size_t transport_messa
...
+    if (pci_doe_header->length >= PCI_DOE_MAX_SIZE_IN_DW) {
+        return LIBSPDM_STATUS_INVALID_MSG_SIZE;
+    } else if (pci_doe_header->length == 0) {
+        length = PCI_DOE_MAX_SIZE_IN_BYTE;
+    } else {
+        length = pci_doe_header->length * sizeof(uint32_t);
+    }
+    if (length != transport_message_size) {
+        return LIBSPDM_STATUS_INVALID_MSG_SIZE;
+    }
```
alistair23, line 365: It seems a bit unnecessary to do all of these checks again. Is it not worth modifying ` libspdm_pci_doe_decode_discovery_request()` to support the `version` field?

Or even just calling ` libspdm_pci_doe_decode_discovery_request()` from `libspdm_pci_doe_decode_discovery_request_version()` so we don't have to duplicate all of these checks

mxu9: > It seems a bit unnecessary to do all of these checks again. Is it not worth modifying ` libspdm_pci_doe_decode_discovery_request()` to support the `version` field?
> 
To support the version field in libspdm_pci_doe_decode_discovery_request() will add a new in/out parameter. That will break the existing code which calls libspdm_pci_doe_decode_discovery_request(). Furthermore, "version" field doesn't exist in DOE Discovery Request before PCIE Spec 6.1.

> Or even just calling ` libspdm_pci_doe_decode_discovery_request()` from `libspdm_pci_doe_decode_discovery_request_version()` so we don't have to duplicate all of these checks
>
I am OK to call libspdm_pci_doe_decode_discovery_request()` from libspdm_pci_doe_decode_discovery_request_version()`

alistair23 (approved):

PR #2795, opened 2024-08-06
Make buffers void instead of uint8_t

steven-bellock:
If pointer arithmetic is needed then they can be cast to the appropriate type.


Wenxing-hou (approved):
LGTM.

Zhiqiang520 (approved):
Looks good to me.

PR #2797, opened 2024-08-08
Fix default build options for initialize_secured_message_context()

kelvin-cao:
In initialize_secured_message_context(), two for-loops access two arrays
with LIBSPDM_MAX_AEAD_KEY_SIZE number of elements and with
LIBSPDM_MAX_AEAD_IV_SIZE number of elements separately. Both iterations
go beyond the size of the arrays when the required macro
LIBSPDM_AEAD_AES_256_GCM_SUPPORT for the test is undefined.

Gcc 11.4.0 reports this and aborts the compilation when -DTARGET=Release
is specified with cmake for the build. Check macro define in the unit
test to avoid compile error.

### DIFF
--- a/unit_test/test_spdm_secured_message/encode_decode.c
+++ b/unit_test/test_spdm_secured_message/encode_decode.c
@@ -7,6 +7,8 @@
 #include "spdm_unit_test.h"
 #include "internal/libspdm_secured_message_lib.h"
 
+#if LIBSPDM_AEAD_AES_256_GCM_SUPPORT
+
 static uint8_t m_secured_message[0x1000];
 static uint8_t m_app_message[0x1000];
 static libspdm_secured_message_context_t m_secured_message_context;
@@ -925,3 +927,5 @@ int libspdm_secured_message_encode_decode_test_main(void)
                                   libspdm_unit_test_group_setup,
                                   libspdm_unit_test_group_teardown);
 }
+
+#endif /* LIBSPDM_AEAD_AES_256_GCM_SUPPORT */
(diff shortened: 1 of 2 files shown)


steven-bellock:
`libspdm_secured_message_encode_decode_test_main` requires `LIBSPDM_AEAD_AES_256_GCM_SUPPORT`. If that algorithm is not present then the test should not be run. You can put `#if LIBSPDM_AEAD_AES_256_GCM_SUPPORT` around https://github.com/DMTF/libspdm/blob/b28434aa3620d45a859164253d5bdbf1a8376ccd/unit_test/test_spdm_secured_message/test_spdm_secured_message.c#L12-L16 and https://github.com/DMTF/libspdm/blob/b28434aa3620d45a859164253d5bdbf1a8376ccd/unit_test/test_spdm_secured_message/encode_decode.c#L9 That way the test will vacuously pass if executed and will resolve the compiler warning.

kelvin-cao:
It makes sense. Updated.

steven-bellock:
@kelvin-cao did you add a compilation flag to get this warning?

steven-bellock (approved):

kelvin-cao:
> @kelvin-cao did you add a compilation flag to get this warning?

No, just turned off some algorithm macros in include/library/spdm_lib_config.h, and used the command for build as following:
cmake -DARCH=x64 -DTOOLCHAIN=GCC -DTARGET=Release -DCRYPTO=openssl -DENABLE_BINARY_BUILD=0 ..
make

PR #2802, opened 2024-08-14
Mask capability flags based on negotiated version

### LINKED ISSUE #2796: Capability flags should be masked with version mask
https://github.com/DMTF/libspdm/blob/3a69c7bb0faccb2c5ac75c52d2d905fcf92d5510/library/spdm_responder_lib/libspdm_rsp_capabilities.c#L250C5-L250C73

Capability flags actually need to be masked with version mask(e.g. SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_12_MASK).
For example, if the responder supports multiple versions and different/new capabilities, the integrator would expect to set all the supported flags and rely on libspdm to report the flags according to version negotiation results.

steven-bellock:
Fix #2796.

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -1711,4 +1711,17 @@ static inline uint64_t libspdm_le_to_be_64(uint64_t value)
             ((value & 0xff00000000000000) >> 56));
 }
 
+/**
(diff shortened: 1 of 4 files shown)


steven-bellock:
@rw8896

PR #2807, opened 2024-08-16
Move unit test context inside _main

### LINKED ISSUE #2780: the vendor_cmds requester unit_test has some error
When any requester unit_test is called after the vendor_cmds requester unit_test, the GCC release build failed.

There is a example PR:
https://github.com/Wenxing-hou/libspdm/pull/17

steven-bellock:
Fix #2780.

This was tested locally.


steven-bellock:
So that fixed GCC on Linux but broke GCC on Windows. Back to the drawing board.

PR #2809, opened 2024-08-16
Setup nightly Coverity scanning

### LINKED ISSUE #1007: Add Coverity to CI
Perhaps we can use https://github.com/marketplace/actions/unofficial-coverity-scan to perform nightly / weekly scans.

steven-bellock:
Fix #1007.


steven-bellock:
The run is failing at downloading the libspdm-specific build file. I regenerated the token so maybe it takes a while to propagate? Will try again tomorrow.

PR #2823, opened 2024-09-02
Refactor unit tests

steven-bellock:
1. Move the test context inside the _main function, as it is not needed outside the function.
2. Make functions static where appropriate.


steven-bellock:
For the tests to pass #2822 needs to go in first.

steven-bellock:
@jyao1 branch has been rebased.

PR #2827, opened 2024-09-05
transport: storage: add support for SPDM over the storage binding (DSP0286)

twilfredo:
This series adds `libspdm` transport support for SPDM over Storage Binding Specification [1]. This allows for example NVMe or SCSI devices that conform to [1] to communicate with a host requester using the storage transport medium as defined in [1].

[1] https://www.dmtf.org/sites/default/files/standards/documents/DSP0286_1.0.0WIP90.pdf


twilfredo:
@alistair23

steven-bellock:
A couple of high level comments.
1. This pull request will not be able to be merged until DSP0286 has been published, but that looks to be soonish.
2. The WDC-authored files have a WDC copyright. I believe (not 100 % sure) that per the "DMTF Members’ Software Submission Policy" it should have the DMTF copyright. If you want to keep the WDC copyright we can talk to the DMTF open source folks to clarify things.

twilfredo:
> A couple of high level comments.
> 
>     1. This pull request will not be able to be merged until DSP0286 has been published, but that looks to be soonish.
Yeah that sounds good, just wanted to get comments early on
> 
>     2. The WDC-authored files have a WDC copyright. I believe (not 100 % sure) that per the "DMTF Members’ Software Submission Policy" it should have the DMTF copyright. If you want to keep the WDC copyright we can talk to the DMTF open source folks to clarify things.

I've updated the license header with an author tag, is that sufficient?

steven-bellock:
Presumably authorship is recorded in the commit log, but if you want to put it in the file itself:
```
/**
 *  Copyright Notice:
 *  Copyright 2024 DMTF. All rights reserved.
 *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
 **/
 
 /**
  *  Author: Wilfred Mallawa <wilfred.mallawa@wdc.com>
  *          Alistair Francis <alistair.francis@wdc.com>
  **/
 ```

**Review by steven-bellock — CHANGES_REQUESTED** [review]

File: include/industry_standard/storage.h [hunk 1]
```diff
@@ -0,0 +1,50 @@
...
+
+#ifndef STORAGE_BINDING_H
+#define STORAGE_BINDING_H
+
+#pragma pack(1)
+
+typedef struct {
+    uint16_t data_length;
+    uint16_t storage_binding_version;
+    uint8_t max_connection_id : 2;
```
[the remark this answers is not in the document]
twilfredo: okay sounds good, do we want to wait on this until the spec is updated then?

include/industry_standard/storage.h
```diff
@@ -0,0 +1,50 @@
...
+
+#define STORAGE_SECURITY_BINDING_VERSION 0x1000
+#define STORAGE_SECURITY_PROTOCOL_DMTF 0xE8
+
+#define STORAGE_OPERATION_CODE_DISCOVERY        0x01
+#define STORAGE_OPERATION_CODE_PENDING_INFO     0x02
+#define STORAGE_OPERATION_CODE_MESSAGE          0x05
+#define STORAGE_OPERATION_CODE_SECURED_MESSAGE  0x06
+
+#define STORAGE_MAX_SIZE_IN_BYTE 0x00100000
```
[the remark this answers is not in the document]
twilfredo: This is not from the spec, we used the DoE transport for layer as a reference for this. https://github.com/DMTF/libspdm/blob/b327ead227f14998789bca7d045af17b99dd6d50/include/industry_standard/pcidoe.h#L40

library/spdm_transport_storage_lib/libspdm_storage.c
```diff
@@ -0,0 +1,816 @@
...
+    if (transport_message_size == 0) {
+        return LIBSPDM_STATUS_INVALID_MSG_SIZE;
+    }
+
+    storage_header = transport_message;
+    if (storage_header->security_protocol != STORAGE_SECURITY_PROTOCOL_DMTF) {
+        return LIBSPDM_STATUS_INVALID_MSG_FIELD;
+    }
+
+#if CMAKE_C_BYTE_ORDER == LITTLE_ENDIAN
```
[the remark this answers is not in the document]
twilfredo: It was to be consistent with the byte ordering defined in the SCSI Primary Commands specification (SPC) where BE is used for the command blocks. The SPC is referred to by the NVMe spec for the Security In/Out implementation also.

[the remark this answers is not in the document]
twilfredo: For context in `qemu`, we will generate the header as such: https://github.com/twilfredo/qemu/blob/d40afa3ad43417d01cf7a6e758a434e3858d0200/hw/nvme/ctrl.c#L7124

include/industry_standard/storage.h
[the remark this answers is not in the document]
steven-bellock: Yes, we'll probably discuss this issue at the VF2F meeting today.

include/industry_standard/storage.h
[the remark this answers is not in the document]
steven-bellock: The PCIe specification explicitly defines those values. If they are not in DSP0286 or an underlying specification then should be removed.

library/spdm_transport_storage_lib/libspdm_storage.c
```diff
@@ -0,0 +1,819 @@
...
+ *          Alistair Francis <alistair.francis@wdc.com>
+ **/
+
+#include "library/spdm_transport_storage_lib.h"
+#include "industry_standard/storage.h"
+#include "internal/libspdm_common_lib.h"
+#include "hal/library/debuglib.h"
+#include "hal/library/memlib.h"
+
+# define BSWAP32(x)  ((((x) & 0x000000ff) << 24) |      \
```
steven-bellock, line 18: https://github.com/DMTF/libspdm/blob/b327ead227f14998789bca7d045af17b99dd6d50/include/internal/libspdm_common_lib.h#L1705 already exists in a header. I would rename these to `libspdm_byte_swap_*` and place them there. `libspdm_le_to_be_64` can be renamed to `libspdm_byte_swap_64` in a subsequent pull request to make things consistent.

library/spdm_transport_storage_lib/libspdm_storage.c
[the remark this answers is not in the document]
twilfredo: The ATA Spec specifies the `trusted send/recv` commands to be a 28bit command block in LE. Since this is a virtual header we are parsing here, would it be best to use LE to stay consistent with SPDM message fields? What do we think?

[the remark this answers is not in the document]
steven-bellock: @twilfredo we can discuss that in https://github.com/DMTF/SPDM-WG/issues/3625.

jyao1:
Usually, the implementation for WIP should NOT be posted to main, but a WIP specific branch
For example, StorageBinding_WIP, in this case.

twilfredo:
> Usually, the implementation for WIP should NOT be posted to main, but a WIP specific branch For example, StorageBinding_WIP, in this case.
> 
> The WIP branch can be merged into main, after the WIP spec becomes official standard.

Okay noted, are you happy to keep this up for now, since review is in progress.

jyao1:
> Okay noted, are you happy to keep this up for now? since review is in progress.

That is fine. We can keep reviewing and address all feedback in this PR.

Whenever you think it is done and ready for merge, you can switch to branch merge request.

include/industry_standard/storage.h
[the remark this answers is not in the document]
jyao1: right. Using bitfield is not recommended.

I recommend you just using `uint8_t max_connection_id`, then use bitmask MACRO, such as `SPDM_STORAGE_MAX_CONNECTION_ID_MASK 0x3` to get the value.

include/industry_standard/storage.h
```diff
@@ -0,0 +1,50 @@
...
+#pragma pack(1)
+
+typedef struct {
+    uint16_t data_length;
+    uint16_t storage_binding_version;
+    uint8_t max_connection_id : 2;
+    uint8_t reserved1[3];
+    uint8_t supported_operations[8];
+    uint8_t reserved2[16];
+} storage_discovery_response_t;
```
jyao1, line 28: If this is SPDM storage binding spec, please always use `spdm_storage_` as prefix, as namespace.

include/industry_standard/storage.h
```diff
@@ -0,0 +1,50 @@
...
+
+typedef struct {
+    uint16_t data_length;
+    uint16_t storage_binding_version;
+    uint32_t pending_info_flag;
+    uint32_t response_length;
+} storage_pending_info_response_t;
+
+#define STORAGE_SECURITY_BINDING_VERSION 0x1000
```
jyao1, line 37: If this is SPDM storage binding spec, please always use `SPDM_STORAGE_` as prefix, as namespace.

include/industry_standard/storage.h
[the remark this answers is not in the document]
jyao1: right. only SPDM storage binding spec defined data should be here.

Any other implementation definition should NOT be added here.

include/industry_standard/storage.h
```diff
@@ -0,0 +1,50 @@
...
+/**
+ *  Author: Wilfred Mallawa <wilfred.mallawa@wdc.com>
+ *          Alistair Francis <alistair.francis@wdc.com>
+ **/
+
+/** @file
+ * Definitions of SPDM over the Storage as defined in DSP0286
+ **/
+
+#ifndef STORAGE_BINDING_H
```
jyao1, line 16: Can we add `spdm_` as prefix for the file name, such as `spdm_storage_binding.h`?

include/industry_standard/storage.h
```diff
@@ -0,0 +1,50 @@
+/**
+ *  Copyright Notice:
+ *  Copyright 2024 DMTF. All rights reserved.
+ *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
+ **/
+
+/**
+ *  Author: Wilfred Mallawa <wilfred.mallawa@wdc.com>
```
jyao1, line 8: Usually, we dont put author name in this project.

library/spdm_transport_storage_lib/libspdm_storage.c
```diff
@@ -0,0 +1,819 @@
...
+ *  Copyright 2024 DMTF. All rights reserved.
+ *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
+ **/
+
+/**
+ *  Author: Wilfred Mallawa <wilfred.mallawa@wdc.com>
+ *          Alistair Francis <alistair.francis@wdc.com>
+ **/
+
+#include "library/spdm_transport_storage_lib.h"
```
jyao1, line 12: Do you have any unit test associated to prove the correctness of this function?

twilfredo: I haven't written any explicit test cases for all the functions in this files. We have tested the logic by testing with QEMU + `spdm-utils` + linux (Kernel SPDM) though. I'm waiting for the spec to become ratified so we can add some of the changes discussed above.

I can add some tests if they are required? Are there exists tests for other transports?

jyao1: Usually, we will need two level of test:
1) Unit test in libspdm project.
2) System test in spdm-emu project.

In history, both tests can catch some error.

Please consider to add both.

jyao1:
@twilfredo , do you want to keep it in PR? or do you want to push to a branch, and keep updating the branch?

I am OK either way. But I just feel a branch might be more helpful.

twilfredo:
> @twilfredo , do you want to keep it in PR? or do you want to push to a branch, and keep updating the branch?
> 
> I am OK either way. But I just feel a branch might be more helpful.

I'll stick to the PR is that's okay with everyone. I'm just waiting for the spec to be ratified so I can address all the changes in one go.

twilfredo:
What is the correct way to generate a seed for the fuzzing tests to go here `unit_test/fuzzing/seeds` mapping to the respective tests.

As requested by @jyao1 I have added some unit tests to test the storage transport api, but I'm unsure about how the seeds are generated (hence why some of the the CI is failing also).

twilfredo:
ping!

twilfredo:
> @twilfredo , would you please:
> 
>     1. fix CI failure.
> 
>     2. Check if it aligns to latest published [DSP0286](https://www.dmtf.org/sites/default/files/standards/documents/DSP0286_1.0.0.pdf)

Good to see the spec being released! and yeah I will work on getting this upto date. Thanks!

include/library/spdm_transport_storage_lib.h
```diff
@@ -0,0 +1,244 @@
...
+ * This structure is publicly defined to provide transport encoding information
+ * to the caller from transport_message buffer(s).
+ */
+#pragma pack(1)
+typedef struct {
+    uint8_t security_protocol;
+    uint16_t security_protocol_specific;
+    bool inc_512;
+    uint32_t length;
+} storage_spdm_transport_header;
```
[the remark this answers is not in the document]
twilfredo: This was defined to allow upper level applications that use `libspdm` to extract the needed information to make storage API calls. That is these fields are required when using the NVMe/SCSI/ATA api to interact with devices. It is not part of the SPDM message.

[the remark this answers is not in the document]
jyao1: Current storage_spdm_transport_header is only align to SCSI (table 17). What about NVME (table 13) and ATA (table 22)?

unit_test/fuzzing/test_transport/test_spdm_transport_storage_encode_message/spdm_transport_storage_encode_message.c
```diff
@@ -0,0 +1,317 @@
...
+    LIBSPDM_ASSERT(ret == LIBSPDM_STATUS_SUCCESS);
+    LIBSPDM_ASSERT(allocation_len == 0);
+
+    /* Trivial Compliance Assertions */
+    hdr = transport_message;
+    LIBSPDM_ASSERT(hdr->security_protocol == SPDM_STORAGE_SECURITY_PROTOCOL_DMTF);
+    LIBSPDM_ASSERT((hdr->security_protocol_specific >> 8) == 0); /* SPSP1 */
+    LIBSPDM_ASSERT((hdr->security_protocol_specific & 0xFF) != 0); /* SPSP0 */
+    LIBSPDM_ASSERT(hdr->inc_512 == false);
+    LIBSPDM_ASSERT(hdr->length == sizeof(storage_spdm_transport_header));
```
jyao1, line 146: if storage_spdm_transport_header is not part of message, why we need sizeof here?

library/spdm_transport_storage_lib/libspdm_storage.c
```diff
@@ -0,0 +1,792 @@
...
+    LIBSPDM_ASSERT(transport_message_size >= sizeof(storage_spdm_transport_header));
+    if (transport_message_size <= sizeof(storage_spdm_transport_header)) {
+        return LIBSPDM_STATUS_INVALID_MSG_SIZE;
+    }
+
+    if (transport_message_size == 0) {
+        return LIBSPDM_STATUS_INVALID_MSG_SIZE;
+    }
+
+    storage_header = transport_message;
```
jyao1, line 99: I am confused here. If you say "It is not part of the SPDM message.", then why it points to transport_message and fill data.

include/library/spdm_transport_storage_lib.h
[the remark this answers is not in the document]
twilfredo: For NVME and ATA, this header still contains the relevant information. For example:
For NVMe, SPS1/0 will be in `security_protocol_specific` and `inc 512` shall be ignored.
For ATA, SP SPECIFIC is again in `security_protocol_specific`. Allocation length will at minimum `length` but that is upto a higher layer decide. `inc 512` shall be ignored.

This way we have a common header, and an upper layer can extract the necessary information for NVMe/SCSI and ATA as required.

library/spdm_transport_storage_lib/libspdm_storage.c
twilfredo: Ah let me clarify, by "It is not part of the SPDM message" I meant for example, for a request, that header does not go through spdm data buffer in the storage api, but instead the storage transport layer. For reference: [this](https://github.com/westerndigitalcorporation/spdm-utils/pull/139/files#diff-90dda305f29d7ba03588448fa907041edf34fae37e66558e2465dd555b00272cR360) is how it would be used by an upper level application using `libspdm`

include/library/spdm_transport_storage_lib.h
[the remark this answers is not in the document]
jyao1: that is fine. If the purpose of the `storage_spdm_transport_header` structure is to provide common info, then
1) We should not put pack(1) around the structure. (Also, you can keep bool.)
2) We should not use sizeof().
3) We should not point to a memory buffer with this structure and fill data there.

See my rest comment.

library/spdm_transport_storage_lib/libspdm_storage.c
```diff
@@ -0,0 +1,792 @@
...
+                spdm_context, *session_id);
+        if (secured_message_context == NULL) {
+            return LIBSPDM_STATUS_UNSUPPORTED_CAP;
+        }
+
+        /* Message to secured message*/
+        sec_trans_header_size = is_request_message ?
+                                sizeof(storage_spdm_transport_header): 0;
+        /* DSP0286 Specifies 4 Reserved bytes at the start of a secured message */
+        sec_trans_header_size += sizeof(uint32_t);
```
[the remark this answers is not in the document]
twilfredo: That will need to be added. I have missed that.

include/library/spdm_transport_storage_lib.h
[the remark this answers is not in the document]
twilfredo: > that is fine. If the purpose of the `storage_spdm_transport_header` structure is to provide common info, then
> 
>     1. We should not put pack(1) around the structure. (Also, you can keep bool.)
> 
>     2. We should not use sizeof().
> 
>     3. We should not point to a memory buffer with this structure and fill data there.
> 
What alternatives are you suggesting? In that, how best otherwise to have this information be returned back to a callee appropriately to generate API calls, and also make sure we can decode incoming transport headers.
> 
> See my rest comment.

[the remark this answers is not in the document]
twilfredo: > Maybe I am still not clear on what the purpose of this `storage_spdm_transport_header` structure.
> 
> Let's clarify the purpose at first:
> 
>     1. The structure is used for a message sent in the wire. It must be packed and must align with data structure definition in spec.
> 
>     2. The structure is NOT for a message sent in the wire. It does not need to be packed, and it can be defined by libspdm implementation, and can be changed anytime if we want.
> 
> 
> Please help me understand which one in this case,

Point 2 is correct. One of the contributing reasons to having this packed is:

- With QEMU, SPDM uses a socket infrastructure to communicate data between and emulated SPDM responder and requester through the SPDM storage(NVMe/ATA support exists, yet to be up-streamed and this will likely be the first user of SPDM Storage) and PCIe DoE (upstream support exists) transports. That means the `storage_spdm_transport_header` is transmitted over sockets and can be **directly** passed into `libspdm`  for parsing.  See [here](https://patchew.org/QEMU/20250116010856.95115-2-wilfred.mallawa@wdc.com/20250116010856.95115-3-wilfred.mallawa@wdc.com/) for the RFC series. Note, that series is an RFC and will also need to be updated to match any new spec changes since.

- My thinking was that there maybe other users who use a similar implementations as well. In which case, it seemed best to keep it packed and to treat it as a memory buffer. However, if you think there's a better approach I'm happy to change things.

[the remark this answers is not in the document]
twilfredo: > Thanks for the clarification. I appreciate that.
> 
> Here is my understanding based on your information. Please correct me if I am wrong.
> 
>     1. `storage_spdm_transport_header` is NOT for the data packet defined in any industry storage (NVME, SCSI, ATA) spec, or SPDM storage binding spec (DSP0286).
> 
Correct, not for the data packet. But required by those storage specifications at the transport level. For example, NVMe security send/recv commands require the fields that exist in  `storage_spdm_transport_header` to generate those commands from user space to a **real** device. Similarly for SCSI/ATA.

>     2. `storage_spdm_transport_header` is an **extra header** data packet on the wire for emulation purpose only. The vsock header is used for communication between emulated responder and emulated requester.
> 
Not exactly, although the emulation layer in QEMU uses this header `storage_spdm_transport_header`, the information within `storage_spdm_transport_header` is **also** used to communicate with **real devices**. That is, for example, the API calls in `libnvme` for NVMe devices or SgCmds for SCSI devices, all require the information passed with  `storage_spdm_transport_header` to _correctly_ generate the security commands for SPDM to a **real** device. The reasoning I gave in my previous response is to why it is `packed`, that facilitates the usage case for emulation over sockets in addition.
> 
> To me, it is like TCP header we added in `spdm-emu` project, see https://github.com/DMTF/spdm-emu/blob/main/spdm_emu/spdm_emu_common/command.h#L14-L34.
> 
> If this condition is TRUE, then I feel that we should separate `storage_spdm_transport_header` from `spdm_transport_storage_lib.h` and `spdm_transport_storage_lib`. The reason is that `libspdm` is for the **real device** usage. We need to ensure the real device can work without this `storage_spdm_transport_header`. NOTE: The real device will not have such `storage_spdm_transport_header`.

`libspdm` cannot interact directly with a **real** device. An application using `libspdm` for storage must make the storage transport specific API calls using `libnvme` in the case of NVMe (similarly for SCSI/ATA). This **cannot** be done without the information that is passed through `storage_spdm_transport_header`. So it is not exactly the same as the TCP header in `spdm-emu` it is **essential** for communicating with even real devices. For an application using `libspdm` to do SPDM Storage (like `spdm-utils`), these fields should be specified by the SPDM layer, and that app shall facilitate transport (API calls to real devices).
> 
> **The principle is that the `spdm_transport_storage_lib.h` and `spdm_transport_storage_lib` should only touch the real physical device layer, defined in SPDM storage binding spec (DSP0286) or industry storage (NVME, SCSI, ATA) spec.** If we need wrap another header for emulated device for SOCKET, I recommend we do it in a separate emulation wrapper.

Thanks for that, this makes sense to me. But note that this header required for communicating with real devices also. Hence, why this 'virtual' header exists. What are your thoughts considering the above?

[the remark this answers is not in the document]
jyao1: > libspdm cannot interact directly with a real device. An application using libspdm for storage must make the storage transport specific API calls using libnvme in the case of NVMe (similarly for SCSI/ATA). This cannot be done without the information that is passed through storage_spdm_transport_header. 

> But note that this header required for communicating with real devices also.

Ah, I see your point. If this is required for real device, then we should consider how to add it.

Do you have an example how it is used on a real device (not emulation)? E.g, how you use libnvme with libspdm in this case? Any reference?
Once I can study how it is used in POC today, I might have some idea on design.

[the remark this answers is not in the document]
jyao1: I studied the code.

In the real device driver, the POC removes the  `storage_spdm_transport_header` and adds a real device specific header, such as SCSI or NVMe, then send it out. 

For receive, I noticed that the POC code uses encode() in receive() function. For example:
NVMe: https://github.com/twilfredo/spdm-utils/blob/5be6539fe483f489c720965eb805f227fd1784f0/src/nvme.rs#L484. It uses `libspdm_transport_storage_encode_message` for `nvme_security_receive_args` receive function.
SCSI: https://github.com/twilfredo/spdm-utils/blob/5be6539fe483f489c720965eb805f227fd1784f0/src/scsi.rs#L422. It uses `libspdm_storage_encode_message` for `gen_libspdm_response` receive function.

It seems that the current POC code uses `storage_spdm_transport_header` structure be a **generic fake header** to **pass the information** between libspdm and the real device driver, when send or receive.

Please let me know if that is right?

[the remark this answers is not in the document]
twilfredo: > It seems that the current POC code uses storage_spdm_transport_header structure be a generic fake header to pass the information between libspdm and the real device driver, when send or receive

Essentially yes, we are using `libspdm` to **publicly** define this _virtual header_  at the transport layer (as `libspdm` is the only current implementer of DSP0286). As you noticed, the data from this structure is then used to generate the _real_ device storage API (SCSI/NVMe) specific headers.

Note that, it is also used to encode management commands using `libspdm_transport_storage_encode_management_cmd()`. Which will generate the SPDM specific datum and pass that over to the device driver above, similarly to above, then used to generate the _real_ device storage API (SCSI/NVMe) specific headers.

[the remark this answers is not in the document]
jyao1: Thanks to confirm that. Here is my thought:

We have 2 directions:
 * define one common virtual header for all storage devices.
 * define multiple device specific virtual headers for NVMe, SCSI, ATA to match the NVMe, SPC-6, ATA8-ACS spec.

Since they do have common attribute {security_protocol, security_protocol_specific_data}, I would prefer one common virtual header, like what you have proposed in the code.
I think we can simply use following code:
```
#pragma pack(1)
typedef struct {
    uint8_t security_protocol;
    uint16_t security_protocol_specific_data;
} spdm_storage_transport_virtual_header_t;
#pragma pack()
```

Now, there are several opens:
1) do we need `length` field?
I think it is NOT needed, because we already have message_size in the parameter. The length of the final data can be determined by (message_size - sizeof(spdm_storage_transport_virtual_header_t)).
Adding `length` here seems redundant to me.

2) do we need `inc_512` field?
I think it is NOT needed, because `inc_512` is SCSI specific way to handle message size internally. It is not common, and it should be invisible to outside.

3) Another thing, I am not sure if we need storage specific field, e.g. NVMe Security Specific Field (NSSF), SCSI Control field, ATA Device field. If that is needed, then we can add descriptor_types and device_specific_data;
```
#pragma pack(1)
typedef struct {
    uint8_t device_type; // Table 9 — SPDM Storage Secured Message descriptor types (NVME=1, SCSI=2, ATA=3)
    uint8_t security_protocol;
    uint16_t security_protocol_specific_data;
    uint32_t device_specific_data_size; // device specific. Currently, it should be 0.
    uint8_t device_specific_data[]; // device specific. Currently, it should be N/A.
} spdm_storage_transport_virtual_header_t;
#pragma pack()
```

Please let me know your thought.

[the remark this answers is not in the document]
twilfredo: Thanks for the suggestions. I agree with having a single virtual header, this would simplify the logic.

1. Yeah I think we could omit this.
2. Agreed
3. I am not too sure, I will try to implement it come back with a better idea. My initial thought is, we can handle it in `libspdm` and it seems only relevant for secured messages so it _likely_ doesn't need to be defined in this header.

So perhaps the below should suffice.
```
#pragma pack(1)
typedef struct {
    uint8_t security_protocol;
    uint16_t security_protocol_specific_data;
} spdm_storage_transport_virtual_header_t;
#pragma pack()

```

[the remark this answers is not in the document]
jyao1: Yes, let's start from this simple version.

twilfredo:
> @twilfredo , the next release (Q2) will be end of June. Please let us know if this can be completed at that time windows.

As an update: I have all of the changes discussed above added and testing with the respective modifications to the NVMe driver, things are working in emulation (QEMU). 

Just need to support and test the ` SPDM Storage Secured Message Descriptors`. So hopefully I will have something ready next week and we can try to get this out before the next release.

library/spdm_transport_storage_lib/libspdm_storage.c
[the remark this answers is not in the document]
twilfredo: Addressed in: https://github.com/DMTF/libspdm/pull/2827/commits/ea2bc758ac4b6abbe81026ff008689832fefb18d

include/library/spdm_transport_storage_lib.h
```diff
@@ -0,0 +1,281 @@
...
+ * |                 |Protocol| Protocol Specific |
+ * +-----------------+--------+-------------------+
+ * |Security Protocol|    1   |         2         |
+ * +-----------------+--------+-------------------+
+ */
+#pragma pack(1)
+typedef struct {
+    uint8_t security_protocol;
+    uint16_t security_protocol_specific;
+} spdm_storage_transport_virtual_header_t;
```
jyao1: we should use `libspdm_storage_transport_virtual_header_t` - the prefix is `libspdm`, which means libspdm implementation.

include/library/spdm_transport_storage_lib.h
```diff
@@ -0,0 +1,281 @@
...
+#pragma pack(1)
+typedef struct {
+    uint8_t rsvd1;
+    uint8_t desc_type;
+    uint8_t status;
+    uint8_t rsvd2;
+    uint32_t length;
+    uint32_t offset;
+    uint32_t rsvd3;
+} spdm_storage_secured_message_descriptor;
```
jyao1: If this align to DSP0286, it should be moved to industry_standard/spdm_storage_binding.h

include/library/spdm_transport_storage_lib.h
```diff
@@ -0,0 +1,281 @@
...
+#pragma pack()
+
+typedef enum
+{
+    SECURED_MSG_DESCRIPTOR_NVME = 0x01,
+    SECURED_MSG_DESCRIPTOR_SCSI = 0x02,
+    SECURED_MSG_DESCRIPTOR_ATA = 0x03,
+    SECURED_MSG_DESCRIPTOR_SPDM = 0x04,
+    SECURED_MSG_DESCRIPTOR_DATA_BUFFER = 0x40
+} spmd_storage_secured_message_descriptor_t;
```
jyao1: If this align to DSP0286, it should be moved to industry_standard/spdm_storage_binding.h

Also it should be started with `SPDM_STORAGE_` - e.g. `SPDM_STORAGE_SECURED_MSG_DESCRIPTOR_xxx`

Typo: spmd -> spdm.

include/library/spdm_transport_storage_lib.h
```diff
@@ -0,0 +1,281 @@
...
+
+typedef enum
+{
+    SECURED_MSG_ENCAPSULATED_STATUS_SUCCESS = 0x00,
+    SECURED_MSG_ENCAPSULATED_STATUS_INVALID_CMD = 0x02
+} spdm_storage_secured_message_encapsulated_status_t;
```
jyao1: If this align to DSP0286, it should be moved to industry_standard/spdm_storage_binding.h

Also it should be started with SPDM_STORAGE_ - e.g. SPDM_STORAGE_SECURED_MSG_xxx

Spec also defines 01, 03, FF. I suggest we define all of them, when you move.

library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -126,7 +126,7 @@ libspdm_return_t libspdm_receive_response(void *spdm_context, const uint32_t *se
     libspdm_return_t status;
     uint8_t *message;
     size_t message_size;
-    uint32_t *message_session_id;
+    uint32_t *message_session_id, message_id;
```
jyao1: Would you please split to 2 different lines?

jyao1 (approved):

steven-bellock:
@steven-bellock to review this week.

include/industry_standard/spdm_storage_binding.h
```diff
@@ -0,0 +1,77 @@
...
+typedef enum
+{
+    SPDM_STORAGE_SECURED_MSG_ENCAPSULATED_STATUS_SUCCESS = 0x00,
+    SPDM_STORAGE_SECURED_MSG_ENCAPSULATED_STATUS_GENERAL_ERROR = 0x01,
+    SPDM_STORAGE_SECURED_MSG_ENCAPSULATED_STATUS_INVALID_CMD = 0x02,
+    SPDM_STORAGE_SECURED_MSG_ENCAPSULATED_STATUS_INVALID_FIELD = 0x03,
+    SPDM_STORAGE_SECURED_MSG_ENCAPSULATED_STATUS_VENDOR_DEFINED = 0xFF
+} spdm_storage_secured_message_encapsulated_status_t;
+
+#define LIBSPDM_STORAGE_SEQUENCE_NUMBER_COUNT 2
```
steven-bellock: `SPDM_STORAGE_SEQUENCE_NUMBER_COUNT`

include/industry_standard/spdm_storage_binding.h
```diff
@@ -0,0 +1,77 @@
...
+#define SPDM_STORAGE_OPERATION_CODE_DISCOVERY        0x01
+#define SPDM_STORAGE_OPERATION_CODE_PENDING_INFO     0x02
+#define SPDM_STORAGE_OPERATION_CODE_MESSAGE          0x05
+#define SPDM_STORAGE_OPERATION_CODE_SECURED_MESSAGE  0x06
+
+#define SPDM_STORAGE_MAX_CONNECTION_ID_MASK 0x3
+
+#pragma pack()
+
+#endif /* STORAGE_BINDING_H */
```
steven-bellock: `SPDM_STORAGE_BINDING_H`

include/library/spdm_transport_storage_lib.h
```diff
@@ -0,0 +1,254 @@
+/**
+ *  Copyright Notice:
+ *  Copyright 2025 DMTF. All rights reserved.
+ *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
+ **/
+
+#ifndef STORAGE_TRANSPORT_LIB_H
+#define STORAGE_TRANSPORT_LIB_H
+
+#include "library/spdm_common_lib.h"
```
steven-bellock, line 10: It is not clear to me why these headers are needed, but they are present in, for example the PCI DOE headers so this is a separate issue.

library/spdm_transport_storage_lib/libspdm_storage.c
```diff
@@ -0,0 +1,999 @@
...
+ *
+ * @return size in byte of the sequence_number_buffer.
+ *        It shall be no greater than 8.
+ *        0 means no sequence number is required.
+ **/
+uint8_t libspdm_storage_get_sequence_number(uint64_t sequence_number,
+                                            uint8_t *sequence_number_buffer)
+{
+    libspdm_copy_mem(sequence_number_buffer, LIBSPDM_STORAGE_SEQUENCE_NUMBER_COUNT,
+                     &sequence_number, LIBSPDM_STORAGE_SEQUENCE_NUMBER_COUNT);
```
steven-bellock: `sizeof(uint64_t)`

library/spdm_transport_storage_lib/libspdm_storage.c
```diff
@@ -0,0 +1,999 @@
...
+    return LIBSPDM_STORAGE_SEQUENCE_NUMBER_COUNT;
+}
+
+/**
+ * Used to keeps track of SPDM Storage Secured Message Descriptor Encapsulated errors.
+ * As this is an error to be proccsessed at the trasnport layer, when an error
+ * is detected in Secured Message Descriptor, this variable shall be updated.
+ * By the subsequent SPDM error response, this value shall be read and cleared.
+ **/
+uint8_t libspdm_storage_secured_message_encap_status = 0;
```
steven-bellock: This is not thread-safe.

twilfredo: Yeah, my thinking was that this is relevant only for a responder. In which case the integration should be a single threaded instance (?). Alternatively, I we can add a mutex or something here. But I'm not sure if there's nicer way to do it outside of that? Any suggestions?

jyao1: Ah, global variable is forbidden and not accepted, even with mutex.

I am not sure why it is needed. Would you please elaborate the usage?

I found it is used in encode() and decode() function. But how encode() is related to decode()?
If the integrator want to use those two function, can we leave the work to integrator?

We can just remove `libspdm_storage_secured_message_encap_status` in libspdm, and let integrator save and check/clear.

twilfredo: Table 11 of DSP0286: Encapsulated Response Status Codes. Which map to the `Status` field of an `SPDM Storage Secured Message descriptor` for an SPDM storage
secured message response (this field is reserved for SPDM requests). This field should be processed at the transport level as that is where we handle descriptors.

For example if we are decoding a request and detect an unsupported message, the subsequent error response should contain the `Invalid Command StatusCode`. But there's no way to convey this information between the encode and the decode functions within the transport layer, hence the global. What are your thoughts?

twilfredo: > Is this global for SPDM connection, or to a specific SPDM session?

This will be used only by the responder in a secured session.

jyao1: >  But there's no way to convey this information between the encode and the decode functions within the transport layer, hence the global. 

Please never never use global variable. It does not work for the ROM code only. Writable data must be in stack or heap.

In this case, the architecture way is to use `last_spdm_error`. See https://github.com/DMTF/libspdm/blob/main/include/internal/libspdm_common_lib.h#L570.

This `->transport_decode_message()` error is returned to `libspdm_return_t`, then it should be saved to the context.
See https://github.com/DMTF/libspdm/blob/main/library/spdm_responder_lib/libspdm_rsp_receive_send.c#L208-L211 and
https://github.com/DMTF/libspdm/blob/main/library/spdm_responder_lib/libspdm_rsp_receive_send.c#L251-L258

Then later `->transport_encode_message()` function is called after inspecting the error status.
See https://github.com/DMTF/libspdm/blob/main/library/spdm_responder_lib/libspdm_rsp_receive_send.c#L489 and https://github.com/DMTF/libspdm/blob/main/library/spdm_responder_lib/libspdm_rsp_receive_send.c#L536-L538.

Also we have `libspdm_get_last_spdm_error_struct()` and `libspdm_set_last_spdm_error_struct()` defined in https://github.com/DMTF/libspdm/blob/main/include/library/spdm_common_lib.h.

twilfredo: Thanks for that! I have removed the global variable usage and instead resorted to using the `libspdm_get_last_spdm_error_struct()` / `libspdm_set_last_spdm_error_struct()` functions.

jyao1: LGTM

jyao1:
@steven-bellock , any other feedback?
I plan to merge.

library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -156,7 +157,22 @@ libspdm_return_t libspdm_receive_response(void *spdm_context, const uint32_t *se
         return status;
     }
 
-    message_session_id = NULL;
+    /*
```
steven-bellock, line 160: We are discussing this in the Code Task Force meeting and trying to map this to the `SPDM Storage Operation Codes`. Will continue to review to make sure we understand it.

jyao1: @twilfredo , would you please share the integrator which is consuming the latest code?

twilfredo: Current development branch for `spdm-utils`: https://github.com/westerndigitalcorporation/spdm-utils/commits/wilfred/spdm-storage-v4/

1. [transport: storage: add nvme support](https://github.com/westerndigitalcorporation/spdm-utils/commit/167f891d6a5c75d4e678bdcd1af62f2e44cd58df) NVMe requester is compatible with the latest libspdm changes here
2. [qemu-server: add SCSI/ATA and NVMe support](https://github.com/westerndigitalcorporation/spdm-utils/commit/929cb4b8ebce850cdff3b2ef5ee30c131f055e2c) Implements the responder server for QEMU
3. [transport: storage: add scsi support](https://github.com/westerndigitalcorporation/spdm-utils/commit/8c3ddb2f0e92bf020ef095dac2479322cacc6e97) SCSI requester is WIP

jyao1: @steven-bellock 
> We are discussing this in the Code Task Force meeting and trying to map this to the SPDM Storage Operation Codes. Will continue to review to make sure we understand it.

My understanding is that: The `SPDM Storage Message` (0x05) or `SPDM Storage Secured Message` (0x06) in "Table 2 — SPDM Storage Operation Codes" is only present in `IF-SEND` or `IF-RECV`, as part of the `SECURITY PROTOCOL SPECIFIC` (`SP Specific`). As such, the info is only present when the host sends a command (send buffer or receive command) to the device. But when host receives the data (send ACK, or receive buffer) from the device, there is no info in the returned buffer, to indicate if the received buffer is normal SPDM or secured SPDM.

This mechanism is different from MCTP or PCI_DOE, when the receive buffer includes such normal SPDM or secured SPDM indicator. The key difference is shown in DSP0286 "Figure 2 — Storage message flow" - The `Response with DIGEST data` does not include `SPDMOperation`. Only `IF-RECV` includes `SPDMOperation`.

twilfredo: That is correct, hence why the response path has some additional logic to differentiate a secured session.

jyao1: @twilfredo 

To be specific, I am commenting for this sentence - `Storage response messages are not transport encoded`. I think it should be transport encoded.

I read the code again on the usage of `libspdm_storage_transport_virtual_header_t`, which includes `security_protocol_specific` for `SPDMOperation`.

I believe in current code, the `libspdm_storage_transport_virtual_header_t` is only included in the transport_message when we send command to the storage (such as IF-SEND or IF-RECV). But it is NOT included in the transport_message when we receive data buffer from the storage.
Because of the asymmetry, we have to use `is_request_message` to handle request/response message in different ways in `libspdm_transport_storage_decode_message()`.

However, What if we also __add `libspdm_storage_transport_virtual_header_t` to the receive data buffer from the storage, as part of transport_message__ ?
With this, we can simplify the code flow, and allow libspdm to use the consistent way to handle the transport_message.

My current thought is that, if we can handle `libspdm_storage_transport_virtual_header_t` in device level for send() and receive(), then the libspdm transport layer can always assume the presence of `libspdm_storage_transport_virtual_header_t`. That means the `SPDMOperation` is also known not only for send() but also for receive().

Do you think it is feasible?

twilfredo: @jyao1 I did consider the following:

> "add libspdm_storage_transport_virtual_header_t to the receive data buffer from the storage, as part of transport_message ?"

However I do not think it is feasible for the following reasons.

1. `libspdm_storage_transport_virtual_header_t` shall not be transferred within the data buffer. It should only be used by an integrator to setup the storage protocol (NVMe/ATA/SCSI) API(s). If a `libspdm` based responder encodes an SPDM response with a `libspdm_storage_transport_virtual_header_t` it would not be spec compliant.

2. An SPDM responder integrator may use `libspdm`, but the requestor side may be non `libspdm` based. For example the `linux` kernel. In such a case, using `libspdm_storage_transport_virtual_header_t` to wrap the response is erroneous. For context, I was developing this series with the WIP kernel integration of SPDM, and we can't have an arbitrary header (such as `libspdm_storage_transport_virtual_header_t`) in response messages transferred through the response data buffer.

steven-bellock (approved):
I am fine if this is included in the upcoming release.

PR #2829, opened 2024-09-09
Cleanup DOE header

steven-bellock:
In particular clarify that the PCIe specification defines the size of a double word as four bytes.

### DIFF
--- a/include/industry_standard/pcidoe.h
+++ b/include/industry_standard/pcidoe.h
@@ -17,16 +17,15 @@
 
 #pragma pack(1)
 
-
-/* DOE header*/
-
+/* DOE header
+ * The PCIe specification defines the size of a DW as four bytes. */
 typedef struct {
     uint16_t vendor_id;
     uint8_t data_object_type;
     uint8_t reserved;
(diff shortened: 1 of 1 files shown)


include/industry_standard/pcidoe.h
```diff
@@ -38,11 +37,9 @@ typedef struct {
 #define PCI_DOE_DATA_OBJECT_TYPE_SECURED_SPDM 0x02
 
 #define PCI_DOE_MAX_SIZE_IN_BYTE 0x00100000
-#define PCI_DOE_MAX_SIZE_IN_DW 0x00040000
-
-
-/* DOE Discovery*/
+#define PCI_DOE_MAX_SIZE_IN_DW (4 * PCI_DOE_MAX_SIZE_IN_BYTE)
```
jyao1, line 40: This should be `(PCI_DOE_MAX_SIZE_IN_BYTE / 4)` ?

steven-bellock: Yes.

twilfredo (approved):

jyao1 (approved):

PR #2832, opened 2024-09-10
Skip certificate algorithm check

### LINKED ISSUE #2830: Set Cert: need to skip checking base_asym_algo check when cert_model is AliasCert
https://github.com/DMTF/libspdm/blob/b327ead227f14998789bca7d045af17b99dd6d50/library/spdm_crypt_lib/libspdm_crypt_cert.c#L1113

When set_cert is true and cert model is AliasCert, this function probably shouldn't use base_asym_algo and base_hash_algo to check the leaf certificate as now the leaf is device cert which doesn't necessarily to use the same algorithm as negotiated.

steven-bellock:
Fix #2830.


steven-bellock:
@rw8896

jyao1 (approved):

PR #2837, opened 2024-09-15
Skip keyUsage check for AliasCert and SET_CERTIFICATE

### LINKED ISSUE #2833: Set Cert: incorrect key usage check for AliasCert Model
https://github.com/DMTF/libspdm/blob/b327ead227f14998789bca7d045af17b99dd6d50/library/spdm_crypt_lib/libspdm_crypt_cert.c#L1193-L1197

When set_cert is true and cert model is SPDM_CERTIFICATE_INFO_CERT_MODEL_ALIAS_CERT, the leaf cert is device CA cert which doesn't need to have the digitalSignature usage. Probably could check keyCertSign instead.

steven-bellock:
Fix #2833.

### DIFF
--- a/library/spdm_crypt_lib/libspdm_crypt_cert.c
+++ b/library/spdm_crypt_lib/libspdm_crypt_cert.c
@@ -1057,7 +1057,6 @@ bool libspdm_x509_common_certificate_check(const uint8_t *cert, size_t cert_size
     size_t asn1_buffer_len;
     bool status;
     size_t cert_version;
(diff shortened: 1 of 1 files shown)


steven-bellock:
@rw8896

jyao1 (approved):

jyao1:
@rw8896 would you please review the fix?

rw8896 (approved):
Change looks good to me.
Thanks for helping solve the issue.

PR #2839, opened 2024-09-17
Add EventAllPolicy to key exchange

### LINKED ISSUE #2566: Add `EventAllPolicy` to key exchange

steven-bellock:
Fix #2566.

This pull request adds two parameters to `libspdm_event_subscribe`. `session_id` is needed because event subscription is based on sessions, and `subscribe_type` specifies whether to subscribe/unsubscribe to all events, or to use the event list. Previously `subscribe_event_group_count == 0` was used to unsubscribe from all events. However that does not work for subscribing to all events. Ie, `subscribe_event_group_count == 0xff` is legal with the event list.


steven-bellock:
@Xiaohanjlll

library/spdm_responder_lib/libspdm_rsp_key_exchange.c
```diff
@@ -291,17 +291,25 @@ libspdm_return_t libspdm_get_response_key_exchange(libspdm_context_t *spdm_conte
...
-            return libspdm_generate_error_response(
-                spdm_context, SPDM_ERROR_CODE_INVALID_REQUEST,
-                0, response_size, response);
+    if (spdm_request->header.spdm_version >= SPDM_MESSAGE_VERSION_13) {
+        if (spdm_context->connection_info.multi_key_conn_rsp && (slot_id != 0xff)) {
+            if ((spdm_context->local_context.local_key_usage_bit_mask[slot_id] &
+                 SPDM_KEY_USAGE_BIT_MASK_KEY_EX_USE) == 0) {
+                return libspdm_generate_error_response(
+                    spdm_context, SPDM_ERROR_CODE_INVALID_REQUEST, 0, response_size, response);
+            }
```
jyao1, line 300: what is this change?

library/spdm_requester_lib/libspdm_req_key_exchange.c
```diff
@@ -305,6 +305,10 @@ static libspdm_return_t libspdm_try_send_receive_key_exchange(
     LIBSPDM_ASSERT(measurement_hash_type == SPDM_KEY_EXCHANGE_REQUEST_NO_MEASUREMENT_SUMMARY_HASH ||
                    measurement_hash_type == SPDM_KEY_EXCHANGE_REQUEST_TCB_COMPONENT_MEASUREMENT_HASH ||
                    measurement_hash_type == SPDM_KEY_EXCHANGE_REQUEST_ALL_MEASUREMENTS_HASH);
+    LIBSPDM_ASSERT(((session_policy & SPDM_KEY_EXCHANGE_REQUEST_SESSION_POLICY_EVENT_ALL_POLICY)
+                    == 0) ||
+                   libspdm_is_capabilities_flag_supported(
+                       spdm_context, true, 0, SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_EVENT_CAP));
```
jyao1, line 311: This check should only for SPDM 1.3 where the event is introduced.
Otherwise, the reserved bit should be ignored.

library/spdm_responder_lib/libspdm_rsp_key_exchange.c
steven-bellock: Separates out the
```
if (spdm_request->header.spdm_version >= SPDM_MESSAGE_VERSION_13)
```
so that the session policy check can use it too.

jyao1: Got it.

jyao1 (approved):

Issue #2844, opened 2024-09-18
Invalid `SessionRequired` messages


steven-bellock:
The `SessionRequired` error code was introduced in SPDM 1.2. However libspdm will send this error if the version is 1.1 for the following messages
- `END_SESSION`
- `FINISH`
- `HEARTBEAT`
- `KEY_UPDATE`
- `PSK_FINISH`

This error needs to be conditional based on the negotiated version. For 1.1 the error code should probably be `Unspecified` since none of the other errors fit in this situation. In addition `SessionRequired` is never exercised in the unit tests, and test cases should be added.

Example
https://github.com/DMTF/libspdm/blob/abbb1a4d25027171a703c80484939d0b513de3f5/library/spdm_responder_lib/libspdm_rsp_end_session.c#L46-L50

steven-bellock:
This was introduced in #690 when the error code was `InvalidRequest`.

jyao1:
Maybe do the check and convert in libspdm_generate_error_response().

steven-bellock:
Keep the error code the same in `libspdm_rsp_end_session.c` but "downgrade" the error code in `libspdm_generate_error_response()`? That might be misleading.

steven-bellock:
Decided in meeting to add
```
if spdm_version == 1.2 use SessionRequired else use Unspecified
```
in each function.

steven-bellock:
Briefly re-opening due to https://github.com/DMTF/libspdm/pull/2848#issuecomment-2672919324.

PR #2848, opened 2024-09-24
Do not use SPDM_ERROR_CODE_SESSION_REQUIRED for SPDM 1.1.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2844


steven-bellock (approved):

steven-bellock:
This pull request missed https://github.com/DMTF/libspdm/blob/c6e90cb0ca50b9ba8404930d8a68e55c2ed6ff32/library/spdm_requester_lib/libspdm_req_encap_key_update.c#L52-L56 presumably because it has `context` instead of the normal `spdm_context`.

PR #2850, opened 2024-09-24
Enforce mandatory CA in SET_CERTIFICATE

### LINKED ISSUE #2731: `libspdm_verify_set_cert_leaf_cert_basic_constraints` needs to be updated for SPDM 1.3
Follow-up to #2657.

`libspdm_verify_set_cert_leaf_cert_basic_constraints` currently allows the existence of a device certificate's CA field to be optional. This is mandatory for SPDM 1.3 and its value must be true. In addition the function should handle generic certificates the same as device certificates. The current logic treats generic certificates the same as alias certificates, which is incorrect.

steven-bellock:
Fix #2731.

### DIFF
--- a/include/library/spdm_crypt_lib.h
+++ b/include/library/spdm_crypt_lib.h
@@ -917,7 +917,10 @@ bool libspdm_x509_certificate_check_ex(const uint8_t *cert, size_t cert_size,
                                        uint32_t base_asym_algo, uint32_t base_hash_algo,
                                        bool is_requester, uint8_t cert_model);
 /**
- * Certificate Check for SPDM leaf cert when set_cert.
+ * Certificate Check for SPDM leaf cert when set_cert. It is used for SPDM 1.2.
+ *
+ * This function differs from libspdm_x509_set_cert_certificate_check_ex in that in SPDM 1.2 the
+ * BasicConstraints CA field is optional while in SPDM 1.3 and later it is mandatory.
  *
  * @param[in]  cert                  Pointer to the DER-encoded certificate data.
  * @param[in]  cert_size             The size of certificate data in bytes.
@@ -935,7 +938,10 @@ bool libspdm_x509_set_cert_certificate_check(const uint8_t *cert, size_t cert_si
                                              bool is_requester, bool is_device_cert_model);
 
 /**
- * Certificate Check for SPDM leaf cert when set_cert. It is used for SPDM 1.3.
+ * Certificate Check for SPDM leaf cert when set_cert. It is used for SPDM 1.3 and later.
+ *
+ * This function differs from libspdm_x509_set_cert_certificate_check in that in SPDM 1.2 the
+ * BasicConstraints CA field is optional while in SPDM 1.3 and later it is mandatory.
  *
  * @param[in]  cert                  Pointer to the DER-encoded certificate data.
  * @param[in]  cert_size             The size of certificate data in bytes.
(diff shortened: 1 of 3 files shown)


library/spdm_responder_lib/libspdm_rsp_set_certificate.c
```diff
@@ -245,7 +260,8 @@ libspdm_return_t libspdm_get_response_set_certificate(libspdm_context_t *spdm_co
 
 #if LIBSPDM_CERT_PARSE_SUPPORT
         /*check the cert_chain*/
-        result = libspdm_set_cert_verify_certchain(cert_chain, cert_chain_size,
+        result = libspdm_set_cert_verify_certchain(spdm_version,
+                                                   cert_chain, cert_chain_size,
                                                    spdm_context->connection_info.algorithm.base_asym_algo,
                                                    spdm_context->connection_info.algorithm.base_hash_algo,
                                                    is_device_cert_model);
```
jyao1, line 267: The last parameter is changed to `uint8_t cert_model`.

library/spdm_responder_lib/libspdm_rsp_set_certificate.c
```diff
@@ -39,11 +40,22 @@ static bool libspdm_set_cert_verify_certchain(const uint8_t *cert_chain, size_t
...
-                                                 false, is_device_cert_model)) {
-        return false;
+    if (spdm_version == SPDM_MESSAGE_VERSION_12) {
+        const bool is_device_cert_model =
+            (cert_model == SPDM_CERTIFICATE_INFO_CERT_MODEL_DEVICE_CERT);
+
+        /*verify leaf cert*/
+        if (!libspdm_x509_set_cert_certificate_check(leaf_cert_buffer, leaf_cert_buffer_size,
+                                                     base_asym_algo, base_hash_algo,
+                                                     false, is_device_cert_model)) {
```
jyao1, line 50: Why not use the `libspdm_x509_set_cert_certificate_check_ex` directly?

library/spdm_responder_lib/libspdm_rsp_set_certificate.c
```diff
@@ -39,11 +40,22 @@ static bool libspdm_set_cert_verify_certchain(const uint8_t *cert_chain, size_t
...
+            return false;
+        }
+    } else {
+        if (!libspdm_x509_set_cert_certificate_check_ex(leaf_cert_buffer, leaf_cert_buffer_size,
+                                                        base_asym_algo, base_hash_algo,
+                                                        false, cert_model)) {
```
jyao1, line 56: It looks weird to only support 1.3 in ex version and only support 1.2 in non-ex version.
I feel it will be easily mis-used, if it is designed in such way.

But I don't have good idea now. Need to think through.

include/library/spdm_crypt_lib.h
```diff
@@ -917,7 +917,7 @@ bool libspdm_x509_certificate_check_ex(const uint8_t *cert, size_t cert_size,
                                        uint32_t base_asym_algo, uint32_t base_hash_algo,
                                        bool is_requester, uint8_t cert_model);
 /**
- * Certificate Check for SPDM leaf cert when set_cert.
+ * Certificate Check for SPDM leaf cert when set_cert. It is used for SPDM 1.2.
```
jyao1, line 920: We had better add some reason for the ex/non-ex version difference.
For example, the `basic constrain` is not checked in 1.2, but will be checked in 1.3.

library/spdm_responder_lib/libspdm_rsp_set_certificate.c
steven-bellock: Wow. Thank you for type checking when the C compiler could not. 😆

jyao1 (approved):

PR #2853, opened 2024-09-30
Accomodate pathLen when checking certificate

### LINKED ISSUE #2831: Set Cert: should not check BasicConstraint CA: True with hardcoded value
https://github.com/DMTF/libspdm/blob/b327ead227f14998789bca7d045af17b99dd6d50/library/spdm_crypt_lib/libspdm_crypt_cert.c#L840-L846

Device Cert could specify pathLenConstraint in BasicConstraints and hence the hardcoded BASIC_CONSTRAINTS_STRING_TRUE_CASE can't cover all the true cases.
https://github.com/DMTF/libspdm/blob/b327ead227f14998789bca7d045af17b99dd6d50/library/spdm_crypt_lib/libspdm_crypt_cert.c#L98

steven-bellock:
Fix #2831.


steven-bellock:
@rw8896 as well.

jyao1:
Question: How the logic is tested?

library/spdm_crypt_lib/libspdm_crypt_cert.c
```diff
@@ -813,13 +813,13 @@ static bool libspdm_verify_set_cert_leaf_cert_basic_constraints(
     const uint8_t *cert, size_t cert_size, uint8_t cert_model, bool need_basic_constraints)
 {
     bool status;
-    /* basic_constraints from certificate. */
-    uint8_t cert_basic_constraints[LIBSPDM_MAX_BASIC_CONSTRAINTS_CA_LEN];
+    /* basic_constraints from certificate. Add space for pathLen. */
+    uint8_t cert_basic_constraints[LIBSPDM_MAX_BASIC_CONSTRAINTS_CA_LEN + 10];
```
jyao1, line 817: How 10 is calculated? Is that enough?

rw8896: Both mbedtls and openssl seem to provide APIs to check if CA is true.
https://docs.openssl.org/3.0/man3/X509_check_ca/#copyright
https://github.com/Mbed-TLS/mbedtls/blob/development/include/mbedtls/x509_crt.h#L920

Maybe we could invent another libspdm_x509 API to return if CA is true instead of parsing the basic constraints raw bytes?

steven-bellock:
> Question: How the logic is tested?

Assuming `LIBSPDM_CERT_PARSE_SUPPORT` is enabled then https://github.com/DMTF/libspdm/blob/525ba87b2f167350a622d9c997f5ea48ced65b52/unit_test/test_spdm_responder/set_certificate_rsp.c#L543-L547 will exercise the logic, as shown in https://dmtf.github.io/libspdm/coverage_log/library/spdm_crypt_lib/libspdm_crypt_cert.c.gcov.html#853.

library/spdm_crypt_lib/libspdm_crypt_cert.c
steven-bellock: >How 10 is calculated? Is that enough?

It's probably more than enough, but let me check what the largest ASN.1 DER integer size is. I'll make a macro for it too.

steven-bellock: > Both mbedtls and openssl seem to provide APIs to check if CA is true. https://docs.openssl.org/3.0/man3/X509_check_ca/#copyright https://github.com/Mbed-TLS/mbedtls/blob/development/include/mbedtls/x509_crt.h#L920
> 
> Maybe we could invent another libspdm_x509 API to return if CA is true instead of parsing the basic constraints raw bytes?

The benefit of having the logic in `libspdm/library` is that the Integrator doesn't need to provide new functions / functionality as the SPDM specification adds or removes certificate fields. But yes, maybe for libspdm 4.0 we could have a basic certificate parsing interface where the Integrator / cryptography library handles most of the logic, but then also have a catch-all interface where the Integrator provides the raw ASN.1 DER to libspdm.

jyao1:
LGTM
@rw8896 any feedback?

library/spdm_crypt_lib/libspdm_crypt_cert.c
rw8896: Thank Steven. Current change looks good and yes, I think the enhancement could wait to be implemented in 4.0

PR #2862, opened 2024-10-16
Fix openssl build failure for AARCH64_GCC

htafr:
Trying to compile with toolchain AARCH64_GCC generates the following error:

```bash
$WORKSPACE/libspdm/os_stub/openssllib/openssl/crypto/o_str.c: In function ‘openssl_strerror_r’:
$WORKSPACE/libspdm/os_stub/openssllib/openssl/crypto/o_str.c:329:13: error: implicit declaration of function ‘strerror_r’; did you mean ‘strerror’? [-Werror=implicit-function-declaration]
  329 |     return !strerror_r(errnum, buf, buflen);
      |             ^~~~~~~~~~
```

Fix: just add the flag `-std=c11` in OPENSSL_FLAGS for AARCH64_GCC.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -387,7 +387,7 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
         if(GCOV STREQUAL "ON")
         add_compile_options(--coverage -fprofile-arcs -ftest-coverage)
         endif()
-        set(OPENSSL_FLAGS -include base.h -Wno-error=maybe-uninitialized -Wno-error=format -Wno-format -Wno-error=unused-but-set-variable -Wno-cast-qual)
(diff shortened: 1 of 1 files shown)


steven-bellock:
The code in question is
```
#elif (defined(_POSIX_C_SOURCE) && _POSIX_C_SOURCE >= 200112L) || \
      (defined(_XOPEN_SOURCE) && _XOPEN_SOURCE >= 600)
    /*
     * We can use "real" strerror_r. The OpenSSL version differs in that it
     * gives 1 on success and 0 on failure for consistency with other OpenSSL
     * functions. Real strerror_r does it the other way around
     */
    return !strerror_r(errnum, buf, buflen);
```

steven-bellock (approved):

PR #2870, opened 2024-10-23
Return `ResetRequired` for certificate retrieval

steven-bellock:
If a `SET_CERTIFICATE` request results in `ResetRequired` then subsequent `GET_CERTIFICATE` requests for that slot, and `GET_DIGESTS`, should result in `ResetRequired`.

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -90,6 +90,7 @@ typedef struct {
     const void *local_cert_chain_provision[SPDM_MAX_SLOT_COUNT];
     size_t local_cert_chain_provision_size[SPDM_MAX_SLOT_COUNT];
     uint8_t local_supported_slot_mask;
+    uint8_t cert_slot_reset_mask;
     spdm_key_pair_id_t local_key_pair_id[SPDM_MAX_SLOT_COUNT];
     spdm_certificate_info_t local_cert_info[SPDM_MAX_SLOT_COUNT];
     spdm_key_usage_bit_mask_t local_key_usage_bit_mask[SPDM_MAX_SLOT_COUNT];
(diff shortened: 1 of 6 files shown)


steven-bellock:
@rw8896 to review.

Note that this reset mask should be sticky between `GET_VERSIONS`, although I haven't tested that yet.
1. `SET_CERTIFICATE` -> `ResetRequired`
2. `VCA`
3. `GET_CERTIFICATE` ->`ResetRequired`

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -3078,6 +3078,8 @@ void libspdm_reset_context(void *spdm_context)
     context->mut_auth_cert_chain_buffer_size = 0;
     context->current_dhe_session_count = 0;
     context->current_psk_session_count = 0;
+
+    context->local_context.cert_slot_reset_mask = 0;
```
rw8896, line 3082: We should not reset cert_slot_reset_mask as you mentioned " reset mask should be sticky between GET_VERSIONS"?

steven-bellock: Ah yeah forgot to delete that. Thanks.

rw8896 (approved):
Thanks! It looks good to me.

jyao1:
Question:
Do we need to update device secret lib sample?
Do we need to update spdm-emu?

steven-bellock:
>Do we need to update device secret lib sample?

Maybe once `libspdm_write_certificate_to_nvm` has the `need_reset` parameter.
>Do we need to update spdm-emu?

I will check if this breaks `spdm-emu`.

steven-bellock:
> I will check if this breaks spdm-emu.

It appears to work OK with this change.

jyao1:
>> Do we need to update device secret lib sample?

> Maybe once libspdm_write_certificate_to_nvm has the need_reset parameter.

Could you please add that?

steven-bellock:
>Could you please add that?

That's a separate issue / pull request.

jyao1 (approved):

PR #2880, opened 2024-11-01
Make variables static in get_key_pair_info.c

ShitalJumbad:


steven-bellock (approved):

jyao1:
Would you please update all files, which include `m_*`?

Issue #2869, opened 2024-10-23
`test_spdm_responder` fails when run multiple times


steven-bellock:
To reproduce run
```
./test_spdm_responder > /dev/null
./test_spdm_responder > /dev/null
```
The second time it is run will produce
```
[  ERROR   ] --- 0x7d != 0x7f
[   LINE   ] --- libspdm/unit_test/test_spdm_responder/set_key_pair_info_ack.c:191: error: Failure!
[  ERROR   ] --- 0x7d != 0x7f
[   LINE   ] --- libspdm/unit_test/test_spdm_responder/set_key_pair_info_ack.c:347: error: Failure!
```
Tests will pass again if the file `libspdm/build/bin/cached_last_set_key_pair_info_request` is deleted.

steven-bellock:
@Xiaohanjlll as well.

Xiaohanjlll:
I'll check it.

Xiaohanjlll:
When the unit test is executed for the second time, the file "cached_last_set_key_pair_info_request" exists. Enter other branches.

https://github.com/DMTF/libspdm/blob/803849f3ab69c0fbc16dad31aa53696c4c9d5f06/os_stub/spdm_device_secret_lib_sample/lib.c#L2591-L2601
https://github.com/DMTF/libspdm/blob/803849f3ab69c0fbc16dad31aa53696c4c9d5f06/os_stub/spdm_device_secret_lib_sample/lib.c#L2632-L2689

PR #2887, opened 2024-11-06
Fix the failure of test_spdm_responder when it is run multiple times.

Xiaohanjlll:
Fix: #2869


steven-bellock:
Seems to work alright.

steven-bellock (approved):

Issue #2890, opened 2024-11-13
Debug library passing x86 asm mnemonic to ARM compiler


wmaroneAMD:
https://github.com/DMTF/libspdm/blob/d6f4b0b6b0a37f53031b7707cafa349b1512dc0d/os_stub/debuglib/debuglib.c#L42

If LIBSPDM_DEBUG_LIBSPDM_ASSERT_CONFIG is set to LIBSPDM_DEBUG_LIBSPDM_ASSERT_BREAKPOINT and the architecture is ARM, the compiler will be fed an x86 assembly instruction.

```
[ 44%] Building C object lib/spdm/libspdm/external/libspdm/os_stub/debuglib/CMakeFiles/debuglib.dir/debuglib.c.o
/tmp/ccCXgt0Y.s: Assembler messages:
/tmp/ccCXgt0Y.s:37: Error: unknown mnemonic `int' -- `int $3'
```

steven-bellock:
@wmaroneAMD do you need this functionality or would just like a cleaner error message? Ie
>Debug breakpoints for the target architecture are not supported.

wmaroneAMD:
A cleaner error message. I can put together a patch that restricts the mnemonic to x86/x86-64 and throws an error if the arch is anything else.

PR #2899, opened 2024-11-15
Restrict x86 mnemonic to x86-64

wmaroneAMD:
This resolves an issue where setting the assert config to breakpoint on ARM will cause the build to fail with an accurate but unhelpful compiler error.

This restricts the mnemonic to x86-64 architectures, and provides a more explanatory error.

Addresses issue: #2890

### DIFF
--- a/os_stub/debuglib/debuglib.c
+++ b/os_stub/debuglib/debuglib.c
@@ -37,9 +37,10 @@ void libspdm_debug_assert(const char *file_name, size_t line_number, const char 
 #elif (LIBSPDM_DEBUG_LIBSPDM_ASSERT_CONFIG == LIBSPDM_DEBUG_LIBSPDM_ASSERT_BREAKPOINT)
 #if defined(_MSC_EXTENSIONS)
     __debugbreak();
-#endif
-#if defined(__GNUC__)
+#elif defined(__GNUC__) && (defined(__i386__) || defined(__x86_64__))
     __asm__ __volatile__ ("int $3");
+#else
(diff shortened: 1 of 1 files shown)


os_stub/debuglib/debuglib.c
```diff
@@ -38,8 +38,10 @@ void libspdm_debug_assert(const char *file_name, size_t line_number, const char
 #if defined(_MSC_EXTENSIONS)
     __debugbreak();
 #endif
-#if defined(__GNUC__)
+#if defined(__GNUC__) && defined(__x86_64__)
```
steven-bellock, line 41: This interrupt is also available on `__i386__`?

os_stub/debuglib/debuglib.c
```diff
@@ -38,8 +38,10 @@ void libspdm_debug_assert(const char *file_name, size_t line_number, const char
...
     __asm__ __volatile__ ("int $3");
+#elif defined(__GNUC__)
```
steven-bellock, line 43: I would combine `_MSC_EXTENSIONS` as part of the `#if - #elif` chain, and just have the `#else` be
>Breakpoint asserts are not available on this architecture and/or compiler

os_stub/debuglib/debuglib.c
wmaroneAMD: Oops, let me add that in.

os_stub/debuglib/debuglib.c
wmaroneAMD: Done

jyao1 (approved):

Issue #2893, opened 2024-11-14
GET_CSR: responder is not able to return InvalidRequest if any RequesterInfo content is not supported


rw8896:
The spec mentioned:
"The Responder shall return an ERROR message with error code InvalidRequest if it cannot support a field included in the RequesterInfo , or if the value of a requested field is not supported and the Responder cannot alter the value of the field."

Current implementation did check the syntax of RequesterInfo as shown below.
https://github.com/DMTF/libspdm/blob/d6f4b0b6b0a37f53031b7707cafa349b1512dc0d/library/spdm_responder_lib/libspdm_rsp_csr.c#L145-L151
But this can't verify whether requester contents are supported. It will need libspdm_gen_csr() and libspdm_gen_csr_ex() to verify the contents and provide more information.

steven-bellock:
In general, all the `hal/` functions need to return `libspdm_return_t` instead of `bool`. That can be a libspdm 4.0 enhancement. But it looks like https://github.com/DMTF/libspdm/blob/8cf9fab59a68b5f42fc9cfe94a2342cedbee8d67/library/spdm_responder_lib/libspdm_rsp_csr.c#L246-L249 can be changed to `InvalidRequest` for libspdm 3.x.

PR #2901, opened 2024-11-18
Add parameters to libspdm_write_certificate_to_nvm

### LINKED ISSUE #2873: Add `need_reset` parameter to `libspdm_write_certificate_to_nvm`
From https://github.com/DMTF/libspdm/issues/2864#issuecomment-2424326846.

Similar to `libspdm_gen_csr`, `libspdm_write_certificate_to_nvm` needs the `need_reset` parameter so that the Integrator can tell libspdm that the device needs to be reset. Currently it is conditional based on `CERT_INSTALL_RESET_CAP`:
https://github.com/DMTF/libspdm/blob/00a71559223b9079c6a716908be6b22b79928ce7/library/spdm_responder_lib/libspdm_rsp_set_certificate.c#L302-L309

steven-bellock:
Fix #2873.


steven-bellock:
@rw8896 as well.

include/hal/library/responder/setcertlib.h
```diff
@@ -48,8 +53,11 @@ extern bool libspdm_write_certificate_to_nvm(
 #endif
     uint8_t slot_id, const void * cert_chain,
     size_t cert_chain_size,
-    uint32_t base_hash_algo, uint32_t base_asym_algo);
-
+    uint32_t base_hash_algo, uint32_t base_asym_algo
+#if LIBSPDM_SET_CERT_PARAMS
+    , bool *need_reset, bool *is_busy
```
jyao1, line 58: I am wondering if we need `is_busy` for CSR API as well.

os_stub/spdm_device_secret_lib_sample/lib.c
```diff
@@ -2221,65 +2222,75 @@ bool libspdm_write_certificate_to_nvm(
...
+                printf("Write output file error %s\n", file_name);
+                fclose(fp_out);
+                return false;
+            }
         }
-    } else {
-        if ((fp_out = open(file_name, O_WRONLY | O_TRUNC)) == -1) {
-            printf("Unable to open file %s\n", file_name);
-            return false;
+
```
rw8896, line 2266: Do we need to handle the (cert_chain == NULL) case?

steven-bellock: That's a separate issue. I just added
```
    if (g_set_cert_is_busy) {
        *is_busy = true;
```
and everything got adjusted. But yes, it looks like the "sample" implementation doesn't support erasing a certificate slot.

include/hal/library/responder/setcertlib.h
steven-bellock: Yes probably. I can do that in a separate pull request, but let me rename `LIBSPDM_SET_CERT_PARAMS` to something more generic so that CSR can leverage it as well.

os_stub/spdm_device_secret_lib_sample/lib.c
rw8896: Yes, the handling could be done in a separate change.

rw8896 (commented):
Looks good to me, thanks!

include/hal/library/responder/setcertlib.h
jyao1: Can we file a separate issue? I feel #2893 only handles INVALID_REQUEST.

jyao1 (approved):

include/library/spdm_lib_config.h
```diff
@@ -442,4 +442,11 @@
 #define LIBSPDM_ADDITIONAL_CHECK_CERT 0
 #endif
 
+/* Enable passing additional parameters to libspdm_write_certificate_to_nvm and libspdm_gen_csr*.
+ * This macro will be removed when libspdm 4.0 is released.
+ */
+#ifndef LIBSPDM_SET_CERT_CSR_PARAMS
+#define LIBSPDM_SET_CERT_CSR_PARAMS 1
```
steven-bellock, line 449: This should not be `1`. I'll fix it in the pull request that fixes #2908.

steven-bellock: Actually this needs a separate pull request, as it is breaking builds.

PR #2902, opened 2024-11-18
Remove unnecessary length check for encap message size

jyao1:
Transport may append more bytes for alignment and cause failure. Also, the smaller header is already checked before. So just remove it here.

### DIFF
--- a/unit_test/test_spdm_requester/encap_request.c
+++ b/unit_test/test_spdm_requester/encap_request.c
@@ -679,53 +679,6 @@ void libspdm_test_requester_encap_request_case2(void **State)
 
 void libspdm_test_requester_encap_request_case3(void **State)
 {
-    libspdm_return_t status;
-    libspdm_test_context_t *spdm_test_context;
-    libspdm_context_t *spdm_context;
-    void *data;
-    size_t data_size;
-    void *hash;
-    size_t hash_size;
-
-    spdm_test_context = *State;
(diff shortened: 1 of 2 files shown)


unit_test/test_spdm_requester/encap_request.c
```diff
@@ -1079,14 +1079,10 @@ int libspdm_requester_encap_request_test_main(void)
         cmocka_unit_test(libspdm_test_requester_encap_request_case1),
         /* Success Case ,func :libspdm_get_encap_response_digest*/
         cmocka_unit_test(libspdm_test_requester_encap_request_case2),
-        /* Error response: When spdm_encapsulated_response_ack_response versions are inconsistent*/
```
steven-bellock, line 1082: Why are these tests removed?

jyao1: The comment is completely wrong.

The test case just input large message size, and expect invalid msg size.

The SPM version is same.

steven-bellock: Please remove the code as well. For example : https://github.com/DMTF/libspdm/blob/6652fc3bbbe444e53637a22522a3fc9d45bd8c78/unit_test/test_spdm_requester/negotiate_algorithms.c#L1605-L1607

jyao1: done

PR #2903, opened 2024-11-18
Update error code for unsupported RequesterInfo in CSR requests

### LINKED ISSUE #2893: GET_CSR: responder is not able to return InvalidRequest if any RequesterInfo content is not supported

ShitalJumbad:
Fix #2893

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_csr.c
+++ b/library/spdm_responder_lib/libspdm_rsp_csr.c
@@ -245,7 +245,7 @@ libspdm_return_t libspdm_get_response_csr(libspdm_context_t *spdm_context,
         } else {
             return libspdm_generate_error_response(
                 spdm_context,
-                SPDM_ERROR_CODE_UNEXPECTED_REQUEST, 0,
+                SPDM_ERROR_CODE_INVALID_REQUEST, 0,
                 response_size, response);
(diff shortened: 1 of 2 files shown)


jyao1 (approved):

steven-bellock:
@rw8896 to review as well.

rw8896 (commented):
Current change looks good to me as the final solution is expected to be done after we change the return value of libspdm_gen_csr().

steven-bellock (approved):
I'll make a follow-up issue / pull request for https://github.com/DMTF/libspdm/issues/2893#issuecomment-2484381881.

Issue #2845, opened 2024-09-20
Standardize Copyright/License Headers


HeikeGilg:
**Problem**
• Inconsistent Headers: Different copyright/license headers are currently used across the repository.
• Missing Headers: Some files do not contain any copyright/license headers.
 
**Proposal for Improvement**
I propose the following solutions to address these issues:
1. Adopting the SPDX Header Format
[SPDX ](https://spdx.dev/learn/handling-license-info/) provides a standardized format for copyright and license headers. This format is widely recognized and makes it easier to understand, automate, and analyze the licensing information in the code.
 2. Utilizing the [REUSE](https://reuse.software/) Tool
The REUSE tool helps make copyright and license headers **machine-readable**. It also offers **automatic validation** to ensure that all files have appropriate and consistent headers.
 
**Why Uniform Headers Are Necessary**
• Maintainability: Using a single format simplifies the maintenance of the repository. It avoids the confusion and potential errors that can arise from having multiple header formats.
• Consistency: A uniform approach ensures that all files are treated equally, avoiding the complexity and potential issues of having different rules for different parts of the codebase.
• Accuracy: Having one consistent source of information minimizes the risk of discrepancies and errors in the copyright/license information.

**Current Discussion State**
The current discussion revolves around maintaining the existing copyright/license headers and adding new SPDX headers in addition. However, this approach is only proposed for a subset of files, with no clear plan for the remaining files in the repository. This partial solution risks increasing complexity and inconsistency, making the codebase harder to maintain and potentially leading to errors. 
 
**My Request**
I recommend adopting a uniform approach to the copyright/license headers across the entire repository. Also I strongly recommend using SPDX header format.
This is a call of the maintainers to first clarify a strategy for all files in the repository before implementing a partial change.

Thank you for considering this proposal.

steven-bellock:
>Inconsistent Headers: Different copyright/license headers are currently used across the repository.

@HeikeGilg can you give an example of this?
>Missing Headers: Some files do not contain any copyright/license headers.

Adding copyright/license to scripts, configuration files, and build files is fine. I see Reuse advises that
>The reason that these files must be licensed is because reasonable minds can differ about the threshold of originality. You may consider a file so insignificant that it does not merit licensing information, but downstream may disagree and believe that the file in question falls under copyright, effectively disallowing them from using it if there is no license.

>Utilizing the [REUSE](https://reuse.software/) Tool

REUSE would be added to CI/CD?

[...]

PR #2905, opened 2024-11-19
Add a SBOM file in CycloneDX format

hughsie:
Hi,

My name is Richard Hughes and I'm a developer at Red Hat. I'm the maintainer of fwupd and LVFS, and am trying to improve software supply chain security by encouraging OEMs, ODMs and IBVs to ship Software Bill of Materials with each firmware binary blob (SBOMs).

I'm working alongside lots of other companies proactively trying to do the right thing. The reason I've opened this pull request is because your project is either used in the *build process* of a firmware we care about (e.g. EDK II, or coreboot, or both) or is built *into the firmware binary* itself. Although my personal focus is on firmware, the SBOM file is in CycloneDX format (one of the most popular industry standards) which makes it also useful when building containers or OS images too.

I would like to contribute this template SBOM file into your project that gets included into source control with substituted values that get populated automatically. I'm not super familiar with libspdm, and so I've done my best populating the project values -- but please point out any that are incorrect and I'll fix them up. I've also put the `sbom.cdx.json` file in what I feel is the right place, but please say if you want me to put it somewhere different or name it a different thing; the directory and `sbom` prefix are unimportant.

The various firmware build tools will take these incomplete SBOM files and then build them into a complete composite SBOM to represent the firmware. Having an upstream reference to what the PURL and CPE values should be means we have something we can trust; I could quite easily spin up a web-service that we say "what CPE do we use for X" -> "cpe:2.3:a:Y:Z::::::::` but we don't actually know if that's still true, up to date, or what the maintainer actually wants them to be. Putting the template upstream means we can trust the values we find in the checked out code during the build process.

Also, if you’re uncomfortable with being labelled a *supplier* (which seems more appropriate from a SBOM point of view, but makes some open source maintainers uncomfortable) we can change the tag to “authors” as the tool that ingests this data can accept either.

I've written a bit more about this proposal here https://blogs.gnome.org/hughsie/2024/11/14/firmware-sboms-for-open-source-projects/ and there's also lot more information about firmware SBOMs here: https://lvfs.readthedocs.io/en/latest/sbom.html – many thanks for your time.


sbom.cdx.json
```diff
@@ -0,0 +1,38 @@
...
+  "components": [
+    {
+      "type": "library",
+      "bom-ref": "pkg:github/DMTF/libspdm@@VCS_TAG@",
+      "cpe": "cpe:2.3:a:DMTF:libspdm:@VCS_TAG@:*:*:*:*:*:*:*",
+      "name": "libspdm",
+      "version": "@VCS_VERSION@",
+      "description": "Sample implementation of DMTF SPDM",
+      "supplier": {
+        "name": "libspdm developers"
```
steven-bellock, line 21:
```suggestion
        "name": "DMTF"
```

sbom.cdx.json
```diff
@@ -0,0 +1,38 @@
...
+    ]
+  },
+  "components": [
+    {
+      "type": "library",
+      "bom-ref": "pkg:github/DMTF/libspdm@@VCS_TAG@",
+      "cpe": "cpe:2.3:a:DMTF:libspdm:@VCS_TAG@:*:*:*:*:*:*:*",
+      "name": "libspdm",
+      "version": "@VCS_VERSION@",
+      "description": "Sample implementation of DMTF SPDM",
```
steven-bellock, line 19:
```suggestion
      "description": "Reference implementation of SPDM",
```

steven-bellock (changes requested):

steven-bellock:
Semi-related issue: https://github.com/DMTF/libspdm/issues/2845 and @HeikeGilg for visibility.

@hughsie how does the CycloneDX distinguish between development branches and "official" releases?

jyao1:
@hughsie, CI failed due to no "Signed-off-by" tag. Would you please to add it?

hughsie:
> So ideally we'll have an automated tag / release process that auto-fills the appropriate tokens.

Ahh, this is done, magically. If you check out tag `3.5.0` as part of the firmware build process the `VCS_VERSION` in the libspdm SBOM file will be exactly that; if you check out main then the VCS tag will be something like `3.5.0-12-d0da67f3`. There's a load more detail in https://docs.google.com/presentation/d/1OPBHYZAr9SWDrmXpistJrVqd8wdN94oOfDfFXoEjTxg/edit?usp=sharing if you'd like to see the bigger picture.

> CI failed due to no "Signed-off-by" tag. Would you please to add it?

The original commit had it, I think the github "accept suggestion" probably confused it. I've rebased and updated the PR, thanks all!

steven-bellock:
GitHub releases that are downloaded, for example https://github.com/DMTF/libspdm/archive/refs/tags/3.5.0.zip, do not contain a `.git` directory; it's just the files, and so presumably the release process would need to fill in any fields that depend on Git.

hughsie:
@steven-bellock right, that's true. In the two cases I personally care about (coreboot and EDK II) libspdm is checked out as a git submodule, and I'm not sure how many people use tarballs like that now. What we do in fwupd is additionally use the uswid tool to populate the values at release time, i.e. https://github.com/fwupd/fwupd/blob/main/contrib/meson.build#L15 -- I'm not entirely sure how to do the same thing with cmake, but I think it should be broadly similar. The only fly in the ointment is that the .zip file generated by github won't have this _generated file_, but will have the _source file_. I'm open for ideas really.

steven-bellock:
For SBOMs, this becomes a larger question for the DMTF. In particular the entire DMTF should standardize on an SBOM format(s). I'll bring this up at the larger DMTF working groups like SPDM and the Technical Committee.

hughsie:
> For SBOMs, this becomes a larger question for the DMTF.

Probably true.

> In particular the entire DMTF should standardize on an SBOM format(s). 

DMTF does have an SBOM format: SWID -- it's just a shame hardly nobody uses it. If they do want to use SWID out of principal rather than SPDX (ISO) or CycloneDX let me know as the uswid tool can import/export in SWID format too and I can re-do the pull request to use that.

>I'll bring this up at the larger DMTF working groups like SPDM and the Technical Committee.

Great, thanks. I'm happy to be malleable; the **trusted** data is the bit I care about, the format less so.

PR #2906, opened 2024-11-19
Change "sample" to "reference"

steven-bellock:


jyao1:
@steven-bellock , is that discussed in SPDM WG?

steven-bellock:
Yes, we can discuss it today.

PR #2907, opened 2024-11-21
Update mbedtls to 3.6.2

os-d:
Mbed TLS 3.6.2 is the latest release in the long-term support 3.6.x series. It brings in some critical security fixes and is the recommended release of Mbed TLS from that community.

One of the fixes updates the certifi version referenced by Mbed TLS to one that fixes this CVE:
https://nvd.nist.gov/vuln/detail/CVE-2023-37920. CVE scanning tools pick up the version of certifi in Mbed TLS 3.6.0 and report libspdm in violation, even though it is not as libspdm does not use these mbedtls pip modules, as they are used only to build mbedtls's ReadTheDocs site.


steven-bellock:
Not sure why the builds are failing. 3.6.2 should still have the `mbedcrypto` target. [They only recently changed](https://github.com/Mbed-TLS/mbedtls/pull/9760) `mbedcrypto` to `tfpsacrypto`.

os-d:
> Not sure why the builds are failing. 3.6.2 should still have the `mbedcrypto` target. [They only recently changed](https://github.com/Mbed-TLS/mbedtls/pull/9760) `mbedcrypto` to `tfpsacrypto`.

Yeah, I'm looking into it now. It may be part of the deprecation of the mbedtls_sha3_* functions: https://github.com/Mbed-TLS/mbedtls/blob/v3.6.2/docs/psa-transition.md#summary-of-api-modules. If it does involve changing apisets, I will file an issue for this, as that is beyond the scope of what I am able to contribute here. I'm hoping the old apiset can continue to be used. Regardless, it does look like this is a change that libspdm will have to consume by MbedTLS 4.0.

os-d:
@steven-bellock , it was just added the SHA3 define to config, but it is strange that it wasn't needed before, because v3.6.0 had the same define: https://github.com/Mbed-TLS/mbedtls/blob/v3.6.0/library/sha3.c#L15

steven-bellock:
@psrisoontorn you confirmed that SHA3 "really" works with your pull requests, correct?

steven-bellock (approved):

jyao1 (approved):

psrisoontorn:
> @psrisoontorn you confirmed that SHA3 "really" works with your pull requests, correct?

Yes, I've tested SHA3 with mbedtls. I think the reason why I didn't see any issue is because 'MBEDTLS_SHA3_C' is already defined in "config_adjust_legacy_from_psa.h". There is a comment in the code indicating that the corresponding MBEDTLS_XXX settings would be enabled automatically if PSA_WANT_XXX symbols are defined. 
Note: PSA_WANT_ALG_SHA3_XXX symbols are defined in crypto_config.h. 

image

Anyway, I think adding the define for 'MBEDTLS_SHA3_C' to config.h is the right thing to do.

PR #2910, opened 2024-11-26
Reorganize `spdm_device_secret_lib_sample`

jyao1:
The original code is hard to maintain because we put so many features in one lib.c.

split lib.c to multiple smaller files.
cert.c -> read_pub_cert.c

Only code move, zero code change.


os_stub/spdm_device_secret_lib_sample/chal.c
```diff
@@ -0,0 +1,80 @@
+/**
+ *  Copyright Notice:
+ *  Copyright 2024 DMTF. All rights reserved.
+ *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
+ **/
+
+/** @file
```
steven-bellock, line 7: I would just delete this comment block from each file. It doesn't add anything and the filenames are probably descriptive enough.

jyao1: yes. agree. done.

steven-bellock (approved):

PR #2911, opened 2024-11-26
Clarify timeout and RTT documentation

### LINKED ISSUE #2904: libspdm_device_send_message_func should be called with the timeout parameter
Some BMC MCTP implementations are atomic, with send and receive as a single "transaction".

libspdm libspdm_register_device_io_func() splits these into two operations, each of which has a timeout parameter.  However, the timeout provided to the **send** libspdm_device_send_message_func is only the RTT time, whereas the timeout provided to the **receive** libspdm_device_receive_message_func is the full calculated CT exponent time.

Instead, the full RTT+(ST1 or CTexp) time should be provided to the libspdm_device_send_message_func.  For MCTP implementations which split send and receive, this should have no effect, as sends do not even use timeouts.

steven-bellock:
Fix #2904.

### DIFF
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -491,6 +491,9 @@ size_t libspdm_get_context_size_without_secured_context(void);
  *                       If called in a Requester context then timeout is equal to RTT.
  *                       If called in a Responder context then timeout is equal to 0 and Responder
  *                       should not timeout when sending the message.
+ *
+ * @retval LIBSPDM_STATUS_SUCCESS   The message was successfully sent to the receiving endpoint.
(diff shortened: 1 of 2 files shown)


steven-bellock:
@lordaule to review as well.

include/library/spdm_common_lib.h
lordaule, line 1: Was the removal of the retval documentation intentional?  The caller does check LIBSPDM_STATUS_IS_ERROR() after calling send or receive.

steven-bellock: These specific error codes no longer exist, but since send/receive is implemented by the Integrator they should know what error codes libspdm expects. I'll add them back in with the correct codes.

lordaule: Looks good to me

lordaule (approved):

Issue #2913, opened 2024-11-27
Do we really need `LIBSPDM_DATA_TOTAL_KEY_PAIRS`?


jyao1:
Currently, all key pair info is inside of spdm_device_secret_lib_sample.
The library hardcoded first 9 algorithms, and assert the total key pairs no greater than 16.

Also, it hardcoded the last (10, total_key_pairs) to same EDDSA 448. I am not sure why to do that.
And I am not sure why we need to expose `LIBSPDM_DATA_TOTAL_KEY_PAIRS`?

Why not just return `total_key_pairs` from spdm_device_secret_lib_sample?

steven-bellock:
See https://github.com/DMTF/libspdm/pull/2771#discussion_r1706099910.

jyao1:
Ok. If the purpose is to just **return** total pairs, I am fine with that.

But now the problem is the **set** total pairs. Do we really want to let integrator **set** key pairs number?
I think it is problematic, since it is already defined in spdm_device_secret_lib_sample().

I propose to remove the **set** feature https://github.com/DMTF/libspdm/blob/main/library/spdm_common_lib/libspdm_com_context_data.c#L792-L800.

steven-bellock:
>Do we really want to let integrator set key pairs number?

Yes, they need to be able to set it, else it will be `0`.

jyao1:
keep it in this version. Update device_secret_lib_sample to make it easy to use.

PR #2919, opened 2024-11-29
Key pairs number

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2913


steven-bellock:
This is a breaking change.

jyao1:
> This is a breaking change.

Yes. right.
If we keep SET_DATA, then I would think to change `LIBSPDM_ASSERT(total_key_pairs <= LIBSPDM_MAX_KEY_PAIR_COUNT);` to `LIBSPDM_ASSERT(total_key_pairs == LIBSPDM_MAX_KEY_PAIR_COUNT);`

That is to ensure the integrator use SET_DATA correctly, matching the one in `device_secret_lib_sample`.

We can remove it in libspdm 4.0, when we add SPDM 1.4 PQC support.
`

jyao1:
updated according to the discussion in the meeting.

steven-bellock (approved):

PR #2927, opened 2024-12-06
Make libspdm_get_certificate_choose_length[_ex]() public APIs

### LINKED ISSUE #2909: Make libspdm_get_certificate_choose_length[_ex]() a public API
spdm_requester_lib.h currently exports two get_certificate functions:
  libspdm_get_certificate() and libspdm_get_certificate_ex()

Neither accept a desired length.  There are functions which accept a length parameter, namely:
  libspdm_get_certificate_choose_length() and libspdm_get_certificate_choose_length_ex()
however these functions are "internal" and not exported.

I have now found two different SPDM responders which require specific length fields - one is an early SPDM version that requires the use of 0xFFFF to get the entire certificate, and the other which requires GET_CERTIFICATE calls to have a length less than  LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN (1024)

To support these responders, libspdm should move the libspdm_get_certificate_choose_length[_ex]() API functions from 
  include/internal/libspdm_requester_lib.h
to
  include/library/spdm_requester_lib.h

ShitalJumbad:
fix #2909


include/library/spdm_requester_lib.h
```diff
@@ -141,6 +141,60 @@ libspdm_return_t libspdm_get_certificate_ex(void *spdm_context,
...
+ **/
+libspdm_return_t libspdm_get_certificate_choose_length(void *spdm_context,
+                                                       const uint32_t *session_id,
+                                                       uint8_t slot_id,
+                                                       uint16_t length,
+                                                       size_t *cert_chain_size,
+                                                       void *cert_chain);
+
+/**
+ * This function sends GET_CERTIFICATE to get certificate chain in one slot from device.
```
steven-bellock, line 171: >and allows the Integrator to specify the size of the certificate chain blocks.

steven-bellock (dismissed):

include/library/spdm_requester_lib.h
```diff
@@ -141,6 +141,62 @@ libspdm_return_t libspdm_get_certificate_ex(void *spdm_context,
...
+ * @param  slot_id          The number of slot for the certificate chain.
+ * @param  length           The length of the certificate chain block to retrieve.
+ * @param  cert_chain_size  On input, indicate the size in bytes of the destination buffer to store the digest buffer.
+ *                          On output, indicate the size in bytes of the certificate chain.
+ * @param  cert_chain       A pointer to a destination buffer to store the certificate chain.
+ **/
+libspdm_return_t libspdm_get_certificate_choose_length(void *spdm_context,
+                                                       const uint32_t *session_id,
+                                                       uint8_t slot_id,
+                                                       uint16_t length,
```
jyao1, line 167: Can we use `uint32_t`, this is for new SPDM spec?

include/library/spdm_requester_lib.h
```diff
@@ -141,6 +141,62 @@ libspdm_return_t libspdm_get_certificate_ex(void *spdm_context,
...
+ * @param  cert_chain_size    On input, indicate the size in bytes of the destination buffer to store the digest buffer.
+ *                            On output, indicate the size in bytes of the certificate chain.
+ * @param  cert_chain         A pointer to a destination buffer to store the certificate chain.
+ * @param  trust_anchor       A buffer to hold the trust_anchor which is used to validate the peer certificate, if not NULL.
+ * @param  trust_anchor_size  A buffer to hold the trust_anchor_size, if not NULL.
+ **/
+libspdm_return_t libspdm_get_certificate_choose_length_ex(void *spdm_context,
+                                                          const uint32_t *session_id,
+                                                          uint8_t slot_id,
+                                                          uint16_t length,
```
jyao1, line 195: Can we use `uint32_t`, this is for new SPDM spec?

include/library/spdm_requester_lib.h
```diff
@@ -141,6 +141,62 @@ libspdm_return_t libspdm_get_certificate_ex(void *spdm_context,
...
+ * @param  spdm_context     A pointer to the SPDM context.
+ * @param  session_id       Indicates if it is a secured message protected via SPDM session.
+ *                          If session_id is NULL, it is a normal message.
+ * @param  slot_id          The number of slot for the certificate chain.
+ * @param  length           The length of the certificate chain block to retrieve.
+ * @param  cert_chain_size  On input, indicate the size in bytes of the destination buffer to store the digest buffer.
+ *                          On output, indicate the size in bytes of the certificate chain.
+ * @param  cert_chain       A pointer to a destination buffer to store the certificate chain.
+ **/
+libspdm_return_t libspdm_get_certificate_choose_length(void *spdm_context,
```
jyao1, line 164: If we have `_ex` version, do we still need this non `_ex` version?

include/library/spdm_requester_lib.h
steven-bellock: So if
- negotiated version is <= 1.3 then `length` must must be `<= 65535`
- negotiated version is >= 1.4 then libspdm uses `LargeOffset` and `LargeLength`.
?

include/library/spdm_requester_lib.h
steven-bellock: Yes, could probably just use the `_ex` version and integrator can set trust anchor to `NULL` if they don't want it.

include/library/spdm_requester_lib.h
jyao1: Please just check 0xFFFF in this version.

We can add more check for 1.4 later.

include/library/spdm_requester_lib.h
steven-bellock: @lordaule says it's ok too.

include/library/spdm_requester_lib.h
ShitalJumbad: For versions <= 1.3, even though we use uint32_t here, we will need to typecast it to uint16_t before passing it to libspdm_try_get_certificate in the function, after checking that length <= 0xFFFF. Trying to understand the motivation behind changing this to uint32 for the current version?

include/library/spdm_requester_lib.h
ShitalJumbad: Updated change to expose only _ex version.

include/library/spdm_requester_lib.h
ShitalJumbad: Updated.

steven-bellock (approved):

PR #2930, opened 2024-12-12
Clean up minor x.509 issues

steven-bellock:
- Add explicit NULL check.
- Fix typos.
- Make documentation and implementation consistent.

### DIFF
--- a/os_stub/cryptlib_mbedtls/pk/x509.c
+++ b/os_stub/cryptlib_mbedtls/pk/x509.c
@@ -197,23 +197,25 @@ bool libspdm_x509_construct_certificate_stack(uint8_t **x509_stack, ...)
 /**
  * Release the specified X509 object.
  *
- * If x509_cert is NULL, then return false.
+ * If x509_cert is NULL, then return early.
  *
  * @param[in]  x509_cert  Pointer to the X509 object to be released.
  *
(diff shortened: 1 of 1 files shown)


steven-bellock:
These issues were found by the NVIDIA CI/CD system.

os_stub/cryptlib_mbedtls/pk/x509.c
```diff
@@ -197,23 +197,25 @@ bool libspdm_x509_construct_certificate_stack(uint8_t **x509_stack, ...)
...
 {
-    if (x509_cert) {
-        mbedtls_x509_crt_free(x509_cert);
-        free_pool(x509_cert);
+    if (x509_cert == NULL) {
+        return;
     }
+
+    mbedtls_x509_crt_free(x509_cert);
+    free_pool(x509_cert);
```
jyao1, line 212: I don't understand the change.

What is the problem of original code?

steven-bellock: The codebase uses explicit `NULL` checks instead of treating the pointer as a boolean. In addition `mbedtls_x509_crt_free` does not specify that `x509_cert` can be `NULL`, and so the function explicitly returns early.

Issue #2934, opened 2024-12-18
error in calculation opaque data element


jyao1:
https://github.com/DMTF/libspdm/blob/main/library/spdm_common_lib/libspdm_com_opaque_data.c#L191-L192

```
        current_element_len = sizeof(secured_message_opaque_element_table_header_t) +
                              opaque_element_table_header->opaque_element_data_len;
```

This is wrong, because `secured_message_opaque_element_table_header_t` has assumption that it is DMTF ID only. The Vendor_len is 0.
But it is not the case for other ID, which has additional 2 bytes `vendor_id` after `vendor_len`.

```
typedef struct {
    uint8_t id; /* SPDM_REGISTRY_ID_DMTF*/
    uint8_t vendor_len;
    uint16_t opaque_element_data_len;
    /*uint8_t    sm_data_version;
     * uint8_t    sm_data_id;
     * uint8_t    sm_data[];*/
} secured_message_opaque_element_table_header_t;
```

The right way is to use `opaque_element_table_header_t` and add `vendor_id_len`.

The check below should be removed.

https://github.com/DMTF/libspdm/blob/main/library/spdm_common_lib/libspdm_com_opaque_data.c#L187

```
            (opaque_element_table_header->vendor_len != 0)) {
            return false;
        }
```

PR #2935, opened 2024-12-18
Allow non-zero vendor_id for libspdm_get_element_from_opaque_data()

jyao1:
Fix https://github.com/DMTF/libspdm/issues/2934

### DIFF
--- a/include/industry_standard/spdm.h
+++ b/include/industry_standard/spdm.h
@@ -478,6 +478,65 @@ typedef struct {
 #define SPDM_REGISTRY_ID_IANA_CBOR 0xa
 #define SPDM_REGISTRY_ID_MAX  0xa
 
+typedef struct {
+    uint8_t id;
+    uint8_t vendor_id_len;
+    /* uint8_t vendor_id[vendor_id_len]; */
+} spdm_svh_header_t;
+
+typedef struct {
+    spdm_svh_header_t header; /* SPDM_REGISTRY_ID_DMTF */
+} spdm_svh_dmtf_header_t;
+
+typedef struct {
+    spdm_svh_header_t header; /* SPDM_REGISTRY_ID_TCG */
+    uint16_t vendor_id;
+} spdm_svh_tcg_header_t;
+
+typedef struct {
+    spdm_svh_header_t header; /* SPDM_REGISTRY_ID_USB */
+    uint16_t vendor_id;
+} spdm_svh_usb_header_t;
+
+typedef struct {
+    spdm_svh_header_t header; /* SPDM_REGISTRY_ID_PCISIG */
+    uint16_t vendor_id;
+} spdm_svh_pcisig_header_t;
+
+typedef struct {
+    spdm_svh_header_t header; /* SPDM_REGISTRY_ID_IANA */
+    uint32_t vendor_id;
+} spdm_svh_iana_header_t;
+
+typedef struct {
+    spdm_svh_header_t header; /* SPDM_REGISTRY_ID_HDBASET */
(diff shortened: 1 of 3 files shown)


include/industry_standard/spdm.h
```diff
@@ -478,6 +478,76 @@ typedef struct {
...
 #define SPDM_REGISTRY_ID_MAX  0xa
 
+typedef struct {
+    uint8_t id;
+    uint8_t vendor_id_len;
+    /* uint8_t vendor_id[vendor_id_len]; */
+} spdm_svh_header_t;
+
+typedef struct {
+    uint8_t id; /* SPDM_REGISTRY_ID_DMTF */
```
steven-bellock, line 488: Can use `spdm_svh_header_t` since that's already defined.

unit_test/test_spdm_common/context_data.c
```diff
@@ -1542,6 +1542,129 @@ static void libspdm_test_max_session_count_case21(void **state)
...
+    opaque_data.dmtf_header.id = SPDM_REGISTRY_ID_DMTF;
+    opaque_data.dmtf_header.vendor_id_len = 0;
+    opaque_data.dmtf_opaque_len = sizeof(opaque_data.dmtf_opaque);
+    opaque_data.dmtf_opaque.sm_data_version = SECURED_MESSAGE_OPAQUE_ELEMENT_SMDATA_DATA_VERSION;
+    opaque_data.dmtf_opaque.sm_data_id = SECURED_MESSAGE_OPAQUE_ELEMENT_SMDATA_ID_VERSION_SELECTION;
+    opaque_data.dmtf_opaque.selected_version = SECURED_SPDM_VERSION_12 << 8;
+
+    opaque_data_ptr = (uint8_t *)&opaque_data;
+    opaque_data_size = sizeof(opaque_data);
+    status = libspdm_get_element_from_opaque_data(spdm_context,
```
steven-bellock, line 1659: Does `libspdm_get_element_from_opaque_data` only get the first `element_id`? What if opaque data contains multiple opaque elements with the same `element_id`?

include/industry_standard/spdm.h
jyao1: yes.

unit_test/test_spdm_common/context_data.c
jyao1: libspdm_get_element_from_opaque_data() will get first (element_id, and sm_data_id) pair. Although the name is generic, but it has assumption that element_id is DMTF, because sm_data_id is only under DMTF.

To me, this function is mess, I would not define in this way, if I can redo the work.

I plan to define two functions:
* `libspdm_get_element_from_opaque_data_with_element_id (element_id, element_index)` - that is to get element with ID and index to support multiple elements.
* `libspdm_get_sm_data_element_from_opaque_data (sm_data_id)` - that is to get SM_DATA only, with assumption that element_id is DMTF.

The first function can be used for SPDM 1.4 with DMTF-DSP, and support DSP0289 Authorization AODS.
`libspdm_get_aods_element_from_opaque_data (aods_id)` - that is to get AODS with assumption that element_id is DMTF-DSP.

I will file a new issue for the new function.

steven-bellock: >I will file a new issue for the new function.

That will be in a separate pull request?

jyao1: yes, separate PR definitely.
Most likely, I will submit after I finish DSP0289 Auth POC. Then we can know how to handle different element ID.

steven-bellock (approved):

Issue #2875, opened 2024-10-31
`ChunkSeqNo` should not wrap


steven-bellock:
I believe a reading of the Large SPDM message transfer mechanism implies that `ChunkSeqNo` should not wrap. Ie, if the message is too large for `65536` chunks then it should not be sent. However libspdm assumes wrapping, for example

https://github.com/DMTF/libspdm/blob/30014a822710d58a0387afdce5bce4ec144bcd74/library/spdm_responder_lib/libspdm_rsp_chunk_get.c#L157

I filed https://github.com/DMTF/SPDM-WG/issues/3703 to increase the size of `ChunkSeqNo` so that we don't have to think about this for SPDM 1.4 and greater.

jyao1:
Good finding.

Read the spec, the LargeMessageSize is present when ChunkSeqNo is 0. But it is not considered on wrap case.

 I think we should prevent wrap from happening.

steven-bellock:
The SPDM Working Group confirmed that it does not wrap. This will be made more explicit in the specification.

Li-Aaron:
I'll take this.

steven-bellock:
@Li-Aaron if the Responder is sending a message and it's too large for `65536` chunks then it can return `ResponseTooLarge`.

PR #2952, opened 2025-01-13
Fix chunk_seq_no wrap issue.

Li-Aaron:
Fix https://github.com/DMTF/libspdm/issues/2875

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_handle_error_response.c
+++ b/library/spdm_requester_lib/libspdm_req_handle_error_response.c
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2021-2024 DMTF. All rights reserved.
+ *  Copyright 2021-2025 DMTF. All rights reserved.
  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
  **/
 
@@ -225,6 +225,7 @@ libspdm_return_t libspdm_handle_error_large_response(
     size_t large_response_capacity;
     size_t large_response_size;
     size_t large_response_size_so_far;
+    uint64_t max_chunk_data_transfer_size;
 
     if (libspdm_get_connection_version(spdm_context) < SPDM_MESSAGE_VERSION_12) {
         return LIBSPDM_STATUS_UNSUPPORTED_CAP;
@@ -262,6 +263,10 @@ libspdm_return_t libspdm_handle_error_large_response(
     message = scratch_buffer + libspdm_get_scratch_buffer_sender_receiver_offset(spdm_context);
     message_size = libspdm_get_scratch_buffer_sender_receiver_capacity(spdm_context);
 
+    max_chunk_data_transfer_size =
+        ((size_t) spdm_context->local_context.capability.data_transfer_size
+         - sizeof(spdm_chunk_response_response_t)) * 65536 - sizeof(uint32_t);
+
     libspdm_zero_mem(large_response, large_response_capacity);
     large_response_size = 0;
     large_response_size_so_far = 0;
@@ -279,6 +284,12 @@ libspdm_return_t libspdm_handle_error_large_response(
         spdm_request->header.param2 = chunk_handle;
         spdm_request->chunk_seq_no = chunk_seq_no;
         spdm_request_size = sizeof(spdm_chunk_get_request_t);
+
+        if (chunk_seq_no == 0 && large_response_size_so_far != 0) {
+            /* chunk_seq_no wrapped */
+            status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
+            break;
+        }
 
         LIBSPDM_DEBUG((LIBSPDM_DEBUG_INFO,
                        "CHUNK_GET Handle %d SeqNo %d\n", chunk_handle, chunk_seq_no));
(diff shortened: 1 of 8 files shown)


library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -330,6 +331,19 @@ libspdm_return_t libspdm_handle_large_request(
...
+    /* Fail if exceed max chunks */
+    min_data_transfer_size = LIBSPDM_MIN(
+        spdm_context->connection_info.capability.data_transfer_size,
+        spdm_context->local_context.capability.sender_data_transfer_size);
+
+    max_chunk_data_transfer_size =
+        ((size_t) min_data_transfer_size - sizeof(spdm_chunk_send_request_t)) * 65536 -
+        sizeof(uint32_t);
+
+    if (request_size > max_chunk_data_transfer_size) {
```
Li-Aaron, line 343: also compare with max spdm message size

library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -330,6 +331,19 @@ libspdm_return_t libspdm_handle_large_request(
         return LIBSPDM_STATUS_ERROR_PEER;
     }
 
+    /* Fail if exceed max chunks */
+    min_data_transfer_size = LIBSPDM_MIN(
+        spdm_context->connection_info.capability.data_transfer_size,
+        spdm_context->local_context.capability.sender_data_transfer_size);
+
+    max_chunk_data_transfer_size =
```
jyao1, line 339: we should also consult max_spdm_msg_size.

library/spdm_requester_lib/libspdm_req_handle_error_response.c
```diff
@@ -280,6 +280,12 @@ libspdm_return_t libspdm_handle_error_large_response(
         spdm_request->chunk_seq_no = chunk_seq_no;
         spdm_request_size = sizeof(spdm_chunk_get_request_t);
 
+        if (chunk_seq_no == 0 && large_response_size_so_far != 0) {
```
Li-Aaron, line 283: add a unit test for this.

library/spdm_responder_lib/libspdm_rsp_chunk_get.c
```diff
@@ -103,12 +104,24 @@ libspdm_return_t libspdm_get_response_chunk_get(
...
     }
 
-    libspdm_zero_mem(response, *response_size);
-
     min_data_transfer_size = LIBSPDM_MIN(
         spdm_context->connection_info.capability.data_transfer_size,
         spdm_context->local_context.capability.sender_data_transfer_size);
 
+    /* Fail if exceed max chunks */
+    max_chunk_data_transfer_size =
```
jyao1, line 112: we should also consult max_spdm_msg_size.

Li-Aaron (commented):
note.

steven-bellock:
When the Requester receives the first `CHUNK_RESPONSE` it can use `LargeMessageSize` to determine whether `ChunkSeqNo` will wrap, assuming the size of each chunk is equal to the Requester's `DataTransferSize`. A similar calculation can take place on the Responder side when it receives the first `CHUNK_SEND`.

Li-Aaron:
> When the Requester receives the first `CHUNK_RESPONSE` it can use `LargeMessageSize` to determine whether `ChunkSeqNo` will wrap, assuming the size of each chunk is equal to the Requester's `DataTransferSize`. A similar calculation can take place on the Responder side when it receives the first `CHUNK_SEND`.

In Spec line 812:
> `ChunkSize` shall be a value where the total size of `CHUNK_SEND` or `CHUNK_RESPONSE` shall be from `MinDataTransferSize` to the `DataTransferSize` of the receiving SPDM endpoint. 

Which means the Responder could send chunk messages with chunk size less than Requester's `DataTransferSize`.

steven-bellock:
Are you saying that even if the check passes then the `ChunkSeqNo` may still wrap? That is correct. However to determine that the Requester has to retrieve 65536 messages to determine that, whereas if the check fails then the Requester knows from the beginning the `ChunkSeqNo` would wrap and not spend the time retrieving all those messages only for `ChunkSeqNo` to wrap. It's a performance improvement. The original check is still needed.

Li-Aaron:
Got it, it make sense to add additional check to improve the performance.

library/spdm_requester_lib/libspdm_req_handle_error_response.c
```diff
@@ -262,6 +263,10 @@ libspdm_return_t libspdm_handle_error_large_response(
     message = scratch_buffer + libspdm_get_scratch_buffer_sender_receiver_offset(spdm_context);
     message_size = libspdm_get_scratch_buffer_sender_receiver_capacity(spdm_context);
 
+    max_chunk_data_transfer_size =
```
steven-bellock, line 266: This calculation can overflow if `sizeof(size_t) == 4`. I would make `max_chunk_data_transfer_size` a `uint64_t`.

library/spdm_requester_lib/libspdm_req_send_receive.c
steven-bellock: The Responder's `max_spdm_msg_size` is already checked in `libspdm_send_spdm_request`. Checking the Requester's `max_spdm_msg_size` doesn't make sense, as the message already exists at this point and so must be small enough to fit.

library/spdm_responder_lib/libspdm_rsp_chunk_get.c
steven-bellock: Same feedback as in https://github.com/DMTF/libspdm/pull/2952/files#r1943830275.

steven-bellock:
Re-opened two comments.

library/spdm_requester_lib/libspdm_req_send_receive.c
jyao1: Ok. If it is already checked, just add a comment to say : it is already checked.

steven-bellock (approved):

PR #2953, opened 2025-01-13
Add mask for `AlgSupported` reserved bits

### LINKED ISSUE #2950: `AlgSupported` checks reserved bits
For example https://github.com/DMTF/libspdm/blob/5ebe5e3946b9439928fa3a7548268c29cccc1b16/library/spdm_responder_lib/libspdm_rsp_algorithms.c#L457-L462

where DHE only uses seven bits. A mask needs to be applied so that reserved bits are not checked.

steven-bellock:
Fix #2950.


steven-bellock:
@jyao1 I think you're right, I should do them all at once.

PR #2957, opened 2025-01-15
PQC proof-of-concept from SPDM 1.4

jyao1:


.github/workflows/cifuzz.yml
```diff
@@ -4,6 +4,9 @@ on:
     paths-ignore:
       - 'doc/**'
       - '**/*.md'
+    branches:
+      - main
+      - release-2.3
```
steven-bellock, line 9: Just `main`. I can delete `2.3` from the other CI/CD files, since that branch is now stale.

include/hal/library/cryptlib/cryptlib_cert_pqc.h
```diff
@@ -0,0 +1,31 @@
+/**
+ *  Copyright Notice:
+ *  Copyright 2021-2024 DMTF. All rights reserved.
```
steven-bellock, line 3: All new files need `Copyright 2025 DMTF`.

CMakeLists.txt
```diff
@@ -939,6 +939,10 @@ else()
         add_subdirectory(os_stub/spdm_device_secret_lib_null)
         add_subdirectory(os_stub/spdm_cert_verify_callback_sample)
 
+        add_subdirectory(os_stub/oqslib)
+        add_subdirectory(os_stub/pqccryptlib_oqs)
```
steven-bellock, line 943: `pqccryptlib_oqs` -> `cryptlib_liboqs`

README.md
```diff
@@ -171,6 +174,8 @@ For other architectures, refer to [build](https://github.com/DMTF/libspdm/blob/m
 
 2) [OpenSSL](https://www.openssl.org) as cryptography library. Version 3.0.14.
 
+3) [OQS](https://github.com/open-quantum-safe/liboqs) as PQC cryptography library. Version 0.12.0.
```
steven-bellock, line 177: Is this intended to be a long-lived cryptography library, or just one that will be used for development and eventually MbedTLS and OpenSSL will support PQC in production?

steven-bellock (commented):
Need to go over how this should be split up into smaller commits and pull requests.

.github/workflows/cifuzz.yml
jyao1: OK

include/hal/library/cryptlib/cryptlib_cert_pqc.h
jyao1: OK

README.md
jyao1: I would say: If MbedTLS and OpenSSL support PQC in production, we can remove OQS for sure.

But before that, we have to use OQS for development. And I don't know **WHEN** the mbedTLS or OpenSSL will be ready.

jyao1: I added clarification.

library/spdm_responder_lib/libspdm_rsp_algorithms.c
```diff
@@ -631,12 +801,46 @@ libspdm_return_t libspdm_get_response_algorithms(libspdm_context_t *spdm_context
         }
     }
 
+    /* if both PQC and tradition are enabled, choose PQC and disable tradition */
```
xiangfeima, line 804: @jyao I don't find any wording about priority between pqc and traditional algorithms in SPDM1.4-WIP70 spec. Just out of curiosity where the does principle come from?

jyao1: It is implementation choice.

The purpose to use 1.4 is to enable PQC. If no PQC requirement, there is no need to use 1.4.

PR #2960, opened 2025-01-15
Fix `ENCAP_CAP` checks to accommodate SPDM 1.2

### LINKED ISSUE #2959: Key exchange fails on 'encap_cap_both' check when running SPDM 1.2
When running SPDM 1.2 even that encap_cap is deprecated in DMTF spec [SPDM1.2](https://www.dmtf.org/sites/default/files/standards/documents/DSP0274_1.2.1.pdf#page=51&zoom=100,0,417):
Still there is a check for the responder and requester encap_cap capability values to be true (_source/fw/tools/libspdm_emulator/src/library/spdm_requester_lib/libspdm_req_key_exchange.c_) - what cause a failure in the flow, and it's wrong.
Need to add there a check what's the SPDM version that configured.

steven-bellock:
Fix #2959.

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_key_exchange.c
+++ b/library/spdm_requester_lib/libspdm_req_key_exchange.c
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2021-2024 DMTF. All rights reserved.
+ *  Copyright 2021-2025 DMTF. All rights reserved.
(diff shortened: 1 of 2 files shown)


steven-bellock:
@Miritq as well.

steven-bellock:
For reference `libspdm_is_encap_supported` is implemented as https://github.com/DMTF/libspdm/blob/87d6c6c938d34b3b01c5440cb03f5b36879bd47c/library/spdm_common_lib/libspdm_com_context_data.c#L2367-L2398

jyao1 (approved):

PR #2968, opened 2025-01-25
CAPABILITIES: Add SupportedAlgorithms

### LINKED ISSUE #2279: [SPDM 1.3] CAPABILITIES: Add SupportedAlgorithms

ShitalJumbad:
fix #2279


library/spdm_responder_lib/libspdm_rsp_capabilities.c
```diff
@@ -264,10 +264,72 @@ libspdm_return_t libspdm_get_response_capabilities(libspdm_context_t *spdm_conte
...
     }
 
-    if (spdm_response->header.spdm_version >= SPDM_MESSAGE_VERSION_12) {
+    if (spdm_response->header.spdm_version >= SPDM_MESSAGE_VERSION_13 &&
+        (spdm_request->header.param1 & 0x01) &&
+        ((spdm_request->flags &
+          SPDM_GET_CAPABILITIES_REQUEST_FLAGS_CHUNK_CAP) == 0)) {
         *response_size = sizeof(spdm_capabilities_response_t);
+
+        spdm_response->supported_algorithms.param1 = spdm_context->local_context.algorithm.param1;
```
steven-bellock, line 273: Need to check if `dhe_named_group`, `aead_cipher_suite`, etc, is non-zero to populate this.

library/spdm_responder_lib/libspdm_rsp_capabilities.c
```diff
@@ -264,10 +264,72 @@ libspdm_return_t libspdm_get_response_capabilities(libspdm_context_t *spdm_conte
...
+        spdm_response->supported_algorithms.ext_hash_count = 0;
+        spdm_response->supported_algorithms.mel_specification =
+            spdm_context->local_context.algorithm.mel_spec;
+
+        if (spdm_response->supported_algorithms.param1 > 0) {
+            size_t index = 0;
+            spdm_response->supported_algorithms.length +=
+                spdm_response->supported_algorithms.param1*
+                sizeof(spdm_negotiate_algorithms_common_struct_table_t);
+            if (spdm_context->local_context.algorithm.dhe_named_group) {
```
steven-bellock, line 294: `!= 0`. See #2954.

steven-bellock (changes requested):
Still need to figure out how this information is conveyed to the Integrator.

include/industry_standard/spdm.h
```diff
@@ -161,6 +161,36 @@ typedef struct {
     uint32_t max_spdm_msg_size;
 } spdm_get_capabilities_request_t;
 
+typedef struct {
+    uint8_t alg_type;
+    uint8_t alg_count;
+    uint16_t alg_supported;
+} spdm_negotiate_algorithms_common_struct_table_t;
```
jyao1, line 168: I am not sure if we should define this. This definition is not so common.

include/industry_standard/spdm.h
```diff
@@ -161,6 +161,36 @@ typedef struct {
...
+    uint32_t base_hash_algo;
+    uint8_t reserved2[12];
+    uint8_t ext_asym_count;
+    uint8_t ext_hash_count;
+    uint8_t reserved3;
+    uint8_t mel_specification;
+    /* Followed by dynamic arrays for ext_asym and ext_hash, if needed
+     * spdm_extended_algorithm_t ext_asym[ext_asym_count];
+     * spdm_extended_algorithm_t ext_hash[ext_hash_count]; */
+    spdm_negotiate_algorithms_common_struct_table_t struct_table[
```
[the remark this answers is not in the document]
jyao1: agree

include/internal/libspdm_common_lib.h
```diff
@@ -43,6 +43,7 @@ typedef struct {
 } libspdm_device_version_t;
 
 typedef struct {
+    uint8_t param1;
```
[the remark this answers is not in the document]
jyao1: I suggest to add a new function.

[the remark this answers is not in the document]
steven-bellock: Actually the function should be called `libspdm_get_supported_algorithms` since that's the field name.

include/industry_standard/spdm.h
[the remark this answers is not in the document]
ShitalJumbad: Updated

include/internal/libspdm_common_lib.h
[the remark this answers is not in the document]
ShitalJumbad: Added a new function.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -251,6 +251,10 @@ static libspdm_return_t libspdm_try_get_capabilities(libspdm_context_t *spdm_con
     }
     spdm_request->header.request_response_code = SPDM_GET_CAPABILITIES;
     spdm_request->header.param1 = 0;
+    if (spdm_request->header.spdm_version >= SPDM_MESSAGE_VERSION_13 &&
```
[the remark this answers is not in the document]
ShitalJumbad: Done.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -308,22 +312,36 @@ static libspdm_return_t libspdm_try_get_capabilities(libspdm_context_t *spdm_con
...
+        if (spdm_response_size < (sizeof(spdm_capabilities_response_t) -
+                                  sizeof(spdm_supported_algorithms_block_t) -
+                                  sizeof(spdm_response->data_transfer_size) -
+                                  sizeof(spdm_response->max_spdm_msg_size))) {
             status = LIBSPDM_STATUS_INVALID_MSG_SIZE;
             goto receive_done;
         }
     }
-    if (spdm_request->header.spdm_version >= SPDM_MESSAGE_VERSION_12) {
+
```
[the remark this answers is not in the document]
ShitalJumbad: Updated the change, please let me know if it looks okay.

library/spdm_responder_lib/libspdm_rsp_capabilities.c
```diff
@@ -264,10 +264,72 @@ libspdm_return_t libspdm_get_response_capabilities(libspdm_context_t *spdm_conte
             spdm_context->local_context.capability.max_spdm_msg_size;
     }
 
-    if (spdm_response->header.spdm_version >= SPDM_MESSAGE_VERSION_12) {
+    if (spdm_response->header.spdm_version >= SPDM_MESSAGE_VERSION_13 &&
+        (spdm_request->header.param1 & 0x01) &&
+        ((spdm_request->flags &
+          SPDM_GET_CAPABILITIES_REQUEST_FLAGS_CHUNK_CAP) == 0)) {
```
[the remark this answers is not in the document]
ShitalJumbad: Removed it.

include/internal/libspdm_common_lib.h
```diff
@@ -55,6 +56,7 @@ typedef struct {
 } libspdm_device_capability_t;
 
 typedef struct {
+    uint8_t param1;
```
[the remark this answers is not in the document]
ShitalJumbad: removed

library/spdm_responder_lib/libspdm_rsp_capabilities.c
ShitalJumbad: updated

library/spdm_responder_lib/libspdm_rsp_capabilities.c
ShitalJumbad: done

library/spdm_responder_lib/libspdm_rsp_capabilities.c
```diff
@@ -271,10 +271,98 @@ libspdm_return_t libspdm_get_response_capabilities(libspdm_context_t *spdm_conte
             spdm_context->local_context.capability.max_spdm_msg_size;
     }
```
steven-bellock, line 273: As a basic validation step, need to check that if negotiated version is >= 1.3 and `Param1[0]` is set then the Requester's `CHUNK_CAP` is also set.

library/spdm_responder_lib/libspdm_rsp_capabilities.c
```diff
@@ -271,10 +271,98 @@ libspdm_return_t libspdm_get_response_capabilities(libspdm_context_t *spdm_conte
...
 
-    if (spdm_response->header.spdm_version >= SPDM_MESSAGE_VERSION_12) {
+    if (spdm_response->header.spdm_version >= SPDM_MESSAGE_VERSION_13 &&
```
steven-bellock, line 274: This is also conditional on both Requester and Responder supporting `CHUNK_CAP`.

library/spdm_responder_lib/libspdm_rsp_capabilities.c
```diff
@@ -271,10 +271,98 @@ libspdm_return_t libspdm_get_response_capabilities(libspdm_context_t *spdm_conte
...
+        spdm_response->supported_algorithms.base_asym_algo=
+            spdm_context->local_context.algorithm.base_asym_algo;
+        spdm_response->supported_algorithms.base_hash_algo=
+            spdm_context->local_context.algorithm.base_hash_algo;
+        spdm_response->supported_algorithms.ext_asym_count = 0;
+        spdm_response->supported_algorithms.ext_hash_count = 0;
+        spdm_response->supported_algorithms.mel_specification =
+            spdm_context->local_context.algorithm.mel_spec;
+
+        uint8_t index = 0;
```
steven-bellock, line 293: Need to declare this at line 277 before any statements. The libspdm codebase tries to be as C89-compatible as possible.

library/spdm_responder_lib/libspdm_rsp_capabilities.c
```diff
@@ -271,10 +271,98 @@ libspdm_return_t libspdm_get_response_capabilities(libspdm_context_t *spdm_conte
...
+
+        spdm_negotiate_algorithms_common_struct_table_t *struct_table =
+            (spdm_negotiate_algorithms_common_struct_table_t*)(&spdm_response->supported_algorithms
+                                                               +1);
+
+        if ((spdm_context->connection_info.capability.flags &
```
steven-bellock, line 299: On the Responder side I don't think the capability checks are needed. If Integrator has a non-zero algorithm then just send it. `libspdm_check_context` can be used by the Integrator to check if they have configured things correctly.

steven-bellock (changes requested):

doc/api/requester_api.md
```diff
@@ -31,6 +31,29 @@ returns early with value not equal to `LIBSPDM_STATUS_SUCCESS` then the SPDM con
 before attempting establish a new connection.
 <br/><br/>
 
+---
+### libspdm_get_supported_algorithms
+---
+
+### Description
+Sends the `GET_VERSION` and `GET_CAPABILITIES` requests, where GET_CAPABILITIES.Param1[0] is set.
```
hiption, line 39: Is it planned that the next version of SPDM will also support `extended capabilities` in `GET_VERSION`?

steven-bellock: That is a question for https://github.com/DMTF/SPDM-WG.

include/library/spdm_requester_lib.h
```diff
@@ -491,6 +491,19 @@ libspdm_return_t libspdm_start_session(void *spdm_context, bool use_psk,
...
+ * Sends the `GET_VERSION` and `GET_CAPABILITIES` requests, where GET_CAPABILITIES.Param1[0] is set
+ * If the Responder supports this extended capability, the Responder will include the Supported
+ * Algorithms Block in its CAPABILITIES response.
+ *
+ * @param spdm_context The SPDM context.
+ *
+ * @retval RETURN_SUCCESS               The connection is initialized successfully.
+ * @retval RETURN_DEVICE_ERROR          A device error occurs when communicates with the device.
+ **/
+libspdm_return_t libspdm_get_supported_algorithms(
```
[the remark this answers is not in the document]
ShitalJumbad: Updated.

library/spdm_responder_lib/libspdm_rsp_capabilities.c
ShitalJumbad: Updated.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -417,3 +466,23 @@ libspdm_return_t libspdm_get_capabilities(libspdm_context_t *spdm_context)
 
     return status;
 }
+
+libspdm_return_t libspdm_get_supported_algorithms(void *spdm_context)
```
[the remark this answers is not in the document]
ShitalJumbad: Updated

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -417,3 +466,23 @@ libspdm_return_t libspdm_get_capabilities(libspdm_context_t *spdm_context)
...
+    libspdm_return_t status;
+    libspdm_context_t *context;
+
+    context = spdm_context;
+
+    status = libspdm_get_version(context, NULL, NULL);
+    if (LIBSPDM_STATUS_IS_ERROR(status)) {
+        return status;
+    }
+
```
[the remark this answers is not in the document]
ShitalJumbad: updated

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -417,3 +466,23 @@ libspdm_return_t libspdm_get_capabilities(libspdm_context_t *spdm_context)
 
     return status;
 }
+
+libspdm_return_t libspdm_get_supported_algorithms(void *spdm_context)
+{
+    libspdm_return_t status;
+    libspdm_context_t *context;
+
```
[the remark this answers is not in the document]
ShitalJumbad: Done

library/spdm_responder_lib/libspdm_rsp_capabilities.c
ShitalJumbad: Done

library/spdm_responder_lib/libspdm_rsp_capabilities.c
ShitalJumbad: Removed.

include/library/spdm_requester_lib.h
```diff
@@ -491,6 +491,19 @@ libspdm_return_t libspdm_start_session(void *spdm_context, bool use_psk,
...
+    void *spdm_context, void *responder_supported_algorithms);
```
jyao1, line 505: I think we need have `responder_supported_algorithms_buffer_length` parameter to indicate the input buffer size.

include/industry_standard/spdm.h
```diff
@@ -161,8 +161,61 @@ typedef struct {
...
+    uint8_t registry_id;
+    uint8_t reserved;
+    uint16_t algorithm_id;
+} spdm_extended_algorithm_t;
+
+typedef struct {
+    uint8_t alg_type;
+    uint8_t alg_count;
+    uint16_t alg_supported;
+} spdm_negotiate_algorithms_common_struct_table_t;
```
jyao1, line 175: This is not correct.
2 bytes alg_supported field is NOT a common one.

include/industry_standard/spdm.h
```diff
@@ -161,8 +161,61 @@ typedef struct {
...
+    uint8_t mel_specification;
+    /* Followed by dynamic arrays for ext_asym, ext_hash, and struct_tableif needed
+     * spdm_extended_algorithm_t ext_asym[ext_asym_count];
+     * spdm_extended_algorithm_t ext_hash[ext_hash_count];
+     * spdm_negotiate_algorithms_common_struct_table_t struct_table[
+     *  SPDM_NEGOTIATE_ALGORITHMS_MAX_NUM_STRUCT_TABLE_ALG];*/
+} spdm_supported_algorithms_block_t;
+
+/* Specification states that total Extended algorithms count is less than or equal to 20*/
+#define SPDM_ALGORITHMS_MAX_NUM_EXT_ASYM_COUNT 20
```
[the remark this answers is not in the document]
jyao1: I am not sure if we need to add SPDM version number here.

The MAX counter value maybe changed in different version.

include/industry_standard/spdm.h
```diff
@@ -174,6 +227,8 @@ typedef struct {
     /* Below field is added in 1.2.*/
     uint32_t data_transfer_size;
     uint32_t max_spdm_msg_size;
+    /* Below field is added in 1.3.*/
+    spdm_supported_algorithms_block_t supported_algorithms;
```
jyao1, line 231: maybe just comment this field?

This size is NOT accurate anyway.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -315,22 +326,60 @@ static libspdm_return_t libspdm_try_get_capabilities(libspdm_context_t *spdm_con
...
     }
-    if (spdm_response->header.spdm_version >= SPDM_MESSAGE_VERSION_12) {
+    if (spdm_response->header.spdm_version >= SPDM_MESSAGE_VERSION_13) {
         if (spdm_response_size < sizeof(spdm_capabilities_response_t)) {
             status = LIBSPDM_STATUS_INVALID_MSG_SIZE;
             goto receive_done;
         }
+    } else if (spdm_response->header.spdm_version >= SPDM_MESSAGE_VERSION_12) {
+        if (spdm_response_size < (sizeof(spdm_capabilities_response_t) -
+                                  sizeof(spdm_supported_algorithms_block_t))) {
```
jyao1, line 336: If you remove spdm_supported_algorithms_block_t from definition, then you dont need to change here.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -315,22 +326,60 @@ static libspdm_return_t libspdm_try_get_capabilities(libspdm_context_t *spdm_con
...
+            status = LIBSPDM_STATUS_INVALID_MSG_SIZE;
+            goto receive_done;
+        }
     } else {
-        if (spdm_response_size < sizeof(spdm_capabilities_response_t) -
-            sizeof(spdm_response->data_transfer_size) - sizeof(spdm_response->max_spdm_msg_size)) {
+        if (spdm_response_size < (sizeof(spdm_capabilities_response_t) -
+                                  sizeof(spdm_supported_algorithms_block_t) -
```
jyao1, line 342: If you remove spdm_supported_algorithms_block_t from definition, then you dont need to change here.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -315,22 +326,60 @@ static libspdm_return_t libspdm_try_get_capabilities(libspdm_context_t *spdm_con
...
+            index++;
+        }
+        if ((spdm_context->connection_info.capability.flags &
+             SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_KEY_UPD_CAP) ||
+            (spdm_context->connection_info.capability.flags &
+             SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_PSK_CAP)) {
+            index++;
+        }
+        spdm_response_size = sizeof(spdm_capabilities_response_t)+
+                             index*sizeof(spdm_negotiate_algorithms_common_struct_table_t);
```
jyao1, line 375: I dont understand. Why index is calculated according to the CAP???

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -383,6 +432,61 @@ static libspdm_return_t libspdm_try_get_capabilities(libspdm_context_t *spdm_con
...
+        }
+
+        if (src->param1 > 0) {
+            size_t copy_size = src->param1 *
+                               sizeof(spdm_negotiate_algorithms_common_struct_table_t);
+            libspdm_copy_mem(dst->struct_table,
+                             sizeof(dst->struct_table),
+                             src_bytes + offset,
+                             copy_size);
+        }
```
jyao1, line 485: I am not sure why we need to copy by bytes.

Whey not just calculate the whole buffer size, then use one copy_mem() for the whole buffer?

library/spdm_responder_lib/libspdm_rsp_capabilities.c
```diff
@@ -270,11 +270,90 @@ libspdm_return_t libspdm_get_response_capabilities(libspdm_context_t *spdm_conte
...
+        }
+
+        if (spdm_context->local_context.algorithm.key_schedule != 0) {
+            struct_table[index].alg_type =
+                SPDM_NEGOTIATE_ALGORITHMS_STRUCT_TABLE_ALG_TYPE_KEY_SCHEDULE;
+            struct_table[index].alg_count = 0x20;
+            struct_table[index].alg_supported =
+                spdm_context->local_context.algorithm.key_schedule;
+            index++;
+        }
```
jyao1, line 336: I am thinking if we need re-evaluate spdm_context->local_context.capability.flags based on the negotiated version. Not just the provisioned algorithm.

E.g. if the negotiated version is low, then the provisioned algo is NOT supported, we should NOT return.

include/industry_standard/spdm.h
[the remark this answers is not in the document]
steven-bellock: Filed https://github.com/DMTF/libspdm/issues/3019.

include/industry_standard/spdm.h
steven-bellock: This is consistent with the other `struct` where they expand with each SPDM version.
>This size is NOT accurate anyway.

This uses `spdm_supported_algorithms_block_t` which is accurate. `spdm_responder_supported_algorithms_t`, which is returned to the Integrator, would be inaccurate.

include/library/spdm_requester_lib.h
jyao1: I don't think so. `spdm_responder_supported_algorithms_t` does not include the extended algorithms and struct table, which are variable length.

include/industry_standard/spdm.h
jyao1: > This uses spdm_supported_algorithms_block_t which is accurate.

I don't think so, the `spdm_supported_algorithms_block_t` does not include the extended algorithms and struct table, which are variable length.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
jyao1: Ah, I think that I overlooked `spdm_responder_supported_algorithms_t`.

I feel this definition is problematic, because we will keep changing the data structure whenever we have new fields, e.g. in SPDM 1.4.

To keep the interface simple, we can
1) return everything as is to the integrator.
2) (or) record individual inside of context, then allow the integrator retrieve based on the need.

library/spdm_responder_lib/libspdm_rsp_capabilities.c
jyao1: Sure, I am not worried about reserved bit check in 2950.

I am more thinking about version check. Yes, it supports all for SPDM 1.3.
If we add SPDM 1.4, then we need to add version check.

I am OK with current one, then.

include/industry_standard/spdm.h
```diff
@@ -161,8 +161,61 @@ typedef struct {
...
+    uint32_t base_asym_algo;
+    uint32_t base_hash_algo;
+    uint8_t ext_asym_count;
+    uint8_t ext_hash_count;
+    uint8_t mel_specification;
+    spdm_extended_algorithm_t ext_asym[SPDM_ALGORITHMS_MAX_NUM_EXT_ASYM_COUNT];
+    spdm_extended_algorithm_t ext_hash[SPDM_ALGORITHMS_MAX_NUM_EXT_HASH_COUNT];
+    spdm_negotiate_algorithms_common_struct_table_t struct_table[
+        SPDM_ALGORITHMS_MAX_NUM_STRUCT_TABLE_ALG];
+} spdm_responder_supported_algorithms_t;
```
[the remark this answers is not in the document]
steven-bellock: >I think it should be moved to requester lib

Agreed.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
steven-bellock: >I feel this definition is problematic, because we will keep changing the data structure whenever we have new fields

That's fine, we do that already with the `_ex` functions. 
>record individual inside of context, then allow the integrator retrieve based on the need.

The `spdm_context`? This is ephemeral data and should not increase the size of the context. Once the Integrator has consumed it then they don't need to retrieve it again, at least from the connection's context.

include/industry_standard/spdm.h
```diff
@@ -161,8 +161,61 @@ typedef struct {
...
+} spdm_supported_algorithms_block_t;
+
+/* Specification states that total Extended algorithms count is less than or equal to 20*/
+#define SPDM_ALGORITHMS_MAX_NUM_EXT_ASYM_COUNT 20
+#define SPDM_ALGORITHMS_MAX_NUM_EXT_HASH_COUNT 20
+#define SPDM_ALGORITHMS_MAX_NUM_STRUCT_TABLE_ALG 4
+
+/* SPDM supported algorithms by responder */
+typedef struct {
+    uint8_t measurement_specification;
```
jyao1, line 205: Suggest to add param1 and param2.
At least param1 means the Number of Algorithms Structure Tables
Otherwise, this information is lost.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
jyao1: > That's fine, we do that already with the _ex functions.

But current _ex function only add more parameter, we did not change any data structure. Here, you must change a data structure.
Also, with current data structure, the integrator cannot know which field is valid or invalid. They should be based on the version.

If you really like to use this style API, here is my thought (combining other feedback):
1) Move the structure definition to request lib (because it is NOT defined in SPDM spec)
2) Add a SPDM version as the first field for the data structure, to indicate the structure version.
3) Add SPDM version as prefix or suffix for the data structure name, as indicated by the first SPDM version field.
4) Add Length field in the function as input or output (which allows the integrator or libspdm checking the structure size.)

Something like:
```
libspdm_get_supported_algorithms (
void spdm_context,
// on input: indicate the max size of the responder_supported_algorithms_buffer
// on output: indicate the valid size of the responder_supported_algorithms_buffer
size_t *responder_supported_algorithms_length,
// on output: the consumer must get the spdm_version to know the version of the data structure
void *responder_supported_algorithms_buffer);

typedef struct {
    uint8 spdm_version; // must be 0x13
    ...... // rest version 1.3 specific field
} libspdm_responder_supported_algorithms_13_t;

typedef struct {
    uint8 spdm_version; // must be 0x14
    ...... // rest version 1.4 specific field
} libspdm_responder_supported_algorithms_14_t;
```

doc/api/requester_api.md
```diff
@@ -31,6 +31,29 @@ returns early with value not equal to `LIBSPDM_STATUS_SUCCESS` then the SPDM con
...
+If the Responder supports this extended capability, the Responder will include the Supported
+Algorithms Block in its CAPABILITIES response. If the requester wishes to continue with the
+connection, they can call `libspdm_init_connection` to restart the connection process.
+
+### Parameters
+
+**spdm_context**<br/>
```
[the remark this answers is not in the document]
ShitalJumbad: Done.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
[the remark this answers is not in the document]
ShitalJumbad: Fixed.

library/spdm_responder_lib/libspdm_rsp_capabilities.c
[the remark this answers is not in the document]
ShitalJumbad: Fixed.

include/internal/libspdm_common_lib.h
```diff
@@ -67,6 +67,30 @@ typedef struct {
...
+typedef struct {
+    uint8_t measurement_specification;
+    uint8_t other_params_support;
+    uint32_t base_asym_algo;
+    uint32_t base_hash_algo;
+    uint8_t mel_specification;
+    uint8_t ext_asym_count;
+    spdm_extended_algorithm_t ext_asym[LIBSPDM_MAX_EXT_ALG_COUNT];
+    uint8_t ext_hash_count;
+    spdm_extended_algorithm_t ext_hash[LIBSPDM_MAX_EXT_ALG_COUNT];
```
jyao1, line 88: libspdm does not support ext_asym.
Not sure if we really need to report ext_.

include/internal/libspdm_common_lib.h
```diff
@@ -67,6 +67,30 @@ typedef struct {
...
+
+#pragma pack(1)
+typedef struct {
+    spdm_negotiate_algorithms_common_struct_table_t alg_struct;
+    spdm_extended_algorithm_t alg_external[LIBSPDM_MAX_EXT_ALG_COUNT];
+} libspdm_supported_algorithms_alg_struct_t;
+#pragma pack()
+
+typedef struct {
+    uint8_t measurement_specification;
```
jyao1, line 80: Need to add version of the structure.

include/internal/libspdm_common_lib.h
```diff
@@ -67,6 +67,30 @@ typedef struct {
...
+    uint32_t base_hash_algo;
+    uint8_t mel_specification;
+    uint8_t ext_asym_count;
+    spdm_extended_algorithm_t ext_asym[LIBSPDM_MAX_EXT_ALG_COUNT];
+    uint8_t ext_hash_count;
+    spdm_extended_algorithm_t ext_hash[LIBSPDM_MAX_EXT_ALG_COUNT];
+    uint8_t struct_table_count;
+    libspdm_supported_algorithms_alg_struct_t
+        struct_table[SPDM_NEGOTIATE_ALGORITHMS_MAX_NUM_STRUCT_TABLE_ALG];
+} libspdm_responder_supported_algorithms_t;
```
jyao1, line 92: Need to add version to the name.

include/internal/libspdm_requester_lib.h
```diff
@@ -141,7 +142,9 @@ libspdm_return_t libspdm_get_version(libspdm_context_t *spdm_context,
  * @retval RETURN_SUCCESS               The GET_CAPABILITIES is sent and the CAPABILITIES is received.
  * @retval RETURN_DEVICE_ERROR          A device error occurs when communicates with the device.
  **/
-libspdm_return_t libspdm_get_capabilities(libspdm_context_t *spdm_context);
+libspdm_return_t libspdm_get_capabilities(libspdm_context_t *spdm_context,
+                                          bool get_supported_algorithms,
+                                          libspdm_responder_supported_algorithms_t *supported_algs);
```
jyao1, line 147: Need to use `void *` for supported_algs, because the structure may be changed in the future version.
Also add length for the structure.

include/library/spdm_requester_lib.h
```diff
@@ -491,6 +492,19 @@ libspdm_return_t libspdm_start_session(void *spdm_context, bool use_psk,
...
+ * If the Responder supports this extended capability, the Responder will include the Supported
+ * Algorithms Block in its CAPABILITIES response.
+ *
+ * @param spdm_context The SPDM context.
+ * @param responder_supported_algorithms The responder supported algorithms.
+ * @retval RETURN_SUCCESS               The connection is initialized successfully.
+ * @retval RETURN_DEVICE_ERROR          A device error occurs when communicates with the device.
+ **/
+libspdm_return_t libspdm_get_supported_algorithms(
+    void *spdm_context, libspdm_responder_supported_algorithms_t *responder_supported_algorithms);
```
jyao1, line 506: Need to use `void *` for responder_supported_algorithms, because the structure may be changed in the future version.
Also add length for the function.

include/internal/libspdm_common_lib.h
steven-bellock: Integrator is free to ignore it if they want, but the library should not hide that information from them.

library/spdm_responder_lib/libspdm_rsp_capabilities.c
ShitalJumbad: Added above.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -211,7 +211,9 @@ static bool validate_responder_capability(uint32_t capabilities_flag, uint8_t ve
  * @retval LIBSPDM_STATUS_BUFFER_FULL
  *         The buffer used to store transcripts is exhausted.
  **/
-static libspdm_return_t libspdm_try_get_capabilities(libspdm_context_t *spdm_context)
+static libspdm_return_t libspdm_try_get_capabilities(libspdm_context_t *spdm_context,
+                                                     bool get_supported_algorithms,
+                                                     spdm_responder_supported_algorithms_t *supported_algs)
```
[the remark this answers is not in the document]
ShitalJumbad: Done.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
ShitalJumbad: corrected.

include/library/spdm_requester_lib.h
ShitalJumbad: new definition of libspdm_responder_supported_algorithms_t provides all the space needed

library/spdm_requester_lib/libspdm_req_get_capabilities.c
ShitalJumbad: spdm_supported_algorithms_block_t  should be part of the definition for specificati1.3 versions as per the specifications?

include/industry_standard/spdm.h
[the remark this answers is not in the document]
ShitalJumbad: Done.

include/industry_standard/spdm.h
ShitalJumbad: Added struct_table_count instead of param1 to make it more intuitive to the integrator.

ShitalJumbad: param2 is probably not necessary to add?

include/internal/libspdm_common_lib.h
ShitalJumbad: I agree.

include/internal/libspdm_common_lib.h
ShitalJumbad: Could you please help me understand the motivation for this and adding version to the name?

library/spdm_responder_lib/libspdm_rsp_capabilities.c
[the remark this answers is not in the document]
ShitalJumbad: fixed.

include/internal/libspdm_common_lib.h
jyao1: See https://github.com/DMTF/libspdm/pull/2968#discussion_r2015409783

include/internal/libspdm_common_lib.h
jyao1: See https://github.com/DMTF/libspdm/pull/2968#discussion_r2015409783

include/internal/libspdm_requester_lib.h
jyao1: See https://github.com/DMTF/libspdm/pull/2968#discussion_r2015409783

include/library/spdm_requester_lib.h
jyao1: See https://github.com/DMTF/libspdm/pull/2968#discussion_r2015409783

library/spdm_requester_lib/libspdm_req_get_capabilities.c
steven-bellock: `spdm_version` should be an output parameter from `libspdm_get_supported_algorithms`.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
jyao1: The data structure definition should consider optional or variable length structure, and might not include them.

I dont see the need to include `spdm_supported_algorithms_block_t` in this case, because 1) it is just a special usage, 2) it has variable length.

include/industry_standard/spdm.h
```diff
@@ -177,6 +210,8 @@ typedef struct {
     /* Below field is added in 1.2.*/
     uint32_t data_transfer_size;
     uint32_t max_spdm_msg_size;
+    /* Below field is added in 1.3.*/
+    spdm_supported_algorithms_block_t supported_algorithms;
```
jyao1, line 214: I gave feedback before to remove `spdm_supported_algorithms_block_t` from `spdm_capabilities_response_t`.

You can comment this line here.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -384,6 +425,65 @@ static libspdm_return_t libspdm_try_get_capabilities(libspdm_context_t *spdm_con
...
+        (spdm_request->header.param1 & 0x01)) {
+
+        spdm_supported_algorithms_block_t *src = &spdm_response->supported_algorithms;
+        /* TODO: When SPDM 1.4 is supported, add version check and handling:
+         * if (spdm_response->header.spdm_version >= SPDM_MESSAGE_VERSION_14) {
+         *     // Handle SPDM 1.4 specific algorithm structure
+         *     libspdm_responder_supported_algorithms_14_t *dst = (libspdm_responder_supported_algorithms_14_t *)supported_algs;
+         *     // Copy SPDM 1.4 specific algorithm data
+         * } else
+         */
```
jyao1: please remove the comment here.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -384,6 +425,65 @@ static libspdm_return_t libspdm_try_get_capabilities(libspdm_context_t *spdm_con
...
+            size_t copy_size = src->ext_hash_count * sizeof(spdm_extended_algorithm_t);
+            libspdm_copy_mem(dst->ext_hash,
+                             sizeof(dst->ext_hash),
+                             src_bytes + offset,
+                             copy_size);
+            offset += copy_size;
+        }
+
+        alg_size = sizeof(spdm_negotiate_algorithms_common_struct_table_t);
+        for (size_t i = 0; i < src->param1; i++) {
```
jyao1: please always define variable in beginning of a function.

include/internal/libspdm_requester_lib.h
```diff
@@ -141,7 +168,10 @@ libspdm_return_t libspdm_get_version(libspdm_context_t *spdm_context,
  * @retval RETURN_SUCCESS               The GET_CAPABILITIES is sent and the CAPABILITIES is received.
  * @retval RETURN_DEVICE_ERROR          A device error occurs when communicates with the device.
  **/
-libspdm_return_t libspdm_get_capabilities(libspdm_context_t *spdm_context);
+libspdm_return_t libspdm_get_capabilities(libspdm_context_t *spdm_context,
+                                          bool get_supported_algorithms,
+                                          size_t *supported_algs_length,
+                                          void *supported_algs);
```
jyao1: I am not sure if it is the best idea to update directly. It involved lots of unnecessary change.

We can keep the original `libspdm_get_capabilities()` and add `libspdm_get_capabilities_with_supported_algs()`

`libspdm_get_capabilities()` can call `libspdm_get_capabilities_with_supported_algs()` internally.

include/internal/libspdm_requester_lib.h
```diff
@@ -129,10 +129,37 @@ libspdm_return_t libspdm_get_version(libspdm_context_t *spdm_context,
                                      uint8_t *version_number_entry_count,
                                      spdm_version_number_t *version_number_entry);
 
+#define LIBSPDM_MAX_EXT_ALG_COUNT 5
```
jyao1, line 132: libspdm does not support ext_alg. I think we can ignore them totally.

include/internal/libspdm_requester_lib.h
```diff
@@ -129,10 +129,37 @@ libspdm_return_t libspdm_get_version(libspdm_context_t *spdm_context,
...
+
+#pragma pack(1)
+typedef struct {
+    spdm_negotiate_algorithms_common_struct_table_t alg_struct;
+    spdm_extended_algorithm_t alg_external[LIBSPDM_MAX_EXT_ALG_COUNT];
+} libspdm_supported_algorithms_alg_struct_t;
+#pragma pack()
+
+typedef struct {
+    uint8_t measurement_specification;
```
jyao1, line 143: I still prefer to add `uint8_t spdm_version` to be the first field of `libspdm_responder_supported_algorithms_13_t` to make the data structure self-descriptive.

include/internal/libspdm_common_lib.h
```diff
@@ -68,6 +68,8 @@ typedef struct {
     uint16_t key_schedule;
 } libspdm_device_algorithm_t;
 
+
+
```
jyao1: why add 2 blank line here? Typo?

include/library/spdm_requester_lib.h
```diff
@@ -525,6 +525,33 @@ libspdm_return_t libspdm_start_session(void *spdm_context, bool use_psk,
...
+ * @retval RETURN_UNSUPPORTED                    The operation is not supported by the device.
+ * @retval RETURN_SECURITY_VIOLATION             Any verification fails.
+ *
+ * @note   The buffer must be at least sizeof(libspdm_responder_supported_algorithms_13_t) bytes.
+ *         The function will assert if responder_supported_algorithms_buffer is NULL.
+ */
+libspdm_return_t libspdm_get_supported_algorithms(void *spdm_context,
+                                                  size_t *responder_supported_algorithms_length,
+                                                  void *responder_supported_algorithms_buffer,
+                                                  uint8_t *spdm_version);
```
jyao1, line 553: If you include `spdm_version` to `libspdm_responder_supported_algorithms_13_t`, then you can remove `spdm_version` parameter here.

include/internal/libspdm_requester_lib.h
steven-bellock: See https://github.com/DMTF/libspdm/pull/2968#discussion_r2035798604.

include/internal/libspdm_requester_lib.h
steven-bellock: That is fine.

include/library/spdm_requester_lib.h
steven-bellock: >then you can remove spdm_version parameter here.

The Integrator still needs to know what `libspdm_responder_supported_algorithms_*_t` to apply to `responder_supported_algorithms_buffer`. Without this parameter they would need to `libspdm_get_data` `LIBSPDM_DATA_SPDM_VERSION`, which is not the end of the world, but this makes it more convenient.

include/library/spdm_requester_lib.h
```diff
@@ -12,6 +12,7 @@ extern "C" {
 #endif
 
 #include "library/spdm_common_lib.h"
+#include "internal/libspdm_common_lib.h"
```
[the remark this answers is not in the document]
ShitalJumbad: Removed.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
ShitalJumbad: Removed.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
ShitalJumbad: Thanks, changed it to the beginning of the code block.

include/internal/libspdm_requester_lib.h
ShitalJumbad: Great suggestion! I’ve made the change as per your feedback..

include/internal/libspdm_requester_lib.h
ShitalJumbad: Updated.

include/internal/libspdm_common_lib.h
ShitalJumbad: good catch, removed blank lines.

include/library/spdm_requester_lib.h
ShitalJumbad: Thanks for the suggestion! I think having the spdm_version parameter improves the convenience and usability of the API. Let me know if you have a strong preference here.

include/library/spdm_requester_lib.h
ShitalJumbad: updated, thanks!

include/internal/libspdm_requester_lib.h
ShitalJumbad: Updated.

include/internal/libspdm_common_lib.h
ShitalJumbad: Added version to the name.

include/internal/libspdm_common_lib.h
ShitalJumbad: Updated, Thanks!

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -315,22 +326,60 @@ static libspdm_return_t libspdm_try_get_capabilities(libspdm_context_t *spdm_con
...
+            SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_MUT_AUTH_CAP) {
+            index++;
+        }
+        if ((spdm_context->connection_info.capability.flags &
+             SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_KEY_UPD_CAP) ||
+            (spdm_context->connection_info.capability.flags &
+             SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_PSK_CAP)) {
+            index++;
+        }
+        spdm_response_size = sizeof(spdm_capabilities_response_t)+
```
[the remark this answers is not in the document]
ShitalJumbad: Incorporated.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -382,6 +423,61 @@ static libspdm_return_t libspdm_try_get_capabilities(libspdm_context_t *spdm_con
         spdm_context->connection_info.capability.max_spdm_msg_size = 0;
     }
 
+    /* Copy algorithms if requested and received */
+    if (get_supported_algorithms && supported_algs != NULL &&
+        spdm_response->header.spdm_version >= SPDM_MESSAGE_VERSION_13 &&
```
ShitalJumbad, line 428: Instead of using >= SPDM_MESSAGE_VERSION_13 here, should I use == SPDM_MESSAGE_VERSION_13 since I am typecasting to libspdm_responder_supported_algorithms_13_t on line 436?

include/industry_standard/spdm.h
ShitalJumbad: Thank you for your feedback and for referencing the earlier discussion.t - https://github.com/DMTF/libspdm/pull/2968/commits/cbda82ce03f022aa29609858b8002d220c75113b#r2151455599 

To summarize, I understand the concern about including spdm_supported_algorithms_block_t in spdm_capabilities_response_t, especially since it’s a variable-length structure and may be considered a special case.

However, I added it here because, based on my reading of the SPDM 1.3 specification, it seems that spdm_supported_algorithms_block_t was introduced specifically for version 1.3, and including it in the structure aligns with the spec for that version.  That said, I may be wrong, so I’m just trying to understand how we make this decision.

I’m also looping in @steven-bellock to get his perspective.

include/internal/libspdm_requester_lib.h
jyao1: No, I don't suggest to keep the complexity.

As we already said, libspdm does not support ext_alg.
What is the value to return ext_alg to the integrator?
The integrator cannot use ext_algo with libspdm anyway.

include/industry_standard/spdm.h
jyao1: No, I don't there is any requirement to add anything defined in the spec.
If a structure is variable size, then it is OK to comment it here.

I prefer to not include it for convenience.

include/library/spdm_requester_lib.h
jyao1: If we can guarantee that `spdm_version` is already the first byte of `responder_supported_algorithms_buffer`, then the integrator can just check the first byte.

Or we can define following `libspdm_responder_supported_algorithms_common_header` structure to ensure that every `libspdm_responder_supported_algorithms_13_t` or `libspdm_responder_supported_algorithms_14_t` will use the same common header.

```
typedef struct {
    uint8_t spdm_version;
    /* the rest of the field is spdm_version specific */
} libspdm_responder_supported_algorithms_common_header;
```

I really cannot see the need to have `uint8_t *spdm_version`

library/spdm_requester_lib/libspdm_req_get_capabilities.c
jyao1: good question!

I think it is OK to use `>=` now.
If change is needed for new version, we need to update the code anyway.

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -382,6 +423,61 @@ static libspdm_return_t libspdm_try_get_capabilities(libspdm_context_t *spdm_con
...
+        size_t i;
+        spdm_supported_algorithms_block_t *src = &spdm_response->supported_algorithms;
+
+        libspdm_responder_supported_algorithms_13_t *dst =
+            (libspdm_responder_supported_algorithms_13_t *)supported_algs;
+        uint8_t *src_bytes = (uint8_t *)src;
+
+        *supported_algs_length = sizeof(libspdm_responder_supported_algorithms_13_t);
+        dst->spdm_version = spdm_response->header.spdm_version;
+        dst->struct_table_count = src->param1;
```
jyao1, line 442: I think we need to check if this `param1` exceeds the allowed `struct_table_count`.
Otherwise, it may cause buffer overflow.

PR #2971, opened 2025-01-29
CMakeLists.txt: Support x86_64 ARCH

alistair23:
Yocto/OpenEmbedded usses "x86_64" for the 64-bit x86 architecture, which currently isn't supported by libspdm. As it's a common name for the architecture let's add support for "x86_64" as the ARCH.


steven-bellock:
@alistair23 this is so that in Yocto/OpenEmbedded it can pass its "ARCH" field to libspdm without having to transform it?

steven-bellock (approved):

Issue #3003, opened 2025-03-10
Encapsulated key update needs to know session id


steven-bellock:
From https://github.com/DMTF/libspdm/issues/2568#issuecomment-2711996273.

`libspdm_init_key_update_encap_state` is used by the Integrator to inform libspdm to perform an encapsulated `KEY_UPDATE`. However `KEY_UPDATE` is sent within a session and the session ID is not specified. The `session_info` should have its own encapsulated context.

steven-bellock:
I will file a separate issue for parallel key exchange sessions.

steven-bellock:
This also gates encapsulated `SEND_EVENT` and `GET_ENDPOINT_INFO`.

PR #3004, opened 2025-03-11
Refactor Responder's key exchange code

steven-bellock:
Refactor the code so that it is easier to read. There are no functional changes.


steven-bellock:
This commit helps pave the way to resolve #3003.

PR #3025, opened 2025-04-01
CMakeLists.txt: Support ppc64le ARCH

manojkiraneda:
libspdm currently fails to build on a ppc64le machine, add ppc architecture in the supported list.

Tested by :
cmake -DARCH=ppc64le -DTOOLCHAIN=GCC -DBUILD_LINUX_SHARED_LIB=ON -DTARGET=Debug -DCRYPTO=mbedtls .. make -j$(nproc)


steven-bellock:
Not sure what other CMake projects do, but if the compilation is native (HOST == TARGET) then `-DARCH` should not be needed at all. Maybe that can be a 4.0 "feature".

@manojkiraneda what type of system did you test this out on?

steven-bellock (approved):

manojkiraneda:
> Not sure what other CMake projects do, but if the compilation is native (HOST == TARGET) then `-DARCH` should not be needed at all. Maybe that can be a 4.0 "feature".
> 
Yeah, the majority of the projects don't really need ARCH variable, we need to be a little explicit about it in the documentation as well.

And even for cross compilation, I think what cmake suggests is to have a separate CMake(tool chain) file that describes the target and use that to compile (https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html#toolchain-files)

> @manojkiraneda what type of system did you test this out on?

IBM Power servers are all of ppc64 architecture & we use them for all development activities here. I have tested it on one of those machines to build libspdm natively & with this patch, the repository builds without any issues.

PR #3030, opened 2025-04-07
Add unit test coverage for chunk sequence no wrapped in chunk get.

### LINKED ISSUE #3023: Coverity 479566 - logically dead code
In the below code, Coverity is flagging
```
if (chunk_seq_no == 0 && large_response_size_so_far != 0)
```
as dead code that is always false. I believe that is because Coverity doesn't think that `chunk_seq_no` can wrap around to `0`. I am pretty sure we can waive this since it is a false positive, however [code coverage](https://dmtf.github.io/libspdm/coverage_log/library/spdm_requester_lib/libspdm_req_handle_error_response.c.gcov.html) is showing that this condition is not being hit. @Li-Aaron is one of your unit tests supposed to trigger this error?

Image

https://github.com/DMTF/libspdm/blob/09d535bc29adcca782b6ad1a25be659be7774d4b/library/spdm_requester_lib/libspdm_req_handle_error_response.c#L272-L292

Li-Aaron:
Fix #3023

### DIFF
--- a/unit_test/test_spdm_requester/chunk_get.c
+++ b/unit_test/test_spdm_requester/chunk_get.c
@@ -263,7 +263,7 @@ void libspdm_requester_chunk_get_test_case4_build_digest_response(
     spdm_response->header.param2 |= (0xFF << 0);
 }
 
-void libspdm_requester_chunk_get_test_case5_build_vendor_response(
+void libspdm_requester_chunk_get_test_case5_case6_build_vendor_response(
     void* context, void* response, size_t* response_size)
 {
     spdm_vendor_defined_response_msg_t *spdm_response;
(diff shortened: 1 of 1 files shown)


unit_test/test_spdm_requester/chunk_get.c
```diff
@@ -917,6 +931,59 @@ static void libspdm_test_requester_chunk_get_case5(void **state)
...
+
+    libspdm_set_mem(data, sizeof(data), 0xAA);
+
+    status = libspdm_vendor_send_request_receive_response(spdm_context, NULL,
+                                                          standard_id, vendor_id_len, vendor_id,
+                                                          data_len, data,
+                                                          &standard_id, &vendor_id_len, vendor_id,
+                                                          &data_len, data);
+
+    assert_int_equal(status, LIBSPDM_STATUS_RECEIVE_FAIL);
```
steven-bellock, line 984: The status returned to the Integrator is inconsistent between messages, but that is another issue.

steven-bellock:
Yes, all those `libspdm_receive_spdm_response`s do slow things down when printing to the terminal. However that's only in Debug mode, so it isn't too horrible.

Issue #3026, opened 2025-04-02
switch to openssl 3.5 which support PQC.


jyao1:
See https://openssl-library.org/post/2025-02-04-release-announcement-3.5/

PR #3033, opened 2025-04-12
Backport API introduced in openssl 3.5.0

kraj:
This commit [1] introduced this new API which is used by pem.h from openssl and it should be provided by asn1.h which is vendored here so it goes out of sync and causes build errors e.g

In file included from ./os_stub/cryptlib_openssl/pk/x509.c:18: /usr/include/openssl/pem.h:399:28: error: unknown type name 'OSSL_i2d_of_void_ctx'
  399 | int PEM_ASN1_write_bio_ctx(OSSL_i2d_of_void_ctx *i2d, void *vctx,

[1] https://github.com/openssl/openssl/commit/35f6e7ea02b599d5aaf220b4720cbadd946d8023

### DIFF
--- a/os_stub/openssllib/openssl_gen/openssl/asn1.h
+++ b/os_stub/openssllib/openssl_gen/openssl/asn1.h
@@ -368,6 +368,7 @@ typedef struct ASN1_VALUE_st ASN1_VALUE;
 
 typedef void *d2i_of_void(void **, const unsigned char **, long);
 typedef int i2d_of_void(const void *, unsigned char **);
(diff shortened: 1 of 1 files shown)


alistair23 (approved):

alistair23:
In the long run it's probably worth removing the vendored in openssl in `os_stub/openssllib/openssl_gen`, but this at least fixes the build issue now

jyao1 (approved):

jyao1:
@kraj , is that the only change required for openssl 3.5.0 ?

kraj:
> @kraj , is that the only change required for openssl 3.5.0 ? See #3026

for getting. build to work, thats the only one.

PR #3057, opened 2025-05-02
Handle "error" if no encapsulated request

### LINKED ISSUE #3031: Requester does not handle "encapsulated" errors from Responder
As detailed in https://github.com/DMTF/SPDM-WG/issues/3992, if the Requester issues `GET_ENCAPSULATED_REQUEST` and the Responder does not have any encapsulated requests, the Responder can reply with `UnexpectedRequest` or `NoPendingRequests`. However the libspdm Requester does not accommodate this, and expects only an `ENCAPSULATED_REQUEST` from the Responder.

steven-bellock:
Fix #3031.

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_encap_request.c
+++ b/library/spdm_requester_lib/libspdm_req_encap_request.c
@@ -235,13 +235,39 @@ libspdm_return_t libspdm_encapsulated_request(libspdm_context_t *spdm_context,
             libspdm_release_receiver_buffer (spdm_context);
             return status;
         }
-        if (libspdm_encapsulated_request_response->header.request_response_code !=
-            SPDM_ENCAPSULATED_REQUEST) {
+        if (libspdm_encapsulated_request_response->header.spdm_version !=
+            spdm_get_encapsulated_request_request->header.spdm_version) {
             libspdm_release_receiver_buffer (spdm_context);
             return LIBSPDM_STATUS_INVALID_MSG_FIELD;
         }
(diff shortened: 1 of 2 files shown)


library/spdm_requester_lib/libspdm_req_encap_request.c
```diff
@@ -235,13 +235,39 @@ libspdm_return_t libspdm_encapsulated_request(libspdm_context_t *spdm_context,
...
+            const uint8_t error_code = libspdm_encapsulated_request_response->header.param1;
+
+            libspdm_release_receiver_buffer(spdm_context);
+
+            if (spdm_response_size < sizeof(spdm_error_response_t)) {
+                libspdm_release_receiver_buffer (spdm_context);
+                return LIBSPDM_STATUS_INVALID_MSG_SIZE;
+            }
+
+            if (mut_auth_requested == 0) {
```
jyao1, line 253: Is that enough?

Should we also consider basic_mut_auth here?

steven-bellock: `mut_auth_requested == 0` is also used for the basic mutual authentication flow. In particular only digests, certificates, and challenge should be allowed in the basic mutual authentication flow, but it does not look like libspdm explicitly enforces that. I will file a separate issue for that.

steven-bellock:
https://github.com/DMTF/libspdm/issues/3058 should be resolved after this is merged to handle the basic mutual authentication cases.

jyao1 (approved):

PR #3073, opened 2025-05-16
Use const-correctness in libspdm_set_data function

### LINKED ISSUE #3071: Use const-correctness for `libspdm_set_data` function `data` parameter
The function does not appear to write to the "data" argument and there are no other compiler errors from setting the arg to const.

In one of our usecases, the data is coming from .rodata, so we would like to keep this const.

rjoleary:
The function does not appear to write to the "data" argument and there are no other compiler errors from setting the arg to const.

In one of our usecases, the data is coming from .rodata, so we would like to keep this const.

Fixes #3071


steven-bellock (approved):

jyao1:
Have you tried spdm-emu?
Is there any change required?

steven-bellock:
libspdm unit tests have lots of calls to `set_data` already, so it should be fine.

PR #3085, opened 2025-06-10
Add mandatory mutual authentication support in responder.

### LINKED ISSUE #2121: Shall responder skip the mutual authentication if requester does not have the capability?
In the implementation of libspdm, responder sets the value of `mut_auth_requested` of the key exchange response according to the capability of requester and responder:
https://github.com/DMTF/libspdm/blob/04baaa92bc59e03aa528988edd46bfc15ce01e2b/library/spdm_responder_lib/libspdm_rsp_key_exchange.c#L386-L399
If responder or requester does not have such capability, responder will skip the mutual authentication and run the rest of the process and finally establish the secure session, which may not be safe.

I tried the spdm-emu to run a responder with argument `--mut_auth DIGESTS` and a requester without `MUT_AUTH` capability using below commands. The result shows they skip the mutual authentication and successfully running to the end.
```
./spdm_responder_emu --trans PCI_DOE --mut_auth DIGESTS --req_asym ECDSA_P384 --basic_mut_auth NO

./spdm_requester_emu --trans PCI_DOE --req_asym ECDSA_P384 --exe_conn DIGEST,CERT,CHAL --exe_session KEY_EX,PSK,KEY_UPDATE,HEARTBEAT,DIGEST,CERT --cap CERT,CHAL,ENCRYPT,MAC,KEY_EX,PSK,HBEAT,KEY_UPD,HANDSHAKE_IN_CLEAR
```

My thought is that responder should end the conversation and return an error message to requester if responder **wants to verify the requester** but requester does not support the `mut_auth` capability.

Li-Aaron:
Fix #2121


jyao1:
why we need to opaque?

I think a SetData for Responder should be enough,

library/spdm_responder_lib/libspdm_rsp_key_exchange.c
```diff
@@ -438,7 +438,12 @@ libspdm_return_t libspdm_get_response_key_exchange(libspdm_context_t *spdm_conte
             SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_MUT_AUTH_CAP)) {
         spdm_response->mut_auth_requested =
             spdm_context->local_context.mut_auth_requested;
+    } else if (spdm_context->local_context.mandatory_mut_auth) {
```
steven-bellock, line 441: SPDM 1.3 introduced the "InvalidPolicy" error code which is more fitting.

Li-Aaron: Modified.
Will use "InvalidPolicy" when version >= 1.3

steven-bellock: Can also add an assertion that `SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_MUT_AUTH_CAP` is set.

PR #3086, opened 2025-06-11
scripts: format_nix: allow newer versions of uncrustify

twilfredo:
Use `sort -v` to sort versions and allow versions greater than 0.78. For example, Fedora 42 ships 0.80, which fails to work since this script currently checks exactly for 0.78.

### DIFF
--- a/script/format_nix.sh
+++ b/script/format_nix.sh
@@ -17,8 +17,8 @@ fi
 VERSION=$(uncrustify --version 2>/dev/null | sed -E 's/Uncrustify-([0-9]+\.[0-9]+).*/\1/')
 REQUIRED_VERSION="0.78"
 
-if [ "$VERSION" != "$REQUIRED_VERSION" ]; then
-    echo "ERROR: uncrustify version $REQUIRED_VERSION is required, but found $VERSION."
+if [ "$(printf '%s\n' "$REQUIRED_VERSION" "$VERSION" | sort -V | head -n1)" != "$REQUIRED_VERSION" ]; then
+    echo "ERROR: uncrustify version $REQUIRED_VERSION or higher is required, but found $VERSION."
     exit 1
 fi
 


script/format_nix.sh
```diff
@@ -17,8 +17,8 @@ fi
 VERSION=$(uncrustify --version 2>/dev/null | sed -E 's/Uncrustify-([0-9]+\.[0-9]+).*/\1/')
 REQUIRED_VERSION="0.78"
 
-if [ "$VERSION" != "$REQUIRED_VERSION" ]; then
-    echo "ERROR: uncrustify version $REQUIRED_VERSION is required, but found $VERSION."
+if [ "$(printf '%s\n' "$REQUIRED_VERSION" "$VERSION" | sort -V | head -n1)" != "$REQUIRED_VERSION" ]; then
+    echo "ERROR: uncrustify version $REQUIRED_VERSION or higher is required, but found $VERSION."
```
manojkiraneda, line 21: The CI pipeline runs within an Ubuntu container. However, even when developers use a newer version of uncrustify (e.g., 0.80) locally, the CI continues to run with version 0.78. This version mismatch can lead to confusion, especially when formatting differences cause CI failures. Hence, I feel it is better to have the stricter version check for making it easy to figure out what is needed for CI to pass.

Originally, the strict version check was introduced because we had encountered issues in the past where newer versions of uncrustify deprecated certain configuration keys or altered the behavior of existing ones. These changes led to unintended formatting modifications and numerous unnecessary code diffs. Additionally, Ubuntu tends to lag behind Fedora in adopting the latest uncrustify releases, further contributing to inconsistencies across development and CI environments.

steven-bellock: I agree with @manojkiraneda. At this point the style checker is mainly for CI/CD. In particular if you run it locally it will change a lot of files. Maybe the file should be moved to https://github.com/DMTF/libspdm/tree/main/.github to show that it should not be run locally.

PR #3087, opened 2025-06-12
Add TCP discovery support

manojkiraneda:
This PR would add the necessary support needed to enable encoding & decoding of TCP discovery commands.


library/spdm_transport_tcp_lib/libspdm_tcp_tcp.c
```diff
@@ -165,6 +165,93 @@ libspdm_return_t libspdm_tcp_decode_message(uint32_t **session_id,
...
+ * @param[in]  transport_message_size   Size of the incoming buffer.
+ * @param[out] message_type             On success, receives the validated message type.
+ *
+ * @retval LIBSPDM_STATUS_SUCCESS             Decoding successful.
+ * @retval LIBSPDM_STATUS_INVALID_MSG_SIZE    Buffer too small for TCP header.
+ * @retval LIBSPDM_STATUS_UNSUPPORTED_CAP     Unsupported binding version.
+ * @retval LIBSPDM_STATUS_INVALID_MSG_FIELD   Payload length is non-zero or message type invalid.
+ **/
+libspdm_return_t libspdm_tcp_decode_discovery_message(const void *transport_message,
+                                                      size_t transport_message_size,
```
jyao1: Can we put `transport_message_size` to be the first parameter, and `transport_message` to be the second?

I recommend to make order consistent with other APIs.

jyao1:
We have similar API for PCI_DOE (https://github.com/DMTF/libspdm/blob/main/library/spdm_transport_pcidoe_lib/libspdm_doe_pcidoe.c) They are used by spdm_emu (https://github.com/DMTF/spdm-emu/blob/main/library/pci_doe_responder_lib/pci_doe_dispatch.c) 

Please try to enable spdm_emu to use those API to ensure they are useful.

library/spdm_transport_tcp_lib/libspdm_tcp_tcp.c
manojkiraneda: done

PR #3091, opened 2025-06-18
Check certificate size before continuing

### LINKED ISSUE #2327: Check certificate size before continuing
Follow-up to #2304.

After the first `CERTIFICATE` response the total size of the certificate chain is known to the Requester. https://github.com/DMTF/libspdm/blob/567b1c8ea731fe42650d43ede50a105b772dc7aa/library/spdm_requester_lib/libspdm_req_get_certificate.c#L255-L257 However the Requester will continue to fetch portions of the certificate chain even if it is too large for its internal buffer(s). The Requester should error out immediately if the certificate chain is too large.

Li-Aaron:
Fix #2327


unit_test/test_spdm_requester/get_certificate.c
```diff
@@ -4342,6 +4408,85 @@ void libspdm_test_requester_get_certificate_case30(void **state)
     free(m_libspdm_local_certificate_chain);
 }
 
+/**
+ * Test 2: Normal case, request a certificate chain
```
steven-bellock: Description needs to be updated.

steven-bellock (changes requested):

unit_test/test_spdm_requester/get_certificate.c
Li-Aaron: Updated

PR #3106, opened 2025-07-02
Requester should handler error before version check.

### LINKED ISSUE #3022: `RequestResync` Error Code Handling
When checking the SPDM functionality using storage controller, synchronization is not reestablished if the responder experiences a reset and loses the last negotiated state.
The sequence of events is as follows:

1. The requester and responder successfully negotiate the SPDM version to 1.1.
2. A controller reset takes place, and the responder's last negotiated state is lost.
3. Despite the reset, the requester continues to believe that the SPDM version is 1.1.
4. The requester sends a GET_DIGEST command with a version of 1.1.
5. Since the responder has not been in a negotiated state following the reset, it responds with an error message indicating a request for resynchronization (0x43), but with an SPDM version of 1.0.
  `spdm_response->header.spdm_version = libspdm_get_connection_version (spdm_context);
   if (spdm_response->header.spdm_version == 0) {
        /* if version is not negotiated, then use default version 1.0 */
        spdm_response->header.spdm_version = SPDM_MESSAGE_VERSION_10;
    }`
7. On the requester's side, the version is validated before the error message is checked, and a version mismatch is identified, resulting in the return of an "InvalidMessage" value.
`/* -=[Validate Response Phase]=- */
    if (spdm_response_size < sizeof(spdm_message_header_t)) {
        status = LIBSPDM_STATUS_INVALID_MSG_SIZE;
        goto receive_done;
    }
    if (spdm_response->header.spdm_version != spdm_request->header.spdm_version) {
        goto receive_done;
    }
    if (spdm_response->header.request_response_code == SPDM_ERROR) {
        status = libspdm_handle_error_response_main(
            spdm_context, session_id,
            &spdm_response_size,
            (void **)&spdm_response, SPDM_GET_DIGESTS, SPDM_DIGESTS);
            goto receive_done;
        }
    }`

Li-Aaron:
Fix #3022

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_send_receive.c
+++ b/library/spdm_requester_lib/libspdm_req_send_receive.c
@@ -458,6 +458,14 @@ libspdm_return_t libspdm_handle_large_request(
             status = LIBSPDM_STATUS_INVALID_MSG_SIZE;
             break;
         }
+
+        if (spdm_response->header.request_response_code == SPDM_ERROR
+            && spdm_response->header.param1 != SPDM_ERROR_CODE_LARGE_RESPONSE) {
+            status = libspdm_handle_simple_error_response(spdm_context,
+                                                          spdm_response->header.param1);
+            break;
+        }
+
         if (spdm_response->header.spdm_version != libspdm_get_connection_version(spdm_context)) {
             status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
             break;
--- a/library/spdm_requester_lib/libspdm_req_challenge.c
+++ b/library/spdm_requester_lib/libspdm_req_challenge.c
@@ -177,10 +177,6 @@ static libspdm_return_t libspdm_try_challenge(libspdm_context_t *spdm_context,
         status = LIBSPDM_STATUS_INVALID_MSG_SIZE;
         goto receive_done;
     }
-    if (spdm_response->header.spdm_version != spdm_request->header.spdm_version) {
-        status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
-        goto receive_done;
-    }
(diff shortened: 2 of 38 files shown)


library/spdm_requester_lib/libspdm_req_send_receive.c
```diff
@@ -438,6 +438,14 @@ libspdm_return_t libspdm_handle_large_request(
             status = LIBSPDM_STATUS_INVALID_MSG_SIZE;
             break;
         }
+
+        if (spdm_response->header.request_response_code == SPDM_ERROR
+            && spdm_response->header.param1 == SPDM_ERROR_CODE_REQUEST_RESYNCH) {
```
Li-Aaron: Shall we handle all error code here for `SPDM_ERROR` response or just request resynch here?
The original check will return `LIBSPDM_STATUS_INVALID_MSG_FIELD` when not receive `CHUNK_SEND_ACK` response.
https://github.com/DMTF/libspdm/blob/2420fef507551555c7f3d6677e12dc7442a01dc6/library/spdm_requester_lib/libspdm_req_send_receive.c#L467-L470

jyao1: we had better handle all other param1 case before version check.

maybe change to ` != SPDM_ERROR_CODE_LARGE_RESPONSE`

jyao1 (approved):

steven-bellock:
I will be able to review this next week.

steven-bellock (approved):
I am somewhat confused on buffer management and
```
(spdm_response->header.spdm_version != spdm_request->header.spdm_version)
```
but that is a separate issue.

PR #3114, opened 2025-07-07
Increment `SPDM_MAX_VERSION_COUNT`

### LINKED ISSUE #3113: Increment `SPDM_MAX_VERSION_COUNT`
Now that SPDM 1.4 has been published, increment `SPDM_MAX_VERSION_COUNT` from `4` to `5`.

steven-bellock:
Fix #3113.

### DIFF
--- a/include/industry_standard/spdm.h
+++ b/include/industry_standard/spdm.h
@@ -13,8 +13,8 @@
 
 #pragma pack(1)
 
-/* 4 means SPDM spec 1.0, 1.1, 1.2, 1.3 */
-#define SPDM_MAX_VERSION_COUNT 4
+/* Covers SPDM specifications from 1.0 to 1.4. */
+#define SPDM_MAX_VERSION_COUNT 5
 #define SPDM_MAX_SLOT_COUNT 8
 #define SPDM_MAX_OPAQUE_DATA_SIZE 1024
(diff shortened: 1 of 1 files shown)


include/industry_standard/spdm.h
```diff
@@ -13,8 +13,8 @@
 
 #pragma pack(1)
 
-/* 4 means SPDM spec 1.0, 1.1, 1.2, 1.3 */
-#define SPDM_MAX_VERSION_COUNT 4
+/* Covers SPDM specifications from 1.0 to 1.4. */
+#define SPDM_MAX_VERSION_COUNT 5
```
jyao1, line 17: Wait.

That will cause uninitialized data in https://github.com/DMTF/libspdm/blob/main/library/spdm_common_lib/libspdm_com_context_data.c#L3281-L3289

This patch may be fine. But it is NOT enough.

jyao1: NOTE: We already have `#define LIBSPDM_MAX_VERSION_COUNT 5` at https://github.com/DMTF/libspdm/blob/main/include/library/spdm_lib_config.h#L168.

I think we can handle the peer with 1.4.

steven-bellock: Ok we can wait until we support 1.4.

PR #3121, opened 2025-07-11
Add Coverity to CI/CD

steven-bellock:
For now it must be manually triggered.


steven-bellock:
See https://github.com/steven-bellock/libspdm/actions/runs/16227358955/job/45822263857 for a successful run.

PR #3122, opened 2025-07-11
Add macros for key update and heartbeat code

### LINKED ISSUE #2977: `KEY_UPDATE` should be enclosed in `LIBSPDM_ENABLE_CAPABILITY_KEY_EX_CAP`
Most files related to `KEY_UPDATE` are unconditionally compiled. For example : https://github.com/DMTF/libspdm/blob/main/library/spdm_requester_lib/libspdm_req_key_update.c

steven-bellock:
Fix #2977.

### DIFF
--- a/include/library/spdm_requester_lib.h
+++ b/include/library/spdm_requester_lib.h
@@ -678,6 +678,7 @@ libspdm_return_t libspdm_send_receive_data(void *spdm_context,
                                            const void *request, size_t request_size,
                                            void *response, size_t *response_size);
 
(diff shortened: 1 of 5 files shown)


jyao1 (approved):

steven-bellock:
This may break `spdm-emu`. I will check.

steven-bellock:
It looks like https://github.com/DMTF/spdm-emu/blob/main/spdm_emu/spdm_requester_emu/spdm_requester_session.c has the correct macros, so this should not break `spdm-emu`.

PR #3124, opened 2025-07-14
1.4 PQC - Part 1: Add PQC crypto library

jyao1:
This patch set add ML-DSA/ML-KEM/SLH-DSA.

This is first patch set for SPDM 1.4 PQC support.

It is based on https://github.com/DMTF/libspdm/pull/3136.


unit_test/test_crypt/mldsa_verify.c
```diff
@@ -0,0 +1,95 @@
+/**
+ *  Copyright Notice:
+ *  Copyright 2021-2022 DMTF. All rights reserved.
```
steven-bellock: `2025`

unit_test/test_crypt/mlkem_verify.c
```diff
@@ -0,0 +1,105 @@
+/**
+ *  Copyright Notice:
+ *  Copyright 2021-2022 DMTF. All rights reserved.
```
steven-bellock: `2025`

unit_test/test_crypt/slhdsa_verify.c
```diff
@@ -0,0 +1,95 @@
+/**
+ *  Copyright Notice:
+ *  Copyright 2021-2022 DMTF. All rights reserved.
```
steven-bellock: `2025`

library/spdm_crypt_lib/libspdm_crypt_asym.c
```diff
@@ -102,7 +102,7 @@ static bool libspdm_sm2_dsa_sign_wrap (void *context, size_t hash_nid,
  * @param  is_requester                         indicate if the signing is from a requester
  * @param  context_size                         SPDM signing context size
  **/
-static const void *libspdm_get_signing_context_string (
+const void *libspdm_get_signing_context_string (
```
steven-bellock, line 105: Why do these need to be semi-public?

jyao1: It will be called by libspdm_crypt_pqc_asym.c

unit_test/test_crypt/mldsa_verify.c
jyao1: Fixed.

Not sure why it is NOT caught in CI.

unit_test/test_crypt/mlkem_verify.c
jyao1: Fixed.

unit_test/test_crypt/slhdsa_verify.c
jyao1: Fixed.

steven-bellock:
For example, https://datatracker.ietf.org/doc/draft-ietf-lamps-dilithium-certificates is still a draft and unratified.

steven-bellock (approved):

PR #3126, opened 2025-07-14
1.4 PQC - Part 2: Add PQC context

jyao1:
This patch set add PQC to SPDM HAL / Context / Device Sample

This is second patch set for SPDM 1.4 PQC support.

It is based on https://github.com/DMTF/libspdm/pull/3124


jyao1:
@steven-bellock, this patch is rebased. Would you please review?

include/hal/library/requester/reqasymsignlib.h
```diff
@@ -43,6 +43,36 @@ extern bool libspdm_requester_data_sign(
     const uint8_t *message, size_t message_size,
     uint8_t *signature, size_t *sig_size);
 
+/**
+ * Sign an SPDM message data.
+ *
+ * @param  spdm_context      A pointer to the SPDM context.
+ * @param  spdm_version      Indicates the negotiated s version.
```
steven-bellock: Typo `s`.

include/hal/library/responder/asymsignlib.h
```diff
@@ -74,4 +74,32 @@ extern bool libspdm_responder_data_sign(
     const uint8_t *message, size_t message_size,
     uint8_t *signature, size_t *sig_size);
 
+/**
+ * Sign an SPDM message data.
+ *
+ * @param  spdm_context    A pointer to the SPDM context.
+ * @param  spdm_version    Indicates the negotiated s version.
```
steven-bellock: Typo `s`.

include/hal/library/requester/reqasymsignlib.h
```diff
@@ -43,6 +43,36 @@ extern bool libspdm_requester_data_sign(
...
+ * @param  signature         A pointer to a destination buffer to store the signature.
+ * @param  sig_size          On input, indicates the size, in bytes, of the destination buffer to
+ *                           store the signature.
+ *                           On output, indicates the size, in bytes, of the signature in the
+ *                           buffer.
+ *
+ * @retval true  signing success.
+ * @retval false signing fail.
+ **/
+extern bool libspdm_requester_data_pqc_sign(
```
steven-bellock: This is a breaking change. If the intent is to allow a breaking change then `libspdm_requester_data_sign` can add `req_pqc_asym_alg` as a function parameter. This applies to all the other new functions in `include/hal/` as well.

include/internal/libspdm_secured_message_lib.h
```diff
@@ -15,6 +15,7 @@
 
 typedef struct {
     uint8_t dhe_secret[LIBSPDM_MAX_DHE_KEY_SIZE];
+    uint8_t kem_secret[LIBSPDM_MAX_KEM_SS_KEY_SIZE];
```
steven-bellock: Since `dhe_secret` and `kem_secret` are mutually exclusive, can just name it `shared_secret` or something, whose size is the larger of `LIBSPDM_MAX_DHE_KEY_SIZE` and `LIBSPDM_MAX_KEM_SS_KEY_SIZE`.

This is true for other code as well, like `libspdm_secured_message_import_kem_secret`.  Once the secret has been established, downstream users of the secret don't really care whether it was established via DHE or KEM.

include/hal/library/responder/key_pair_info.h
```diff
@@ -13,6 +13,11 @@
 
 #if LIBSPDM_ENABLE_CAPABILITY_GET_KEY_PAIR_INFO_CAP
 
+/**
+ * read the total key pairs number.
+ **/
+extern uint8_t libspdm_read_total_key_pairs();
```
steven-bellock: Why is this needed for PQC?

include/hal/library/requester/reqasymsignlib.h
jyao1: fixed.

include/hal/library/responder/asymsignlib.h
jyao1: fixed

include/hal/library/requester/reqasymsignlib.h
jyao1: That is fine. I can add `req_pqc_asym_alg` as parameter.

Since this is a breaking change, we have to add more commit for the consumer change.

include/hal/library/responder/key_pair_info.h
jyao1: I notice it is only used by unit_test. I can remove it.

But I feel it is burden to use set_data(LIBSPDM_DATA_TOTAL_KEY_PAIRS). It is an easier to just use libspdm_read_total_key_pairs(). I will use a separate issue to discuss it. (https://github.com/DMTF/libspdm/issues/3153)

include/internal/libspdm_secured_message_lib.h
jyao1: Changing `xxx_import_dhe_secret()` to `xxx_import_shared_secret()` is a big breaking change.
I am fine with it, if we decide to upgrade to 4.0.

include/hal/library/requester/reqasymsignlib.h
jyao1: done

include/hal/library/responder/key_pair_info.h
jyao1: done

include/internal/libspdm_secured_message_lib.h
jyao1: done

PR #3138, opened 2025-07-15
encap_cert test: fix context buffer usage.

jyao1:


steven-bellock:
@jyao1 how was this issue discovered?

jyao1:
> @jyao1 how was this issue discovered?

when I enable PQC.

steven-bellock (approved):

PR #3140, opened 2025-07-15
1.4 Add DMTF-DSP vendor ID definition.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/3134

### DIFF
--- a/library/spdm_common_lib/libspdm_com_support.c
+++ b/library/spdm_common_lib/libspdm_com_support.c
@@ -478,6 +478,8 @@ bool libspdm_validate_svh_vendor_id_len(uint8_t id, uint8_t vendor_id_len)
         return ((vendor_id_len == 0) || (vendor_id_len == 4));
     case SPDM_REGISTRY_ID_IANA_CBOR:
         return true;
+    case SPDM_REGISTRY_ID_DMTF_DSP:
+        return (vendor_id_len == 2);
     default:
         return false;
     }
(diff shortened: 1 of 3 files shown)


steven-bellock (commented):
https://github.com/DMTF/libspdm/blob/962fafe2972447e32817c10dfc21cc13c24e3d9b/library/spdm_common_lib/libspdm_com_support.c#L463 should be updated as well.

library/spdm_common_lib/libspdm_com_support.c
```diff
@@ -472,6 +472,7 @@ bool libspdm_validate_svh_vendor_id_len(uint8_t id, uint8_t vendor_id_len)
     case SPDM_REGISTRY_ID_MIPI:
     case SPDM_REGISTRY_ID_CXL:
     case SPDM_REGISTRY_ID_JEDEC:
+    case SPDM_REGISTRY_ID_DMTF_DSP:
```
steven-bellock: >When using this ID, the VendorID field shall always be present with a valid DSP value and the VendorIDLen **shall always be 2**.

jyao1: Please see #3142. We should remove all `(vendor_id_len == 0) ||`.
Why we allow them before?

steven-bellock: Regardless of the outcome of #3142, this pull request can use
```
return (vendor_id_len == 2);
```
for `SPDM_REGISTRY_ID_DMTF_DSP`, and, if needed, the other values can be fixed in another pull request.

jyao1: OK. Updated.

PR #3143, opened 2025-07-15
Fix copyright text

steven-bellock:
For some reason there was a line break.


unit_test/test_spdm_responder/encap_get_digests.c
```diff
@@ -1,9 +1,9 @@
 /**
  *  Copyright Notice:
  *  Copyright 2021-2022 DMTF. All rights reserved.
```
jyao1: fix copy right year?

Just curious why CI cannot catch this issue?

steven-bellock: >Just curious why CI cannot catch this issue?

It currently only checks in
- https://github.com/DMTF/libspdm/tree/main/include
- https://github.com/DMTF/libspdm/tree/main/library

but it can be expanded to `unit_test/test_spdm_*` directories.

jyao1 (approved):

unit_test/test_spdm_responder/encap_get_digests.c
jyao1: Got it.
We may consider extending to `os_stub` directory.

PR #3148, opened 2025-07-18
Autodetect ARCH if not set

manojkiraneda:
For native build, we should automatically detect the architecture rather than forcing users to provide that. ARCH is probably meant for cross compilation.


steven-bellock:
Maybe remove `arch` from
- https://github.com/DMTF/libspdm/blob/main/.github/workflows/code_coverage.yml
- https://github.com/DMTF/libspdm/blob/main/.github/workflows/codeql-analysis.yml
- https://github.com/DMTF/libspdm/blob/main/.github/workflows/coverity.yml

since those are always Linux / x64.

PR #3129, opened 2025-07-14
1.4 PQC - Part 4: Add large cert support

jyao1:
This patch set add PQC large cert support

This is 4th patch set for SPDM 1.4 PQC support.

It is based on https://github.com/DMTF/libspdm/pull/3128

Issue #2936, opened 2024-12-19
add new function for opaque data parsing.


jyao1:
libspdm_get_element_from_opaque_data() will get first (element_id, and sm_data_id) pair. Although the name is generic, but it has assumption that element_id is DMTF, because sm_data_id is only under DMTF.

I suggest to define two functions:
* `libspdm_get_element_from_opaque_data_with_element_id (element_id, element_index)` - that is to get element with ID and index to support multiple elements.
* `libspdm_get_sm_data_element_from_opaque_data (sm_data_id)` - that is to get SM_DATA only, with assumption that element_id is DMTF.

The first function can be used for SPDM 1.4 with DMTF-DSP, and support DSP0289 Authorization AODS.
* `libspdm_get_aods_element_from_opaque_data (aods_id)` - that is to get AODS with assumption that element_id is DMTF-DSP in the future.

Ref: https://github.com/DMTF/libspdm/pull/2935#discussion_r1891151023

PR #3150, opened 2025-07-18
1.4 Add opaque to FINISH/PSK_FINISH

jyao1:
Fix https://github.com/DMTF/libspdm/issues/3130

NOTE: This PR is based on https://github.com/DMTF/libspdm/pull/3129.

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_finish.c
+++ b/library/spdm_responder_lib/libspdm_rsp_finish.c
@@ -407,6 +407,9 @@ libspdm_return_t libspdm_get_response_finish(libspdm_context_t *spdm_context, si
     uint8_t th2_hash_data[LIBSPDM_MAX_HASH_SIZE];
     libspdm_return_t status;
     libspdm_session_state_t session_state;
+    uint8_t *ptr;
+    size_t opaque_data_entry_size;
+    size_t opaque_data_size;
 
     spdm_request = request;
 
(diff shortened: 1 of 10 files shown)


library/spdm_responder_lib/libspdm_rsp_finish.c
```diff
@@ -603,8 +674,15 @@ libspdm_return_t libspdm_get_response_finish(libspdm_context_t *spdm_context, si
     spdm_response->header.param1 = 0;
     spdm_response->header.param2 = 0;
 
+    ptr = (uint8_t *)spdm_response + sizeof(spdm_finish_response_t);
+    if (libspdm_get_connection_version(spdm_context) >= SPDM_MESSAGE_VERSION_14) {
```
steven-bellock, line 674: Why doesn't the Responder generate opaque data?

jyao1: No OpaqueData is required in FINISH by SPDM spec. It is same as today. No additional device specific OpaqueData is required in KEY_EXCHANGE.
This is for future Auth Spec.

The new OpaqueData handling will be addressed in https://github.com/DMTF/libspdm/issues/2936

jyao1:
> > NOTE: This PR is based on #3129.
> 
> Why does this need PQC? Opaque data should be completely independent of that.


This does not need PQC, but this does need SPDM 1.4.
Opaque is only present in 1.4+

steven-bellock (approved):

Issue #3108, opened 2025-07-03
Define a Standard (lib)SPDM Measurement Report


steven-bellock:
From https://github.com/DMTF/SPDM-WG/issues/4127.

1. Define a Standard SPDM Measurement Report that other specifications and Verifiers can easily reference and point to. In particular it would include only one `GET_MEASUREMENTS` request that requests all measurement indices.
2. Implement the `libspdm_get_std_measurement_report` function that provides a wrapper around `libspdm_get_measurement`.

xiaoyuruan:
Is Requester in Verifier’s TCB? i.e., does Verifier trust Requester?
 
- If yes, then why send full L1/L2 to Verifier. Requester should verify Responder’s signature and only send a set of “index + measurement” to Verifier.
     * At least no need to send VCA. May also trim off other fields of GET_MEASUREMENTS and MEASUREMENTS.
     * No need for the complication of two cases (get all, and, get one index after another)
- If no, then L1/L2 is not enough. You also need:
    * Responder’s cert chain sent to Verifier
    * Verifier must specify nonce in GET_MEASUREMENTS.

@steven-bellock @jyao1

steven-bellock:
>Verifier must specify nonce in GET_MEASUREMENTS.

This can either come from the Verifier itself or the Relying Party. In the TDISP case, the TVM (either kernel or user-mode) may generate the nonce, and send it to both to the TSM and the Verifier.

steven-bellock:
It looks like @wmaroneAMD and https://github.com/OpenPRoT want to use this, so let us revisit this at the next meeting. 

https://github.com/OpenPRoT/openprot/blob/3235e95d4fd4c90a8b3b8f840208b9f7dd000b49/docs/src/specification/services/attestation.md?plain=1#L371-L375

steven-bellock:
Looping in @mathieupoirier and @thomas-fossati as well since they are in this space with their https://github.com/rats-device-attestation/draft-poirier-rats-eat-da draft.

steven-bellock:
We have had plans for helper functions to generate and operate on SPDM data structures, including measurements. Those are obviously lower priority in favor of getting SPDM messages and features into libspdm first. And like I said in https://github.com/DMTF/libspdm/issues/3108#issue-3200746784 libspdm Requesters could call `libspdm_get_std_measurement_report` so that this is all automated.

PR #3155, opened 2025-07-25
Standard measurement format

jyao1:
This is based on https://github.com/steven-bellock/libspdm/tree/fix-3108, and add One-by-One Measurement Report format.

Ref https://github.com/DMTF/libspdm/issues/3108

### DIFF
--- a/doc/standard_measurement_report.md
+++ b/doc/standard_measurement_report.md
@@ -13,9 +13,12 @@ SPDM Requester and consumed by a Verifier while all agents are operating in prod
 particular, a production Verifier may only support a measurement report of this type and can point
 to this document to advertise that restriction to other agents.
 
-## Standard Measurement Report Definition
+There are 2 types of Standard Measurement Report. A verifier should support All-Measurements Report
+and may support One-by-One-Measurements Report.
 
-The standard measurement report is a byte buffer that consists of the L1/L2 transcript along with
+## Standard All-Measurements Report Definition
+
+The standard all-measurements report is a byte buffer that consists of the L1/L2 transcript along with
 the signature over the transcript if the Responder supports signing. It is comprised of a single
 `GET_MEASUREMENTS` request and a single `MEASUREMENTS` response.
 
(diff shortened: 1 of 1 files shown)


steven-bellock:
@jyao1 since this is not an official DMTF specification the document should not have normative language like "should", "shall", and "must"; it is a descriptive document. A Verifier's documentation can use normative language if it wants, to specify that this must be followed.

I might be able to discuss this with a wider audience at OCP Security on 29-July-2025. In particular to get feedback on where a document like this should reside, as well as the `One-by-One-Measurements Report` and buffer sizes.

jyao1:
> @jyao1 since this is not an official DMTF specification the document should not have normative language like "should", "shall", and "must"; it is a descriptive document. A Verifier's documentation can use normative language if it wants, to specify that this must be followed.
> 
> I might be able to discuss this with a wider audience at OCP Security on 29-July-2025. In particular to get feedback on where a document like this should reside, as well as the `One-by-One-Measurements Report` and buffer sizes.

I changed `shall` and `must` to `should`.
`should` is weak, it should be OK for a white paper document.

doc/standard_measurement_report.md
```diff
@@ -0,0 +1,150 @@
...
+The standard one-by-one-measurements report is a byte buffer that consists of the L1/L2 transcript along with
+the signature over the transcript if the Responder supports signing. It is comprised of multiple
+`GET_MEASUREMENTS` requests and multiple `MEASUREMENTS` responses.
+
+For SPDM 1.0 and 1.1, the byte buffer is {`GET_MEASUREMENTS`(0), `MEASUREMENTS`(0),
+`GET_MEASUREMENTS`(1), `MEASUREMENTS`(1), ..., `GET_MEASUREMENTS`(n), `MEASUREMENTS`(n)}.
+For SPDM 1.2 and later, the byte buffer is {`VCA`, `GET_MEASUREMENTS`(0), `MEASUREMENTS`(0),
+`GET_MEASUREMENTS`(1), `MEASUREMENTS`(1), ..., `GET_MEASUREMENTS`(n), `MEASUREMENTS`(n)}.
+
+The `GET_MEASUREMENTS`(0) request has the following properties:
```
steven-bellock, line 48: Why put this in the L1/L2 transcript? Presumably the Verifier knows exactly how many measurement indices a device reports.

jyao1: What do you mean "this"? Would you please be specific?

If you are asking why we need to send command to get Total Number (n), the answer is: The requester needs to know when to send `SignatureRequested` - that is after (n-1) successful response is received.

steven-bellock: >What do you mean "this"? Would you please be specific?

`GET_MEASUREMENTS`(0) and `MEASUREMENTS`(0).

>If you are asking why we need to send command to get Total Number (n),

No, the question is why is it present in the measurement report when the Verifier already knows how many measurement indices a device has? The Requester can use this information, but, barring a good reason to keep it, it probably can be omitted from the report.

jyao1: The logic is as follows:

- First, the requester must send it, in order to send the last `SignatureRequested`.
- Second, if the requester sends it, then it is in the L1/L2 transcript for signature calculation.
- Third, if it is in L1/L2 transcript for signature calculation, then it must be passed to verifier. As such, the verifier can reproduce the L1/L2 signature.

jyao1: I think it is good, because it passed what happened honestly.
For example, the verifier can check param to know what it is. The verifier can compare the total number to see if the total is expected, and fail immediately. Or if verifier does not care, then verifier can just skip.
What is the concern to add this message pair?

steven-bellock: >What is the concern to add this message pair?

There needs to be some rationale for delivering this information to the Verifier, else it should be omitted. In particular the Verifier can parse the messages without it, as it does with the "All-Measurements Report". 

If it is omitted from the Measurement Report then Requester would query the number of indices and have that signed. Once it has determined the number of indices, then it would begin a new L1/L2 transcript that then becomes the Measurement Report.

jyao1: It is omitted in "All-Measurement Report" because the requester does not send it.
It is totally different from the "One-by-one Measurement Report", where the requester does send it.

Omitting message means the verifier must reconstruct the `GetTotalNumber` to L1/L2, in order to reproduce the signature.
I don't think it is good idea to let verifier make assumption on how the requester sends message, and let verifier modify the transcript - that overloads too much on the verifier.

steven-bellock: Discussed this at the 3-Nov-2025 meeting. Will keep this in the report. @jyao1 will file an issue against the specification regarding `Content changed` and the `NumberOfBlocks` field.

jyao1: See https://github.com/DMTF/SPDM-WG/issues/4394

doc/standard_measurement_report.md
```diff
@@ -0,0 +1,150 @@
...
+to `GET_MEASUREMENTS`(n) is non-sequentially incremental.
+A device may implement non-sequentially increased measurement index.
+For example, a device has 3 measurement blocks. The index is 1, 4 and 6.
+Then the `Param2` of `GET_MEASUREMENTS`(1) is 1, the `Param2` of `GET_MEASUREMENTS`(2) is 4,
+and the `Param2` of `GET_MEASUREMENTS`(3) is 6.
+The requester may send a `GET_MEASUREMENTS` with `Param2` 2, but it will get `ERROR` response.
+As such, the `GET_MEASUREMENTS` with `Param2` 2 and `ERROR` response are NOT included
+in the measurement report.
+Once the successfully received number of measurement block is `n`-1, the requester should send
+the next `GET_MEASUREMENTS` with `SignatureRequested` set.
```
mcsmonk, line 120: https://github.com/DMTF/libspdm/pull/3341#pullrequestreview-3440404374
> Under the current logic, an invalid or missing index (e.g., gap in sequence) triggers an error, leading to re-initialization of the L1/L2 transcript — thereby breaking attestation.
> (DSP0274 describes that "An error response with any error code other than ResponseNotReady or LargeResponse
shall re-initialize L1/L2 to null".)

Does this cuase a problem in this process?

steven-bellock: Yes. Also need to check if the Responder-validator assumes this.

jyao1: Yes, for 1-by-1, we need two rounds.
Round 1 is to collect valid index.
Round 2 is to collect measurement based on the valid index.

PR #3165, opened 2025-08-14
1.4 add PQC_first policy.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/3158


jyao1:
@steven-bellock , could you please review this one?

PR #3175, opened 2025-08-22
1.4 PQC enlarge LIBSPDM_MAX_CERT_CHAIN_SIZE in spdm_lib_config

jyao1:


include/library/spdm_lib_config.h
```diff
@@ -431,6 +424,33 @@
 #define LIBSPDM_ML_KEM_1024_SUPPORT 1
 #endif
 
+/* This value specifies the maximum size, in bytes, of a certificate chain that can be stored in a
+ * libspdm context.
+ */
```
steven-bellock, line 429: Can be conditional on `LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT == 1`.

jyao1: Maybe. But it is irrelevant to this patch. This patch just increases the size to pass PQC cert test.

I filed https://github.com/DMTF/libspdm/issues/3186 to track the investigation.

steven-bellock (approved):

Issue #3170, opened 2025-08-18
Coverity 488780 - uninitialized variable


steven-bellock:
Coverity is confused by
https://github.com/DMTF/libspdm/blob/6f831282e021bda6c4957bbea8947ca0510f93a0/library/spdm_requester_lib/libspdm_req_encap_endpoint_info.c#L83-L92 https://github.com/DMTF/libspdm/blob/6f831282e021bda6c4957bbea8947ca0510f93a0/library/spdm_requester_lib/libspdm_req_encap_endpoint_info.c#L156-L162 and thinks that `signature_size` is uninitialized. Maybe change
```
if (context->connection_info.algorithm.req_pqc_asym_alg != 0)
```
to just `else`?

PR #3176, opened 2025-08-22
avoid uninitialized variable for signature_size.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/3170


library/spdm_requester_lib/libspdm_req_encap_endpoint_info.c
```diff
@@ -82,13 +82,12 @@ libspdm_return_t libspdm_get_encap_response_endpoint_info(void *spdm_context,
...
     if ((spdm_request->request_attributes &
          SPDM_GET_ENDPOINT_INFO_REQUEST_ATTRIBUTE_SIGNATURE_REQUESTED) != 0) {
-        if (context->connection_info.algorithm.req_base_asym_alg != 0) {
-            signature_size = libspdm_get_req_asym_signature_size(
-                context->connection_info.algorithm.req_base_asym_alg);
-        }
         if (context->connection_info.algorithm.req_pqc_asym_alg != 0) {
             signature_size = libspdm_get_req_pqc_asym_signature_size(
                 context->connection_info.algorithm.req_pqc_asym_alg);
+        } else {
```
steven-bellock, line 88: Might want to add assertion that `connection_info.algorithm.req_base_asym_alg != 0` on this path.

jyao1: I don't think that is needed.
0 will be returned if unsupported.

jyao1: I filed https://github.com/DMTF/libspdm/issues/3185 to deal with that, to ensure if SIG is required, then ASYM_ALGO is non-zero.

steven-bellock (approved):

Issue #3174, opened 2025-08-20
1.4 ​​Remove 2-byte length restriction in `libspdm_get_certificate_choose_length_ex()​​`


Li-Aaron:
The function `libspdm_get_certificate_choose_length_ex()` enforces an assertion that `length <= SPDM_MAX_CERTIFICATE_CHAIN_SIZE` while 1.4 support 4-byte length.
https://github.com/DMTF/libspdm/blob/befb49f327ee0c188ca083723b3ff634ec133d27/library/spdm_requester_lib/libspdm_req_get_certificate.c#L576-L585

PR #3177, opened 2025-08-22
remove restriction in libspdm_get_certificate_choose_length_ex

jyao1:
Fix https://github.com/DMTF/libspdm/issues/3174

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_get_certificate.c
+++ b/library/spdm_requester_lib/libspdm_req_get_certificate.c
@@ -572,17 +572,14 @@ libspdm_return_t libspdm_get_certificate_choose_length_ex(void *spdm_context,
     uint64_t retry_delay_time;
     libspdm_return_t status;
 
-    /* -=[Check Parameters Phase]=- */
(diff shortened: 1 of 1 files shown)


library/spdm_requester_lib/libspdm_req_get_certificate.c
```diff
@@ -572,17 +572,14 @@ libspdm_return_t libspdm_get_certificate_choose_length_ex(void *spdm_context,
     uint64_t retry_delay_time;
     libspdm_return_t status;
 
-    /* -=[Check Parameters Phase]=- */
-    LIBSPDM_ASSERT(length <= SPDM_MAX_CERTIFICATE_CHAIN_SIZE);
```
steven-bellock, line 576: This is conditional on the negotiated SPDM version and the Responder's `LARGE_RESP_CAP`.

jyao1: yes, that will be checked later, in libspdm_try_get_large_certificate().

The API uses `uint32_t length`, and we should allow to input a uint32_t value for 1.4.

steven-bellock (approved):

Issue #3179, opened 2025-08-22
Unit test build failure with `LIBSPDM_ENABLE_CAPABILITY_ENDPOINT_INFO_CAP`


steven-bellock:
The combination of `LIBSPDM_ENABLE_CAPABILITY_ENDPOINT_INFO_CAP == 0` and `LIBSPDM_SEND_GET_ENDPOINT_INFO_SUPPORT == 1` causes unit test build failures due to the absence of `libspdm_generate_device_endpoint_info`. For example https://github.com/DMTF/libspdm/blob/b13c7afcdca06c52fb5d0bcb72d60aff16181d25/unit_test/test_spdm_requester/get_endpoint_info.c#L98-L101 Unit tests will need an alternative way to generate endpoint info and not rely on the `hal/` function.

PR #3181, opened 2025-08-22
Fix build breaks in 3.8 release

steven-bellock:


steven-bellock:
Note that while https://github.com/DMTF/libspdm/pull/3181/commits/ff5d8ac59a3ade75bc7fa25cea972c89b6f5a88c will be brought to the `main` branch, https://github.com/DMTF/libspdm/pull/3181/commits/d23bb48147492fe13955149b0f43b862561fef1b is only for the 3.8 release. https://github.com/DMTF/libspdm/issues/3179 will correctly resolve https://github.com/DMTF/libspdm/pull/3181/commits/d23bb48147492fe13955149b0f43b862561fef1b at `main`.

include/hal/library/requester/reqasymsignlib.h
```diff
@@ -10,7 +10,8 @@
 #include "hal/base.h"
 #include "internal/libspdm_lib_config.h"
 
-#if LIBSPDM_ENABLE_CAPABILITY_MUT_AUTH_CAP
+#if (LIBSPDM_ENABLE_CAPABILITY_ENCAP_CAP) && \
+    ((LIBSPDM_ENABLE_CAPABILITY_MUT_AUTH_CAP) || (LIBSPDM_ENABLE_CAPABILITY_ENDPOINT_INFO_CAP))
```
jyao1: Supporting ENDPOINT_INFO does not mean to support signing.
The ENDPOINT_INFO could be unsigned.

jyao1:
Maybe let's fix main at first. Then sync back to 3.8.

include/hal/library/requester/reqasymsignlib.h
steven-bellock: As discussed in the meeting, the capabilities macros do not make that distinction, and if the Integrator only supports unsigned messages then they still need to provide a stub implementation.

steven-bellock:
> Maybe let's fix main at first. Then sync back to 3.8.

There are additional fixes in `main` that will not be in 3.8. It is easier to go from 3.8 to `main`.

PR #3182, opened 2025-08-22
Fix build when capability macros are disabled

### LINKED ISSUE #3178: declaration of libspdm_requester_data_sign() not found when LIBSPDM_ENABLE_CAPABILITY_MUT_AUTH_CAP is not enabled
After support of libspdm_generate_endpoint_info_signature() to support SPDM 1.3, when LIBSPDM_ENABLE_CAPABILITY_MUT_AUTH_CAP is disabled libspdm_requester_data_sign() declaration (and definition in OS stubs) is not available.

Either the API call to libspdm_requester_data_sign() (from libspdm_generate_endpoint_info_signature() API) should be under the flag LIBSPDM_ENABLE_CAPABILITY_MUT_AUTH_CAP or the declaration and definition in os_stubs for libspdm_requester_data_sign() API should be out of this header.

steven-bellock:
Fix #3178.


include/hal/library/requester/reqasymsignlib.h
```diff
@@ -10,7 +10,8 @@
 #include "hal/base.h"
 #include "internal/libspdm_lib_config.h"
 
-#if LIBSPDM_ENABLE_CAPABILITY_MUT_AUTH_CAP
+#if (LIBSPDM_ENABLE_CAPABILITY_ENCAP_CAP) && \
+    ((LIBSPDM_ENABLE_CAPABILITY_MUT_AUTH_CAP) || (LIBSPDM_ENABLE_CAPABILITY_ENDPOINT_INFO_CAP))
```
jyao1: Supporting ENDPOINT_INFO does not mean to support signing.
The ENDPOINT_INFO could be unsigned.

Why we need to "AND" ENCAP?
A requester may support MUT_AUTH without ENCAP, it still need signing capability.

jyao1:
we had better to enable CI.

include/hal/library/requester/reqasymsignlib.h
steven-bellock: >Why we need to "AND" ENCAP?

That has been removed.

PR #3209, opened 2025-09-11
1.4 add ExtFlags support to SPDM context

### LINKED ISSUE #3167: 1.4 add ExtFlags support to SPDM context
This is pre-requisite for https://github.com/DMTF/libspdm/issues/3131

Li-Aaron:
fix #3167


include/industry_standard/spdm.h
```diff
@@ -163,7 +163,8 @@ typedef struct {
      * Below field is added in 1.1.*/
     uint8_t reserved;
     uint8_t ct_exponent;
-    uint16_t reserved2;
+    /* reserved2 is replaced by ext_flags in 1.4 */
```
steven-bellock: For the request message this can stay reserved. Though the specification gives it a name, `ExtFlags`, every bit in that field is reserved.

include/industry_standard/spdm.h
```diff
@@ -178,7 +179,8 @@ typedef struct {
      * param2 == RSVD*/
     uint8_t reserved;
     uint8_t ct_exponent;
-    uint16_t reserved2;
+    /* reserved2 is replaced by ext_flags in 1.4 */
```
steven-bellock: >ext_flags is a reserved field for SPDM versions prior to 1.4

library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -279,6 +279,13 @@ static libspdm_return_t libspdm_try_get_capabilities(libspdm_context_t *spdm_con
         spdm_request->max_spdm_msg_size =
             spdm_context->local_context.capability.max_spdm_msg_size;
     }
+    if (spdm_request->header.spdm_version >= SPDM_MESSAGE_VERSION_14) {
+        spdm_request->ext_flags =
+            libspdm_mask_capability_ext_flags(spdm_context, true,
+                                              spdm_context->local_context.capability.ext_flags);
+    } else {
+        spdm_request->ext_flags = 0;
+    }
```
steven-bellock, line 288: Can just delete this as, even in 1.4, it is reserved and `0`.

library/spdm_responder_lib/libspdm_rsp_capabilities.c
```diff
@@ -313,6 +321,13 @@ libspdm_return_t libspdm_get_response_capabilities(libspdm_context_t *spdm_conte
         spdm_context->connection_info.capability.max_spdm_msg_size = 0;
     }
 
+    if (spdm_response->header.spdm_version >= SPDM_MESSAGE_VERSION_14) {
+        spdm_context->connection_info.capability.ext_flags =
+            libspdm_mask_capability_ext_flags(spdm_context, true, spdm_request->ext_flags);
+    } else {
+        spdm_context->connection_info.capability.ext_flags = 0;
+    }
+
```
steven-bellock, line 330: Can just delete this as, even in 1.4, it is reserved and `0`.

steven-bellock (changes requested):

include/industry_standard/spdm.h
jyao1: I think it should be fine to use ext_flags to align with 1.4, although every bit is reserved.

We should follow spec language...

library/spdm_requester_lib/libspdm_req_get_capabilities.c
jyao1: I believe we should setup a framework, to support what is defined in the spec, although the mask is 0.

This is extensible to > 1.4.

library/spdm_responder_lib/libspdm_rsp_capabilities.c
jyao1: I believe we should setup a framework, to support what is defined in the spec, although the mask is 0.

This is extensible to > 1.4.

jyao1:
> `ext_flags` also needs to be cleared in `libspdm_reset_context`.

Yes. I agree. Good catch!

library/spdm_requester_lib/libspdm_req_get_capabilities.c
steven-bellock: I view it as a minor bug in the specification.

PR #3232, opened 2025-09-19
Tidy up public header files

steven-bellock:


steven-bellock:
@jyao1 https://github.com/DMTF/libspdm/pull/3232/files#diff-1163dd3adc64b4fa768fe672a41993b670fd0c7368c97d25ccca3f0b476a784e is needed due to https://github.com/DMTF/libspdm/blob/050cde6139ce0d69fced53db43a9ec1a96689d17/include/industry_standard/cxl_idekm.h#L19.

jyao1 (approved):

PR #3234, opened 2025-09-23
Simplify SPDM version 1.4 handling in chunk send ack logic.

### LINKED ISSUE #3228: Use `header_size` to hide the difference between 1.3 CHUNK header and 1.4 CHUNK header.
Because 1.4 changed the CHUNK header, it caused the code using following programming style everywhere.
```
if (libspdm_get_connection_version(context) < SPDM_MESSAGE_VERSION_14) {
   ... // business logic
} else {
   ... // business logic
}
```

E.g. https://github.com/DMTF/libspdm/pull/3194
E.g. https://github.com/DMTF/libspdm/pull/3227

I think a better way is to define a `header_size`, and only this `header_size` in all business logic function. Then the business logic could be:
```
if (libspdm_get_connection_version(context) < SPDM_MESSAGE_VERSION_14) {
   header_size = 1.3 header size;
} else {
   header_size = 1.4 header size;
}
... // business logic to use header_size only.
```

It is a cleaner way to help code review and avoid mistake.
For example:
```
                if (libspdm_get_connection_version(context) < SPDM_MESSAGE_VERSION_14) {
                    header_size = sizeof(spdm_chunk_send_ack_response_14_t);
                } else {
                    header_size = sizeof(spdm_chunk_send_ack_response_t);
                }
                libspdm_copy_mem(large_buffer, large_buffer_size,
                                     my_response + header_size,
                                     my_response_size - header_size ;
                get_info->large_message = large_buffer;
                get_info->large_message_size =
                        my_response_size - header_size;
```

Li-Aaron:
fix #3228 

there are 2 header contains different size between 1.3 and 1.4
`CHUNK_GET_REQUEST` and `CHUNK_SEND_ACK_RESPONSE`

the header size logic of `CHUNK_GET_REQUEST` is simple.
this pr only enhance the header size logic of `CHUNK_SEND_ACK_RESPONSE`.

verified on emu with chunk enable, both version 1.3 and 1.4.

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_chunk_send_ack.c
+++ b/library/spdm_responder_lib/libspdm_rsp_chunk_send_ack.c
@@ -29,6 +29,7 @@ libspdm_return_t libspdm_get_response_chunk_send(libspdm_context_t *spdm_context
     uint8_t* chunk_response;
     size_t chunk_response_size;
     uint64_t max_chunk_data_transfer_size;
+    uint32_t chunk_seq_no;
 
     spdm_request = (const spdm_chunk_send_request_t*) request;
 
@@ -93,6 +94,13 @@ libspdm_return_t libspdm_get_response_chunk_send(libspdm_context_t *spdm_context
             response_size, response);
     }
 
+    if (libspdm_get_connection_version(spdm_context) < SPDM_MESSAGE_VERSION_14) {
+        chunk_seq_no = spdm_request->chunk_seq_no;
+    } else {
+        spdm_request_14 = (const spdm_chunk_send_request_14_t *) spdm_request;
+        chunk_seq_no = spdm_request_14->chunk_seq_no;
+    }
+
     send_info = &spdm_context->chunk_context.send;
 
     if (!send_info->chunk_in_use) {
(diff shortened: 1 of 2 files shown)


library/spdm_responder_lib/libspdm_rsp_chunk_send_ack.c
```diff
@@ -227,17 +227,14 @@ libspdm_return_t libspdm_get_response_chunk_send(libspdm_context_t *spdm_context
 
     if (libspdm_get_connection_version(spdm_context) < SPDM_MESSAGE_VERSION_14) {
         spdm_response->chunk_seq_no = spdm_request->chunk_seq_no;
-
-        chunk_response = (uint8_t*) (spdm_response + 1);
-        chunk_response_size = *response_size - sizeof(spdm_chunk_send_ack_response_t);
     } else {
         spdm_response_14 = response;
         spdm_response_14->chunk_seq_no = spdm_request->chunk_seq_no;
```
jyao1: I think this should be `spdm_request_14->chunk_seq_no`.

In general, I think we should define `uint32_t chunk_seq_no`.
In the beginning, 
```
if (1.4) {
    chunk_seq_no = spdm_request_14->chunk_seq_no;
} else {
    chunk_seq_no = spdm_request->chunk_seq_no;
}
```

Then always use `chunk_seq_no` in rest of the places.
For example, https://github.com/DMTF/libspdm/pull/3234/files#diff-9261204dcdafa7afb4654fab9c3bc78fd58d90dd4ee26a3788dd85a9cea6181bR202-R206 can be simplified.

jyao1 (approved):

steven-bellock (approved):

PR #3250, opened 2025-09-26
test fuzzing.

Li-Aaron:


unit_test/fuzzing/spdm_unit_fuzzing_common/common.c
```diff
@@ -119,11 +119,16 @@ bool libspdm_read_input_file(const char *file_name, void **file_data,
 {
     FILE *fp_in;
     size_t temp_result;
+    char ci_file_name[UINT8_MAX];
 
     if ((fp_in = fopen(file_name, "rb")) == NULL) {
-        printf("Unable to open file %s\n", file_name);
-        *file_data = NULL;
-        return false;
+        /* CI environment working directory is parent folder */
```
steven-bellock, line 125: Are fuzzing tests able to be run outside of GitHub CI/CD?

Li-Aaron: Yes, fuzzing tests are able to be run outside of GitHub CI/CD even without this change.

PR #3278, opened 2025-10-09
Improve certificate failure reporting

### LINKED ISSUE #3276: Provide more granular certificate chain verification prints
From https://github.com/DMTF/libspdm/discussions/3274 and in the same vein as https://github.com/DMTF/libspdm/issues/2920.

Provide more granular and detailed prints when verifying the peer's certificate chain, especially in `libspdm_x509_common_certificate_check`.

steven-bellock:
Fix #3276.

### DIFF
--- a/library/spdm_crypt_lib/libspdm_crypt_cert.c
+++ b/library/spdm_crypt_lib/libspdm_crypt_cert.c
@@ -1320,11 +1320,13 @@ static bool libspdm_verify_leaf_cert_spdm_eku(const uint8_t *cert, size_t cert_s
     if (is_requester_cert) {
         /* it should not only contain responder auth oid */
         if (!req_auth_oid_find_success && rsp_auth_oid_find_success) {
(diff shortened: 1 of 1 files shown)


library/spdm_crypt_lib/libspdm_crypt_cert.c
```diff
@@ -1465,36 +1469,43 @@ bool libspdm_x509_common_certificate_check(
     end_cert_from_len = 64;
     end_cert_to_len = 64;
 
-    /* 1. version*/
+    /* 1. Version */
     cert_version = 0;
     status = libspdm_x509_get_version(cert, cert_size, &cert_version);
     if (!status) {
```
steven-bellock, line 1475: Should also add print if Version field is not present in the certificate.

jyao1: yes, whatever something that can help debug.

PR #3284, opened 2025-10-13
Support for Darwin clang toolchain

itsManjeet:
Added CMake configurations to support build on darwin clang toolchain Tested for following configurations
- ARCH=aarch64 CRYPTO=openssl TOOLCHAIN=CLANG


os_stub/platform_lib/CMakeLists.txt
```diff
@@ -20,10 +20,16 @@ elseif(CMAKE_SYSTEM_NAME MATCHES "Windows")
         time_win.c
         watchdog.c
     )
+elseif(CMAKE_SYSTEM_NAME MATCHES "Darwin")
+    target_sources(platform_lib
+        PRIVATE
+        time_darwin.c
```
itsManjeet: Do we need separate `time_darwin.c` or we can make a common `time_unix.c` for Darwin and Linux?

steven-bellock: `time_unix.c` makes sense since they are identical code.

.github/workflows/build.yml
```diff
@@ -151,6 +162,16 @@ jobs:
...
+      - name: Build - Darwin
+        if: matrix.os == 'macos-latest'
+        run: |
+          mkdir -p build
+          cd build
+          cmake --version
+          cmake -E env CFLAGS="${{ matrix.configurations }} -DLIBSPDM_DEBUG_LIBSPDM_ASSERT_CONFIG=3" cmake -DARCH=${{ matrix.arch }} -DTOOLCHAIN=${{ matrix.toolchain }} -DTARGET=${{ matrix.target }} -DCRYPTO=${{ matrix.crypto }} ..
+          cp -a ../unit_test/sample_key/* bin/
+          cp -a ../unit_test/fuzzing/* bin/ && rm -rf bin/test_spdm_common bin/test_spdm_crypt
+          make -j1 # process killed with nproc
```
steven-bellock: See, for example, https://github.com/zipper177/SatDump/commit/d9e3b1519db8c09d4d4463ccd499f8d35acc6021.

.github/workflows/build.yml
```diff
@@ -151,6 +162,16 @@ jobs:
...
           make copy_seed
           make -j`nproc`
+      - name: Build - Darwin
+        if: matrix.os == 'macos-latest'
+        run: |
+          mkdir -p build
+          cd build
+          cmake --version
+          cmake -E env CFLAGS="${{ matrix.configurations }} -DLIBSPDM_DEBUG_LIBSPDM_ASSERT_CONFIG=3" cmake -DARCH=${{ matrix.arch }} -DTOOLCHAIN=${{ matrix.toolchain }} -DTARGET=${{ matrix.target }} -DCRYPTO=${{ matrix.crypto }} ..
+          cp -a ../unit_test/sample_key/* bin/
```
steven-bellock: Why doesn't `make copy_sample_key` work?

itsManjeet: Fixed with https://github.com/DMTF/libspdm/pull/3284/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR900

.github/workflows/build.yml
steven-bellock: @itsManjeet this is still unresolved.

itsManjeet: Hardcoding -j1 is intentional. I tried using `$(sysctl -n hw.logicalcpu)` previously to use all cores, but it caused the process killed with `SIGKILL` Error 137 on github actions.

steven-bellock: >I tried using $(sysctl -n hw.logicalcpu)

The comment says
```
# process killed with nproc
```

But if `$(sysctl -n hw.logicalcpu)` does not work then standalone `make` without the `-j1` is preferred.

Issue #3168, opened 2025-08-18
Add support for multikey signing


steven-bellock:
https://github.com/DMTF/libspdm/blob/6f831282e021bda6c4957bbea8947ca0510f93a0/include/hal/library/responder/asymsignlib.h#L67-L75

At a minimum the `KeyPairID` needs to be passed to the signing entity.

steven-bellock:
>If we add KeyPairID, that means libspdm need to translate SlotID to KeyPairID, that means it needs to assume DIGEST or GET_KEY_PAIR_INFO or SLOT_MANAGEMENT invoked and info is cached.

That mapping already exists in
```
spdm_context->local_context.local_key_pair_id[slot_id];
```

jyao1:
> > If we add KeyPairID, that means libspdm need to translate SlotID to KeyPairID, that means it needs to assume DIGEST or GET_KEY_PAIR_INFO or SLOT_MANAGEMENT invoked and info is cached.
> 
> That mapping already exists in
> 
> ```
> spdm_context->local_context.local_key_pair_id[slot_id];
> ```

It is different. This local_context is in **responder** side, not **requester** side.
You cannot provision responder's slot_id<->key_pair_id mapping to the requester.

PR #3290, opened 2025-10-14
Multikey signing

jyao1:
Fix https://github.com/DMTF/libspdm/issues/3168

NOTE: This PR just adds minimal support for multiple key in libspdm.
A full emulation requires to update spdm_device_secret_lib_sample to emulate the multi-key internal configuration. It can be handled in different issue and PR.

Also, the responder signing function can also check the mapping between slot_id and keypairId capability.
That improvement can be done in different issue and PR.

### DIFF
--- a/unit_test/fuzzing/test_requester/test_spdm_requester_challenge/challenge.c
+++ b/unit_test/fuzzing/test_requester/test_spdm_requester_challenge/challenge.c
@@ -104,7 +104,7 @@ libspdm_return_t libspdm_device_receive_message(void *spdm_context, size_t *resp
     libspdm_responder_data_sign(
         spdm_context,
         spdm_response->header.spdm_version << SPDM_VERSION_NUMBER_SHIFT_BIT,
-            SPDM_CHALLENGE_AUTH, m_libspdm_use_asym_algo, m_libspdm_use_pqc_asym_algo,
+            0, SPDM_CHALLENGE_AUTH, m_libspdm_use_asym_algo, m_libspdm_use_pqc_asym_algo,
             m_libspdm_use_hash_algo, false,
             m_libspdm_local_buffer, m_libspdm_local_buffer_size, ptr,
             &sig_size);
--- a/unit_test/fuzzing/test_requester/test_spdm_requester_key_exchange/key_exchange.c
+++ b/unit_test/fuzzing/test_requester/test_spdm_requester_key_exchange/key_exchange.c
@@ -205,7 +205,7 @@ libspdm_return_t libspdm_device_receive_message(void *spdm_context, size_t *resp
         libspdm_responder_data_sign(
             spdm_context,
             spdm_response->header.spdm_version << SPDM_VERSION_NUMBER_SHIFT_BIT,
-                SPDM_KEY_EXCHANGE_RSP,
+                0, SPDM_KEY_EXCHANGE_RSP,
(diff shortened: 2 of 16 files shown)


include/hal/library/responder/asymsignlib.h
```diff
@@ -65,7 +65,8 @@ extern bool libspdm_challenge_opaque_data(
 extern bool libspdm_responder_data_sign(
     void *spdm_context,
     spdm_version_number_t spdm_version,
-    uint8_t op_code, uint32_t base_asym_algo, uint32_t pqc_asym_algo,
+    uint8_t slot_id, uint8_t op_code,
```
steven-bellock: Since libspdm already has a mapping between `SlotID` and `KeyPairID` then this should take `KeyPairID` as the parameter. If multikey is not supported then its value would always be `0` or `1` (TBD).

jyao1: Yes, using KeyPairID should be fine.

I feel we should use `0` because GET_CSR and SET_CERT command use `0` if multikey is not enabled.

steven-bellock: And by "multikey not supported" we mean `MULTI_KEY_CAP` is `0`.

steven-bellock: https://github.com/DMTF/SPDM-WG/issues/3270 is where we discussed values of `KeyPairID`.

jyao1: Look through the long discussion, I feel we may treat both 0 and 1 to be valid value if multi-key is disabled.

jyao1: I changed HAL to use `key_pair_id`, and created an internal function `libspdm_slot_id_to_key_pair_id()`

include/hal/library/requester/reqasymsignlib.h
```diff
@@ -35,7 +35,7 @@
 extern bool libspdm_requester_data_sign(
     void *spdm_context,
     spdm_version_number_t spdm_version,
-    uint8_t op_code,
+    uint8_t key_pair_id, uint8_t op_code,
```
steven-bellock, line 41: Need to specify that the value is `0` if
- SPDM version is < 1.3 OR
- SPDM version is >= 1.3 AND `LIBSPDM_DATA_MULTI_KEY_CONN_REQ` is `false`

include/hal/library/responder/asymsignlib.h
```diff
@@ -65,7 +65,8 @@ extern bool libspdm_challenge_opaque_data(
 extern bool libspdm_responder_data_sign(
     void *spdm_context,
     spdm_version_number_t spdm_version,
-    uint8_t op_code, uint32_t base_asym_algo, uint32_t pqc_asym_algo,
+    uint8_t key_pair_id, uint8_t op_code,
```
steven-bellock, line 71: Need to specify that the value is `0` if
- SPDM version is < 1.3 OR
- SPDM version is >= 1.3 AND `LIBSPDM_DATA_MULTI_KEY_CONN_RSP` is `false`

include/hal/library/requester/reqasymsignlib.h
jyao1: done

include/hal/library/responder/asymsignlib.h
jyao1: done

steven-bellock (approved):

Issue #3263, opened 2025-10-04
`KEY_PAIR_INFO` tests are not being executed


steven-bellock:
https://github.com/DMTF/libspdm/blob/1578aee665d769d797a5c02efb39a6972e915394/unit_test/test_spdm_requester/get_key_pair_info.c#L200-L203

https://github.com/DMTF/libspdm/blob/1578aee665d769d797a5c02efb39a6972e915394/unit_test/test_spdm_requester/get_key_pair_info.c#L579-L584

Also `case3` should be `case2`. This issue is present in the original pull request: https://github.com/DMTF/libspdm/pull/2817.

steven-bellock:
Note also that when the test case is actually run it encounters a failure, so that needs to be fixed as well.

PR #3292, opened 2025-10-14
Keypair info conformance check

jyao1:
Fix https://github.com/DMTF/libspdm/issues/3263

### DIFF
--- a/unit_test/test_spdm_requester/get_key_pair_info.c
+++ b/unit_test/test_spdm_requester/get_key_pair_info.c
@@ -198,7 +198,8 @@ static void req_get_key_pair_info_case1(void **state)
 }
 
 /**
- * Test 2: The collection of multiple sub-cases.
+ * Test 2: The collection of multiple sub-cases for invalid combination.
+ * Expected Behavior: get a LIBSPDM_STATUS_INVALID_MSG_FIELD return code
  **/
 void req_get_key_pair_info_case2(void **state)
 {
@@ -581,7 +582,8 @@ int libspdm_req_get_key_pair_info_test(void)
     const struct CMUnitTest test_cases[] = {
         /* Successful response to get key pair info, key_pair_id is 1*/
         cmocka_unit_test(req_get_key_pair_info_case1),
-        /* cmocka_unit_test(req_get_key_pair_info_case2) */
+        /* The collection of multiple sub-cases for invalid combination.*/
+        cmocka_unit_test(req_get_key_pair_info_case2),
     };
 
     libspdm_test_context_t test_context = {


library/spdm_requester_lib/libspdm_req_get_key_pair_info.c
```diff
@@ -262,24 +280,18 @@ static libspdm_return_t libspdm_try_get_key_pair_info(libspdm_context_t *spdm_co
         *current_pqc_asym_algo = rsp_current_pqc_asym_algo;
     }
 
-    /*If responder doesn't support SET_KEY_PAIR_INFO_CAP,the capabilities should be 0*/
-    if ((!libspdm_is_capabilities_flag_supported(
-             spdm_context, true, 0,
-             SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_SET_KEY_PAIR_INFO_CAP)) &&
-        ((spdm_response->capabilities & SPDM_KEY_PAIR_CAP_MASK) != 0)) {
+    *assoc_cert_slot_mask = spdm_response->assoc_cert_slot_mask;
+    if (!libspdm_onehot0(*assoc_cert_slot_mask) &&
```
steven-bellock, line 284: This also needs to evaluate `SET_KEY_PAIR_INFO_CAP`.
>If `ShareableCap` is not set and `SET_KEY_PAIR_INFO_CAP` is set, no more than one bit shall be set. Otherwise, any number of bits can be set.

jyao1: fixed

steven-bellock: Shouldn't it be `&& libspdm_is_capabilities_flag_supported(`?

jyao1: I read spec again.
`ShareableCap` means "If set, the Responder allows a Requester to associate the given KeyPairID with more than one CertSlot". Since `Capabilities` cannot be changed, it is a fixed cap. That means if a device does not have `ShareableCap`, then more than one bit for CertSlot is a golden rule.

If `SET_KEY_PAIR_INFO_CAP` is set, "Shareable not set" + "more than one bit set" is invalid config when requester SET_KEY_PAIR_INFO. That means, the responder cannot be put in such situation at any time.

If `SET_KEY_PAIR_INFO_CAP` is NOT set, then the config is not changeable. I believe "Shareable not set" + "more than one bit set" is still invalid config.

Now, I am thinking to skip `SET_KEY_PAIR_INFO_CAP` check. I feel it is meaningless. As long as "Shareable not set", the device should never be in condition that "more than one bit set".

I filed https://github.com/DMTF/SPDM-WG/issues/4375

steven-bellock: See https://github.com/DMTF/SPDM-WG/issues/3543 and https://github.com/DMTF/SPDM-WG/pull/3596 that brought in that change.

jyao1: I see. Fixed accordingly.

Issue #3299, opened 2025-10-15
EDDSA 448 Sign FIPS test failed in v3.8.0


jyao1:
It failed in v3.8.0 and till dc7e7c304284cd1c32f94d763c97cfeb4d87d4d5.

But bb851496a400f571ace8d0b9c4754ad8a79d9443 introduced another regression (https://github.com/DMTF/libspdm/issues/3298)

```
[==========] spdm_crypt_lib_tests: Running 1 test(s).
[ RUN      ] libspdm_test_fips
EDDSA 448 sign failed
[  ERROR   ] --- status
[   LINE   ] --- C:\home\DMTF\libspdm\unit_test\test_spdm_fips\test_spdm_fips.c:23: error: Failure!
[  FAILED  ] libspdm_test_fips
[==========] spdm_crypt_lib_tests: 1 test(s) run.
[  PASSED  ] 0 test(s).
[  FAILED  ] spdm_crypt_lib_tests: 1 test(s), listed below:
[  FAILED  ] libspdm_test_fips
```

PR #3300, opened 2025-10-15
Fix EdDSA448 NULL context string handling

jyao1:
Fix https://github.com/DMTF/libspdm/issues/3299


steven-bellock:
@jyao1 feel free to merge as you see fit.

PR #3302, opened 2025-10-16
Add FIPS self-test for ML-DSA

Li-Aaron:
Ref: https://github.com/DMTF/libspdm/issues/3152


library/spdm_crypt_lib/fips/libspdm_selftest_mldsa.c
```diff
@@ -0,0 +1,290 @@
...
+        return false;
+    }
+
+    /* check if run before.*/
+    if ((context->tested_algo & LIBSPDM_FIPS_SELF_TEST_ML_DSA) != 0) {
+        return true;
+    }
+
+#if LIBSPDM_ML_DSA_44_SUPPORT
+    uint8_t signature_44[2420];
```
steven-bellock: This is huge for being stack-allocated. Should probably look into having the Integrator provide a buffer to hold these values.

PR #3306, opened 2025-10-17
Add FIPS self-test for SLH-DSA

Li-Aaron:
Ref: https://github.com/DMTF/libspdm/issues/3152


library/spdm_crypt_lib/fips/libspdm_selftest_slhdsa.c
```diff
@@ -0,0 +1,133 @@
...
+    if ((context->tested_algo & LIBSPDM_FIPS_SELF_TEST_SLH_DSA) != 0) {
+        return true;
+    }
+
+    uint8_t *signature;
+    size_t sig_size;
+    void *slhdsa_context;
+
+    sig_size = 29792;
+    signature = allocate_zero_pool(sig_size);
```
jyao1: Please don't use `allocate()` function.

One possible option is to define a buffer in fips_context and let integrator define it (either in heap or in global data area)

library/spdm_crypt_lib/fips/libspdm_selftest_slhdsa.c
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    /* KAT Vectors */
+    extern const uint8_t message_hash_sha2_256s[1];
+    extern const uint8_t priv_key_sha2_256s[128];
+    extern const uint8_t public_key_sha2_256s[64];
+    extern const uint8_t sign_context_sha2_256s[254];
+    extern const uint8_t expected_signature_sha2_256s[29792];
+
+    slhdsa_context = libspdm_slhdsa_new(LIBSPDM_CRYPTO_NID_SLH_DSA_SHA2_256S);
```
Li-Aaron: will use `SLH-DSA-SHA2-128s` instead, since it requires smaller buffer size.

steven-bellock:
@jyao1 feel free to merge as you see fit.

PR #3318, opened 2025-10-21
Pass request Context field to Integrator

### LINKED ISSUE #3308: Pass request context to Integrator
Inspired by https://github.com/DMTF/SPDM-WG/issues/4378.

SPDM 1.3 added an eight byte `context` field to `CHALLENGE` and `GET_MEASUREMENTS`. At a minimum libspdm should pass that value to the Responder Integrator via the HAL functions.

steven-bellock:
Fix #3308.


include/hal/library/requester/reqasymsignlib.h
```diff
@@ -65,6 +67,7 @@ extern bool libspdm_encap_challenge_opaque_data(
     void *spdm_context,
     spdm_version_number_t spdm_version,
     uint8_t slot_id,
+    uint64_t request_context,
```
jyao1: how about to use `uint8_t *request_context`?

Just in case that, the request_context is enlarged (more than 8 bytes) in next version.

PR #3319, opened 2025-10-21
remove LIBSPDM_MAX_ASYM_KEY_SIZE

jyao1:
LIBSPDM_MAX_ASYM_SIG_SIZE should be used for signature size


library/spdm_crypt_lib/fips/libspdm_selftest_rsa_pss.c
```diff
@@ -32,7 +32,7 @@ bool libspdm_fips_selftest_rsa_pss(void *fips_selftest_context)
         return true;
     }
 
-    uint8_t signature[LIBSPDM_MAX_ASYM_KEY_SIZE];
+    uint8_t signature[512];
```
steven-bellock, line 35: `LIBSPDM_MAX_ASYM_SIG_SIZE` no?

jyao1: No.
I use the value purposely, because, the MAX size is MAX for **all** supported algorithm, while here 512 is good enough for **PSS**.

PR #3325, opened 2025-10-28
Fix : vendor_response_get_id should not exist

### LINKED ISSUE #3045: `vendor_response_get_id` should not exist
`libspdm_get_vendor_defined_response` makes separate calls to https://github.com/DMTF/libspdm/blob/707af3b726707ef3052922355324a04bc92c2a85/library/spdm_responder_lib/libspdm_rsp_vendor_response.c#L180-L184 and https://github.com/DMTF/libspdm/blob/707af3b726707ef3052922355324a04bc92c2a85/library/spdm_responder_lib/libspdm_rsp_vendor_response.c#L191-L197 In the case where a Responder supports multiple VDM types, for the call to `vendor_response_get_id` the Integrator is completely blind as to what data to return, as it needs to examine the request message. Only `vendor_response_callback` is needed and this should include the response VDM's `RegisteryId`, etc.

ShitalJumbad:
fix #3045

### DIFF
--- a/include/library/spdm_common_lib.h
+++ b/include/library/spdm_common_lib.h
@@ -945,24 +945,6 @@ bool libspdm_get_fips_mode(void);
 #if LIBSPDM_ENABLE_VENDOR_DEFINED_MESSAGES
 
 /**
- * Vendor Response Get Vendor ID Callback Function Pointer.
- * Required to be able to compose the Vendor Defined Response correctly
- *
- * @param  spdm_context        A pointer to the SPDM context.
- * @param  session_id          If non-NULL then message is within a secure session.
- *                             If NULL then message is outside a secure session.
- * @param  resp_standard_id    Registry or Standards body used for response
- * @param  resp_vendor_id_len  Length in bytes of the vendor id field for the response
- * @param  resp_vendor_id      Vendor ID assigned by the Registry or Standards Body. Little-endian format
- **/
-typedef libspdm_return_t (*libspdm_vendor_get_id_callback_func)(
-    void *spdm_context,
-    const uint32_t *session_id,
-    uint16_t *resp_standard_id,
-    uint8_t *resp_vendor_id_len,
-    void *resp_vendor_id);
-
-/**
  * Vendor Response Callback Function Pointer.
  *
  * @param  spdm_context       A pointer to the SPDM context.
--- a/library/spdm_responder_lib/libspdm_rsp_vendor_response.c
+++ b/library/spdm_responder_lib/libspdm_rsp_vendor_response.c
@@ -8,15 +8,6 @@
 
 #if LIBSPDM_ENABLE_VENDOR_DEFINED_MESSAGES
 
-libspdm_return_t libspdm_register_vendor_get_id_callback_func(void *spdm_context,
-                                                              libspdm_vendor_get_id_callback_func resp_callback)
-{
-
-    libspdm_context_t *context = (libspdm_context_t *)spdm_context;
-    context->vendor_response_get_id = resp_callback;
-    return LIBSPDM_STATUS_SUCCESS;
-}
-
 libspdm_return_t libspdm_register_vendor_callback_func(void *spdm_context,
                                                        libspdm_vendor_response_callback_func resp_callback)
 {
@@ -76,9 +67,8 @@ libspdm_return_t libspdm_get_vendor_defined_response(libspdm_context_t *spdm_con
         session_id = &session_info->session_id;
     }
 
-    /* Check if caller is using the old Vendor Defined API. */
-    if ((spdm_context->vendor_response_callback == NULL ||
-         spdm_context->vendor_response_get_id == NULL)) {
+    /* Check if vendor callback is registered. */
+    if (spdm_context->vendor_response_callback == NULL) {
         if (spdm_context->get_response_func != NULL) {
             return ((libspdm_get_response_func)spdm_context->get_response_func)(
                 spdm_context,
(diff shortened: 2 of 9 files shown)


include/library/spdm_common_lib.h
```diff
@@ -935,37 +935,25 @@ bool libspdm_get_fips_mode(void);
...
-typedef libspdm_return_t (*libspdm_vendor_get_id_callback_func)(
-    void *spdm_context,
-    const uint32_t *session_id,
-    uint16_t *resp_standard_id,
-    uint8_t *resp_vendor_id_len,
-    void *resp_vendor_id);
-
-/**
- * Vendor Response Callback Function Pointer.
+ * The library may invoke this callback twice:
```
steven-bellock: No, it only needs to be called once.

steven-bellock (changes requested):

include/library/spdm_common_lib.h
```diff
@@ -946,37 +946,25 @@ bool libspdm_get_fips_mode(void);
...
-typedef libspdm_return_t (*libspdm_vendor_get_id_callback_func)(
-    void *spdm_context,
-    const uint32_t *session_id,
-    uint16_t *resp_standard_id,
-    uint8_t *resp_vendor_id_len,
-    void *resp_vendor_id);
-
-/**
- * Vendor Response Callback Function Pointer.
+ * The library invokes this callback once. The integrator should:
```
steven-bellock: >The library invokes this callback once.

Can just delete this.

library/spdm_responder_lib/libspdm_rsp_vendor_response.c
```diff
@@ -11,15 +11,6 @@
 /* expected number of bytes for VENDOR MESSAGE HEADERS */
 #define SPDM_VENDOR_DEFINED_FIXED_HEADER_LEN 7
```
steven-bellock: Pre-existing issue, but I do not see a reason for this `#define` to exist. I will file a separate issue for it.

library/spdm_responder_lib/libspdm_rsp_vendor_response.c
```diff
@@ -182,8 +172,8 @@ libspdm_return_t libspdm_get_vendor_defined_response(libspdm_context_t *spdm_con
     libspdm_reset_message_buffer_via_request_code(spdm_context, NULL,
                                                   spdm_request->header.request_response_code);
 
-    /* length of spdm request/response header before payload start */
-    header_length = sizeof(spdm_vendor_defined_response_msg_t) + spdm_request->len +
+    /* reserve max vendor ID in header */
+    header_length = sizeof(spdm_vendor_defined_response_msg_t) + SPDM_MAX_VENDOR_ID_LENGTH +
```
steven-bellock: @ShitalJumbad We (mostly me) have been a little lazy in defining `SPDM_MAX_VENDOR_ID_LENGTH` as `255` bytes. Let me file an issue, resolve it with a pull request, and then you can rebase with the new value. This should simplify things so that you do not need to
>compact payload after Vendor ID

steven-bellock: https://github.com/DMTF/SPDM-WG/issues/4396 has complicated things. Let me resolve that first.

steven-bellock: After discussing this in the SPDM Working Group, for IANA CBOR the `VendorId` field is the CBOR-encoded tag number. As such its maximum size is nine bytes: `0xdb` plus eight bytes for the tag number. I will update `SPDM_MAX_VENDOR_ID_LENGTH` to this more precise value.

include/library/spdm_common_lib.h
```diff
@@ -985,6 +973,9 @@ typedef libspdm_return_t (*libspdm_vendor_response_callback_func)(
     const void *req_vendor_id,
     uint32_t req_size,
     const void *req_data,
+    uint16_t *resp_standard_id,
+    uint8_t *resp_vendor_id_len,
+    void *resp_vendor_id,
```
steven-bellock: @ShitalJumbad as discussed in the meeting today, please remove these parameters and have libspdm just copy the standard ID and vendor ID from the request to the response in `libspdm_rsp_vendor_response.c`.

include/library/spdm_common_lib.h
ShitalJumbad: Done.

include/library/spdm_common_lib.h
ShitalJumbad: Done.

library/spdm_responder_lib/libspdm_rsp_vendor_response.c
steven-bellock: Filed https://github.com/DMTF/libspdm/issues/3403.

include/library/spdm_common_lib.h
```diff
@@ -977,7 +959,7 @@ typedef libspdm_return_t (*libspdm_vendor_get_id_callback_func)(
  * @param  resp_size          Length of the response
  * @param  resp_data          The vendor defined response
  **/
-typedef libspdm_return_t (*libspdm_vendor_response_callback_func)(
+ typedef libspdm_return_t (*libspdm_vendor_response_callback_func)(
```
jyao1: why change this ? typo?

ShitalJumbad: Correct, thanks for catching this. Updated.

steven-bellock (approved):

PR #3338, opened 2025-11-06
erase the plain-text VDM message after processing it.

### LINKED ISSUE #2426: consider to erase the plain-text VDM message after process it.
The reason is that there might be secret info there, such as IDE key.

Ref: https://github.com/DMTF/spdm-emu/pull/298

Li-Aaron:
fix #2426

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_vendor_request.c
+++ b/library/spdm_requester_lib/libspdm_req_vendor_request.c
@@ -167,6 +167,7 @@ libspdm_return_t libspdm_try_vendor_send_request_receive_response(
         libspdm_release_sender_buffer (spdm_context);
         return LIBSPDM_STATUS_SEND_FAIL;
     }
+    libspdm_zero_mem(message, message_size);
     libspdm_release_sender_buffer (spdm_context);
     spdm_request = (void *)spdm_context->last_spdm_request;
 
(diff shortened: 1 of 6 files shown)


library/spdm_requester_lib/libspdm_req_vendor_request.c
```diff
@@ -296,6 +297,11 @@ libspdm_return_t libspdm_try_vendor_send_request_receive_response(
 
     status = LIBSPDM_STATUS_SUCCESS;
 done:
+    libspdm_zero_mem(spdm_context->last_spdm_request,
+                     libspdm_get_scratch_buffer_last_spdm_request_capacity(spdm_context));
+    spdm_context->last_spdm_request_size = 0;
+    /* spdm_response may not equal to message */
```
jyao1: what does it mean?

Li-Aaron: reciver buffer `message` contains crypted message
`spdm_response` contains the plain-text VDM message
will modify the comments

library/spdm_responder_lib/libspdm_rsp_receive_send.c
```diff
@@ -864,7 +866,19 @@ libspdm_return_t libspdm_build_response(void *spdm_context, const uint32_t *sess
         }
     }
 
-    return LIBSPDM_STATUS_SUCCESS;
+    status = LIBSPDM_STATUS_SUCCESS;
+done:
+    if (spdm_request->request_response_code == SPDM_VENDOR_DEFINED_REQUEST) {
```
steven-bellock: I don't think this needs to be limited to a VDM. Any encrypted message should probably be cleared.

steven-bellock (approved):

PR #3339, opened 2025-11-07
Add SPDM 1.4 in README.md

jyao1:


README.md
```diff
@@ -52,6 +52,8 @@
 
    SPDM 1.3: `GET_KEY_PAIR_INFO`, `SET_KEY_PAIR_INFO`, `SUBSCRIBE_EVENT_TYPE`, `GET_SUPPORTED_EVENT_TYPES`, `GET_ENDPOINT_INFO` and `GET_MEASUREMENT_EXTENSION_LOG`. Additional SPDM 1.3 messages will be implemented in future releases.
 
+   SPDM 1.4: Additional SPDM 1.4 messages will be implemented in future releases.
```
steven-bellock, line 55: Though we do not not need to implement `SLOT_MANAGEMENT` before the next major release, it would be good to look at `SLOT_MANAGEMENT` and see what effect it has on existing `hal/` APIs like `libspdm_write_certificate_to_nvm`. If they need to be changed to support `SLOT_MANAGEMENT` then now is the time to do it.

PR #3341, opened 2025-11-07
Specify SPDM Standard Measurement Transcripts

steven-bellock:
Fix https://github.com/DMTF/libspdm/issues/3108.
Fix https://github.com/DMTF/SPDM-WG/issues/4127.

Of note is that there is no normative text in the document and that it is mechanical and descriptive. 

Co-Authored-By: Jiewen Yao <jiewen.yao@intel.com>

### DIFF
--- a/doc/standard_measurement_transcripts.md
+++ b/doc/standard_measurement_transcripts.md
@@ -49,6 +49,11 @@ For SPDM 1.2 and later, the byte buffer is
 {VCA, GET_MEASUREMENTS(0), MEASUREMENTS(0), GET_MEASUREMENTS(1), MEASUREMENTS(1), ..., GET_MEASUREMENTS(N), MEASUREMENTS(N)}
 ```
 
+For SPDM 1.0 and 1.1 the above indices correspond to the value of `GET_MEASUREMENTS.Param2`. For
+SPDM 1.2 and later, this mapping may not hold, as the Responder can introduce gaps between populated
+measurement indices. For SPDM 1.2 and later, this document does not describe how a Requester
+determines which measurement indices are populated.
+
 The `GET_MEASUREMENTS(0)` request has the following properties:
 * `SignatureRequested` is not set.
 * `Param2 = 0x00`
@@ -57,12 +62,13 @@ The value of `N` is equal to the value of `MEASUREMENTS(0).Param1`.
 
 The `GET_MEASUREMENTS(1)` to `GET_MEASUREMENTS(N - 1)` requests have the following properties:
 * `SignatureRequested` is not set.
-* `Param2` starts with a value of `0x01` and its value increments by `1` with each successive
-  `GET_MEASUREMENTS` request. This is repeated until the number of `MEASUREMENT` responses in this
-  step is equal to `N - 1`. The last measurement index of this step is denoted as `I`.
+* The value of `Param2` increases strictly monotonically with each `GET_MEASUREMENT` request.
+    * For SPDM 1.0 and 1.1, `Param2` starts with a value of `1` and increments by `1` until its
+      value is equal to `N - 1`.
 
 The `GET_MEASUREMENTS(N)` request has the following properties:
 * If the Responder supports signature generation (`MEAS_CAP = 10b`) then `SignatureRequested` is
   set, else it is not set.
-* `Param2` starts with a value of `I + 1` and its value increments by `1` until `MEASUREMENT(N)`
-  is returned.
+* `Param2` is
+    * For SPDM 1.0 and 1.1, equal to `N`.
+    * For SPDM 1.2 and later, greater than `GET_MEASUREMENTS(N - 1).Param2`.


steven-bellock:
From https://github.com/DMTF/SPDM-WG/issues/4127#issuecomment-3506965011.

Consider renaming "Reports" to "Transcripts" to reduce scope of document and temper expectations.

doc/standard_measurement_reports.md
```diff
@@ -0,0 +1,67 @@
+# SPDM Standard Measurement Reports
+
+This document describes two standard measurement reports for an SPDM Responder that are constructed
+by an SPDM Requester and consumed by a Verifier while all agents are operating in production mode. A
+specification for or an implementation of a Verifier or Requester can specify and publish support
```
steven-bellock:
```suggestion
specification for, or an implementation of, a Verifier or Requester can specify and publish support
```

doc/standard_measurement_reports.md
```diff
@@ -0,0 +1,67 @@
...
+
+The `GET_MEASUREMENTS(1)` to `GET_MEASUREMENTS(N - 1)` requests have the following properties:
+* `Param2` starts with a value of `0x01` and its value increments by `1` with each successive
+  `GET_MEASUREMENTS` request. This is repeated until the number of `MEASUREMENT` responses in this
+  step is equal to `N - 1`. The last measurement index of this step is denoted as `I`.
+* `SignatureRequested` is not set.
+
+The `GET_MEASUREMENTS(N)` request has the following properties:
+* `Param2` starts with a value of `I + 1` and its value increments by `1` until a `MEASUREMENT`
+   response is returned.
```
steven-bellock:
```suggestion
* `Param2` starts with a value of `I + 1` and its value increments by `1` until `MEASUREMENT(N)`
  is returned.
```

mcsmonk (commented):
The current version assumes that measurement blocks are indexed contiguously (i.e., increasing index function I(i) = i for i in [1, N-1]). This assumption does not fully comply with SPDM v1.2+, which permits non-contiguous but strictly increasing indices (e.g., I(i) in {1, 2, 3, 0xFE} or extream case I(i) in {1, 3, 5}), as long as they are ordered.

Under the current logic, an invalid or missing index (e.g., gap in sequence) triggers an error, leading to re-initialization of the `L1`/`L2` transcript — thereby breaking attestation.
(`DSP0274` describes that "An error response with any error code other than ResponseNotReady or LargeResponse
shall re-initialize `L1/L2` to null".)

To solve this, the Requester must accumulate measurement data 1-by-1 in a buffer only for indices that are explicitly returned by the Responder.

To safely locate valid indices without assuming contiguity, two practical approaches are feasible:

Exhaustive Search: Probe indices from 0x01 to 0xFE 1-by-1 via individual GET_MEASUREMENT requests, and then process only those blocks that return valid responses.
Manifest-Based Discovery: Leverage the `spdm-indirect` field (as defined in TCG "DICE Concise Evidence Binding for SPDM") to obtain a pre-declared list of valid Measurement Block indices from the Responder’s manifest.
A more scalable and efficient solution would require the Responder to advertise its supported indices in a dedicated field (e.g., a SupportedIndices measurement block), eliminating the need for probing or manifest parsing. Maybe, such an extension would align with SPDM’s extensibility model and significantly improve performance in large-scale attestation scenarios.

doc/standard_measurement_transcripts.md
```diff
@@ -0,0 +1,68 @@
...
+
+A standard measurement transcript is a byte buffer that consists of the L1/L2 transcript along with
+a signature over the transcript if the Responder supports measurement signing. All
+`GET_MEASUREMENT` requests have the following properties:
+* For SPDM 1.2 and later, `RawBitStreamRequested` is not set.
+* For SPDM 1.3 and later, `NewMeasurementRequested` is not set.
+
+## Single-request Measurement Transcript Definition
+
+The single-request measurement transcript is comprised of a single `GET_MEASUREMENTS` request and a
```
thomas-fossati, line 18:
```suggestion
The single-request measurement transcript comprises a single `GET_MEASUREMENTS` request and a
```

doc/standard_measurement_transcripts.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+The `GET_MEASUREMENTS` request has the following properties:
+* If the Responder supports signature generation (`MEAS_CAP = 10b`) then `SignatureRequested` is
+  set, else it is not set.
+* `Param2 = 0xFF`
+
+## Multiple-request Measurement Transcript Definition
+
+The multiple-request measurement transcript is comprised of multiple `GET_MEASUREMENTS` requests and
```
thomas-fossati, line 38:
```suggestion
The multiple-request measurement transcript comprises multiple `GET_MEASUREMENTS` requests and
```

doc/standard_measurement_transcripts.md
```diff
@@ -0,0 +1,68 @@
...
+multiple `MEASUREMENTS` responses. Apart from `MEASUREMENTS(0)`, a single and unique measurement
+block is present in each `MEASUREMENTS` response.
+
+For SPDM 1.0 and 1.1, the byte buffer is
+```
+{GET_MEASUREMENTS(0), MEASUREMENTS(0), GET_MEASUREMENTS(1), MEASUREMENTS(1), ..., GET_MEASUREMENTS(N), MEASUREMENTS(N)}
+```
+
+For SPDM 1.2 and later, the byte buffer is
```
thomas-fossati, line 47: Does _“later”_ here mean that this structure can be guaranteed for any future version of the protocol?

steven-bellock: No. If SPDM 1.5 comes along and disrupts this, then it becomes
>For SPDM versions 1.2 through 1.4...

thomas-fossati: Thanks for the clarification.  I wonder if “later” could be characterised more precisely, then.  E.g.: _“[…] later (1.4 at the time of writing) […]”_.  If not, no worries.

doc/standard_measurement_transcripts.md
steven-bellock: ["comprised of"](https://www.merriam-webster.com/dictionary/comprised%20of) is correct.

doc/standard_measurement_transcripts.md
steven-bellock: ["comprised of"](https://www.merriam-webster.com/dictionary/comprised%20of) is correct.

doc/standard_measurement_transcripts.md
steven-bellock: I imagine this will be a living document with respect to newer specifications, but backwards compatibility will not be broken for already published one. If a Verifier says it supports the Single-request Measurement Transcript for SPDM 1.2, this document will not break that.

doc/standard_measurement_transcripts.md
thomas-fossati: sure, it’s just using more ASCII codepoints :-)

mcsmonk (approved):

mcsmonk (approved):

PR #3329, opened 2025-10-30
Remove deprecated interfaces and internal interfaces, and use modern …

coder-sageres:
…OpenSSL interfaces

PR #3345, opened 2025-11-11
libspdm: enable OSSL_STORE API support

itsManjeet:
This change enables proper integration of the OSSL_STORE API within openssllib. The CMakeLists.txt has been updated to include the necessary source files, and the dummy oss_store.c implementation only used in Windows for now until fixed LIBSPDM_OPENSSL_STDINT_WORKAROUND.

OSSL_STORE provide functionality to retrieve a keys and certs from a repository of any kind, addressable as a filename or as a URI from multiple kind of support hardware, like normal storage drive and/or secure vault like TPM.

https://docs.openssl.org/3.0/man7/ossl_store

### DIFF
--- a/os_stub/openssllib/CMakeLists.txt
+++ b/os_stub/openssllib/CMakeLists.txt
@@ -840,6 +840,22 @@ target_sources(openssllib
         openssl_gen/der_wrap_gen.c
         openssl_gen/params_idx.c
         rand_pool.c
-        ossl_store.c
         timer_wrapper.c
 )
+
+if (CMAKE_SYSTEM_NAME MATCHES "Windows")
+    target_sources(openssllib PRIVATE
+        ossl_store.c
+    )
+else()
+    target_sources(openssllib PRIVATE
+        openssl/crypto/store/store_err.c
+        openssl/crypto/store/store_init.c
+        openssl/crypto/store/store_lib.c
+        openssl/crypto/store/store_register.c
+        openssl/crypto/store/store_result.c
+        openssl/crypto/store/store_strings.c
+        openssl/providers/implementations/storemgmt/file_store.c
+        openssl/providers/implementations/storemgmt/file_store_any2obj.c
+    )
+endif()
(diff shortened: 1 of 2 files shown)


os_stub/openssllib/include/crt_support.h
```diff
@@ -240,15 +240,27 @@ typedef char *LIBSPDM_VA_LIST;
 /* Basic types mapping*/
 
 typedef size_t u_int;
-#if defined(__GNUC__) && !defined(__MINGW64__)
+#if defined(__APPLE__)
+#include <machine/types.h>
+typedef __darwin_time_t time_t;
+#elif defined(__GNUC__) && !defined(__MINGW64__)
 typedef size_t time_t; /* time_t is 4 bytes for 32bit machine and 8 bytes for 64bit machine */
```
steven-bellock, line 247: Pre-existing, but there will (probably) come a time when this is no longer true due to https://en.wikipedia.org/wiki/Year_2038_problem.

os_stub/openssllib/CMakeLists.txt
```diff
@@ -840,6 +840,22 @@ target_sources(openssllib
         openssl_gen/der_wrap_gen.c
         openssl_gen/params_idx.c
         rand_pool.c
-        ossl_store.c
         timer_wrapper.c
-)
\ No newline at end of file
+)
+
```
steven-bellock, line 845: This will have merge conflicts and seems to be incompatible with the proposed changes in https://github.com/DMTF/libspdm/pull/3329/files#diff-ea075edaec541911c99ad77f9ca43518c121157ea3cd69d3307654645c6fe2bc. @coder-sageres if these changes go in first would your pull request be able to maintain the functionality?

itsManjeet: Oh, nice those changes also include building openssl using ExternalProject hence using Openssl provided build system. If that thing is working as expected then We might not event need this change

PR #3350, opened 2025-11-11
Reset non-session transcripts in reset_context

### LINKED ISSUE #3347: `libspdm_reset_context` should also reset transcripts
Follow-up to https://github.com/DMTF/libspdm/issues/467 and https://github.com/DMTF/libspdm/pull/533.

#533 re-arranged the order of `libspdm_reset_message_*` and `libspdm_reset_context`. Presumably `libspdm_reset_context` should also reset transcripts and hashes. Also of note is that, in `GET_VERSION` / `VERSION`,  `libspdm_reset_message_buffer_via_request_code` is called which does nothing if the `RequestResponseCode` is `GET_VERSION`. https://github.com/DMTF/libspdm/blob/ab1af244faad4a4a26d02c784ae1916119313965/library/spdm_common_lib/libspdm_com_context_data.c#L1591-L1628

steven-bellock:
Fix #3347.

### DIFF
--- a/library/spdm_common_lib/libspdm_com_context_data.c
+++ b/library/spdm_common_lib/libspdm_com_context_data.c
@@ -3259,9 +3259,10 @@ void libspdm_reset_context(void *spdm_context)
 
     context = spdm_context;
 
-    /*Clear all info about last connection*/
+    /* Clear all information about previous connection. Local context information is preserved. */
 
-    /*need clear session info to free context before algo is zeroed.*/
+    /* Need to clear session information and message transcripts before negotiated algorithm
+     * information is cleared. */
     for (index = 0; index < LIBSPDM_MAX_SESSION_COUNT; index++)
     {
         libspdm_session_info_init(context,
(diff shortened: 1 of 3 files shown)


library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -3271,6 +3272,13 @@ void libspdm_reset_context(void *spdm_context)
                                   false);
     }
 
+    libspdm_reset_message_a(spdm_context);
+    libspdm_reset_message_d(spdm_context);
+    libspdm_reset_message_b(spdm_context);
+    libspdm_reset_message_c(spdm_context);
+    libspdm_reset_message_mut_b(spdm_context);
+    libspdm_reset_message_mut_c(spdm_context);
```
jyao1, line 3280: what about the non-session part of `libspdm_reset_message_m`, `libspdm_reset_message_e`, `libspdm_reset_message_encap_e`?

steven-bellock: Yes, those should be included as well.

jyao1 (approved):

PR #3367, opened 2025-11-19
Remove deprecated interfaces and internal interfaces, and use modern …

coder-sageres:
…OpenSSL interfaces


steven-bellock:
Ah yeah it looks like there are many languages and tools installed by default on a GitHub runner : https://github.com/actions/runner-images/blob/win25/20251102.77/images/windows/Windows2025-Readme.md. Perl is one of them.

CMakeLists.txt
```diff
@@ -700,6 +700,11 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
         set(CMAKE_EXE_LINKER_FLAGS "")
 
         set(CMAKE_C_LINK_EXECUTABLE "")
+    elseif(TOOLCHAIN STREQUAL "NONE")
```
itsManjeet: Hardcoding `CMAKE_LINKER` for `TOOLCHAIN=NONE` might impact yocto build since it will override the yocto provided cross compiler linker.

itsManjeet: @ArchanaKakani can you confirm it

coder-sageres: I tried to remove it, it was added during the debugging of the OpenSSL build, and I also think it shouldn't be hardcoded.

coder-sageres: This redundant code has been removed.

itsManjeet (approved):
Look good to me

coder-sageres:
> @coder-sageres please write down the prerequisite clearly in readme.
> 
> **You can create another commit or PR for readme update for perl.**
> 
> No need to merge into this commit.

OK

PR #3392, opened 2025-11-26
update script for PQC raw key generation

### LINKED ISSUE #3380: need update script for PQC raw key genration
Current PQC raw pub key is generated by https://github.com/jyao1/libspdm/blob/pqc_oqs/unit_test/gen_pqc_key/ (not checked in).
The traditional raw pub key is generated with https://github.com/DMTF/libspdm/blob/main/unit_test/sample_key/raw_data_key_gen.py.

We should add corresponding raw_data_key_gen_pqc.py.
Also update https://github.com/DMTF/libspdm/blob/main/unit_test/sample_key/readme.txt

Li-Aaron:
fix #3380

### DIFF
--- a/unit_test/sample_key/raw_data_key_gen_pqc.py
+++ b/unit_test/sample_key/raw_data_key_gen_pqc.py
@@ -1,0 +1,91 @@
+# /**
+#  *  Copyright Notice:
+#  *  Copyright 2025 DMTF. All rights reserved.
+#  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
+#  **/
+
+# automatic generate raw private key, raw public key and der format public key file for PQC after
+# regenerating new key in sample_key.
+# The generated raw data key for pqc will be consistent with the regenerating new key in sample_key.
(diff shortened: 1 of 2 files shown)


unit_test/sample_key/raw_data_key_gen_pqc.py
```diff
@@ -0,0 +1,91 @@
+# /**
+#  *  Copyright Notice:
+#  *  Copyright 2025 DMTF. All rights reserved.
+#  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
+#  **/
+
+# automatic generate raw private key, raw public key and der format public key file for PQC after
+# regenerating new key in sample_key.
+# The generated raw data key for pqc will be consistent with the regenerating new key in sample_key.
+# Run in linux: python raw_data_key_gen_pqc.py
```
steven-bellock: >Run in linux:

`python raw_data_key_gen_pqc.py` executes equally well in Windows and macOS. Can change it to
>To execute: python raw_data_key_gen_pqc.py

steven-bellock (approved):

PR #3394, opened 2025-11-26
Responder check for `BasicMutAuthReq`

### LINKED ISSUE #3389: Responder should check `BasicMutAuthReq`
`BasicMutAuthReq` says (emphasis mine)
>When mutual authentication is supported by both Responder and Requester, the Responder shall set this bit to indicate that the Responder wants to authenticate the identity of the Requester using the basic mutual authentication flow. **The Requester shall not set this bit in a basic mutual authentication flow.** See [Basic mutual authentication flow](https://github.com/DMTF/SPDM-WG/blob/master/docs/DSP0274/SPDMSpecification.md#basic-mutual-authentication). If mutual authentication is not supported, this bit shall be zero.

Of note is that this bit is never reserved, so Responder can check it for correctness. Requester already has this check in https://github.com/DMTF/libspdm/blob/c5e7386fa90b92acfc476a9c2cc7f861594ed20e/library/spdm_requester_lib/libspdm_req_challenge.c#L215-L223

steven-bellock:
Fix #3389.

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_encap_challenge.c
+++ b/library/spdm_responder_lib/libspdm_rsp_encap_challenge.c
@@ -144,6 +144,10 @@ libspdm_return_t libspdm_process_encap_response_challenge_auth(
         }
     }
 
+    if ((auth_attribute & SPDM_CHALLENGE_AUTH_RESPONSE_ATTRIBUTE_BASIC_MUT_AUTH_REQ) != 0) {
+        return LIBSPDM_STATUS_INVALID_MSG_FIELD;
+    }
+
     spdm_context->connection_info.peer_used_cert_chain_slot_id =
         spdm_context->encap_context.req_slot_id;
 


library/spdm_responder_lib/libspdm_rsp_encap_challenge.c
```diff
@@ -144,6 +144,10 @@ libspdm_return_t libspdm_process_encap_response_challenge_auth(
         }
     }
 
+    if ((auth_attribute & SPDM_CHALLENGE_AUTH_RESPONSE_ATTRIBUTE_BASIC_MUT_AUTH_REQ) != 0) {
+        return LIBSPDM_STATUS_INVALID_MSG_FIELD;
+    }
+
```
jyao1, line 150: why the responder need to check in encap flow?

This should be set by responder in first CHALLENGE_AUTH, not encap.

PR #3399, opened 2025-12-01
Remove basic mutual auth from context

steven-bellock:
Replace `LIBSPDM_DATA_BASIC_MUT_AUTH_REQUESTED` in the `spdm_context` with a hal call to the Integrator. This is informed by https://github.com/DMTF/libspdm/blob/main/doc/internal/encapsulated_flow.md#mutual-authentication.


doc/api/common_api.md
```diff
@@ -276,14 +276,6 @@ Enumeration value used for the `libspdm_set_data` and/or `libspdm_get_data` func
...
       limit is twice the `HeartbeatPeriod`.
-- `LIBSPDM_DATA_BASIC_MUT_AUTH_REQUESTED`
-    - Specifies whether the Responder requires basic mutual authentication with the Requester via
-      the `CHALLENGE / CHALLENGE_AUTH` messages.
-        - If `true` then Responder requires basic mutual authentication.
-        - If `false` then Responder does not require basic mutual authentication.
-    - Note that basic mutual authentication was deprecated in SPDM 1.2 and should also be considered
-      deprecated in SPDM 1.1. If a Responder requires mutual authentication then it should use
-      session-based mutual authentication via symmetric or asymmetric key exchange.
 - `LIBSPDM_DATA_MUT_AUTH_REQUESTED`
```
jyao1, line 279: Do we need to remove `LIBSPDM_DATA_MUT_AUTH_REQUESTED` as well?

steven-bellock: Yes, but that will be in a separate pull request.

PR #3412, opened 2025-12-04
Remove LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN

### LINKED ISSUE #1655: Remove LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN
In the spirit of #1491, let's remove `LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN` since the size of the response buffer is known.

Li-Aaron:
fix #1655

### DIFF
--- a/include/internal/libspdm_requester_lib.h
+++ b/include/internal/libspdm_requester_lib.h
@@ -172,30 +172,6 @@ libspdm_return_t libspdm_get_capabilities_with_supported_algs(libspdm_context_t 
  **/
 libspdm_return_t libspdm_negotiate_algorithms(libspdm_context_t *spdm_context);
 
-/**
- * This function sends GET_CERTIFICATE to get certificate chain in one slot from device.
- *
- * This function verify the integrity of the certificate chain.
- * root_hash -> Root certificate -> Intermediate certificate -> Leaf certificate.
- *
- * If the peer root certificate hash is deployed,
- * this function also verifies the digest with the root hash in the certificate chain.
- *
- * @param  spdm_context     A pointer to the SPDM context.
- * @param  session_id       Indicates if it is a secured message protected via SPDM session.
- *                          If session_id is NULL, it is a normal message.
- * @param  slot_id          The number of slot for the certificate chain.
- * @param  length           The size of cert chain block.
- * @param  cert_chain_size  On input, indicate the size in bytes of the destination buffer to store the digest buffer.
- *                          On output, indicate the size in bytes of the certificate chain.
- * @param  cert_chain       A pointer to a destination buffer to store the certificate chain.
- **/
-libspdm_return_t libspdm_get_certificate_choose_length(void *spdm_context,
-                                                       const uint32_t *session_id,
-                                                       uint8_t slot_id,
-                                                       uint16_t length,
-                                                       size_t *cert_chain_size,
-                                                       void *cert_chain);
 #if LIBSPDM_ENABLE_CAPABILITY_MEL_CAP
 /**
  * This function sends GET_MEASUREMENT_EXTENSION_LOG to get MEL from device.
--- a/library/spdm_requester_lib/libspdm_req_get_certificate.c
+++ b/library/spdm_requester_lib/libspdm_req_get_certificate.c
@@ -501,27 +501,14 @@ done:
     return status;
 }
 
-static libspdm_return_t libspdm_try_get_certificate(libspdm_context_t *spdm_context,
-                                                    const uint32_t *session_id,
-                                                    uint8_t slot_id,
-                                                    uint16_t length,
-                                                    size_t *cert_chain_size,
-                                                    void *cert_chain,
-                                                    const void **trust_anchor,
-                                                    size_t *trust_anchor_size)
-{
-    return libspdm_try_get_large_certificate(spdm_context, session_id, slot_id, length,
-                                             cert_chain_size, cert_chain,
-                                             trust_anchor, trust_anchor_size);
-}
-
 libspdm_return_t libspdm_get_certificate(void *spdm_context, const uint32_t *session_id,
                                          uint8_t slot_id,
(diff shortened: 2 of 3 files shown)


library/spdm_requester_lib/libspdm_req_get_certificate.c
```diff
@@ -123,6 +123,24 @@ static libspdm_return_t libspdm_try_get_large_certificate(libspdm_context_t *spd
...
+                                               SPDM_GET_CAPABILITIES_REQUEST_FLAGS_CHUNK_CAP,
+                                               SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_CHUNK_CAP);
+
+    /* use default max buffer length */
+    if (length == 0) {
+        if (chunk_enabled) {
+            length = spdm_context->local_context.capability.max_spdm_msg_size - rsp_msg_header_size;
+        } else {
+            length = spdm_context->local_context.capability.data_transfer_size - rsp_msg_header_size;
+        }
```
jyao1: I think we can always use `max_spdm_msg_size` as well in chunk_disable case, because they are same.

library/spdm_responder_lib/libspdm_rsp_encap_get_certificate.c
```diff
@@ -58,14 +61,29 @@ libspdm_return_t libspdm_get_encap_request_get_certificate(libspdm_context_t *sp
...
+        req_msg_length = spdm_context->local_context.capability.max_spdm_msg_size -
+                         sizeof(spdm_deliver_encapsulated_response_request_t) -
+                         rsp_msg_header_size;
+    } else {
+        /* certificate response is encapsulate in deliver encapsulated response msg */
+        req_msg_length = spdm_context->local_context.capability.data_transfer_size -
+                         sizeof(spdm_deliver_encapsulated_response_request_t) -
+                         rsp_msg_header_size;
+    }
+
```
jyao1: ditto, always use max_spdm_msg_size

library/spdm_responder_lib/libspdm_rsp_encap_get_certificate.c
```diff
@@ -151,6 +170,23 @@ libspdm_return_t libspdm_process_encap_response_certificate(
...
+            SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_CHUNK_CAP)) {
+        req_msg_length = spdm_context->local_context.capability.max_spdm_msg_size -
+                         sizeof(spdm_deliver_encapsulated_response_request_t) -
+                         rsp_msg_header_size;
+    } else {
+        /* certificate response is encapsulate in deliver encapsulated response msg */
+        req_msg_length = spdm_context->local_context.capability.data_transfer_size -
+                         sizeof(spdm_deliver_encapsulated_response_request_t) -
+                         rsp_msg_header_size;
+    }
```
jyao1: ditto

library/spdm_requester_lib/libspdm_req_get_certificate.c
Li-Aaron: `data_transfer_size` and `max_spdm_msg_size` are not same.

- The `data_transfer_size` is initialized from the `receiver_buffer_size` parameter provided to `libspdm_register_device_buffer_func`. It is then adjusted downward by `libspdm_register_transport_layer_func` to account for the protocol's `transport_header_size` and `transport_tail_size`.

- The `max_spdm_msg_size` is taken directly from the `max_spdm_msg_size` parameter provided to `libspdm_register_transport_layer_func`.

include/internal/libspdm_requester_lib.h
```diff
@@ -185,7 +185,7 @@ libspdm_return_t libspdm_negotiate_algorithms(libspdm_context_t *spdm_context);
  * @param  session_id       Indicates if it is a secured message protected via SPDM session.
  *                          If session_id is NULL, it is a normal message.
  * @param  slot_id          The number of slot for the certificate chain.
- * @param  length           LIBSPDM_MAX_CERT_CHAIN_BLOCK_LEN.
+ * @param  length           The size of cert chain block.
```
steven-bellock: Does `libspdm_get_certificate_choose_length` even need to exist? It is a non-public function and, from what I can tell, neither `spdm-emu` or `SPDM-Responder-Validator` call it.

library/spdm_requester_lib/libspdm_req_encap_certificate.c
```diff
@@ -27,6 +27,7 @@ libspdm_return_t libspdm_get_encap_response_certificate(void *spdm_context,
     uint32_t req_msg_header_size;
     uint32_t rsp_msg_header_size;
     size_t cert_chain_size;
+    uint32_t max_cert_chain_block_size;
```
steven-bellock, line 30: Can move this declaration to line `127`.

library/spdm_responder_lib/libspdm_rsp_certificate.c
```diff
@@ -27,6 +27,7 @@ libspdm_return_t libspdm_get_response_certificate(libspdm_context_t *spdm_contex
     uint32_t req_msg_header_size;
     uint32_t rsp_msg_header_size;
     size_t cert_chain_size;
+    uint32_t max_cert_chain_block_size;
```
steven-bellock, line 30: Can move this declaration to line `166`.

library/spdm_requester_lib/libspdm_req_encap_certificate.c
jyao1: Why do we have to move?

I think nothing wrong to define it here.

library/spdm_responder_lib/libspdm_rsp_certificate.c
jyao1: Why do we have to move?

I think nothing wrong to define it here.

Issue #3164, opened 2025-08-12
Need a build-time flag to ignore time validity of certificates


lordaule:
libspdm currently defines OPENSSL_CHECK_TIME in os_stub/openssllib/include/openssl/configuration.h

And this is used in os_stub/cryptlib_openssl/pk/x509.c to check the time validity of certificates.  This poses a problem for integrators taking libspdm, as they have to modify the source code if they want to bypass time checks.

This is logically reversed from OPENSSL_IGNORE_CRITICAL, which is un-defined by default.  To enable OPENSSL_IGNORE_CRITICAL, integrators can pass a cmake flag using -DX509_IGNORE_CRITICAL=ON

We need a similar X509_IGNORE_TIME build flag that can be passed to cmake, rather than having to edit any source code in the libspdm library itself.

steven-bellock:
I think the rationale is that if an endpoint is able to run OpenSSL then it has access to a realtime clock. You can always use `libspdm_register_verify_spdm_cert_chain_func` to bypass libspdm's checks and use your own.

As for an enhancement, it seems like this could be a runtime check in `libspdm_x509_common_certificate_check` using a bitmask to disable certain checks, like time, serial number, etc.

PR #3422, opened 2025-12-09
Remove `BASIC_CONSTRAINTS_STRING_FALSE_CASE2`

### LINKED ISSUE #3419: `BASIC_CONSTRAINTS_STRING_FALSE_CASE2` should (probably) not exist
From https://github.com/DMTF/SPDM-WG/issues/4416.

In a DER-encoded certificate chain, if a sequence element is equal to its default value then it is not present in the final encoding. However, `BASIC_CONSTRAINTS_STRING_FALSE_CASE2` violates this when checking the leaf certificate.

https://github.com/DMTF/libspdm/blob/e9cc874d4ff2163a18619b8c99a8ed3848b72aa4/library/spdm_crypt_lib/libspdm_crypt_cert.c#L162-L163

~Will wait first for resolution of https://github.com/DMTF/SPDM-WG/issues/4416 before fixing this, since it is still an open issue.~

steven-bellock:
Fix #3419 and enforce strict DER checks.

### DIFF
--- a/library/spdm_crypt_lib/libspdm_crypt_cert.c
+++ b/library/spdm_crypt_lib/libspdm_crypt_cert.c
@@ -156,14 +156,13 @@
 #define ALGO_SLHDSA_SHAKE_256F_OID {0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x03, 0x1F}
 #endif
 
-/*leaf cert basic_constraints false case1: CA: false and CA object is excluded */
-#define BASIC_CONSTRAINTS_STRING_FALSE_CASE1 {0x30, 0x00}
+/* Leaf certificate basic constraints with cA field set to false. According to RFC5280, false is the
+ * default value, and according to DER encoding a sequence item with a default value must not be
+ * encoded. */
+#define BASIC_CONSTRAINTS_CA_FALSE {0x30, 0x00}
 
-/*leaf cert basic_constraints false case2: CA: false */
-#define BASIC_CONSTRAINTS_STRING_FALSE_CASE2 {0x30, 0x03, 0x01, 0x01, 0x00}
-
-/*leaf cert basic_constraints true case: CA: true */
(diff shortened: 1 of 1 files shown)


library/spdm_crypt_lib/libspdm_crypt_cert.c
```diff
@@ -1147,17 +1145,10 @@ static bool libspdm_verify_leaf_cert_basic_constraints(const uint8_t *cert, size
...
-
-    if ((len == sizeof(basic_constraints_false_case2)) &&
+    if ((len == sizeof(basic_constraints_false_case)) &&
         (libspdm_consttime_is_mem_equal(cert_basic_constraints,
-                                        basic_constraints_false_case2,
-                                        sizeof(basic_constraints_false_case2)))) {
+                                        basic_constraints_false_case,
+                                        sizeof(basic_constraints_false_case)))) {
         return true;
     }
```
jcahill0, line 1154: If I'm reading commit a9e42b0d691076cacbb121ca4ea807b464c4384c correctly, then it appears this would trigger a "false"/failure of any existing/previous deployed leaf certificate with the Basic Constraints OID's CA:FALSE (len!=sizeof(basic_constraints_false_case fallout case), which libspdm_verify_leaf_cert_basic_constraints() used to pass.

Could the empty CA:FALSE check be made optional for SPDM Requesters that choose to handle this case?

steven-bellock: >Could the empty CA:FALSE check be made optional for SPDM Requesters that choose to handle this case?

https://github.com/DMTF/libspdm/issues/3164#issuecomment-3181047348 proposes to let the Integrator selectively disable specific checks, so when/if that gets implemented then they would be able to disable the `Basic Constraints` check if they so desire.

steven-bellock:
Re-opening, as the SPDM WG has confirmed that the certificates need to be DER-encoded.

PR #3423, opened 2025-12-09
Move `local_used_cert_chain_slot_id` to `session_info`

### LINKED ISSUE #3405: `connection_info.local_used_cert_chain_slot_id` should not exist
This issue is similar to https://github.com/DMTF/libspdm/issues/2381.

libspdm currently stores `KEY_EXCHANGE.SlotID` in `connection_info.local_used_cert_chain_slot_id`. This is then referenced in later messages to calculate key exchange transcripts. However the `SlotID` is tied to the secure session and not to the connection, and a Requester can initiate multiple secure sessions using different `SlotID`s. As such, it belongs in the session info.

steven-bellock:
Fix #3405 and move `local_used_cert_chain_slot_id` from `connection_info` to `session_info`.

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -150,11 +150,6 @@ typedef struct {
     libspdm_peer_used_cert_chain_t peer_used_cert_chain[SPDM_MAX_SLOT_COUNT];
     uint8_t peer_used_cert_chain_slot_id;
 
-    /* Local Used CertificateChain (for responder, or requester in mut auth) */
-    const uint8_t *local_used_cert_chain_buffer;
-    size_t local_used_cert_chain_buffer_size;
-    uint8_t local_used_cert_chain_slot_id;
-
     /* Specifies whether the cached negotiated state should be invalidated. (responder only)
      * This is a "sticky" bit wherein if it is set to 1 then it cannot be set to 0. */
     uint8_t end_session_attributes;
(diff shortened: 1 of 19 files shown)


unit_test/test_spdm_responder/finish_rsp.c
```diff
@@ -2035,7 +2035,6 @@ void rsp_finish_rsp_case18(void **state)
 
     spdm_context->encap_context.req_slot_id = 0xFF;
     spdm_context->connection_info.peer_used_cert_chain_slot_id = 0xFF;
-    spdm_context->connection_info.local_used_cert_chain_slot_id = 0xFF;
```
jyao1, line 2038: Question: how about `peer_used_cert_chain_slot_id` ?
should we remove it as well?

include/internal/libspdm_common_lib.h
```diff
@@ -477,6 +477,8 @@ typedef struct {
     /* Register for the last KEY_UPDATE token and operation (responder only)*/
     spdm_key_update_request_t last_key_update_request;
     void *secured_message_context;
+    /* Only present in session info as it is currently only used within a secure session. */
+    uint8_t local_used_cert_chain_slot_id;
```
jyao1, line 476: should we remove `local_used_cert_chain_slot_id` from `libspdm_connection_info_t` ?

steven-bellock: Yes. Thanks for catching that.

unit_test/test_spdm_responder/finish_rsp.c
steven-bellock: That can be used both inside and outside of a session, so it will stay inside `connection_info`. I will add it to `session_info` in a later pull request.

include/library/spdm_common_lib.h
```diff
@@ -898,6 +896,7 @@ bool libspdm_get_local_cert_chain_buffer(void *spdm_context,
  * @retval false Local used certificate chain data without spdm_cert_chain_t header is not found.
  **/
 bool libspdm_get_local_cert_chain_data(void *spdm_context,
+                                       uint8_t slot_id,
```
steven-bellock, line 899: `libspdm_get_local_cert_chain_data` is an unused function, so I am also fine if the entire function is removed.

PR #3450, opened 2025-12-24
provide libspdm_send_receive_spdm_data() API for chunking

### LINKED ISSUE #3448: provide libspdm_send_receive_spdm_data() API for chunking
See https://github.com/DMTF/SPDM-Responder-Validator/issues/184

We need libspdm provide `libspdm_send_receive_spdm_data()` API, which supports chunking.
`libspdm_send_receive_data` may call `libspdm_send_spdm_request` and `libspdm_receive_spdm_response`. (like what `libspdm_try_vendor_send_request_receive_response` does)

To remove duplication and for testing, we may let `libspdm_try_vendor_send_request_receive_response` call the new `libspdm_send_receive_spdm_data()` .

reference: https://github.com/DMTF/libspdm/issues/3343

Li-Aaron:
fix #3448

### DIFF
--- a/include/library/spdm_requester_lib.h
+++ b/include/library/spdm_requester_lib.h
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2021-2025 DMTF. All rights reserved.
+ *  Copyright 2021-2026 DMTF. All rights reserved.
  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
  **/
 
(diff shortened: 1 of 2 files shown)


include/library/spdm_requester_lib.h
```diff
@@ -686,6 +686,32 @@ libspdm_return_t libspdm_send_receive_data(void *spdm_context,
...
+ * This API handles SPDM chunking.
+ *
+ *
+ * @param  spdm_context    A pointer to the SPDM context.
+ * @param  session_id      Indicates if it is a secured message protected via SPDM session.
+ *                         If session_id is NULL, it is a normal message.
+ *                         If session_id is NOT NULL, it is a secured message.
+ * @param  request         A pointer to the request data.
+ * @param  request_size    Size in bytes of the request data.
+ * @param  response        A pointer to the response data.
```
steven-bellock, line 702: Can `request` and `response` point to the same buffer? Ie, the request message would be overwritten by the response message when `libspdm_send_receive_spdm_data` returns?

Li-Aaron: Yes, `request` and `response` can be same buffer.
The API copy the data from `request` buffer to `sender_buffer` before send request; and copy data from `receiver_buffer` to `response` buffer after recieve full response.

steven-bellock: Should probably specify that that is an option.

jyao1: what do you mean "an option" ?

steven-bellock: Assuming that we want to explicitly support this at the API level, then state
>The request and response message buffers can be the same buffer.

steven-bellock (approved):

Issue #3082, opened 2025-05-29
Add self-test for SPDM Key Schedule.


jyao1:
This is for FIPS preparation.

steven-bellock:
See also https://github.com/usnistgov/ACVP-Server/pull/405.

steven-bellock:
>The expected output is: {requester,responder}_{handshake,data}_secret, export_master_secret

Why stop at the secrets? Why not go to keys and IVs?

xiaoyuruan:
>The purpose is to test spdm_key_schedule function is good, according to SPDM specification.

OK. So the selftest for key schedule is NOT a FIPS selftest; it's an SPDM selftest. Then do not name it fips_selftest (current PR has "fips_selftest_key_schedule..." as function name), to avoid confusion.

I agree with @steven-bellock. Now that it's SPDM selftest, we can decide the scope, why not test all the way to keys and IVs.

BTW, I would recommend creating a "selftest spec" to specify requirements of all selftests - some due to FIPS requirement, some as enhancement of SPDM itself, and scopes of each test (such as testing only master secret or keys and IVs.

PR #3463, opened 2026-01-07
extract key schedule logic to spdm_crypt_lib for FIPS CAVP test

Li-Aaron:
Fix: #3082 

In this PR
- Extract key schedule logic from spdm_secure_message_lib to new APIs in spdm_crypt_lib
  - `libspdm_generate_handshake_key`
  - `libspdm_generate_data_key`

EMU test: https://github.com/DMTF/spdm-emu/pull/462

### DIFF
--- a/include/library/spdm_crypt_lib.h
+++ b/include/library/spdm_crypt_lib.h
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2021-2025 DMTF. All rights reserved.
+ *  Copyright 2021-2026 DMTF. All rights reserved.
  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
  **/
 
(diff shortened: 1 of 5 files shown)


include/library/spdm_crypt_lib.h
```diff
@@ -1581,6 +1581,99 @@ bool libspdm_kem_decapsulate(uint32_t kem_alg, void *context,
...
+    bool fips_test);
+
+/**
+ * This function generates SPDM DataKey.
+ *
+ * @param spdm_version                   The SPDM version number.
+ * @param secret                         Pointer to the input secret used for key derivation.
+ *                                       For DHE/KEM, input is handshake_secret
+ *                                       For PSK, input is psk_hint
+ *                                       For PSK (FIPS test), input is psk
```
jyao1: I have 2 high level directions:
1) `psk_hint` is not secret. I prefer split it from `secret`.
2) `secret` is `handshake_secret` in DHE/KEM case, but `psk` (as `shared secret`) in PSK case. I prefer to align them to the same level - either to use `handshake_secret` for both or `shared_secret` for both.

jyao1:
@steven-bellock would you please review?

PR #3466, opened 2026-01-08
Merge libspdm_get_certificate_ex with libspdm_get_certificate_choose_…

### LINKED ISSUE #3426: `GET_CERTIFICATE` public API
From https://github.com/DMTF/libspdm/issues/1655.

There are currently three public APIs to fetch a certificate chain. The `length` parameter is no longer needed now that libspdm automatically sizes the request. The three APIs can then be condensed into one

```
libspdm_return_t libspdm_get_certificate(void *spdm_context,
                                         const uint32_t *session_id,
                                         uint8_t slot_id,
                                         size_t *cert_chain_size,
                                         void *cert_chain,
                                         const void **trust_anchor,
                                         size_t *trust_anchor_size);
```
The Integrator can set `trust_anchor` to `NULL` if they so desire.

ShitalJumbad:
fix #3426

### DIFF
--- a/include/library/spdm_requester_lib.h
+++ b/include/library/spdm_requester_lib.h
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2021-2025 DMTF. All rights reserved.
+ *  Copyright 2021-2026 DMTF. All rights reserved.
  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
  **/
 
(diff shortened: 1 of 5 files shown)


include/library/spdm_requester_lib.h
```diff
@@ -108,63 +108,36 @@ libspdm_return_t libspdm_get_certificate(void *spdm_context,
...
-                                                          const void **trust_anchor,
-                                                          size_t *trust_anchor_size);
+libspdm_return_t libspdm_get_certificate_ex(void *spdm_context,
+                                           const uint32_t *session_id,
+                                           uint8_t slot_id,
+                                           uint32_t length,
+                                           size_t *cert_chain_size,
+                                           void *cert_chain,
+                                           const void **trust_anchor,
+                                           size_t *trust_anchor_size);
```
steven-bellock: These lines need to moved one space to the right. The style checker is currently broke but it will be up with https://github.com/DMTF/libspdm/pull/3469.

ShitalJumbad: Done.

steven-bellock (approved):

PR #3473, opened 2026-01-12
return slot storage size when SLOT_SIZE_REQUESTED

### LINKED ISSUE #3166: In case of SLOT_SIZE_REQUESTED, Responder should return slot storage size, instead of current cert chain size.

Li-Aaron:
fix #3166

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_certificate.c
+++ b/library/spdm_responder_lib/libspdm_rsp_certificate.c
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2021-2025 DMTF. All rights reserved.
+ *  Copyright 2021-2026 DMTF. All rights reserved.
  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
  **/
 #include "internal/libspdm_responder_lib.h"
@@ -198,6 +198,13 @@ libspdm_return_t libspdm_get_response_certificate(libspdm_context_t *spdm_contex
     if (use_large_cert_chain) {
         spdm_response->header.param1 |= SPDM_CERTIFICATE_RESPONSE_LARGE_CERT_CHAIN;
     }
+
+#if LIBSPDM_ENABLE_CAPABILITY_SET_CERT_CAP
+    if ((spdm_request->header.spdm_version >= SPDM_MESSAGE_VERSION_13) &&
+        (spdm_request->header.param2 & SPDM_GET_CERTIFICATE_REQUEST_ATTRIBUTES_SLOT_SIZE_REQUESTED)) {
+        remainder_length = libspdm_get_cert_chain_slot_storage_size(spdm_context, slot_id);
+    }
+#endif /* LIBSPDM_ENABLE_CAPABILITY_SET_CERT_CAP */
 
     if (use_large_cert_chain) {
         spdm_response->portion_length = 0;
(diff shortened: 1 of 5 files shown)


library/spdm_requester_lib/libspdm_req_encap_certificate.c
```diff
@@ -166,6 +167,21 @@ libspdm_return_t libspdm_get_encap_response_certificate(void *spdm_context,
         spdm_response->header.param1 |= SPDM_CERTIFICATE_RESPONSE_LARGE_CERT_CHAIN;
     }
 
+    if ((spdm_request->header.spdm_version >= SPDM_MESSAGE_VERSION_13) &&
+        (spdm_request->header.param2 & SPDM_GET_CERTIFICATE_REQUEST_ATTRIBUTES_SLOT_SIZE_REQUESTED)) {
```
jyao1: param2 is set above to 0. How to set `SPDM_GET_CERTIFICATE_REQUEST_ATTRIBUTES_SLOT_SIZE_REQUESTED`?

steven-bellock: `SlotSizeRequested` should be reserved when the Responder is sending it. There is no reason for the Responder to know this value, as a Requester does not support `SET_CERT_CAP`. I will file an issue against the specification.

steven-bellock: Filed https://github.com/DMTF/SPDM-WG/issues/4443.

steven-bellock: @Li-Aaron we discussed this in the work group, and the consensus was that the easiest thing is for the Requester to just report the size of the certificate chain.

Li-Aaron: I remove the change for requester, only fix responder.

library/spdm_responder_lib/libspdm_rsp_certificate.c
```diff
@@ -199,6 +200,21 @@ libspdm_return_t libspdm_get_response_certificate(libspdm_context_t *spdm_contex
...
+        if (use_large_cert_chain) {
+            spdm_response->portion_length = 0;
+            spdm_response->remainder_length = 0;
+            spdm_response->large_portion_length = 0;
+            spdm_response->large_remainder_length = slot_storage_size;
+        } else {
+            spdm_response->portion_length = 0;
+            spdm_response->remainder_length = (uint16_t)slot_storage_size;
+        }
+        return LIBSPDM_STATUS_SUCCESS;
```
jyao1: I think this is still need to be in `message_b`  transcript.

@steven-bellock , what do you think?

steven-bellock: Yes, that allows for integrity protection.

include/hal/library/responder/getcertlib.h
```diff
@@ -0,0 +1,28 @@
...
+ **/
+
+#ifndef RESPONDER_GETCERTLIB_H
+#define RESPONDER_GETCERTLIB_H
+
+#include "hal/base.h"
+#include "internal/libspdm_lib_config.h"
+#include "industry_standard/spdm.h"
+
+#if LIBSPDM_ENABLE_CAPABILITY_CERT_CAP
```
steven-bellock: For the most part `SlotSizeRequested` is only used in conjunction with `SET_CERTIFICATE` and `SLOT_MANAGEMENT`. So I would say this function should be in https://github.com/DMTF/libspdm/blob/main/include/hal/library/responder/setcertlib.h. If an Integrator does not support `SET_CERTIFICATE` at compile time (`LIBSPDM_ENABLE_CAPABILITY_SET_CERT_CAP == 0`) then libspdm can return the size of the provisioned certificate chain.

steven-bellock (approved):

Issue #3468, opened 2026-01-09
SupportedAlgorithms bit not properly set or checked


wmaroneAMD:
The following issue was brought to my attention:

In libspdm_get_response_capabilities, bit 1 of param1 is never set in the response even if supported_algs_requested is true.

https://github.com/DMTF/libspdm/blob/3b03272963e31f36ed2a3ad4ff5b73914de3d2a4/library/spdm_responder_lib/libspdm_rsp_capabilities.c#L318

 In libspdm_try_get_capabilities, the request bit is checked in the original request but the response is not validated:

https://github.com/DMTF/libspdm/blob/3b03272963e31f36ed2a3ad4ff5b73914de3d2a4/library/spdm_requester_lib/libspdm_req_get_capabilities.c#L429-L431

As such the response that comes back is in violation of the spec. If this understanding is correct, assign this issue to me and I can open up a PR with a fix.

PR #3474, opened 2026-01-12
Fix SupportedAlgorithms bit which is not properly set or checked

wmaroneAMD:
Addresses https://github.com/DMTF/libspdm/issues/3468 ensuring bits for SupportedAlgorithms are set and checked properly.

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_get_capabilities.c
+++ b/library/spdm_requester_lib/libspdm_req_get_capabilities.c
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2021-2025 DMTF. All rights reserved.
+ *  Copyright 2021-2026 DMTF. All rights reserved.
  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
  **/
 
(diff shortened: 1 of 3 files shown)


library/spdm_requester_lib/libspdm_req_get_capabilities.c
```diff
@@ -428,7 +428,8 @@ static libspdm_return_t libspdm_try_get_capabilities(libspdm_context_t *spdm_con
     /* Copy algorithms if requested and received */
     if (supported_algs != NULL &&
         spdm_response->header.spdm_version >= SPDM_MESSAGE_VERSION_13 &&
-        (spdm_request->header.param1 & SPDM_GET_CAPABILITIES_REQUEST_PARAM1_SUPPORTED_ALGORITHMS)) {
+        (spdm_request->header.param1 & SPDM_GET_CAPABILITIES_REQUEST_PARAM1_SUPPORTED_ALGORITHMS) &&
+        (spdm_response->header.param1 & SPDM_CAPABILITIES_RESPONSE_PARAM1_SUPPORTED_ALGORITHMS)) {
```
steven-bellock, line 432: `supported_algs_length` should be `0` if nothing is copied to `supported_algs`.

nitinklohar: Updated the code to set `supported_algs_length` to `0` if nothing is copied to `supported_algs`.

steven-bellock (approved):

PR #3491, opened 2026-01-30
Fix cryptlib_openssl linking for ENABLE_BINARY_BUILD mode

Jino26:


Jino26:
Since spdm-emu submodule reference libspdm commits, PR-only or dangling commits break clean Yocto/BitBake builds.

PR #3497, opened 2026-02-10
Update version for 3.8.2 release

steven-bellock:


steven-bellock:
https://github.com/DMTF/libspdm/pull/3494 needs to first be merged.

Issue #3393, opened 2025-11-26
Intermittent failure in `test_crypt`


steven-bellock:
See https://github.com/DMTF/libspdm/actions/runs/19687979037/job/56397683686?pr=3391 and
```
Crypto DH Engine Testing:
- Context1 ... Context2 ... Generate key1 ... Generate key2 ... Compute key1 ... [Fail]
Error: Process completed with exit code 1.
```
This is on Ubuntu with OpenSSL. Re-running the test resulted in a pass.

steven-bellock:
Also https://github.com/DMTF/libspdm/actions/runs/19934705026/job/57156199313?pr=3412.

steven-bellock:
Also a failure with OpenSSL in `test_spdm_fips`. 
```
[==========] spdm_crypt_lib_tests: Running 1 test(s).
[ RUN      ] libspdm_test_fips
[  ERROR   ] --- status
[   LINE   ] --- D:\a\libspdm\libspdm\unit_test\test_spdm_fips\test_spdm_fips.c:29: error: Failure!
[  PASSED  ] 0 test(s).
[  FAILED  ] libspdm_test_fips
[==========] spdm_crypt_lib_tests: 1 test(s) run.
[  FAILED  ] spdm_crypt_lib_tests: 1 test(s), listed below:
[  FAILED  ] libspdm_test_fips
```

steven-bellock:
`test_crypt` is still intermittently failing: https://github.com/DMTF/libspdm/actions/runs/20623880006/job/59230944464#step:18:8.

steven-bellock:
https://github.com/DMTF/libspdm/pull/3522 has been merged so we will see if this resolves some or all of this issue.

PR #3522, opened 2026-02-19
Fixes: Coverity CID 900514

itsManjeet:
Coverity reported a potential array overrun in
get_evp_compatible_public_key() when converting a raw EC public key (X || Y) into the SEC1 uncompressed format required by OpenSSL EVP (0x04 || X || Y).

The conversion intentionally increases the public key size by one byte to prepend the 0x04 uncompressed point prefix. This size adjustment may have triggered a false positive analysis warning.


steven-bellock:
It looks like this commit has also fixed https://github.com/DMTF/libspdm/issues/3393.

PR #3541, opened 2026-03-05
get element from opaque data with element id

Li-Aaron:
Ref: #2936


include/internal/libspdm_common_lib.h
```diff
@@ -1158,24 +1158,49 @@ bool libspdm_calculate_l1l2_hash(libspdm_context_t *spdm_context,
 #endif /* LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT */
 
 /**
- * Get element from multi element opaque data by element id.
+ * Get element from multi element opaque data by element id and element index
+ *
+ * This function should be called in
```
steven-bellock: Pre-existing, but sentence is incomplete. Can just delete.

steven-bellock (changes requested):

PR #3580, opened 2026-04-02
Refactor libspdm interfaces to decouple spdm-emu

itsManjeet:
This PR refactors parts of the `libspdm` interface to improve modularity and enable proper decoupling between `spdm-emu` and `libspdm`. Currently, `spdm-emu` relies on internal `libspdm` components, leading to tight coupling and making it harder to use external `libspdm` (preferred in external build system like yocto)

This change introduces explicit interface boundaries by exporting required components instead of relying on internal linkage.

Exported `cryptlib_ext`, `spdm_crypt_lib_ext`, `libspdm_device_secret_lib` to allow used by external tools.


jyao1:
I think we purposely did not include them in libspdm, because they are NOT consumed by libspdm library.

The proposal is to move to the os_stub/include.

itsManjeet:
> I think we purposely did not include them in libspdm, because they are NOT consumed by libspdm library.
> 
> The proposal is to move to the os_stub/include.

Got it, that makes sense. I’ve placed the files under os_stub/include accordingly:

spdm_crypt_ext_lib/spdm_crypt_ext_lib.h → library/spdm_crypt_ext_lib.h
spdm_crypt_ext_lib/cryptlib_ext.h → hal/library/cryptlib_ext.h
spdm_device_secret_lib_internal.h → internal/libspdm_device_secret_lib.h

jyao1 (approved):

Issue #3416, opened 2025-12-05
TCBInfo format issue


reasly:
Current 
/os_stub/spdm_cert_verify_callback_sample/spdm_cert_verify_callback.c 
has the following code:
```    /*get DiceTcbInfo*/
    result = libspdm_asn1_get_tag(&ptr, end, &obj_len,
                                  LIBSPDM_CRYPTO_ASN1_SEQUENCE | LIBSPDM_CRYPTO_ASN1_CONSTRUCTED);
    if (!result) {
        *spdm_get_dice_tcb_info_size = 0;
        return false;
    }
    result = libspdm_asn1_get_tag(&ptr, end, &obj_len,
                                  LIBSPDM_CRYPTO_ASN1_SEQUENCE | LIBSPDM_CRYPTO_ASN1_CONSTRUCTED);
    if (!result) {
        *spdm_get_dice_tcb_info_size = 0;
        return false;
    }
```

The code requires the extension has "SEQUENCE (SEQUENCE (xxx))" structure. 
However, according to DICE Attestation Architecture (https://trustedcomputinggroup.org/wp-content/uploads/DICE-Attestation-Architecture-v1.2_pub.pdf) :
```
tcg-dice-TcbInfo OBJECT IDENTIFIER ::= {tcg-dice tcbinfo(1)}
DiceTcbInfo ::== SEQUENCE { ...}
```
The DICETCBInfo should have only one layer of SEQUENCE.

```
tcg-dice-MultiTcbInfo OBJECT IDENTIFIER ::= {tcg-dice multi-tcbinfo(5)}
DiceTcbInfoSeq ::= SEQUENCE SIZE (1..MAX) OF DiceTcbInfo
```
Only MultiDICETCBInfo can have SEQUENCE(SEQUENCE(...)) structure.

Please help chech if our understanding is correct.

steven-bellock:
That is consistent with what @jyao1 said in https://github.com/DMTF/libspdm/pull/2672#issuecomment-2084144021 :
>Only one reference TcbInfo entry is provided by the integrator. (Multiple reference TcbInfo is NOT supported in this PR.)

steven-bellock:
Marking this as not-a-bug, as the code just checks if `DiceTCBInfo` exists twice.

steven-bellock:
@xiaoyuruan and @richkong88 did some testing and confirmed that this is now fixed. Can we add a test to the repository?

PR #3583, opened 2026-04-02
Fix issue #3416

xiaoyuruan:
Fix issue #3416
Remove the lookup for the second SEQUENCE


steven-bellock (approved):

jyao1:
Please clarify how you perform the test to ensure the change is correct.

xiaoyuruan:
@reasly Can you review and potentially test the fix if possible? Thanks.

steven-bellock:
The original issue was marked as not-as-bug, just redundant code. @jyao1 will review with @xiaoyuruan.

jyao1:
It seems OK to just remove redundant code.

PR #3585, opened 2026-04-03
Fix security issue in measurement extension log

steven-bellock:
This also updates the release tag 3.8.2 to include this fix.


steven-bellock:
https://github.com/steven-bellock/libspdm/actions/runs/23956757515 shows that build and test passed for this pull request.

PR #3595, opened 2026-04-10
Add libspdm_parse_measurement_transaction_data function

### LINKED ISSUE #3566: Need a method to decode the SPDM transaction log output from the SPDMGetSignedMeasurements Action in Redfish.
libspdm currently allows for decoding the get measurements request/response itself, but doesn't include a method for the whole transaction log (VCA + GET_MEASUREMENTS + MEASUREMENTS) as returned by Redfish Action.

I would like to add a method such as this:
```
libspdm_return_t libspdm_parse_measurement_transaction_data(void* data, size_t data_size, spdm_measurement_block_common_header_t** measurements, size_t* measurement_count);
```

That would abstract the clients from each having to reinvent parts of libspdm and keep those implementations up to date with new versions.

pboyd04:
Fixes #3566


library/spdm_requester_lib/libspdm_req_parse_measurement_transaction_data.c
```diff
@@ -0,0 +1,470 @@
+/**
+ *  Copyright Notice:
+ *  Copyright 2021-2026 DMTF. All rights reserved.
```
steven-bellock, line 3: `2021-2026` -> `2026`

library/spdm_requester_lib/libspdm_req_parse_measurement_transaction_data.c
```diff
@@ -0,0 +1,470 @@
...
+ *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
+ **/
+
+#include "internal/libspdm_requester_lib.h"
+
+/* Forward declaration - resolved at link time from the integrator's malloc implementation. */
+extern void *allocate_pool(size_t size);
```
steven-bellock, line 10: Core libspdm libraries do not dynamically allocate memory. If memory is needed then it is provided by the API caller. See, for example, `libspdm_get_certificate()`.

steven-bellock (changes requested):

ayj (commented):
Question on how libspdm_parse_measurements_response should handle presence of signature.

ayj (commented):
Should the parser validate signature field presence or expected size? (Lines 237-241)

In `libspdm_parse_measurements_response`:

```c
237:     if (signature_requested) {
238:         /* Signature is the last field. We do not need to know its size
239:          * since this must be the final response in the transcript. */
240:         *next_offset_out = data_size;
241:     }
```

While this helper only parses the transaction log structure and does not cryptographically verify the signature, if the input buffer is truncated right before the signature field (`ptr == data_size` at line 237), the function currently returns `LIBSPDM_STATUS_SUCCESS`. 

* Should we add a check here to verify that the remaining buffer is non-empty (`ptr < data_size`) to ensure structural correctness?
* Or, since the VCA phase is walked at the start of the buffer, should we extract the negotiated measurement signature algorithm from the `ALGORITHMS` response to strictly validate that `data_size - ptr == expected_signature_size`?

jyao1:
@pboyd04. allocate_pool() is NOT allowed in currently libspdm core library. Please fix it according to the feedback.

PR #3606, opened 2026-04-22
Harden SPDM 1.4 `PSK_FINISH` opaque handling with core unit test coverage

### LINKED ISSUE #3592: PSK_FINISH/PSK_FINISH_RSP (SPDM 1.4): Opaque-Length Validation and Response Completeness Gaps
In the SPDM 1.4 PSK_FINISH flow, opaque-data handling is only partially implemented. Validation and completeness around opaque-length processing are inconsistent between request and response paths. This creates a protocol-completeness gap and may allow malformed opaque-length patterns to be handled less strictly than intended.

Observed Behavior

Opaque-length processing in PSK_FINISH/PSK_FINISH_RSP is present but not fully hardened for strict 1.4 handling.
Responder behavior is limited to minimal/empty opaque response payload handling, rather than full opaque-data response support.
Validation behavior is not fully symmetric between requester and responder for opaque-length constraints.

Expected Behavior

Full SPDM 1.4-compliant opaque-length handling for PSK_FINISH/PSK_FINISH_RSP.
Explicit and consistent length-bound validation on both requester and responder paths.
Deterministic handling of non-empty opaque-data response paths where supported.

Impact

Protocol completeness gap in a 1.4 session-finalization path.
Potential interoperability issues with peers that use broader valid opaque-data patterns.
Reduced parser robustness under malformed or edge-case opaque-length inputs.
Severity
Medium (functional/compliance hardening; security relevance depends on malformed-input threat model and deployment policy).

Suggested Fix

Add explicit opaque-length upper-bound and structural validation in responder-side request parsing.
Align requester/responder validation symmetry for opaque-length rules.
Implement or explicitly gate non-empty opaque-data response behavior for 1.4.
Add unit tests covering valid/invalid opaque-length boundary scenarios and non-empty opaque response cases.

czwolak:
Fix #3592.

- Harden requester/responder SPDM 1.4 PSK_FINISH opaque validation paths.
- Align requester SPDM 1.4 opaque format negotiation setup in UT.
- Keep core opaque boundary/error coverage: req cases 18/19 and rsp cases 17/18.
- Remove optional requester cases 20/21 and related harness scaffolding.

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_psk_finish.c
+++ b/library/spdm_requester_lib/libspdm_req_psk_finish.c
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2021-2025 DMTF. All rights reserved.
+ *  Copyright 2021-2026 DMTF. All rights reserved.
  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
  **/
 
@@ -293,6 +293,7 @@ static libspdm_return_t libspdm_try_send_receive_psk_finish(
             status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
             goto receive_done;
         }
+
         /* this message can only be in secured session
          * thus don't need to consider transport layer padding, just check its exact size */
         if (spdm_response_size != sizeof(spdm_psk_finish_response_t) + sizeof(uint16_t) +
@@ -302,7 +303,7 @@ static libspdm_return_t libspdm_try_send_receive_psk_finish(
         }
 
         if ((responder_opaque_data != NULL) && (responder_opaque_data_size != NULL)) {
-            if (opaque_data_size >= *responder_opaque_data_size) {
+            if (opaque_data_size > *responder_opaque_data_size) {
                 status = LIBSPDM_STATUS_BUFFER_TOO_SMALL;
                 goto receive_done;
             }
@@ -325,7 +326,7 @@ static libspdm_return_t libspdm_try_send_receive_psk_finish(
         }
         opaque_data_entry_size = 0;
     }
-    spdm_response_size = sizeof(spdm_finish_response_t) + opaque_data_entry_size;
+    spdm_response_size = sizeof(spdm_psk_finish_response_t) + opaque_data_entry_size;
 
     status = libspdm_append_message_f(spdm_context, session_info, true, spdm_response,
                                       spdm_response_size);
--- a/library/spdm_responder_lib/libspdm_rsp_psk_finish_rsp.c
+++ b/library/spdm_responder_lib/libspdm_rsp_psk_finish_rsp.c
@@ -266,6 +266,12 @@ libspdm_return_t libspdm_get_response_psk_finish(libspdm_context_t *spdm_context
                                                        SPDM_ERROR_CODE_UNSPECIFIED, 0,
                                                        response_size, response);
             }
+
+            if (opaque_data_size > SPDM_MAX_OPAQUE_DATA_SIZE) {
+                return libspdm_generate_error_response(spdm_context,
+                                                       SPDM_ERROR_CODE_UNSPECIFIED, 0,
+                                                       response_size, response);
+            }
         }
         libspdm_write_uint16(ptr, (uint16_t)opaque_data_size);
         opaque_data_entry_size = sizeof(uint16_t) + opaque_data_size;
--- a/unit_test/test_spdm_requester/psk_finish.c
+++ b/unit_test/test_spdm_requester/psk_finish.c
@@ -123,6 +123,8 @@ static libspdm_return_t send_message(
         m_libspdm_local_buffer_size += decoded_message_size;
     }
         return LIBSPDM_STATUS_SUCCESS;
+    case 0x12:
+        return LIBSPDM_STATUS_SUCCESS;
     default:
         return LIBSPDM_STATUS_SEND_FAIL;
     }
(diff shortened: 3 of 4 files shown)


library/spdm_requester_lib/libspdm_req_psk_finish.c
```diff
@@ -170,6 +170,11 @@ static libspdm_return_t libspdm_try_send_receive_psk_finish(
     ptr = (uint8_t *)spdm_request + sizeof(spdm_psk_finish_request_t);
     if (libspdm_get_connection_version(spdm_context) >= SPDM_MESSAGE_VERSION_14) {
         if (requester_opaque_data != NULL) {
+            if (requester_opaque_data_size > SPDM_MAX_OPAQUE_DATA_SIZE) {
+                status = LIBSPDM_STATUS_INVALID_PARAMETER;
+                libspdm_release_sender_buffer (spdm_context);
+                goto error;
+            }
```
jyao1: In current libspdm, we always use ASSERT for SPDM_MAX_OPAQUE_DATA_SIZE check.
You may search others, such as key_exchange, finish.
Why we have to add check here, in this psk_finish?

library/spdm_requester_lib/libspdm_req_psk_finish.c
```diff
@@ -183,6 +188,11 @@ static libspdm_return_t libspdm_try_send_receive_psk_finish(
                              requester_opaque_data, requester_opaque_data_size);
             opaque_data_size = requester_opaque_data_size;
         } else {
+            if (requester_opaque_data_size != 0) {
+                status = LIBSPDM_STATUS_INVALID_PARAMETER;
+                libspdm_release_sender_buffer (spdm_context);
+                goto error;
+            }
```
jyao1: Same question.
I think we need use a consistent way to decide when to check and when not to check.
Please refer to opaque data handling in other command.

library/spdm_requester_lib/libspdm_req_psk_finish.c
```diff
@@ -293,6 +303,14 @@ static libspdm_return_t libspdm_try_send_receive_psk_finish(
             status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
             goto receive_done;
         }
+
+        if (((spdm_context->connection_info.algorithm.other_params_support &
+              SPDM_ALGORITHMS_OPAQUE_DATA_FORMAT_MASK) == SPDM_ALGORITHMS_OPAQUE_DATA_FORMAT_NONE)
+            && (opaque_data_size != 0)) {
+            status = LIBSPDM_STATUS_INVALID_MSG_FIELD;
+            goto receive_done;
+        }
```
jyao1: This is already checked in key_exchange. Why we need to check again?

library/spdm_requester_lib/libspdm_req_psk_finish.c
```diff
@@ -302,7 +320,7 @@ static libspdm_return_t libspdm_try_send_receive_psk_finish(
         }
 
         if ((responder_opaque_data != NULL) && (responder_opaque_data_size != NULL)) {
-            if (opaque_data_size >= *responder_opaque_data_size) {
+            if (opaque_data_size > *responder_opaque_data_size) {
```
jyao1, line 306: Good catch!

I notice finish has same issue.

library/spdm_requester_lib/libspdm_req_psk_finish.c
```diff
@@ -325,7 +343,7 @@ static libspdm_return_t libspdm_try_send_receive_psk_finish(
         }
         opaque_data_entry_size = 0;
     }
-    spdm_response_size = sizeof(spdm_finish_response_t) + opaque_data_entry_size;
+    spdm_response_size = sizeof(spdm_psk_finish_response_t) + opaque_data_entry_size;
```
jyao1, line 329: Good catch!

library/spdm_responder_lib/libspdm_rsp_psk_finish_rsp.c
```diff
@@ -266,6 +269,13 @@ libspdm_return_t libspdm_get_response_psk_finish(libspdm_context_t *spdm_context
                                                        SPDM_ERROR_CODE_UNSPECIFIED, 0,
                                                        response_size, response);
             }
+
+            if ((opaque_data_size > SPDM_MAX_OPAQUE_DATA_SIZE) ||
+                (opaque_data_size > max_opaque_data_size)) {
```
jyao1: I am not sure how it is possible that (opaque_data_size > max_opaque_data_size).

I feel (opaque_data_size > SPDM_MAX_OPAQUE_DATA_SIZE) should be enough.

library/spdm_requester_lib/libspdm_req_psk_finish.c
czwolak: Agree. For outbound requester input-size validation, key_exchange / psk_exchange / finish use LIBSPDM_ASSERT and do not return runtime INVALID_PARAMETER.
I will align psk_finish with existing style by removing the runtime check and keeping ASSERT-only

library/spdm_requester_lib/libspdm_req_psk_finish.c
czwolak: Agree. The requester_opaque_data == NULL && requester_opaque_data_size != 0 runtime error is also not used in peer commands.
I will remove this runtime check for consistency with existing requester command patterns.

library/spdm_requester_lib/libspdm_req_psk_finish.c
czwolak: Agree this is redundant with current negotiated state handling and inconsistent with finish/key_exchange response parsing style.
I will remove this extra check to keep behavior/policy consistent.

library/spdm_requester_lib/libspdm_req_psk_finish.c
czwolak: Thanks. Do you think we better create separate follow-up issue or propose "finish" fix within this PR ?

library/spdm_responder_lib/libspdm_rsp_psk_finish_rsp.c
czwolak: I recommend keeping both checks:
> SPDM_MAX_OPAQUE_DATA_SIZE: protocol/library upper bound
> max_opaque_data_size: output-buffer capacity bound
They protect different constraints, so both are valid.

It can happen when the protocol limit and the actual buffer capacity are different.

SPDM_MAX_OPAQUE_DATA_SIZE is a global maximum allowed by libspdm/protocol.
max_opaque_data_size is how many bytes are currently free in this specific response buffer after headers.
So opaque_data_size may still be valid globally, but too large for the current output buffer.
Example: global max is 1024, but current buffer has only 80 bytes left; function generates 120 bytes. Then:

120 <= 1024 (passes global check)
120 > 80 (fails buffer-capacity check)

library/spdm_requester_lib/libspdm_req_psk_finish.c
jyao1: I am OK to create another PR.
I just want to ensure 1) all issues are fixed, 2) the solution is consistent, especially the similar ones.
That is why I cross-check other commands.

library/spdm_responder_lib/libspdm_rsp_psk_finish_rsp.c
jyao1: I am looking at `libspdm_psk_finish_rsp_opaque_data()` API. `opaque_data_size` is an IN OUT.
`opaque_data_size` is already `max_opaque_data_size` on input.

The output `opaque_data_size` must be equal or smaller than the input size.
That is why I think it is impossible. Otherwise, the buffer overflow already happened.

library/spdm_requester_lib/libspdm_req_psk_finish.c
czwolak: Thank you :) can we confirm also another cases?

1. FINISH - what you found

https://github.com/DMTF/libspdm/blob/538d089fb11ab9f90b3d78f098424ccfebb04416/library/spdm_requester_lib/libspdm_req_finish.c#L611-L614

2. KEY_EXCHANGE

https://github.com/DMTF/libspdm/blob/538d089fb11ab9f90b3d78f098424ccfebb04416/library/spdm_requester_lib/libspdm_req_key_exchange.c#L728-L731

3. PSK_EXCHANGE

https://github.com/DMTF/libspdm/blob/538d089fb11ab9f90b3d78f098424ccfebb04416/library/spdm_requester_lib/libspdm_req_psk_exchange.c#L459-L462

library/spdm_responder_lib/libspdm_rsp_psk_finish_rsp.c
czwolak: OK, I understand we want to assume IN OUT is working here, if it didn't then it is already too late to detect. I will remove this.

czwolak:
@jyao1 @steven-bellock did you have a chance to look into this proposal?

alistair23 (approved):

PR #3610, opened 2026-04-24
libspdm_rsp_algorithms: Add a callback on successfull negotiation

alistair23:
Add a new optional callback that is called when ALGORITHMS are sucessfully negotiated. The idea is that once algorithms are set libspdm can trigger a callback, this allows the main implementation to update the banked certificate chains.

This allows us to support banking, without requiring all of the possible combinations to be set in libspdm arrays.

Note that this doesn't support `SLOT_MGMT_CAP=1`, but that can be added in the future, where the implementer of libspdm can then handle the requests.

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -564,6 +564,8 @@ typedef struct {
     libspdm_transport_encode_message_func transport_encode_message;
     libspdm_transport_decode_message_func transport_decode_message;
 
+    libspdm_device_algs_negotiated_func algs_negotiated;
+
     /* Cached plain text command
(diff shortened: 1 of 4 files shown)


jyao1:
We already have `libspdm_register_connection_state_callback_func`. Once you registered, the `spdm_connection_state_callback` will be called.

It is used by spdm-emu now.
https://github.com/DMTF/spdm-emu/blob/main/spdm_emu/spdm_responder_emu/spdm_responder_spdm.c#L348

alistair23:
Ah! I missed that, `libspdm_register_connection_state_callback_func` does exactly what I want. It's even called at the same place. Thanks for pointing that out

PR #3618, opened 2026-04-30
transport/tcp: expose transport haeder and tail sizes

twilfredo:
Currently the TCP trasnport header does not expose the transport header and tail size unlike the other transports. So expose it.

### DIFF
--- a/include/library/spdm_transport_tcp_lib.h
+++ b/include/library/spdm_transport_tcp_lib.h
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2025 DMTF. All rights reserved.
+ *  Copyright 2026 DMTF. All rights reserved.
  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
  **/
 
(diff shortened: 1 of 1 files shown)


include/library/spdm_transport_tcp_lib.h
```diff
@@ -19,6 +19,9 @@
  *
  */
 
+#define LIBSPDM_TCP_TRANSPORT_HEADER_SIZE  (4 + 4 + 2 + 2)
+#define LIBSPDM_TCP_TRANSPORT_TAIL_SIZE    (16 + 3 + 31)
```
alistair23, line 23: You should comment where these come from and what they are

twilfredo: https://github.com/twilfredo/libspdm/blob/21eee4e38a9c1ed11c423bbf252e04daaa0fa1ba/include/library/spdm_transport_tcp_lib.h#L12

It's based on this, which is just above

alistair23: Whoops!

alistair23 (approved):

jyao1 (approved):

PR #3621, opened 2026-05-11
macro_check: enforce capability dependencies for KEM/DHE and signature algorithms

czwolak:
Require KEY_EX_CAP when DHE/KEM/MUT_AUTH is enabled, and require CERT_CAP/CHAL_CAP/KEY_EX_CAP when any signature algorithm is enabled.
    
Keep checks focused on capability-to-algorithm consistency in libspdm_macro_check.h.

### DIFF
--- a/include/internal/libspdm_macro_check.h
+++ b/include/internal/libspdm_macro_check.h
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2021-2025 DMTF. All rights reserved.
+ *  Copyright 2021-2026 DMTF. All rights reserved.
  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
  **/
 
@@ -35,6 +35,17 @@
 #define LIBSPDM_FIPS_HASH_ALGO_SUPPORT \
     ((LIBSPDM_SHA2_SUPPORT) || (LIBSPDM_SHA3_SUPPORT))
 #endif /*LIBSPDM_FIPS_MODE*/
+
+#if ((LIBSPDM_ML_KEM_SUPPORT) || (LIBSPDM_DHE_ALGO_SUPPORT)) && \
+    !(LIBSPDM_ENABLE_CAPABILITY_KEY_EX_CAP)
+    #error KEM/DHE support requires KEY_EX_CAP capability.
+#endif
+
+#if ((LIBSPDM_ASYM_ALGO_SUPPORT) || (LIBSPDM_ML_DSA_SUPPORT) || (LIBSPDM_SLH_DSA_SUPPORT)) && \
+    !((LIBSPDM_ENABLE_CAPABILITY_CERT_CAP) || (LIBSPDM_ENABLE_CAPABILITY_CHAL_CAP) || \
+    (LIBSPDM_ENABLE_CAPABILITY_KEY_EX_CAP))
+    #error Asymmetric signature support requires CERT_CAP, CHAL_CAP, or KEY_EX_CAP.
+#endif
 
 #if (LIBSPDM_ENABLE_CAPABILITY_KEY_EX_CAP) && !LIBSPDM_ASYM_ALGO_SUPPORT
     #error If KEY_EX_CAP is enabled then at least one asymmetric algorithm must also be enabled.


jyao1:
please clarify the problem statement. Why this is needed?

czwolak:
> please clarify the problem statement. Why this is needed?

Thanks for the feedback.
The need is to make SPDM 1.4 + PQC configuration deterministic and fail early.

Today, requester/responder role selection and PQC capability combinations can lead to invalid macro sets that are only discovered later (integration/CI/runtime).
This PR adds explicit SPDM 1.4 build profiles (both/requester/responder) and compile-time fail-fast checks so invalid combinations are rejected immediately.

So the value is:

- clearer role-focused builds,
- earlier detection of invalid SPDM/PQC configs,
- lower integration/debug cost, while keeping backward compatibility (both remains default).

include/library/spdm_lib_config_profile_requester.h
```diff
@@ -0,0 +1,28 @@
...
+ **/
+
+/** @file
+ *  SPDM 1.4 requester-focused build profile.
+ */
+
+#ifndef SPDM_LIB_CONFIG_PROFILE_REQUESTER_H
+#define SPDM_LIB_CONFIG_PROFILE_REQUESTER_H
+
+#define LIBSPDM_SPDM14_BUILD_PROFILE_ID LIBSPDM_SPDM14_BUILD_PROFILE_REQUESTER
```
jyao1: I am not sure why we need `SPDM14` here.
What about SPDM10, SPDM11, SPDM12, SPDM13, and future SPDM15?
Do we want to add for each specific version?

include/internal/libspdm_macro_check.h
```diff
@@ -36,6 +36,27 @@
...
+    #error LIBSPDM_SPDM14_BUILD_PROFILE_ID is invalid.
+#endif
+
+#if (LIBSPDM_SPDM14_BUILD_PROFILE_ID == LIBSPDM_SPDM14_BUILD_PROFILE_RESPONDER) && \
+    ((LIBSPDM_EVENT_RECIPIENT_SUPPORT) || (LIBSPDM_SEND_GET_ENDPOINT_INFO_SUPPORT))
+    #error Responder profile requires EVENT_RECIPIENT_SUPPORT and SEND_GET_ENDPOINT_INFO_SUPPORT to be disabled.
+#endif
+
+#if (LIBSPDM_ML_KEM_SUPPORT) && !(LIBSPDM_ENABLE_CAPABILITY_KEY_EX_CAP)
+    #error ML-KEM support requires KEY_EX_CAP capability.
```
jyao1: why only check KEM? DHE also need KEY_EX.

include/internal/libspdm_macro_check.h
```diff
@@ -36,6 +36,27 @@
...
+#endif
+
+#if ((LIBSPDM_ML_DSA_SUPPORT) || (LIBSPDM_SLH_DSA_SUPPORT)) && \
+    !((LIBSPDM_ENABLE_CAPABILITY_CERT_CAP) || (LIBSPDM_ENABLE_CAPABILITY_CHAL_CAP) || \
+    (LIBSPDM_ENABLE_CAPABILITY_KEY_EX_CAP) || (LIBSPDM_ENABLE_CAPABILITY_MUT_AUTH_CAP))
+    #error ML-DSA/SLH-DSA support requires CERT_CAP, CHAL_CAP, KEY_EX_CAP, or MUT_AUTH_CAP.
```
jyao1: same question. RSA/ECDSA/EdDSA/... also need similar check.

jyao1:
I notice this PR include 2 features: 1) add consistency check, 2) add PROFILE_ID.

I think those are 2 different features. I suggest we split to 2 different PRs.

include/library/spdm_lib_config_profile_requester.h
```diff
@@ -0,0 +1,28 @@
...
+
+/*
+ * Requester profile keeps requester-originated command support enabled.
+ * The base config already defaults these to enabled, but we set them
+ * explicitly for profile clarity and to avoid ambiguity in downstream builds.
+ */
+#define LIBSPDM_SEND_GET_CERTIFICATE_SUPPORT 1
+#define LIBSPDM_SEND_CHALLENGE_SUPPORT 1
+#define LIBSPDM_EVENT_RECIPIENT_SUPPORT 1
+#define LIBSPDM_SEND_GET_ENDPOINT_INFO_SUPPORT 1
```
jyao1: I do not understand why those should be in profile.

include/library/spdm_lib_config_profile_responder.h
```diff
@@ -0,0 +1,27 @@
...
+#define LIBSPDM_SPDM14_BUILD_PROFILE_ID LIBSPDM_SPDM14_BUILD_PROFILE_RESPONDER
+
+/*
+ * Responder profile disables requester-originated message flows.
+ * This keeps the binary focused on pure responder behavior.
+ */
+#define LIBSPDM_SEND_GET_CERTIFICATE_SUPPORT 1
+#define LIBSPDM_SEND_CHALLENGE_SUPPORT 1
+#define LIBSPDM_EVENT_RECIPIENT_SUPPORT 0
+#define LIBSPDM_SEND_GET_ENDPOINT_INFO_SUPPORT 0
```
jyao1: I do not understand why this is a must.

czwolak:
@jyao1 I think I see your point. Lets close this PR and related task. I will try to introduce 1.4 as 1.1-1.3 was before without additional profiles, checks. Just pure support, OK?

czwolak:
I will try to introduce 1.4 as 1.1-1.3 was before, without additional build profiles, checks.

include/library/spdm_lib_config_profile_requester.h
czwolak: removed

include/library/spdm_lib_config_profile_requester.h
czwolak: removed

include/library/spdm_lib_config_profile_responder.h
czwolak: removed

include/internal/libspdm_macro_check.h
czwolak: added for LIBSPDM_DHE_ALGO_SUPPORT

include/internal/libspdm_macro_check.h
czwolak: added for LIBSPDM_ASYM_ALGO_SUPPORT and removed MUT_AUTH_CAP from the condition. added condition that MUT_AUTH_CAP requires KEY_EX_CAP

jyao1:
please change the descript if the scope is changed.

include/internal/libspdm_macro_check.h
```diff
@@ -36,6 +36,21 @@
...
+
+#if ((LIBSPDM_ASYM_ALGO_SUPPORT) || (LIBSPDM_ML_DSA_SUPPORT) || (LIBSPDM_SLH_DSA_SUPPORT)) && \
+    !((LIBSPDM_ENABLE_CAPABILITY_CERT_CAP) || (LIBSPDM_ENABLE_CAPABILITY_CHAL_CAP) || \
+    (LIBSPDM_ENABLE_CAPABILITY_KEY_EX_CAP))
+    #error Asymmetric signature support requires CERT_CAP, CHAL_CAP, or KEY_EX_CAP.
+#endif
+
+#if (LIBSPDM_ENABLE_CAPABILITY_MUT_AUTH_CAP) && !(LIBSPDM_ENABLE_CAPABILITY_KEY_EX_CAP)
+    #error If MUT_AUTH_CAP is enabled then KEY_EX_CAP must also be enabled.
+#endif
```
jyao1: The SPDM spec does not mention this rule.
I suggest to remove it.

czwolak: done

czwolak:
> I notice this PR include 2 features: 1) add consistency check, 2) add PROFILE_ID.
> 
> I think those are 2 different features. I suggest we split to 2 different PRs.

Only consistency check left

alistair23 (approved):

PR #3623, opened 2026-05-13
TPM: add configurable TPM key handles

itsManjeet:
Generate the TPM key-handle configuration at CMake configure time instead of
using fixed values in keys.h. This lets platforms provide requester/responder
key handles and certificate-chain NV indices without editing the TPM device
secret library.

```sh
-D LIBSPDM_TPM_REQUESTER_HANDLES="0x81000011;0x81000012"
-D LIBSPDM_TPM_REQUESTER_CERTCHAINS="0x1500011;0x1500012"
-D LIBSPDM_TPM_RESPONDER_HANDLES="0x81000021;0x81000022"
-D LIBSPDM_TPM_RESPONDER_CERTCHAINS="0x1500021;0x1500022"
```


os_stub/spdm_device_secret_lib_tpm/CMakeLists.txt
```diff
@@ -1,9 +1,70 @@
 cmake_minimum_required(VERSION 3.5)
 
+# Default values for backward compatibility (2 slots)
+set(LIBSPDM_TPM_REQUESTER_HANDLES "0x81000011;0x81000011" CACHE STRING "Semicolon-separated list of TPM Requester Handle Strings")
+set(LIBSPDM_TPM_REQUESTER_CERTCHAINS "0x1500011;0x1500011" CACHE STRING "Semicolon-separated list of TPM Requester Certificate Chain NV Indices")
+set(LIBSPDM_TPM_RESPONDER_HANDLES "0x81000021;0x81000021" CACHE STRING "Semicolon-separated list of TPM Responder Handle Strings")
+set(LIBSPDM_TPM_RESPONDER_CERTCHAINS "0x1500021;0x1500021" CACHE STRING "Semicolon-separated list of TPM Responder Certificate Chain NV Indices")
```
jyao1: It will be good to add some description in https://github.com/DMTF/libspdm/blob/main/doc/tpm_usage.md.

doc/tpm_usage.md
```diff
@@ -4,6 +4,31 @@ This document describes how to use TPM-backed helper APIs with libspdm for secur
...
+### Configuration Parameters
+
+| Parameter | Description | Format |
+|-----------|-------------|--------|
+| `LIBSPDM_TPM_REQUESTER_HANDLES` | TPM persistent key handles for requester | Semicolon-separated hex values |
+| `LIBSPDM_TPM_REQUESTER_CERTCHAINS` | NV indices for requester certificate chains | Semicolon-separated hex values |
+| `LIBSPDM_TPM_RESPONDER_HANDLES` | TPM persistent key handles for responder | Semicolon-separated hex values |
+| `LIBSPDM_TPM_RESPONDER_CERTCHAINS` | NV indices for responder certificate chains | Semicolon-separated hex values |
+
+**Note:** Multiple handles/indices support multi-key scenarios. First handle maps to slot 0, second to slot 1, etc.
```
jyao1: It seems this design does not support empty slot.
Current libspdm device sample support slot 0, slot 1, and slot 4.

Please make slot 4 work or add comment to indicate the limitation.

ayj (commented):
Build-time CMake generation vs. Runtime Configuration API

While generating `keys.h` via CMake solves the hardcoded handle problem nicely for CMake users, `file(WRITE ...)` code generation creates friction for integrators using other build systems (Makefiles, Bazel, Zephyr, or EDKII) who would need to replicate this generator script in their respective build tools.

Has the team considered replacing `keys.h` generation with a lightweight runtime configuration API on the TPM device stub? This would directly mirror the established precedence in `libspdm` for runtime callback/hook registration (e.g., `libspdm_register_device_io_func`, `libspdm_register_transport_layer_func`) and context provisioning (`libspdm_set_data`). For example:

```c
typedef struct {
    uint8_t slot_id;
    uint32_t key_handle;
    uint32_t certchain_nv_index;
} libspdm_tpm_slot_config_t;

bool libspdm_tpm_register_responder_slots(const libspdm_tpm_slot_config_t *configs, size_t count);
```

* It aligns the OS and TPM device stub with `libspdm`'s core pattern of runtime configuration (`libspdm_register_*`) rather than build-time source generation.
* It requires zero build-time generation scripts across CMake, Make, Bazel, or Zephyr.
* It allows platform initialization code to dynamically discover TPM layouts (e.g., via device tree or flash setup variables) at runtime without compiling separate binaries per platform SKU layout.

Additionally, regarding `sign.c`: both `libspdm_responder_data_sign` and `libspdm_requester_data_sign` currently hardcode Slot 0 (`LIBSPDM_TPM_HANDLE_RESPONDER_HANDLE_SLOT_0`). If multi-slot signing is planned for SPDM 1.2+, a runtime registration table (`libspdm_tpm_register_responder_slots`) would also make it straightforward to look up key handles by `slot_id` during signing.

PR #3629, opened 2026-05-25
Add initial certificate bank support and allow selecting the bank

alistair23:
SPDM 1.4 add support for the banked architecture. Basically allowing a number of banks of certificate slots.

Currently bank support can be handled by the implementer using the CONNECTION_STATE_NEGOTIATED callback registered with `libspdm_register_connection_state_callback_func()`.

The problem with this is it pushes a lot of complexity back to the implementer and it makes supporting the Slot Management commands tricky, as the implementer will need to handle the commands as well.

Instead let's move the bank support into libspdm. Currently we just allow at build time a 2-D array of certificates, for banks and slots. We then allow the implementation to select the bank to use, which defaults to zero.

Future patches will work on the Slot Management commands, which can then operate on specific banks.

At build time we allow users to specify their own bank count, allowing smaller bank counts on size reduced systems.


script/format_nix.sh
```diff
@@ -19,7 +19,6 @@ REQUIRED_VERSION="0.78"
 
 if [ "$VERSION" != "$REQUIRED_VERSION" ]; then
     echo "ERROR: uncrustify version $REQUIRED_VERSION is required, but found $VERSION."
-    exit 1
```
itsManjeet, line 22: Please keep the exit 1 in place. It helps ensure the intended behavior remains clear and consistent.

include/library/spdm_lib_config.h
```diff
@@ -86,6 +86,11 @@
 #define LIBSPDM_ENABLE_CAPABILITY_ENDPOINT_INFO_CAP 1
 #endif
 
+/* SPDM 1.4 capabilities. */
+#ifndef LIBSPDM_MAX_BANK_COUNT
+#define LIBSPDM_MAX_BANK_COUNT SPDM_MAX_BANK_COUNT
```
jyao1: I feel 240 bank is too much for a normal device.

How many banks you have seen or planned in a production usually?

include/internal/libspdm_common_lib.h
```diff
@@ -99,13 +99,14 @@ typedef struct {
...
-    const void *local_cert_chain_provision[SPDM_MAX_SLOT_COUNT];
-    size_t local_cert_chain_provision_size[SPDM_MAX_SLOT_COUNT];
+    const void *local_cert_chain_provision[LIBSPDM_MAX_BANK_COUNT][SPDM_MAX_SLOT_COUNT];
+    size_t local_cert_chain_provision_size[LIBSPDM_MAX_BANK_COUNT][SPDM_MAX_SLOT_COUNT];
     uint8_t local_supported_slot_mask;
     uint8_t cert_slot_reset_mask;
-    spdm_key_pair_id_t local_key_pair_id[SPDM_MAX_SLOT_COUNT];
-    spdm_certificate_info_t local_cert_info[SPDM_MAX_SLOT_COUNT];
-    spdm_key_usage_bit_mask_t local_key_usage_bit_mask[SPDM_MAX_SLOT_COUNT];
+    uint8_t current_bank;
```
jyao1: I am not sure why `current_bank` is in `libspdm_local_context_t`.

I think `current_bank` should be based on negotiated algorithm, right?
Or do I misunderstand the meaning of `current_bank`?

jyao1:
> SPDM 1.4 add support for the banked architecture. Basically allowing a number of banks of certificate slots.
> 
> Currently bank support can be handled by the implementer using the CONNECTION_STATE_NEGOTIATED callback registered with `libspdm_register_connection_state_callback_func()`.
> 
> The problem with this is it pushes a lot of complexity back to the implementer and it makes supporting the Slot Management commands tricky, as the implementer will need to handle the commands as well.
> 
> Instead let's move the bank support into libspdm. Currently we just allow at build time a 2-D array of certificates, for banks and slots. We then allow the implementation to select the bank to use, which defaults to zero.
> 
> Future patches will work on the Slot Management commands, which can then operate on specific banks.
> 
> At build time we allow users to specify their own bank count, allowing smaller bank counts on size reduced systems.

I think we might want a high level design for bank management.
To me, there are too different ways:
1) Put bank management into libspdm. That means, the integrator needs to register everything to libspdm. As such, READ operation does not need to reach integrator, while WRITE operation is write-through from libspdm to the integrator's permanent configuration (e.g. NVRAM).
2) Put bank management out of libspdm. All bank management command (READ and WRITE) needs to pass-thru to integrator.

Bank management is similar to key pair info feature. Today we are using option 2) - see https://github.com/DMTF/libspdm/blob/main/include/hal/library/responder/key_pair_info.h

I feel we can start from option 2.

include/library/spdm_lib_config.h
alistair23: An implementation could overwrite this, 240 is just the maximum.

I would guess maybe 2 banks? One for pre-quantum and one for post-quantum. From my reading of things banks are mostly to support PQC

include/internal/libspdm_common_lib.h
alistair23: `NEGOTIATE_ALGORITHMS` doesn't select the bank. It's up to the implementation to select the bank. So the implementation uses the information about the negotiated algorithm to set `current_bank`.

For example, if an implementation supported 3 banks, A, B, C, if algorithm B was negotiated then it would set `current_bank` to `1`

alistair23:
> * 
> 
> * Put bank management out of libspdm. All bank management command (READ and WRITE) needs to pass-thru to integrator.

The issue here though is it's very complex managing the `libspdm_local_context_t` state in libspdm. The implementor needs to regularly update the state inside libspdm as well as handling the READ and WRITE commands.


> 
> Bank management is similar to key pair info feature. Today we are using option 2) - see https://github.com/DMTF/libspdm/blob/main/include/hal/library/responder/key_pair_info.h

I think key pairs are simpler then banks. Note that the bank approach I'm thinking about here will be similar to https://github.com/DMTF/libspdm/blob/main/include/hal/library/responder/key_pair_info.h, in that we will have to call to the implementer for handling the commands. The main difference is that libspdm understands the banks as well, compared to currently where libspdm doesn't comprehend banks and the implementer has to fake it by manually swapping out the certs

include/internal/libspdm_common_lib.h
jyao1: Yes, that was exactly what I meant. "the implementation uses the information about the negotiated algorithm to set current_bank."

As such, this `current_bank` should be runtime negotiation info (`libspdm_connection_info_t`), but NOT local content info (`libspdm_local_context_t`).

jyao1:
> > * Put bank management out of libspdm. All bank management command (READ and WRITE) needs to pass-thru to integrator.
> 
> The issue here though is it's very complex managing the `libspdm_local_context_t` state in libspdm. The implementor needs to regularly update the state inside libspdm as well as handling the READ and WRITE commands.
> 
> > Bank management is similar to key pair info feature. Today we are using option 2) - see https://github.com/DMTF/libspdm/blob/main/include/hal/library/responder/key_pair_info.h
> 
> I think key pairs are simpler then banks. Note that the bank approach I'm thinking about here will be similar to https://github.com/DMTF/libspdm/blob/main/include/hal/library/responder/key_pair_info.h, in that we will have to call to the implementer for handling the commands. The main difference is that libspdm understands the banks as well, compared to currently where libspdm doesn't comprehend banks and the implementer has to fake it by manually swapping out the certs

Ok. That is fine.
But so far, I do not have a full picture on your solution. As such, I do not feel comfortable to approve and merge this partial solution, because it might not be needed if we choose a different direction. For example, if we go option 2, then we do not need to cache all bank info in local content.

Can you do a design review to compare those 2 directions?

include/library/spdm_lib_config.h
jyao1: if you think 2 banks are typical case, then we should define LIBSPDM_MAX_BANK_COUNT  to be 2 as default. Of course people can override, if they want more.

jyao1:
BTW: I am trying to add SLOT_MANAGEMENT feature using option 2.
I will submit patch soon.

alistair23:
> But so far, I do not have a full picture on your solution. As such, I do not feel comfortable to approve and merge this partial solution, because it might not be needed if we choose a different direction. For example, if we go option 2, then we do not need to cache all bank info in local content.

But option 2 requires a lot more duplicated work each implementer, with a greater chance of bugs in all of them.

My plan is to keep working on this, but I didn't want to spend too much time on the bank management if it wasn't going to be accepted.

> Can you do a design review to compare those 2 directions?

I can, but I'm not sure what that is

jyao1:
@alistair23 
I have submitted draft https://github.com/DMTF/libspdm/pull/3637, which is idea for option 2.
Please take a look and feedback.

ayj (commented):
In `libspdm_check_context`, certificate provision sizes are checked against `max_spdm_msg_size`. When hybrid ECDSA (Bank 0, ~2KB) and PQC (Bank 1, e.g. ML-DSA/Dilithium, ~15KB) are both populated, `max_spdm_msg_size` must either be sized for the greater of the two banks or SPDM 1.2 Chunking (`CHUNK_CAP`) must be strictly enforced whenever Bank 1 or higher contains large PQC chains. 

Should `libspdm_check_context` iterate across all configured banks during startup (rather than checking only `current_bank`) to catch PQC buffer overflow misconfigurations early?

PR #3637, opened 2026-06-04
add Slot management

### LINKED ISSUE #3131: 1.4 Add SLOT_MANAGEMENT

jyao1:
Fix https://github.com/DMTF/libspdm/issues/3131

The spdm-emu update is at https://github.com/DMTF/spdm-emu/pull/496.
The spdm-dump update is at https://github.com/DMTF/spdm-dump/pull/140.

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_slot_management.c
+++ b/library/spdm_responder_lib/libspdm_rsp_slot_management.c
@@ -1,0 +1,1263 @@
+/**
+ *  Copyright Notice:
+ *  Copyright 2026 DMTF. All rights reserved.
+ *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
+ **/
+
+#include "internal/libspdm_responder_lib.h"
+
+#if LIBSPDM_ENABLE_CAPABILITY_SLOT_MGMT_CAP
+
+/**
+ * Validate a SlotAddress request structure.
+ *
+ * Per DSP0274 Table 143, the ReqLength field of a SlotAddress shall be 8 for this version of the
+ * specification.
+ *
+ * @retval true   the SlotAddress is well-formed.
+ * @retval false  the SlotAddress shall be rejected with ERROR(InvalidRequest).
+ **/
+static bool libspdm_slot_management_slot_address_valid(
+    const spdm_slot_management_slot_address_struct_t *slot_address)
+{
+    return slot_address->req_length == SPDM_SLOT_MANAGEMENT_SLOT_ADDRESS_REQ_LENGTH;
+}
+
+/**
+ * Check whether a state-modifying certificate slot management operation on the given SlotID is
+ * permitted.
+ *
+ * Per DSP0274 "Certificate slot management", for slots 1-7 these commands shall only be issued
+ * in a secure session or a trusted environment. (Slot 0 has no such requirement here; it is
+ * recommended to be managed in a trusted environment, which is outside this check.) A secure
+ * session is an accepted alternative to a trusted environment, matching the base SET_CERTIFICATE
+ * handler.
+ *
+ * This applies to the SubCodes that modify state (SetCertificate, ManageSlot, and GetCSR, which
+ * generates/installs a CSR). The read-only SubCodes (GetBankInfo, GetBankDetails,
+ * GetCertificateChain) only retrieve information and, like base GET_CERTIFICATE / GET_DIGESTS,
+ * carry no such requirement.
+ *
+ * @retval true   the operation on slot_id is allowed.
+ * @retval false  the operation shall be rejected with ERROR(UnexpectedRequest).
+ **/
+static bool libspdm_slot_management_access_allowed(libspdm_context_t *spdm_context,
+                                                   uint8_t slot_id)
+{
+    if (slot_id == 0) {
+        return true;
+    }
+    if (libspdm_is_in_trusted_environment(spdm_context)) {
+        return true;
+    }
+    if (spdm_context->last_spdm_request_session_id_valid) {
+        return true;
+    }
+    return false;
+}
+
+/**
+ * Return whether a SubCode value is listed in DSP0274 Table 142 (i.e. a defined SubCode).
+ *
+ * A listed-but-unadvertised SubCode is answered with UnsupportedRequest, whereas an unlisted
+ * (reserved) SubCode is answered with InvalidRequest, so the two cases must be distinguished.
+ **/
+static bool libspdm_slot_management_subcode_is_listed(uint8_t sub_code)
+{
+    switch (sub_code) {
+    case SPDM_SLOT_MANAGEMENT_SUBCODE_SUPPORTED_SUBCODES:
+    case SPDM_SLOT_MANAGEMENT_SUBCODE_GET_BANK_INFO:
+    case SPDM_SLOT_MANAGEMENT_SUBCODE_GET_BANK_DETAILS:
+    case SPDM_SLOT_MANAGEMENT_SUBCODE_GET_CERTIFICATE_CHAIN:
+    case SPDM_SLOT_MANAGEMENT_SUBCODE_GET_CSR:
+    case SPDM_SLOT_MANAGEMENT_SUBCODE_MANAGE_BANK:
+    case SPDM_SLOT_MANAGEMENT_SUBCODE_MANAGE_SLOT:
+    case SPDM_SLOT_MANAGEMENT_SUBCODE_SET_CERTIFICATE:
+        return true;
+    default:
+        return false;
+    }
+}
+
+/**
+ * Process the SLOT_MANAGEMENT SupportedSubCodes SubCode.
+ *
+ * SupportedSubCodes does not use a request structure. The response carries the bit map of
+ * SubCodes the Responder supports.
+ **/
+static libspdm_return_t libspdm_get_response_slot_management_supported_subcodes(
+    libspdm_context_t *spdm_context, size_t request_size, const void *request,
+    size_t *response_size, void *response)
+{
+    const spdm_slot_management_request_t *spdm_request;
+    spdm_slot_management_response_t *spdm_response;
+    spdm_slot_management_supported_subcodes_struct_t *resp_struct;
+    uint8_t sub_code_bitmap[8];
+
+    spdm_request = request;
+
+    /* SupportedSubCodes has no request structure, so per DSP0274 MgmtStructOffset shall be 0. */
+    if (spdm_request->mgmt_struct_offset != 0) {
+        return libspdm_generate_error_response(spdm_context,
+                                               SPDM_ERROR_CODE_INVALID_REQUEST, 0,
+                                               response_size, response);
+    }
+
+    if (!libspdm_read_slot_management_supported_subcodes(spdm_context, sub_code_bitmap)) {
+        return libspdm_generate_error_response(spdm_context,
+                                               SPDM_ERROR_CODE_INVALID_REQUEST, 0,
+                                               response_size, response);
+    }
+
+    LIBSPDM_ASSERT(*response_size >= sizeof(spdm_slot_management_response_t) +
+                   sizeof(spdm_slot_management_supported_subcodes_struct_t));
+    libspdm_zero_mem(response, *response_size);
+    *response_size = sizeof(spdm_slot_management_response_t) +
+                     sizeof(spdm_slot_management_supported_subcodes_struct_t);
+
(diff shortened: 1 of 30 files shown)


os_stub/spdm_device_secret_lib_sample/slot_management.c
```diff
@@ -0,0 +1,698 @@
...
+
+/* The maximum number of Banks this sample reports. The sample maps one Bank per distinct key
+ * pair algorithm, so this must be at least the number of distinct algorithms among the key
+ * pairs. */
+#define LIBSPDM_SAMPLE_SLOT_MANAGEMENT_BANK_COUNT 16
+
+static libspdm_slot_management_sample_bank_t
+    m_slot_management_bank[LIBSPDM_SAMPLE_SLOT_MANAGEMENT_BANK_COUNT];
+static uint8_t m_slot_management_bank_count = 0;
+static bool m_slot_management_bank_initialized = false;
```
alistair23, line 140: This seems really un-ideal. We need a bunch of state that has to be stored as a global variable as we can't store it in the actual SPDM context

os_stub/spdm_device_secret_lib_sample/slot_management.c
```diff
@@ -0,0 +1,698 @@
...
+        return false;
+    }
+
+    if (operation != SPDM_SLOT_MANAGEMENT_MANAGE_SLOT_OPERATION_ERASE) {
+        return false;
+    }
+
+    /* Erase: mark the slot's certificate chain as removed. Subsequent GetBankInfo/
+     * GetBankDetails/GetCertificateChain for this (Bank, slot) will report it as empty. */
+    slot->erased = true;
```
alistair23: But what if this bank is the actual bank in use? We now need to also push this back to libspdm if this bank is in use, based on the algorithms negotiated

os_stub/spdm_device_secret_lib_sample/slot_management.c
```diff
@@ -0,0 +1,698 @@
...
+
+    return true;
+}
+
+bool libspdm_write_slot_management_bank(
+    void *spdm_context,
+    uint8_t bank_id,
+    uint8_t operation,
+    uint32_t select_asym_algo,
+    uint32_t select_pqc_asym_algo)
```
alistair23: What if the responder requires a reset? We have no way to propagate that up

os_stub/spdm_device_secret_lib_sample/slot_management.c
```diff
@@ -0,0 +1,698 @@
...
+    if (slot == NULL) {
+        return false;
+    }
+
+    /* The certificate chain is selected by the Bank's algorithm and the Bank-local SlotID. It
+     * is read on demand and does not need to be provisioned into the SPDM context. */
+    if (!libspdm_slot_management_read_slot_cert_chain(
+            context, bank, slot, &slot_cert_chain, &slot_cert_chain_size)) {
+        return false;
+    }
```
alistair23, line 870: This won't work though. If you are using a `SLOT_MANAGEMENT` command to read the current in use bank it's possible that the chain was set using `SetCertificate`. So you need to check with the SPDM context to ensure it hasn't been changed.

alistair23 (commented):
It looks like there are some issues with syncing the data back to the current SPDM state

os_stub/spdm_device_secret_lib_sample/slot_management.c
alistair23: `libspdm_write_certificate_to_nvm()` is supposed to do this now, as it takes a `bank_id`, but it isn't wired up in the example. The `GET_CSR` helper also isn't wired up to use these management commands.

I understand that this is just a sample, but as it is the slot management doesn't look like it'll work correctly. If someone does some `SLOT_MANAGEMENT` on an active bank (which appears to be allowed in the spec) it will fall apart.

I do feel that managing the banks manually is really tricky (hence https://github.com/DMTF/libspdm/pull/3629)

os_stub/spdm_device_secret_lib_sample/slot_management.c
jyao1: I understand. But that is current design for key_pair info as well.

I feel that slot_management and key_pair_info are tangled together.
It is really hard to put everything into SPDM context, unless we totally redesign them.

But I am open for suggestions, since 4.0 can have incompatible change.

I just thought that we had discussed before in the meeting - **which part is in spdm-context, which part is left to integrator.**
But we have not written down that clearly.
I think it is time for us to **record the design principle in doc** dir.

os_stub/spdm_device_secret_lib_sample/slot_management.c
jyao1: That is good point.
Same question for key_pair_info.

os_stub/spdm_device_secret_lib_sample/slot_management.c
jyao1: Good point.
Will consider that. Maybe use the same way as others reset_required command.

os_stub/spdm_device_secret_lib_sample/slot_management.c
jyao1: In general, I agree with you that it is tricky, especially when we touch the active bank.
I think the current code for multi-key/key_info is also broken. For example, https://github.com/DMTF/libspdm/issues/3638.
SetCert is NOT synced to local context yet. For example, https://github.com/DMTF/libspdm/issues/873.
And SlotState is not fully supported. For example, https://github.com/DMTF/libspdm/issues/3645.

To address your concern and current limitation, I think we need **a clear link between: slot/bank/algo/key_pair, and that involves multiple commands such as DIGEST/CERT/SET_CERT/KEY_PAIR_INFO/SLOT_MANAGEMENT.**
Then we need **ensure one change can be propagated to other setting.**
(It is like a small database.)

The problem of #3629 is just a very small piece of a whole solution.
But I have no idea on what the whole solution looks like.
That is why I asked for a design review to understand more.
Or, please submit a full solution of slot_management for review, then I can understand how it works.

os_stub/spdm_device_secret_lib_sample/slot_management.c
alistair23: > I understand. But that is current design for key_pair info as well.
> 
> I feel that slot_management and key_pair_info are tangled together.

I would argue that doesn't mean we can't improve things here and then improve things with key_pair afterwards.

> It is really hard to put everything into SPDM context, unless we totally redesign them.

We can pretty easily allow the implementation to store a `void *` in the context (if that isn't already supported?), then at least it doesn't have to be a global.

> But I am open for suggestions, since 4.0 can have incompatible change.

https://github.com/DMTF/libspdm/pull/3629 :)

os_stub/spdm_device_secret_lib_sample/slot_management.c
alistair23: Yeah, https://github.com/DMTF/libspdm/pull/3629 is the first step. It's split out to make it easy to review and rebase (it already needs a rebase). I'm a little worried about keeping it rebased on master, as it touches lots of code.

I can work on more features, but I wanted at least a "this looks ok" before I spend more time on it.

I'm also not sure what a design review is.

I can pull some of the slot management out of this PR and add at least an example of how it could work with https://github.com/DMTF/libspdm/pull/3629 next week

jyao1: > I can work on more features, but I wanted at least a "this looks ok" before I spend more time on it.

I have to say honestly, that I cannot say "this looks ok" before I see a full picture of the solution.
For me, it is hard to approve partial of the solution, without understanding the whole solution.
It is dead-lock.

> I'm also not sure what a design review is.

Design review means to describe the whole solution, instead of a piece of it.
Then people can ask question and provide feedback, even before you make implementation.

You can use PPT, you can use pseudo-code, or you can submit a full solution.
The form does not matter. It is content that matters, and that is the full solution in your mind.

os_stub/spdm_device_secret_lib_sample/slot_management.c
jyao1: > I would argue that doesn't mean we can't improve things here and then improve things with key_pair afterwards.

Sure, we can improve the key_pair afterwards. I am not against it.

I point it to ensure that, when the new solution is designed, we consider all parts, instead of just part of it.

Also, there is a restriction. The HAL API cannot be changed in minor version update.
That means, if we want to change HAL for key_pair_info, we should either change now (in 4.0), or much later (in 5.0).

> We can pretty easily allow the implementation to store a void * in the context (if that isn't already supported?), then at least it doesn't have to be a global.

Again, I cannot say good or bad, before I see a full solution.
Please present the whole solution, and how you plan to use void *.

os_stub/spdm_device_secret_lib_sample/slot_management.c
jyao1: This is fixed now. See the new libspdm_write_slot_management_bank() API.

jyao1:
@alistair23 If you created PPT or some pseudo-code to describe the whole design, you may create a https://github.com/DMTF/libspdm/discussions item. Then we can have more discussion for non-code content.

I have created https://github.com/DMTF/libspdm/discussions/3646 for my analysis and description.
https://github.com/jyao1/libspdm/blob/slot_management/doc/slot_management.md is the current one.
https://github.com/jyao1/libspdm/blob/slot_management/doc/slot_management_database.md is the future feature on how we manage the consistency.

Feel free to ask, if you have any question on design review.

os_stub/spdm_device_secret_lib_sample/slot_management.c
alistair23: > > 
> 
> Again, I cannot say good or bad, before I see a full solution. Please present the whole solution, and how you plan to use void *.

`LIBSPDM_DATA_APP_CONTEXT_DATA` already does this. These globals should at least be moved under a `LIBSPDM_DATA_APP_CONTEXT_DATA` struct

unit_test/test_spdm_responder/slot_management.c
```diff
@@ -0,0 +1,1581 @@
...
+        /* Restore detail_request's bank_id for any later use. */
+        slot_address = (void *)(detail_request + sizeof(spdm_slot_management_request_t));
+        slot_address->bank_id = bank_id;
+    }
+
+    /* An all-zero ConfigAlgo (no bit set in SelectAsymAlgo or SelectPqcAsymAlgo) selects no
+     * algorithm. Per the "no more than one" reading, it is accepted -- not rejected -- and CLEARS
+     * the Bank's algorithm (returns it to the "not yet selected" state). The Bank's slots were
+     * erased above, so this configuration change succeeds with a SLOT_MANAGEMENT_RESP (no response
+     * structure). */
```
alistair23, line 831: This is incorrect, as discussed here: https://github.com/DMTF/SPDM-WG/issues/4558

library/spdm_responder_lib/libspdm_rsp_slot_management.c
```diff
@@ -0,0 +1,1263 @@
...
+                                               response_size, response);
+    }
+
+    /* ConfigAlgo selects the Bank's algorithm: the total number of bits set in SelectAsymAlgo and
+     * SelectPqcAsymAlgo shall be no more than one. (DSP0274 Table 145 says "exactly one", which
+     * would reject an all-zero selection; this implementation takes the "no more than one" reading,
+     * consistent with the CurrentAsymAlgo "not yet selected" state. An all-zero selection is
+     * therefore accepted: it clears the Bank's algorithm, returning it to the not-yet-selected
+     * state, which -- like any Bank configuration change -- requires the Bank's slots to be
+     * unprovisioned.) More than one bit total is malformed and rejected. */
```
alistair23, line 547: The spec is unclear, but this is not correct and will be updated in the future: https://github.com/DMTF/SPDM-WG/issues/4558

alistair23:
https://github.com/DMTF/libspdm/pull/3668 has been updated to include the fixes commented on above as well

ayj (commented):
When `ManageBank(ConfigAlgo)` or `ManageSlot(Erase)` successfully modifies a slot or bank that is currently associated with an active SPDM session (or the currently selected `connection_info.current_bank`), does `libspdm` automatically terminate those active sessions (`libspdm_session_state_not_started`) or invalidate cached transcript hashes? 

If session cleanup is currently left to the HAL implementation inside `libspdm_write_slot_management_*`, should the core responder explicitly drop active sessions when the active bank's cryptographic capabilities change to guarantee state machine integrity per SPDM 1.4 rules?

PR #3642, opened 2026-06-05
Fix key pair info compliance

jyao1:


library/spdm_responder_lib/libspdm_rsp_key_pair_info.c
```diff
@@ -142,8 +142,14 @@ libspdm_return_t libspdm_get_response_key_pair_info(libspdm_context_t *spdm_cont
 
     spdm_response = response;
     if (spdm_request->header.spdm_version >= SPDM_MESSAGE_VERSION_14) {
-        LIBSPDM_ASSERT(*response_size >= sizeof(spdm_key_pair_info_response_t) + sizeof(uint32_t)*4);
-        *response_size = sizeof(spdm_key_pair_info_response_t) + public_key_info_len + sizeof(uint32_t)*4;
+        /* The SPDM 1.4 tail is PqcAsymAlgoCapLen (1) + PqcAsymAlgoCapabilities (4) +
+         * CurrentPqcAsymAlgoLen (1) + CurrentPqcAsymAlgo (4) = 10 bytes. */
```
alistair23, line 146: Where do you see 4 bytes for `PqcAsymAlgoCapabilities`? I don't see anything saying 4 bytes

alistair23 (changes requested):

library/spdm_responder_lib/libspdm_rsp_key_pair_info.c
jyao1: right. The spec does not say that.
It is responder's implementation choice.

alistair23 (approved):

PR #3653, opened 2026-06-13
 [issue-3185] When SPDM_GET_CAPABILITIES_REQUEST_FLAGS_EP_INFO_CAP_SIG is negotiated,

### LINKED ISSUE #3185: Need to add (asym_algo != 0) check if (SPDM_GET_CAPABILITIES_RESPONSE_FLAGS_EP_INFO_CAP_SIG != 0), in ALGORITHM function
It should be done in both requester and responder, and align with other SIGNATURE related CAP.

owlHJ:
## Problem

When `SPDM_GET_CAPABILITIES_REQUEST_FLAGS_EP_INFO_CAP_SIG` is negotiated, the encap `GET_ENDPOINT_INFO` flow requires the requester to sign its response using `req_base_asym_alg`. If `NEGOTIATE_ALGORITHMS` results in `req_base_asym_alg == 0` (empty intersection between requester andresponder algorithm sets), the downstream signing path silently computes a zero-length signature, violating the protocol.

The root cause is that `NEGOTIATE_ALGORITHMS` validation already enforces `req_base_asym_alg != 0` for `MUT_AUTH_CAP`, but had no equivalent guard for `EP_INFO_CAP_SIG`.

## Details

`libspdm_rsp_encap_get_endpoint_info.c` uses `req_base_asym_alg` directly to compute `signature_size`. If `req_base_asym_alg == 0`, then `signature_size == 0`, which is a protocol violation
```c
/* libspdm_rsp_encap_get_endpoint_info.c:185 */
signature_size = libspdm_get_req_asym_signature_size(
    spdm_context->connection_info.algorithm.req_base_asym_alg);
/* if req_base_asym_alg == 0 → signature_size == 0 → protocol violation */
```

## Fix

- Extend the existing `req_base_asym_alg` validation block in both the requester (`libspdm_req_negotiate_algorithms.c`) and the responder to also trigger when `EP_INFO_CAP_SIG` is active, mirroring the existing `MUT_AUTH_CAP` treatment.

- `resp_flag` is kept at `0` because `EP_INFO_CAP_SIG` is a requester-only capability. This is consistent with its usage throughout the encap code path (`libspdm_rsp_encap_get_endpoint_info.c` lines 74 and 178).

## Unit Tests

| File | Case | Scenario | Expected |
|------|------|----------|----------|
| `test_spdm_requester/.../negotiate_algorithms_err.c` | 44 | EP_INFO_CAP_SIG set; ALGORITHMS response has `req_base_asym_alg == 0` | `LIBSPDM_STATUS_NEGOTIATION_FAIL` |
| `test_spdm_requester/.../negotiate_algorithms_err.c` | 45 | EP_INFO_CAP_SIG set; ALGORITHMS response has valid `req_base_asym_alg` | `LIBSPDM_STATUS_SUCCESS` |
| `unit_test/test_spdm_responder/algorithms.c` | 33 | EP_INFO_CAP_SIG set; `req_base_asym_alg` intersection with local is 0 | `SPDM_ERROR_CODE_INVALID_REQUEST` |
| `unit_test/test_spdm_responder/algorithms.c` | 34 | EP_INFO_CAP_SIG set; `req_base_asym_alg` intersection with local is non-zero | `SPDM_ALGORITHMS` (success) |
```
Fixes #3185

### DIFF
--- a/unit_test/test_spdm_requester/error_test/negotiate_algorithms_err.c
+++ b/unit_test/test_spdm_requester/error_test/negotiate_algorithms_err.c
@@ -129,6 +129,8 @@ static libspdm_return_t send_message(
         return LIBSPDM_STATUS_SUCCESS;
     case 0x2C:
         return LIBSPDM_STATUS_SUCCESS;
+    case 0x2D:
+        return LIBSPDM_STATUS_SUCCESS;
     default:
         return LIBSPDM_STATUS_SEND_FAIL;
     }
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: unit_test/test_spdm_requester/error_test/negotiate_algorithms_err.c [hunk 1]
```diff
@@ -3046,6 +3140,74 @@ static void libspdm_test_requester_negotiate_algorithms_error_case43(void** stat
...
+    spdm_context->local_context.capability.flags |=
+        SPDM_GET_CAPABILITIES_REQUEST_FLAGS_EP_INFO_CAP_SIG;
+
+    status = libspdm_negotiate_algorithms(spdm_context);
+    assert_int_equal(status, LIBSPDM_STATUS_NEGOTIATION_FAIL);
+}
+
+/**
+ * Test 45: EP_INFO_CAP_SIG is negotiated and req_base_asym_alg in the ALGORITHMS response is valid.
+ * Expected behavior: returns with status LIBSPDM_STATUS_SUCCESS.
```
steven-bellock: This test case should not be in this file, as it is only for error / negative test cases.

unit_test/test_spdm_responder/algorithms.c
```diff
@@ -2887,6 +2887,101 @@ static void rsp_algorithms_case32(void **state)
     assert_int_equal(spdm_context->connection_info.algorithm.measurement_spec, 0);
 }
 
+/**
+ * Test 33: EP_INFO_CAP_SIG is negotiated but req_base_asym_alg negotiation results in 0.
+ * Expected behavior: returns SPDM_ERROR_CODE_INVALID_REQUEST.
```
steven-bellock, line 2892: This test case should not be in this file, as it is only for successful / positive test cases.

steven-bellock (changes requested):

unit_test/test_spdm_requester/error_test/negotiate_algorithms_err.c
owlHJ: Thank you for advice. Test 45 asserts success, so I'll move it to `negotiate_algorithms.c` and keep only the `negative Test 44` here.

unit_test/test_spdm_responder/algorithms.c
owlHJ: Agreed, this is a negative case. I couldn't find an algorithms_err.c under the responder `error_test/` directory. I also noticed that `algorithms.c` already contains a number of negative cases asserting SPDM_ERROR (e.g. around lines 1003, 1228, 2336). 

So should I create a new `error_test/algorithms_err.c` and move `Test 33` there, or would you prefer it stay in `algorithms.c` alongside the existing negative cases?

steven-bellock: Yes, you are right. Let's keep it here and we can move the error tests to the `_err.c` file when it is created in a later pull request.

steven-bellock (approved):

PR #3658, opened 2026-06-19
Add a update local cert chain callback

### LINKED ISSUE #873: [SET_CERTIFICATE] Save certificate into SPDM context

alistair23:
After a successful SET_CERTIFICATE (that doesn't require a reset) this
new callback is called. It's the implementations responsibility to
allocate memory to store the new certificate chain and update
`LIBSPDM_DATA_LOCAL_PUBLIC_CERT_CHAIN` to use the new chain.

After which the old certificate chain can be freed. Unfortunately we can't
handle this all in libspdm as it requires allocation, which is why the
HAL must handle this.

For DEVICE_CERT and GENERIC_CERT this is a simple. All that needs to be done
is memory allocated, the new certificate copied and the old certificate freed.

The ALIAS_CERT is similar, but required combining the new updated certificates
with the existing ones that aren't changed.

While we are at it we need to fix a bug with the current `AliasCert` certificate chains.
The SPDM spec states that in the AliasCert model "the Device Certificate
CA should contain the Hardware identity OID." This was previously
missing, let's ensure it's included and update the certificates to
include it.

Resolves: https://github.com/DMTF/libspdm/issues/873

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_set_certificate_rsp.c
+++ b/library/spdm_responder_lib/libspdm_rsp_set_certificate_rsp.c
@@ -73,11 +73,17 @@ libspdm_return_t libspdm_get_response_set_certificate(libspdm_context_t *spdm_co
     const spdm_cert_chain_t *cert_chain_header;
     size_t cert_chain_size;
     const void * cert_chain;
+    const void *full_cert_chain;
+    size_t full_cert_chain_size;
+    const void *old_local_cert_chain;
+    size_t old_local_cert_chain_size;
 
     libspdm_session_info_t *session_info;
     libspdm_session_state_t session_state;
 
     spdm_request = request;
+    full_cert_chain = NULL;
+    full_cert_chain_size = 0;
 
     /* -=[Check Parameters Phase]=- */
     LIBSPDM_ASSERT(spdm_request->header.request_response_code == SPDM_SET_CERTIFICATE);
@@ -156,6 +162,11 @@ libspdm_return_t libspdm_get_response_set_certificate(libspdm_context_t *spdm_co
     root_cert_hash_size = libspdm_get_hash_size(
         spdm_context->connection_info.algorithm.base_hash_algo);
 
+    old_local_cert_chain =
+        spdm_context->local_context.local_cert_chain_provision[slot_id];
+    old_local_cert_chain_size =
+        spdm_context->local_context.local_cert_chain_provision_size[slot_id];
+
     if (spdm_version >= SPDM_MESSAGE_VERSION_13) {
         const uint8_t key_pair_id = spdm_request->header.param2;
 
@@ -217,9 +228,13 @@ libspdm_return_t libspdm_get_response_set_certificate(libspdm_context_t *spdm_co
         }
 
         /* erase slot_id cert_chain*/
-        result = libspdm_write_certificate_to_nvm(
+        result = libspdm_update_local_cert_chain(
             spdm_context,
-            slot_id, NULL, 0, 0, 0, 0,
+            slot_id, 0, 0, 0, 0,
+            old_local_cert_chain,
+            old_local_cert_chain_size,
+            NULL, 0,
+            set_cert_model,
             &need_reset, &is_busy);
         if (!result) {
             if (is_busy) {
@@ -243,6 +258,8 @@ libspdm_return_t libspdm_get_response_set_certificate(libspdm_context_t *spdm_co
         /*point to full SPDM certificate chain*/
         cert_chain = (const void*)(spdm_request + 1);
         cert_chain_header = cert_chain;
+        full_cert_chain = cert_chain;
+        full_cert_chain_size = cert_chain_header->length;
 
         if (cert_chain_header->length < sizeof(spdm_cert_chain_t) + root_cert_hash_size) {
             return libspdm_generate_error_response(spdm_context,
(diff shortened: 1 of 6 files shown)


unit_test/sample_key/openssl.cnf
```diff
@@ -65,6 +65,11 @@ keyUsage = nonRepudiation, digitalSignature, keyEncipherment, keyCertSign
 subjectKeyIdentifier = hash
 subjectAltName = otherName:1.3.6.1.4.1.412.274.1;UTF8:ACME:WIDGET:1234567890
 extendedKeyUsage = critical, serverAuth, clientAuth, OCSPSigning
+1.3.6.1.4.1.412.274.6 = ASN1:SEQUENCE:id_spdm_cert_oids
+[id_spdm_cert_oids]
+field1 = SEQUENCE:id_spdm_cert_oid
+[id_spdm_cert_oid]
+field1 = OID:1.3.6.1.4.1.412.274.2
```
jyao1, line 72: When you generate the cert, would you please handle cert for slot 4 as well?

jyao1:
Could you please split this one to 3 different PR?

1) Update certs to add Hardware OID 
2) Split out hardware ID OID finding function
3) Support local cert chain update.

I am OK to merge 1) soon. But we need more time to review 2) and 3).

unit_test/sample_key/openssl.cnf
alistair23: Fixed in https://github.com/DMTF/libspdm/pull/3663

alistair23:
> Could you please split this one to 3 different PR?
> 
>     1. Update certs to add Hardware OID
> 
>     2. Split out hardware ID OID finding function
> 
>     3. Support local cert chain update.
> 
> 
> I am OK to merge 1) soon. But we need more time to review 2) and 3).

I have opened https://github.com/DMTF/libspdm/pull/3662 and https://github.com/DMTF/libspdm/pull/3663

jyao1:
I have merged other 2. Please rebase.

alistair23:
Done!

library/spdm_responder_lib/libspdm_rsp_set_certificate_rsp.c
```diff
@@ -326,6 +332,27 @@ libspdm_return_t libspdm_get_response_set_certificate(libspdm_context_t *spdm_co
...
                                                response_size, response);
     } else {
+        /* If no reset is required we call back to have the local cert chain updated */
+        if (full_cert_chain != NULL) {
+            const void *old_local_cert_chain =
+                spdm_context->local_context.local_cert_chain_provision[slot_id];
+            size_t old_local_cert_chain_size =
+                spdm_context->local_context.local_cert_chain_provision_size[slot_id];
+
+            if (!libspdm_update_local_cert_chain(spdm_context, slot_id,
```
jyao1: I am not sure why we have to use this new HAL.

Can libspdm_write_certificate_to_nvm() implementation update the spdm_context->local_context.local_cert_chain_provision directly?

It can use libspdm_set_data (LIBSPDM_DATA_LOCAL_PUBLIC_CERT_CHAIN).

jyao1:
> While we are at it we need to fix a bug with the current AliasCert certificate chains.
The SPDM spec states that in the AliasCert model "the Device Certificate
CA should contain the Hardware identity OID."

BTW: I do not think this is a bug.
"SHOULD" means recommendation, but not a requirement.
I do not mind to add it. But it should be allowed if no Hardware identity OID in the cert chain.

library/spdm_responder_lib/libspdm_rsp_set_certificate_rsp.c
alistair23: Kind of. In theory `libspdm_write_certificate_to_nvm()` could be used to do this. It is not being used though, which I think indicates that it isn't fit for this purpose.

`libspdm_write_certificate_to_nvm()` isn't provided the right information, for example it's missing the full chain (with length and hash header), it's missing the cert model information (alias or device). Again that could in theory be determined using `libspdm_get_data()`, but considering no one is, it's clearly too clunky and confusing.

On top of that there is no `LIBSPDM_DATA_LOCAL_PUBLIC_CERT_CHAIN` support for `libspdm_get_data()`, so today you can't actually check against the existing cert in `libspdm_write_certificate_to_nvm()`, which is important for the AliasCert update.

Based on that, I do think `libspdm_write_certificate_to_nvm()` should be used for only writing the certificate to nvm, as the function describes and this new callback can handle updating the actual internal state.

jyao1: > which I think indicates that it isn't fit for this purpose.

I do not think so. It is a bug.

> Based on that, I do think libspdm_write_certificate_to_nvm() should be used for only writing the certificate to nvm, as the function describes and this new callback can handle updating the actual internal state.

IMHO, the reason to define 2 APIs is that 2 APIs can be used independently, separately.
If 2 APIs must be used together, I do not see the value to define 2 and let the caller invoke twice.

We can do the HAL API change in major release.
If you think the name is confusing, please propose a better name.

alistair23: Fixed

jyao1 (approved):

PR #3659, opened 2026-06-22
spdm_device_secret_lib_sample: Fix build when GET_KEY_PAIR_INFO_CAP is disabled

alistair23:
Commit 1e16c60f6d30 broke builds when
LIBSPDM_ENABLE_CAPABILITY_GET_KEY_PAIR_INFO_CAP is not enabled as libspdm_get_key_pair_id_by_slot() isn't defined. Fix the build by guarding calls of libspdm_get_key_pair_id_by_slot().

Fixes: 1e16c60f6d30 ("Add slot 4 multi-key example with a distinct responder leaf key")

### DIFF
--- a/os_stub/spdm_device_secret_lib_sample/read_priv_key_pem.c
+++ b/os_stub/spdm_device_secret_lib_sample/read_priv_key_pem.c
@@ -67,6 +67,7 @@ bool libspdm_read_responder_private_key(uint32_t base_asym_algo,
 bool libspdm_read_responder_private_key_ex(uint32_t base_asym_algo, uint8_t key_pair_id,
                                            void **data, size_t *size)
 {
+#if LIBSPDM_ENABLE_CAPABILITY_GET_KEY_PAIR_INFO_CAP
     bool res;
     char *file;
 
@@ -114,6 +115,9 @@ bool libspdm_read_responder_private_key_ex(uint32_t base_asym_algo, uint8_t key_
     }
     res = libspdm_read_input_file(file, data, size);
     return res;
+#else
+    return libspdm_read_responder_private_key(base_asym_algo, data, size);
+#endif
 }
 #endif
 
--- a/os_stub/spdm_device_secret_lib_sample/read_priv_key_pem_pqc.c
+++ b/os_stub/spdm_device_secret_lib_sample/read_priv_key_pem_pqc.c
@@ -82,6 +82,7 @@ bool libspdm_read_responder_pqc_private_key(uint32_t pqc_asym_algo,
 bool libspdm_read_responder_pqc_private_key_ex(uint32_t pqc_asym_algo, uint8_t key_pair_id,
                                                void **data, size_t *size)
 {
+#if LIBSPDM_ENABLE_CAPABILITY_GET_KEY_PAIR_INFO_CAP
     bool res;
     char *file;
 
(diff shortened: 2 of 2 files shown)


os_stub/spdm_device_secret_lib_sample/read_priv_key_pem.c
```diff
@@ -64,6 +64,7 @@ bool libspdm_read_responder_private_key(uint32_t base_asym_algo,
     return res;
 }
 
+#if LIBSPDM_ENABLE_CAPABILITY_GET_KEY_PAIR_INFO_CAP
```
jyao1: I doubt if this is a right fix to exclude this API just because GET_KEY_PAIR_INFO_CAP is absent.

If the root-cause is libspdm_get_key_pair_id_by_slot(), then I think we should put `#if LIBSPDM_ENABLE_CAPABILITY_GET_KEY_PAIR_INFO_CAP` around libspdm_get_key_pair_id_by_slot(), and always reture libspdm_read_responder_private_key() if GET_KEY_PAIR_INFO_CAP is absent.

alistair23: The APIs were added as part of https://github.com/DMTF/libspdm/commit/1e16c60f6d307dc20ab7e27b6810bfd6070ce022, so I just assumed they should be gated.

I have changed the checkout to just be around the specific functions

jyao1: Thanks for the update.

It seems the new change always return slot 4.

But I mean to "always return libspdm_read_responder_private_key() if GET_KEY_PAIR_INFO_CAP is absent."

The reason is that: If GET_KEY_PAIR_INFO is not there, then MULTI_KEY should be disabled. As such, all keys should be the key from slot 0, not slot 4.

See the example below: 
```
bool libspdm_read_responder_private_key_ex(uint32_t base_asym_algo, uint8_t key_pair_id,
                                           void **data, size_t *size)
{
#if LIBSPDM_ENABLE_CAPABILITY_GET_KEY_PAIR_INFO_CAP
    ... // original function
#else
        return libspdm_read_responder_private_key(base_asym_algo, data, size);
#endif
}
```

os_stub/spdm_device_secret_lib_sample/read_priv_key_pem.c
```diff
@@ -70,12 +70,14 @@ bool libspdm_read_responder_private_key_ex(uint32_t base_asym_algo, uint8_t key_
     bool res;
     char *file;
 
+#if LIBSPDM_ENABLE_CAPABILITY_GET_KEY_PAIR_INFO_CAP
```
jyao1: Thanks for the update.

It seems the new change always return slot 4.

But I mean to "always return libspdm_read_responder_private_key() if GET_KEY_PAIR_INFO_CAP is absent."

The reason is that: If GET_KEY_PAIR_INFO is not there, then MULTI_KEY should be disabled. As such, all keys should be the key from slot 0, not slot 4.

See the example below: 
```
bool libspdm_read_responder_private_key_ex(uint32_t base_asym_algo, uint8_t key_pair_id,
                                           void **data, size_t *size)
{
#if LIBSPDM_ENABLE_CAPABILITY_GET_KEY_PAIR_INFO_CAP
    ... // original function
#else
        return libspdm_read_responder_private_key(base_asym_algo, data, size);
#endif
}
```

alistair23: Fixed

jyao1 (approved):

PR #3667, opened 2026-06-26
[issue-3590] Allow SlotSizeRequested for an empty certificate slot

### LINKED ISSUE #3590: GET_CERTIFICATE Handling Missing Strict Responder Gate for SLOT_SIZE_REQUESTED Without SET_CERT_CAP
The responder path for GET_CERTIFICATE/CERTIFICATE does not explicitly enforce a strict capability check that ties SLOT_SIZE_REQUESTED usage to negotiated SET_CERT_CAP. In other words, when SET_CERT_CAP is not negotiated, there is no explicit responder-side rejection path for requests using SLOT_SIZE_REQUESTED.

Observed Behavior

Request processing does not include an explicit strict gate for SLOT_SIZE_REQUESTED against negotiated SET_CERT_CAP.
Behavior may rely on tolerant parsing/implicit handling instead of deterministic reject logic.
Expected Behavior

If SET_CERT_CAP is not negotiated, responder should explicitly reject GET_CERTIFICATE requests that set/use SLOT_SIZE_REQUESTED (with protocol-appropriate error handling).
Impact

Spec-compliance and interop ambiguity.
Inconsistent behavior across implementations during negative/strict conformance testing.
Severity
Medium (non-security protocol compliance gap, unless project policy requires strict reject semantics).

Suggested Fix
Add an explicit responder-side validation gate:

Check SLOT_SIZE_REQUESTED usage in GET_CERTIFICATE.
Verify SET_CERT_CAP negotiation state.
Return defined protocol error when the flag is used without capability negotiation.
Add unit tests for both allowed and rejected combinations.

owlHJ:
## Summary

A `GET_CERTIFICATE` request with the `SlotSizeRequested` attribute asks only for the slot size. Per the SPDM-WG discussion behind DSP0274, this result is **independent** of whether the slot currently holds a certificate chain:

| Condition                                | Returned size              |
| ---------------------------------------- | -------------------------- |
| `SET_CERTIFICATE` supported              | maximum provisionable size |
| no `SET_CERTIFICATE`, slot **has** chain | size of the chain          |
| no `SET_CERTIFICATE`, slot **empty**     | 0                          |

The responder previously made the "empty slot" outcomes unreachable.

## Problem

`libspdm_get_response_certificate()` rejected an empty slot **before** ever reaching the size logic:

1. The `local_cert_chain_provision[slot_id] == NULL` check returned `InvalidRequest`.
2. A later `offset >= cert_chain_size` check rejected the same case again. As a result, both the **"empty slot → 0"** outcome and the **"empty slot under SET_CERT_CAP → maximum size"** outcome could never be produced, even though a `SlotSizeRequested` query is valid against an empty slot.

## Changes

- Compute a `slot_size_requested` flag (SPDM 1.3+ and the `SLOT_SIZE_REQUESTED` attribute set).
- Bypass the `NULL` rejection when `slot_size_requested` is set.
- Treat a `NULL` provision as `cert_chain_size = 0`.
- Skip the certificate copy when the portion length is 0 (the source pointer may be `NULL` for an empty slot), avoiding a NULL dereference.

The size still flows through the existing response path, so the `SET_CERT_CAP` HAL override, transcript logging, and connection-state update are unchanged. Non-`SlotSizeRequested` requests behave exactly as before.

## Testing

- Added responder unit test **case 20**: a `SlotSizeRequested` query against an empty slot returns `SUCCESS` with `portion_length == 0` and `remainder_length` equal to the slot storage size — the maximum size with `SET_CERT_CAP`, otherwise 0.
- The existing **case 18** already covers the non-empty slot.

## Related

Fixes #3590

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_certificate.c
+++ b/library/spdm_responder_lib/libspdm_rsp_certificate.c
@@ -24,6 +24,7 @@ libspdm_return_t libspdm_get_response_certificate(libspdm_context_t *spdm_contex
     libspdm_session_info_t *session_info;
     libspdm_session_state_t session_state;
     bool use_large_cert_chain;
+    bool slot_size_requested;
     uint32_t req_msg_header_size;
     uint32_t rsp_msg_header_size;
     size_t cert_chain_size;
@@ -121,13 +122,28 @@ libspdm_return_t libspdm_get_response_certificate(libspdm_context_t *spdm_contex
                                                response_size, response);
     }
 
-    if (spdm_context->local_context.local_cert_chain_provision[slot_id] == NULL) {
+    /* The SlotSizeRequested attribute is only defined for SPDM 1.3 and later; in earlier
+     * versions the param2 bit is reserved and must be ignored. */
+    if (spdm_request->header.spdm_version >= SPDM_MESSAGE_VERSION_13) {
+        slot_size_requested =
+            ((spdm_request->header.param2 &
+              SPDM_GET_CERTIFICATE_REQUEST_ATTRIBUTES_SLOT_SIZE_REQUESTED) != 0);
+    } else {
+        slot_size_requested = false;
+    }
+
+    if (!slot_size_requested &&
+        (spdm_context->local_context.local_cert_chain_provision[slot_id] == NULL)) {
         return libspdm_generate_error_response(
             spdm_context, SPDM_ERROR_CODE_INVALID_REQUEST,
             0, response_size, response);
     }
 
-    cert_chain_size = spdm_context->local_context.local_cert_chain_provision_size[slot_id];
+    if (spdm_context->local_context.local_cert_chain_provision[slot_id] == NULL) {
+        cert_chain_size = 0;
+    } else {
+        cert_chain_size = spdm_context->local_context.local_cert_chain_provision_size[slot_id];
+    }
 
     if ((spdm_request->header.spdm_version >= SPDM_MESSAGE_VERSION_14) &&
         (!use_large_cert_chain) && (cert_chain_size > SPDM_MAX_CERTIFICATE_CHAIN_SIZE)) {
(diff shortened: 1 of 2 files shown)


library/spdm_responder_lib/libspdm_rsp_certificate.c
```diff
@@ -121,13 +122,25 @@ libspdm_return_t libspdm_get_response_certificate(libspdm_context_t *spdm_contex
                                                response_size, response);
     }
 
-    if (spdm_context->local_context.local_cert_chain_provision[slot_id] == NULL) {
+    /* A SlotSizeRequested query returns only a size and is independent of whether the slot
+    * currently holds a certificate chain, so it must not be rejected for an empty slot. */
+    slot_size_requested =
```
steven-bellock: We tend to prefer

```
if (spdm_request->header.spdm_version >= SPDM_MESSAGE_VERSION_13) {
    slot_size_requested = ...;
} else {
    slot_size_requested = false;
}
```

owlHJ: @steven-bellock  thanks. it is so simpler than my first code. 

```
-    /* A SlotSizeRequested query returns only a size and is independent of whether the slot
-    * currently holds a certificate chain, so it must not be rejected for an empty slot. */
-    slot_size_requested =
-        (spdm_request->header.spdm_version >= SPDM_MESSAGE_VERSION_13) &&
-        ((spdm_request->header.param2 &
-          SPDM_GET_CERTIFICATE_REQUEST_ATTRIBUTES_SLOT_SIZE_REQUESTED) != 0);
+    /* The SlotSizeRequested attribute is only defined for SPDM 1.3 and later; in earlier
+     * versions the param2 bit is reserved and must be ignored. */
+    if (spdm_request->header.spdm_version >= SPDM_MESSAGE_VERSION_13) {
+        slot_size_requested =
+            ((spdm_request->header.param2 &
+              SPDM_GET_CERTIFICATE_REQUEST_ATTRIBUTES_SLOT_SIZE_REQUESTED) != 0);
+    } else {
+        slot_size_requested = false;
+    }
```

library/spdm_responder_lib/libspdm_rsp_certificate.c
```diff
@@ -121,13 +122,28 @@ libspdm_return_t libspdm_get_response_certificate(libspdm_context_t *spdm_contex
...
+    } else {
+        slot_size_requested = false;
+    }
+
+    if (!slot_size_requested &&
+        (spdm_context->local_context.local_cert_chain_provision[slot_id] == NULL)) {
         return libspdm_generate_error_response(
             spdm_context, SPDM_ERROR_CODE_INVALID_REQUEST,
             0, response_size, response);
     }
```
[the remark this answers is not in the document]
jyao1: disagree.
Function is still valid without SET_CERT_CAP

PR #3668, opened 2026-06-29
Add Slot management

### LINKED ISSUE #3638: In device sample, `assoc_cert_slot_mask` not consistent with slot certs or `DIGESTS` per-slot `KeyPairID`
**Location:** `os_stub/spdm_device_secret_lib_sample/key_pair.c`
(`libspdm_init_key_pair_info`, the `m_key_pair_info[index].assoc_cert_slot_mask`
assignments) versus `library/spdm_responder_lib/libspdm_rsp_digests.c:155` and
the emu configuration in
`spdm-emu/spdm_emu/spdm_responder_emu/spdm_responder_spdm.c:566`–`:569`.

**Spec basis:** The key-pair⇄slot association is a single device property viewed
from two messages, and the two views must agree:

- `KEY_PAIR_INFO.AssocCertSlotMask` (Table 111, §3346) — "a bit mask
  representing the currently associated certificate slots" for a given
  `KeyPairID`.
- `DIGESTS.KeyPairID[slot]` (Table 9, lines 1473/1475) — "shall be the
  `KeyPairID` of the key pair associated with … `CertSlot[slot]`", present when
  `MULTI_KEY_CONN_RSP` is true.

These are inverse mappings of the same relation, so for every slot X,
`DIGESTS.KeyPairID[X]` must equal the `KeyPairID` whose `AssocCertSlotMask` has
bit X set, and that key pair's `CurrentAsymAlgo` must match the algorithm of the
certificate actually installed in slot X.

**What the sample/emu actually do — three independent, contradictory sources:**

1. **`KEY_PAIR_INFO` (HAL table).** Each key pair is mapped to one distinct slot
   via a single-bit mask whose bit index equals `KeyPairID - 1`, and each key
   pair carries a *different* `CurrentAsymAlgo`:

   | KeyPairID | `CurrentAsymAlgo` | `assoc_cert_slot_mask` | → slot |
   |-----------|-------------------|------------------------|--------|
   | 1 | RSA2048 | `0x01` | 0 |
   | 2 | RSA3072 | `0x02` | 1 |
   | 3 | RSA4096 | `0x04` | 2 |
   | 4 | ECC256  | `0x08` | 3 |
   | 5 | ECC384  | `0x10` | 4 |
   | 6 | ECC521  | `0x20` | 5 |
   | 7 | SM2     | `0x40` | 6 |
   | 8 | Ed25519 | `0x80` | 7 |
   | 9 / 10–12 | Ed448 / ML-DSA | `0x00` | none |

2. **`DIGESTS` (responder).** `libspdm_rsp_digests.c:155` fills each slot's
   `KeyPairID` from `local_context.local_key_pair_id[slot]`, a separate array.
   The responder emu sets it to `0xA0 + slot_index`
   (`spdm_responder_spdm.c:566`–`:569`); the requester emu uses `0xB0 + slot_index`.
   Unset, it defaults to `0`.

3. **Slot certificate contents.** Every populated slot holds a certificate chain
   for the single *negotiated* algorithm (e.g. all ECC384 in a typical run), and
   the emu reports `CertificateInfo = DeviceCert` for each.

**The contradictions:**

- For slot 3, `KEY_PAIR_INFO` claims KeyPairID 4, while `DIGESTS` reports
  KeyPairID `0xA3` — the two never match (and `0xA0`–`0xA7` are not even valid
  indices into the 1-based key-pair table, which only has IDs 1–12).
- `KEY_PAIR_INFO` claims slot 3 ↔ ECC256, slot 0 ↔ RSA2048, etc., but the
  certificate physically installed in those slots uses the negotiated algorithm,
  not the per-key-pair algorithm the table asserts.
- The mask only covers slots 0–7 by aligning bit = `KeyPairID − 1`; key pairs
  9–12 (Ed448, the three ML-DSA PQC pairs) get `0x00`, so they are reported as
  associated with no slot at all.

**Assessment.** The literal values (e.g. `0x08`) are *format*-legal under
Tables 111/112 — each is a valid single-bit mask, and because every sample key
pair sets `SPDM_KEY_PAIR_CAP_MASK` (which includes `ShareableCap`), even
multi-bit masks would be permitted. `GET_KEY_PAIR_INFO` is also defined to report
key-pair data "independent of any negotiated parameters" (§3314), so the handler
returning this canned data is conformant in isolation, and the unit tests (which
check one message at a time) pass.

The defect is **device-model fidelity**, not message-handler conformance: a
Requester that correlates `KEY_PAIR_INFO.AssocCertSlotMask` with
`DIGESTS.KeyPairID[]` and the installed certificates — exactly what the multi-key
feature exists to enable — sees three mutually inconsistent answers. Severity is
**Low** (sample/emu fixture data, no library "shall" violated), but it makes the
sample unsuitable as a reference for the association semantics and can mislead
integrators.

alistair23:
This builds on top of the [pending PR for bank support in libspdm](https://github.com/DMTF/libspdm/pull/3629) and utilises some parts of https://github.com/DMTF/libspdm/pull/3637, while also replacing it.

This replaces https://github.com/DMTF/libspdm/pull/3637 with a few minor fixes, but otherwise mostly just rebases https://github.com/DMTF/libspdm/pull/3637 on https://github.com/DMTF/libspdm/pull/3629 to have libspdm manage the banks instead of the HAL.

As part of the review process this PR has also expanded to fixup the existing data consistency issues with KeyPairs in libspdm.

Resolves: https://github.com/DMTF/libspdm/issues/3638


include/hal/library/responder/setcertlib.h
```diff
@@ -103,6 +104,7 @@ uint32_t libspdm_get_cert_chain_slot_storage_size(
  **/
 extern bool libspdm_update_local_cert_chain(
     void *spdm_context,
+    uint8_t bank_id,
```
jyao1, line 107: I suggest we use `uint8_t *bank_id`, to support legacy SET_CERT API, where bank_id is not provided.

See my HAL update at https://github.com/DMTF/libspdm/pull/3637/changes/613af0a4d9b2a69f97d849ac226530703b7dbd4d.

alistair23:
Ready to go!

include/internal/libspdm_common_lib.h
```diff
@@ -148,6 +148,9 @@ typedef struct {
     /* Peer CertificateChain */
     libspdm_peer_used_cert_chain_t peer_used_cert_chain[SPDM_MAX_SLOT_COUNT];
 
+    /* The local certificate bank to use */
+    uint8_t current_bank;
```
jyao1, line 156: I do not think this is really needed.

This could be managed by integrator internally. No need to expose to spdm_context.

include/internal/libspdm_common_lib.h
```diff
@@ -99,13 +99,13 @@ typedef struct {
     libspdm_secured_message_version_t secured_message_version;
 
     /* My Certificate */
-    const void *local_cert_chain_provision[SPDM_MAX_SLOT_COUNT];
-    size_t local_cert_chain_provision_size[SPDM_MAX_SLOT_COUNT];
+    const void *local_cert_chain_provision[LIBSPDM_MAX_BANK_COUNT][SPDM_MAX_SLOT_COUNT];
+    size_t local_cert_chain_provision_size[LIBSPDM_MAX_BANK_COUNT][SPDM_MAX_SLOT_COUNT];
```
jyao1, line 104: I do not see the need to expose to spdm_context.
The reason we expose local cert for negotiated algo is that it can be used for transcript calculation.

The rest banks could be managed by integrator internally.

include/library/spdm_common_lib.h
```diff
@@ -78,6 +78,10 @@ typedef enum {
     LIBSPDM_DATA_LOCAL_KEY_USAGE_BIT_MASK,
     LIBSPDM_DATA_LOCAL_CURRENT_BANK,
     LIBSPDM_DATA_LOCAL_SLOT_MANAGEMENT_SUBCODES,
+    LIBSPDM_DATA_LOCAL_BANK_ASYM_ALGO,
+    LIBSPDM_DATA_LOCAL_BANK_PQC_ASYM_ALGO,
+    LIBSPDM_DATA_LOCAL_BANK_ASYM_ALGO_CAPABILITIES,
+    LIBSPDM_DATA_LOCAL_BANK_PQC_ASYM_ALGO_CAPABILITIES,
```
jyao1: I am not sure why we have to expose these.
Just like what we did for KeyPair management, the integrator can manage those bank info internally, and return such info in HAL API.

jyao1:
A fundamental question: Since spdm_context does not include key_pair_info, how do you sync key_pair_info with the slot_management/bank information? How can you guarantee they are still in sync?

To me, it is weird design to give key_pair_info to integrator and only expand lots of bank data into spdm_context. The inconsistency worried me a lot.

include/internal/libspdm_common_lib.h
alistair23: The `spdm_context` can keep track of the currently used bank. Otherwise it needs to call out to the integrator all of the time to determine bank details

include/internal/libspdm_common_lib.h
alistair23: This way libspdm can handle the banks, as we don't need to put the complexity back onto users

include/library/spdm_common_lib.h
alistair23: This is a lot simpler. You set it once at startup (it can then be updated as well) and libspdm handles it for you. It makes it a lot easier to use and integrate

alistair23:
> A fundamental question: Since spdm_context does not include key_pair_info, how do you sync key_pair_info with the slot_management/bank information? How can you guarantee they are still in sync?
> 
> To me, it is weird design to give key_pair_info to integrator and only expand lots of bank data into spdm_context. The inconsistency worried me a lot.

Yeah, key_pair_info will need to be fixed and included in libspdm. Then all of the sync issues will be solved!

I agree it's a bit weird at the moment, but key_pair_info can be updated. It provides a huge simplification for the implementation

include/library/spdm_common_lib.h
alistair23: It should then avoid any need for something like https://github.com/jyao1/libspdm/blob/9bbad4f15a652d2f89e53c38c72427539d358f89/doc/slot_management.md

jyao1:
> Yeah, key_pair_info will need to be fixed and included in libspdm. Then all of the sync issues will be solved!

I think the is the fundamental design choice.
 * Option 1: Put key_pair_info + slot_management into libspdm context (sync by libspdm)
 * Option 2: Leave key_pair_info + slot_management to the integrator (sync by device integrator. libspdm can provide sample lib for managing those data - see https://github.com/DMTF/libspdm/discussions/3646)

I prefer option 2.
The reason is that this management work belongs to device non-volatile state change, while libspdm_context is design to carry the required info for this connection. Even if we choose option 1 to extend the libspdm context includes these info, the device vendor still need to carry these non-volatile info somewhere. I treat it as duplicated effort.

include/library/spdm_common_lib.h
jyao1: OK. please provide a full solution for evaluation.
So far, I think this is just an partial solution, still having lots of gap unresolved.

alistair23: Done! This fully implements the sample device and unit tests, without any gaps or catastrophic sync issues

alistair23:
> The reason is that this management work belongs to device non-volatile state change, while libspdm_context is design to carry the required info for this connection.

Couldn't agree more. Which is why we need to store the bank information in libspdm_context, so that libspdm can handle the connection and slot management.

> Even if we choose option 1 to extend the libspdm context includes these info, the device vendor still need to carry these non-volatile info somewhere. I treat it as duplicated effort.

The HAL will need to handle reading/writing the information from/to something non-volatile. There isn't much we can do about that. But with this approach that is **all** the HAL needs to do. Instead of managing the connection as well, like you try to do here: https://github.com/DMTF/libspdm/pull/3637/changes#diff-f223c3bf81dcc56487fc5e8f7b3f8538bdeacb2d7e3e20a04922582cbadd74b7R702 (which doesn't work as there are a range of sync issues).

alistair23:
KeyPairID is now working in libspdm. I fixed some of the bugs in the unit tests, so they should be checking more now as well

include/internal/libspdm_common_lib.h
```diff
@@ -99,13 +99,22 @@ typedef struct {
...
-    spdm_key_pair_id_t local_key_pair_id[SPDM_MAX_SLOT_COUNT];
-    spdm_certificate_info_t local_cert_info[SPDM_MAX_SLOT_COUNT];
-    spdm_key_usage_bit_mask_t local_key_usage_bit_mask[SPDM_MAX_SLOT_COUNT];
+    spdm_key_pair_id_t local_key_pair_id[LIBSPDM_MAX_BANK_COUNT][SPDM_MAX_SLOT_COUNT];
+    spdm_certificate_info_t local_cert_info[LIBSPDM_MAX_BANK_COUNT][SPDM_MAX_SLOT_COUNT];
+    spdm_key_usage_bit_mask_t local_key_usage_bit_mask[LIBSPDM_MAX_BANK_COUNT][SPDM_MAX_SLOT_COUNT];
+    uint8_t local_slot_management_subcodes[8];
+
+    uint8_t total_key_pairs;
+    const libspdm_key_pair_info_t *local_key_pair_info[LIBSPDM_MAX_KEY_PAIR_COUNT];
```
jyao1: why use a pointer?

include/internal/libspdm_common_lib.h
```diff
@@ -99,13 +99,22 @@ typedef struct {
...
+
+    uint32_t local_bank_asym_algo[LIBSPDM_MAX_BANK_COUNT];
+    uint32_t local_bank_pqc_asym_algo[LIBSPDM_MAX_BANK_COUNT];
+    uint32_t local_bank_asym_algo_capabilities[LIBSPDM_MAX_BANK_COUNT];
+    uint32_t local_bank_pqc_asym_algo_capabilities[LIBSPDM_MAX_BANK_COUNT];
```
jyao1: It seems the design between key_pair_info and bank_management is NOT consistent.

For key_pair_info, it is a big chunk.
For bank_management, it is multiple small fields.

That is very weird to me. I suggest we use a consistent design.

include/library/spdm_common_lib.h
```diff
@@ -76,6 +76,15 @@ typedef enum {
...
     LIBSPDM_DATA_LOCAL_KEY_USAGE_BIT_MASK,
+    LIBSPDM_DATA_TOTAL_KEY_PAIRS,
+    LIBSPDM_DATA_LOCAL_KEY_PAIR_INFO,
+
+    LIBSPDM_DATA_LOCAL_CURRENT_BANK,
+    LIBSPDM_DATA_LOCAL_SLOT_MANAGEMENT_SUBCODES,
+    LIBSPDM_DATA_LOCAL_BANK_ASYM_ALGO,
+    LIBSPDM_DATA_LOCAL_BANK_PQC_ASYM_ALGO,
+    LIBSPDM_DATA_LOCAL_BANK_ASYM_ALGO_CAPABILITIES,
+    LIBSPDM_DATA_LOCAL_BANK_PQC_ASYM_ALGO_CAPABILITIES,
```
jyao1: It seems the design between key_pair_info and bank_management is NOT consistent.

For LIBSPDM_DATA_LOCAL_KEY_PAIR_INFO, it is a big chunk.
For LIBSPDM_DATA_LOCAL_BANK_*, it is multiple small fields.

That is very weird to me. I suggest we use a consistent design.

include/hal/library/responder/key_pair_info.h
```diff
@@ -11,84 +11,56 @@
...
+ *                                                On output, whether the device needs to be reset
+ *                                                to complete the SET_KEY_PAIR_INFO operation.
  *
- * @retval true  set key pair info successfully.
- * @retval false set key pair info failed.
+ * @retval true   the operation was applied and written to non-volatile memory successfully.
+ * @retval false  the operation could not be applied.
  **/
-extern bool libspdm_write_key_pair_info(
+extern bool libspdm_update_local_key_pair_info(
```
jyao1, line 54: OK. this is a write-through HAL function for key_pair.

But I do not see a write-through HAL function for bank management.

It seems the design is not consistent.

How do you store the updated bank info to the platform, after the context is destroyed?

library/spdm_responder_lib/libspdm_rsp_set_key_pair_info_ack.c
```diff
@@ -111,31 +111,27 @@ libspdm_return_t libspdm_get_response_set_key_pair_info_ack(libspdm_context_t *s
...
                                                SPDM_ERROR_CODE_UNSPECIFIED, 0,
                                                response_size, response);
     }
+    capabilities = key_pair_info->capabilities;
+    key_usage_capabilities = key_pair_info->key_usage_capabilities;
+    asym_algo_capabilities = key_pair_info->asym_algo_capabilities;
+    current_asym_algo = key_pair_info->current_asym_algo;
+    pqc_asym_algo_capabilities = key_pair_info->pqc_asym_algo_capabilities;
+    current_pqc_asym_algo = key_pair_info->current_pqc_asym_algo;
+    assoc_cert_slot_mask = key_pair_info->assoc_cert_slot_mask;
```
jyao1, line 134: How do you cross check the key_pair_id in DIGEST ?

This solution does NOT resolve the slot/bank/algo/key_pair syncing problem at all. That would be a big problem.
I have pointed out before at https://github.com/DMTF/libspdm/discussions/3646.

jyao1:
I feel this new update does not resolve the slot/bank/algo/key_pair syncing problem at all. That would be a big problem.
I have pointed out before at https://github.com/DMTF/libspdm/discussions/3646.

Originally, we let the device handle that in sample lib. Code in libspdm is good.
Now, with this approach, libspdm itself is buggy because it does not have consistent state internally.

Please also clarify: What test you have designed and run to ensure the slot/bank/algo/key_pair are in sync in any cases.

include/internal/libspdm_common_lib.h
alistair23: `libspdm_key_pair_info_t` is huge, the public key allocates 65535 bytes. That's a lot to store in the libspdm context

include/internal/libspdm_common_lib.h
alistair23: The bank management API is based on your implementation here: https://github.com/DMTF/libspdm/pull/3637/changes#diff-c114df7308b1cc4616e9bc11f40eb014444e96839d0a3dd11d1a00233c704a66R99

The KeyPairInfo is based on the existing one here: https://github.com/DMTF/libspdm/blob/84121c55e058391dab1eb126b8c9c0656794debd/os_stub/spdm_device_secret_lib_sample/key_pair.c#L41

Hence the inconsistency. I don't mind either approach, do you have a preference which way to go?

include/library/spdm_common_lib.h
alistair23: Will fix, just waiting on confirmation of which method you prefer

library/spdm_responder_lib/libspdm_rsp_set_key_pair_info_ack.c
alistair23: I don't follow, `local_key_pair_id` in GET_DIGET is unchanged, so the same as before

include/hal/library/responder/key_pair_info.h
alistair23: Good point. I have added a simple `libspdm_write_slot_management_manage_bank()` callback to store the configuration to NVM

library/spdm_responder_lib/libspdm_rsp_set_key_pair_info_ack.c
alistair23: I think you mean you want me to fix https://github.com/DMTF/libspdm/issues/3638 as part of this PR. Done!

alistair23:
> Originally, we let the device handle that in sample lib. Code in libspdm is good.
> Now, with this approach, libspdm itself is buggy because it does not have consistent state internally.

What is inconsistent?

> Please also clarify: What test you have designed and run to ensure the slot/bank/algo/key_pair are in sync in any cases.

You don't think the current tests do that? What would you propose?

library/spdm_responder_lib/libspdm_rsp_set_key_pair_info_ack.c
jyao1: I do not get it. How you fix that?

jyao1:
> > Originally, we let the device handle that in sample lib. Code in libspdm is good.
> > Now, with this approach, libspdm itself is buggy because it does not have consistent state internally.
> 
> What is inconsistent?

@alistair23 
Since SPDM spec has different way to return same/similar info, the device must maintain the consistent state between slot/bank/algo/key_pair, if the libspdm uses different way to save the info.
For example:
DIGEST.KeyPairID is linked to KEY_PAIR_INFO.AssocCertSlotMask
DIGEST.KeyUsageMask is linked to KEY_PAIR_INFO.CurrentKeyUsage

Just 2 examples, the SPDM spec requires more. Please read the spec carefully to ensure you understand the link between them. There are others in slot_management command as well.

Optionally, you can refer to my document at https://github.com/jyao1/libspdm/blob/slot_management/doc/slot_management_database.md. (where I did consider the state sync.)

Anyway, if you just update one commend one time, then you break the state consistency across state.
Even worse, if you are not aware of this consistency requirement in SPDM spec, you even do not know you are making mistake.
That is exactly what I am worried about.

> 
> > Please also clarify: What test you have designed and run to ensure the slot/bank/algo/key_pair are in sync in any cases.
> 
> You don't think the current tests do that? What would you propose?

This patch is very big. As such, I am not sure what test you have covered and I might miss something.

Since you are the owner to add the test, I would like to ask you. Please list all of cross state consistency related test for evaluation.

Again, please read the SPDM spec, to ensure the tests have fully covered all the scenarios for slot/bank/algo/key_pair change and the tests have fully covered the changed state check, since you are updating libspdm core code.
Just checking the command response according to the request is far from enough.

That will also help me to evaluate if there is something missing.

include/internal/libspdm_common_lib.h
jyao1: I do not think this is a right understanding. You CANNOT take my partial design as a base, because of the assumption between us is different.

My API design is based on the assumption that device will maintain an internal state for state sync.
While your assumption is that the libspdm will maintain an internal state for state sync.

I think you need to consider how to design an libspdm API for hold the internal state for consistency. But current proposal in libspdm_context is NOT. I do not know how you achieve that.

It is NOT about which style to choose. It is more about what the consistency design in your mind.

include/internal/libspdm_common_lib.h
jyao1: I do not feel that is right way. If you group many thing together, then this data structure will become big.
But if you split to small pieces, then the data structure will become small.

It is weird design to me, that some fields are split and assigned as value, while some fields are combined and assigned as pointer.

Even worse, some of the key_pair_info_t will eventually impact other field (as consistency consideration). What if the integrator set different value by mistake?

I still prefer a design that the integrator can only set one value, and the rest response is derived from that unique source. That is a way to prevent consistency issue.

alistair23: > I do not feel that is right way. If you group many thing together, then this data structure will become big.
> But if you split to small pieces, then the data structure will become small.

I don't follow. The keys will be large, no matter how you split it up

> It is weird design to me, that some fields are split and assigned as value, while some fields are combined and assigned as pointer.

We already have pointers in the libspdm context, such as the certificates `local_cert_chain_provision`. They are pointers, it doesn't make sense to store the value for there. This doesn't change the fact that some things are stored as values and some are stored as pointers.

> Even worse, some of the key_pair_info_t will eventually impact other field (as consistency consideration). What if the integrator set different value by mistake?

Yes, `key_pair_info_t` will impact the operation. I don't know what consistency consideration there is, as `key_pair_info_t` is the source of truth for KeyPairs.

If an integrator sets the wrong values then things will break. It's the exact same problem if they set the wrong certificate chain. This PR does improve things, as currently there is a `LIBSPDM_DATA_LOCAL_KEY_PAIR_ID` that the integrator needs to set and then ensure the HAL callbacks match that, making it easy to mess up consistency. My PR removes this so there is only a single source of truth.

> I still prefer a design that the integrator can only set one value, and the rest response is derived from that unique source. That is a way to prevent consistency issue.

Agreed, that is what I'm doing here....

`LIBSPDM_DATA_LOCAL_KEY_PAIR_INFO` and `LIBSPDM_DATA_TOTAL_KEY_PAIRS` are now the single source of truth

library/spdm_responder_lib/libspdm_rsp_set_key_pair_info_ack.c
alistair23: `local_supported_slot_mask` and `local_key_pair_id` have been removed. So there is no longer some hardcoded state that needs to be updated. Instead the state can all be derived from the KeyPairs `libspdm_key_pair_info_t`

jyao1: OK. I see. That is good.

include/internal/libspdm_common_lib.h
jyao1: > The keys will be large, no matter how you split it up

I am not sure what do you mean. Keys will be large. Yes.
But here what we need is key_info, not key.

> If an integrator sets the wrong values then things will break

To be clear, I am not worried about that the integrator sets wrong value for a specific item.
**What I am worried about is that the integrator sets 2 conflicted value by mistake.** That is why we have libspdm_context check to early detect such configuration.

Or libspdm_context has 2 fields to save same / similar info. If they are out of sync, then we will be in trouble.

> Agreed, that is what I'm doing here....

> LIBSPDM_DATA_LOCAL_KEY_PAIR_INFO and LIBSPDM_DATA_TOTAL_KEY_PAIRS are now the single source of truth

Yes. Thanks. It is necessary, But I am not sure if that is enough.

For example, how do you ensure the consistency between bank info and key_pair_info for a given slot. Then algo of bank and key_pair_info must be same.

It is unclear to me, in this huge PR. I hope there is a design doc to clarify that.

alistair23:
> Since SPDM spec has different way to return same/similar info, the device must maintain the consistent state between slot/bank/algo/key_pair, if the libspdm uses different way to save the info. For example: DIGEST.KeyPairID is linked to KEY_PAIR_INFO.AssocCertSlotMask DIGEST.KeyUsageMask is linked to KEY_PAIR_INFO.CurrentKeyUsage
> 
> Just 2 examples, the SPDM spec requires more. Please read the spec carefully to ensure you understand the link between them. There are others in slot_management command as well.

It wasn't clear in the spec that DIGEST.KeyUsageMask is updated directly from the KeyPair.

The confusing part is that it isn't updated in libspdm today from the KeyPair and instead is set separately, so it seems like it is always out of date already. I have updated this PR to fix that issue by removing `local_key_usage_bit_mask` entirely.

I'm on holidays for two weeks. Maybe it would be best if you could read through the PR and comment explicitly on any areas where you think there are consistency issues? It's hard to address or respond to generic "consistency issues", but I can easily fix any bugs that you see.

AFAIK this PR doesn't add any consistency issues, in fact it removes a lot as it consolidates a bunch of competing information into a single source of truth. KeyPairs today for example are practically always out of sync and this PR does a pretty good job of cleaning that up.

> This patch is very big. As such, I am not sure what test you have covered and I might miss something.

Yeah, I wanted to do it step by step, a little bit at a time to make it easy to review and test, but you didn't like that approach. So we now have a giant PR. I have added some more tests to compare consistency issues, hopefully that helps

include/internal/libspdm_common_lib.h
alistair23: > I am not sure what do you mean. Keys will be large. Yes. But here what we need is key_info, not key.

Yeah, the key itself could just be a pointer. I stuck with the existing layout to keep things simple, but I can update it if we want?

> What I am worried about is that the integrator sets 2 conflicted value by mistake

Isn't that the case today? Currently in upstream nothing checks that `local_key_usage_bit_mask` matches the values set by `libspdm_read_key_pair_info()`. They can easily conflict (and probably often do).

> Or libspdm_context has 2 fields to save same / similar info. If they are out of sync, then we will be in trouble.

But that is the case TODAY! In upstream libspdm right now there are multiple fields storing redundant information that are out of sync. `local_key_usage_bit_mask` and `desired_key_usage` set in `libspdm_get_response_set_key_pair_info_ack()` is a clear example.

This PR addresses a lot of those sync issues. You seem very concerned about the possibility of this introducing a bug, but it does fix numerous current existing bugs.

It feels like the goal posts keep moving. To me this seems like a clear improvement on improving consistency and removing redundant information.

jyao1:
> I'm on holidays for two weeks. Maybe it would be best if you could read through the PR and comment explicitly on any areas where you think there are consistency issues? It's hard to address or respond to generic "consistency issues", but I can easily fix any bugs that you see.

That is exactly my concern. I feel like you are not aware of what consistency requirement in SPDM spec, and you rely my to find out that in your code.  E.g. You fixed DIGEST/KEY_PAIR syncing problem after I point it out. But BANK and KEY_PAIR also need to keep syncing, I am not sure how you are doing that. (Please correct me if I am wrong.)
The development process should be: 1) read spec and understand the requirement, 2) propose a design and write test to ensure the spec requirement is followed, 3) then submit PR and ask for code review for final check.

In general, I do not feel it is a good idea to ask other people to find bug for you, then you fix it.
To me, the code review should be the last step, assuming everything is already correct, and just let other people confirm you are doing right thing.
So, please read SPDM spec and design test to cover the corner case for consistency between bank/slot/algo/key_pair.
I would like to review your design doc and test design description, before jump into the code.


> AFAIK this PR doesn't add any consistency issues, in fact it removes a lot as it consolidates a bunch of competing information into a single source of truth. KeyPairs today for example are practically always out of sync and this PR does a pretty good job of cleaning that up.

I cannot agree with this part. And that is exactly my concern.
Original the consistency issue is in device sample. I agree it is problem (as I have submitted the issue), but at least Libspdm code lib is clean.
With your design, you moved the consistency issue into libspdm core lib. That changed the position completely. That is the reason that I am worried and ask for evidence of validation and production quality.

> Yeah, I wanted to do it step by step, a little bit at a time to make it easy to review and test, but you didn't like that approach. So we now have a giant PR. I have added some more tests to compare consistency issues, hopefully that helps

OK. usually, for a clear architecture or a clear design, I prefer step by step. A small patch to resolve one specific problem.
But here, not the case, because you reject my ask for design review, then I have no way to understand your design. I have to ask you to provide everything.
A full design cannot be step by step. That will cause problem easily. Just like how easy we introduce the syncing issue for DIGEST and KeyPair. Now we have to deal with it.
And I want to prevent such thing from happening again for bank / slot_management.

Let me clarify my goal again: **a clean architecture to resolve any data syncing issue between slot/bank/algo/key_pair according to SPDM specification.**

include/internal/libspdm_common_lib.h
jyao1: > > I am not sure what do you mean. Keys will be large. Yes. But here what we need is key_info, not key.
> 
> Yeah, the key itself could just be a pointer. I stuck with the existing layout to keep things simple, but I can update it if we want?

Here I provide comment for "local_key_pair_info" and ask why you use a pointer.
But you are talking that the "The keys will be large", so you have to use point for a key.
To me, a big key size is totally irrelevant to a key_pair_info. I am not sure if we are talking the same thing.

> 
> > What I am worried about is that the integrator sets 2 conflicted value by mistake
> 
> Isn't that the case today? Currently in upstream nothing checks that `local_key_usage_bit_mask` matches the values set by `libspdm_read_key_pair_info()`. They can easily conflict (and probably often do).

Here is my logic:
1) Yes, I agree with you that it is a problem today. I aware of this issue.
2) That is why I proposed a solution to resolve the problem. See https://github.com/jyao1/libspdm/blob/slot_management/doc/slot_management_database.md
3) You think my proposal is too complicated. You have simple way.
4) Then I am asking you to provide your proposal.
5) But I notice your PR does not resolve the problem (until I point out)
5) Now What worried me is that it seems you are not aware of this problem, before I point out in the PR.

> 
> > Or libspdm_context has 2 fields to save same / similar info. If they are out of sync, then we will be in trouble.
> 
> But that is the case TODAY! In upstream libspdm right now there are multiple fields storing redundant information that are out of sync. `local_key_usage_bit_mask` and `desired_key_usage` set in `libspdm_get_response_set_key_pair_info_ack()` is a clear example.
> 
> This PR addresses a lot of those sync issues. You seem very concerned about the possibility of this introducing a bug, but it does fix numerous current existing bugs.
> 
> It feels like the goal posts keep moving. To me this seems like a clear improvement on improving consistency and removing redundant information.

My goal is NOT moving from day 1. But it seems you have some misunderstanding on the final goal.
I am aware that there is a lot of sync issue today. I hope there will be **a clean architecture to resolve the data syncing issue between slot/bank/algo/key_pair**. We need to think through before proposing more stuff and introducing more sync issue. 

That is why I proposed a new design.
But I feel you are so eagle to add slot_management stuff without considering data syncing issue. just because there is some syncing issue today.
That action did worry me, because that will introduce more consistency issues potentially, and make code base worse.

Let me clarify my goal again: **a clean architecture to resolve any data syncing issue between slot/bank/algo/key_pair according to SPDM specification.**.

alistair23: I believe that this approach is a clean architecture to fixup the sync issues. The current approach of discovering sync issues between the numerous duplicated information and fixing each one-by-one isn't viable.

I noticed you just fixed https://github.com/DMTF/libspdm/commit/34203598746e414fb354fabe92025a86b130a8d7 (which this PR would have fixed as well). But how many others are there? You only noticed that bug in libspdm because I pointed it out, which I only noticed because I saw it was broken while writing this code. I suspect there are numerous other issues in the core libspdm library right now.

Trying to keep multiple duplicates of information in sync in a complex code base is like playing whack-a-mole.

The approach in this PR removes the duplicate information. So there is now nothing to keep in sync.

Yes, my aim was to do this to support Slot Management, but I don't think that matters. The end result is still substantially better as there is now a lot less opportunities for issues.

I would have preferred to make small incremental changes, but here we are with a big overhaul to ensure everything is updated at once.

Your proposal at https://github.com/jyao1/libspdm/blob/slot_management/doc/slot_management_database.md isn't really an alternative. It doesn't fix the issues in libspdm. The issue in `libspdm_get_response_set_key_pair_info_ack()` would still be there as will other ones. You seem to ignore the fact that libspdm today has duplicate information that is out of sync and leads to bugs.

I'm not claiming there aren't any bugs in this PR. There might be, all software has bugs. But the overall design simplifies things. The aim to reduce the number of bugs and make fixing any easier in the future. The fact that you found a bug in review doesn't mean this is broke, it's the point of reviewing code!

My hope is that this can be merged, to cleanup and simplify the code base. That gives a place to build from

include/internal/libspdm_common_lib.h
alistair23: I went with the existing KeyPairInfo type design

alistair23:
This feels like moving goal posts.

The development proces was
1. read the spec
2. propose design https://github.com/DMTF/libspdm/pull/3629
3. submit PR and ask for code review

Now you want a design doc. I don't see those for other features that are added.

You could probably point an LLM at this PR and have it write one for you if you would like?

> Original the consistency issue is in device sample. I agree it is problem (as I have submitted the issue), but at least Libspdm code lib is clean.
> With your design, you moved the consistency issue into libspdm core lib. That changed the position completely. That is the reason that I am worried and ask for evidence of validation and production quality.

The current consistency issues are in libspdm core. This PR identified bugs and fixes them. Some of the bugs you have fixed independently (see https://github.com/DMTF/libspdm/commit/34203598746e414fb354fabe92025a86b130a8d7 and https://github.com/DMTF/libspdm/pull/3717). But libspdm today has consistency issues. It has issues today in both the core library and the reference implementations include in this repo.

I believe that this PRs approach is a clean architecture to fixup the sync issues. It removes a lot of duplicate information that currently exists. This PR might have bugs (what code doesn't), but the overall design is simplification and cleanup of the existing code.

> OK. usually, for a clear architecture or a clear design, I prefer step by step. A small patch to resolve one specific problem. But here, not the case, because you reject my ask for design review, then I have no way to understand your design. I have to ask you to provide everything. A full design cannot be step by step. That will cause problem easily. Just like how easy we introduce the syncing issue for DIGEST and KeyPair. Now we have to deal with it. And I want to prevent such thing from happening again for bank / slot_management.

I tried! I opened https://github.com/DMTF/libspdm/pull/3629 with a clear description of goals and next steps. It was a small incremental step.  You said instead you wanted a ["complete solution"](https://github.com/DMTF/libspdm/pull/3637#discussion_r3362711829).

Now I have a complete solution, that resolves some long standing bugs, specifically the existing "syncing issue for DIGEST and KeyPair".

Maybe everyone should take a step back and look at the general architecture of the existing code and this PR. I believe this PR simplifies and improves the general design, by removing duplicate state data (and also adding slot management support). Can't we all agree that is a big improvement?

alistair23:
@jyao1 @steven-bellock  Ping!

steven-bellock:
@alistair23 I am focusing on the 4.0 RC2 release, and actually know very little about slot management. That will change, for the better, once 4.0 is released.

PR #3669, opened 2026-06-30
spdm_responder_lib: Enforce RESPOND_IF_READY session context

### LINKED ISSUE #3664: Need to check if RESPONSE_IF_READY command is per session
If it is sent not in the same session, error should be returned, instead of getting a real response.

jyao1:
Fix: https://github.com/DMTF/libspdm/issues/3664

Per DSP0274 (RESPOND_IF_READY request message format), "the validity of the RESPOND_IF_READY request ... is defined by the original request that caused the RESPOND_IF_READY flow." The responder must therefore only honor a RESPOND_IF_READY that arrives in the same session context as the original deferred request; otherwise the cached response would be returned outside the context it belongs to.

The handler previously validated only the SPDM version, the original request code, and the token -- not the session context -- so a RESPOND_IF_READY sent in a different session (or in a session when the original was on the connection, or vice versa) was wrongly accepted and the cached response replayed.

Capture the original request's session context when the ResponseNotReady flow is entered (libspdm_responder_handle_response_state), and reject a RESPOND_IF_READY whose session context does not exactly match -- same in-session / out-of-session state, and the same session_id when in a session -- with ERROR(UnexpectedRequest), before replaying the cached request.

- libspdm_common_lib.h: add cache_spdm_request_session_id_valid / cache_spdm_request_session_id to the context (RESPOND_IF_READY support).
- libspdm_rsp_handle_response_state.c: record the original request's session context alongside the cached request.
- libspdm_rsp_respond_if_ready.c: reject a session-context mismatch with UnexpectedRequest.
- test_spdm_responder/respond_if_ready.c: add case 15 (original on the connection, RESPOND_IF_READY in a session -> UnexpectedRequest); seed the new cache session fields in the existing session/connection cases so they do not depend on cross-case state.

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -627,6 +627,12 @@ typedef struct {
 #if LIBSPDM_RESPOND_IF_READY_SUPPORT
     void *cache_spdm_request;
     size_t cache_spdm_request_size;
+    /* Session context of the original request that triggered the ResponseNotReady flow. The validity
+     * of the RESPOND_IF_READY request is defined by that original request, so the RESPOND_IF_READY
+     * must arrive in the same session context (same in-session/out-of-session, and if in a session
+     * the same session_id). */
+    bool cache_spdm_request_session_id_valid;
+    uint32_t cache_spdm_request_session_id;
 #endif
     uint8_t current_token;
 
--- a/library/spdm_responder_lib/libspdm_rsp_handle_response_state.c
+++ b/library/spdm_responder_lib/libspdm_rsp_handle_response_state.c
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2021-2025 DMTF. All rights reserved.
+ *  Copyright 2021-2026 DMTF. All rights reserved.
  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
  **/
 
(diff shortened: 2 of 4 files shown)


alistair23 (approved):

include/internal/libspdm_common_lib.h
```diff
@@ -627,6 +627,12 @@ typedef struct {
 #if LIBSPDM_RESPOND_IF_READY_SUPPORT
     void *cache_spdm_request;
     size_t cache_spdm_request_size;
+    /* Session context of the original request that triggered the ResponseNotReady flow. Per DSP0274,
```
steven-bellock: >Per DSP0274

The whole of libspdm is per DSP0274. This is not needed.

library/spdm_responder_lib/libspdm_rsp_handle_response_state.c
```diff
@@ -31,13 +31,31 @@ libspdm_return_t libspdm_responder_handle_response_state(libspdm_context_t *spdm
         return LIBSPDM_STATUS_SUCCESS;
     #if LIBSPDM_RESPOND_IF_READY_SUPPORT
     case LIBSPDM_RESPONSE_STATE_NOT_READY:
-        /*do not update ErrorData if a previous request has not been completed*/
+        /* Do not update ErrorData if a previous request has not been completed.
+         *
+         * Only one deferred (ResponseNotReady) request is tracked per connection. Per DSP0274
```
steven-bellock: >Per DSP0274

Same thing here.

PR #3674, opened 2026-07-01
spdm_responder_lib: Sync slot context after SET_KEY_PAIR_INFO

### LINKED ISSUE #3673: Need sync DIGESTS resolve after SET_KEY_PAIR_INFO_ACK
The slot/keyPair info mapping might be changed after SET_KEY_PAIR_INFO_ACK.
We need to sync them for DIGESTS response later.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/3673

DIGESTS reports exactly one KeyPairID and one KeyUsageMask per certificate slot (DSP0274 Table 42), sourced from local_context.local_key_pair_id[] and local_context.local_key_usage_bit_mask[]. A successful no-reset SET_KEY_PAIR_INFO changes a key pair's certificate slot association inside the device (via the libspdm_write_key_pair_info HAL), but the responder never updated these connection-level arrays. As a result a later DIGESTS could report a stale KeyPairID / KeyUsageMask for a slot whose association had changed, disagreeing with the key pair actually used for signing.

After a successful no-reset SET_KEY_PAIR_INFO, the responder re-reads the key pair info (libspdm_read_key_pair_info) and syncs the per-slot context from the authoritative post-write AssocCertSlotMask and CurrentKeyUsage. Re-reading (rather than inferring the new state from the request inputs) keeps the context correct even when the device normalizes the request or applies defaults on a successful write. The pre-write mask is kept only to identify slots this KeyPairID no longer owns, which are then cleared.

- libspdm_rsp_set_key_pair_info_ack.c: re-read key pair info and sync local_key_pair_id[] and local_key_usage_bit_mask[] for the affected slots in the no-reset success path.
- test_spdm_responder/set_key_pair_info_ack.c: add case 6, which seeds a sentinel KeyPairID in the slot context, performs a no-reset association change, and asserts the context reflects the new association (and is cleared when the association is removed).

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_set_key_pair_info_ack.c
+++ b/library/spdm_responder_lib/libspdm_rsp_set_key_pair_info_ack.c
@@ -394,6 +394,38 @@ libspdm_return_t libspdm_get_response_set_key_pair_info_ack(libspdm_context_t *s
                                                SPDM_ERROR_CODE_RESET_REQUIRED, 0,
                                                response_size, response);
     } else {
+        /* Update context with new key pair information if a reset is not needed. Re-read the key
+         * pair info so the authoritative post-write association and key usage are used (the device
+         * may normalize or apply defaults on a successful write). */
+        uint8_t new_assoc_cert_slot_mask = 0;
+        uint16_t new_current_key_usage = 0;
+        uint8_t slot_index;
+
+        result = libspdm_read_key_pair_info(
+            spdm_context, key_pair_id, &total_key_pairs, &capabilities, &key_usage_capabilities,
+            &new_current_key_usage, &asym_algo_capabilities, &current_asym_algo,
+            &pqc_asym_algo_capabilities, &current_pqc_asym_algo, &new_assoc_cert_slot_mask,
+            NULL, NULL);
+        if (!result) {
+            return libspdm_generate_error_response(spdm_context,
+                                                   SPDM_ERROR_CODE_UNSPECIFIED, 0,
+                                                   response_size, response);
+        }
+
+        for (slot_index = 0; slot_index < SPDM_MAX_SLOT_COUNT; slot_index++) {
+            if ((new_assoc_cert_slot_mask & (1 << slot_index)) != 0) {
+                /* Slot is (still) associated with this KeyPairID. */
+                spdm_context->local_context.local_key_pair_id[slot_index] = key_pair_id;
+                spdm_context->local_context.local_key_usage_bit_mask[slot_index] =
+                    new_current_key_usage;
+            } else if (spdm_context->local_context.local_key_pair_id[slot_index] == key_pair_id) {
+                /* Context still points at this KeyPairID but the slot is no longer in the
+                 * authoritative mask (removed or stale); clear it. */
+                spdm_context->local_context.local_key_pair_id[slot_index] = 0;
+                spdm_context->local_context.local_key_usage_bit_mask[slot_index] = 0;
+            }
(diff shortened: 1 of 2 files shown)


library/spdm_responder_lib/libspdm_rsp_set_key_pair_info_ack.c
```diff
@@ -342,6 +342,51 @@ libspdm_return_t libspdm_get_response_set_key_pair_info_ack(libspdm_context_t *s
                                                SPDM_ERROR_CODE_RESET_REQUIRED, 0,
                                                response_size, response);
     } else {
+        /* The change has been applied without a reset. Keep the connection-level context that
```
steven-bellock: This is too verbose. Could just be
>Update context with new key pair information if a reset is not needed.

jyao1: done

library/spdm_responder_lib/libspdm_rsp_set_key_pair_info_ack.c
```diff
@@ -342,6 +342,45 @@ libspdm_return_t libspdm_get_response_set_key_pair_info_ack(libspdm_context_t *s
...
+            &total_key_pairs,
+            &capabilities,
+            &key_usage_capabilities,
+            &new_current_key_usage,
+            &asym_algo_capabilities,
+            &current_asym_algo,
+            &pqc_asym_algo_capabilities,
+            &current_pqc_asym_algo,
+            &new_assoc_cert_slot_mask,
+            NULL, NULL);
```
alistair23, line 408: Why do we need to call this? We know `libspdm_write_key_pair_info()` succeeded, can't we just use `desired_assoc_cert_slot_mask` and friends

alistair23 (changes requested):

library/spdm_responder_lib/libspdm_rsp_set_key_pair_info_ack.c
jyao1: fair statement.
I removed the HAL call and just used the old/new value.

jyao1: wait. See suggestion from copilot - https://github.com/DMTF/libspdm/pull/3674#discussion_r3643299284
I think it is reasonable. I will keep it.

library/spdm_responder_lib/libspdm_rsp_set_key_pair_info_ack.c
```diff
@@ -394,6 +394,40 @@ libspdm_return_t libspdm_get_response_set_key_pair_info_ack(libspdm_context_t *s
...
+                /* Slot is (still) associated with this KeyPairID. */
+                spdm_context->local_context.local_key_pair_id[slot_index] = key_pair_id;
+                spdm_context->local_context.local_key_usage_bit_mask[slot_index] =
+                    new_current_key_usage;
+            } else if (((old_assoc_cert_slot_mask & (1 << slot_index)) != 0) &&
+                       (spdm_context->local_context.local_key_pair_id[slot_index] == key_pair_id)) {
+                /* Slot was associated with this KeyPairID and has now been removed. */
+                spdm_context->local_context.local_key_pair_id[slot_index] = 0;
+                spdm_context->local_context.local_key_usage_bit_mask[slot_index] = 0;
+            }
```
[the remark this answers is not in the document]
jyao1: ok

PR #3675, opened 2026-07-01
spdm_responder_lib: Reject same-algorithm slot collision in SET_KEY_PAIR_INFO

### LINKED ISSUE #3672: Need to check the per algorithm uniqueness of keyPair/slot mapping in SET_KEY_PAIR_INFO_ACK
According to DIGESTS response, one slot can only have 1 key pair.
SET_KEY_PAIR_INFO allows the requester to set keyPair/slot mapping. The responder need to check if a new keypair is set to a slot which already has a KeyPair associated with the same algorithm.

Clarify expectation at https://github.com/DMTF/SPDM-WG/issues/4628

jyao1:
Fix https://github.com/DMTF/libspdm/issues/3672

Within one connection a certificate slot resolves to a single KeyPairID for the negotiated algorithm: DIGESTS (DSP0274 Table 42) returns exactly one KeyPairID per slot. Two different KeyPairIDs of the SAME asymmetric algorithm therefore must not both be associated with the same certificate slot, otherwise that slot's KeyPairID would be ambiguous.

The SET_KEY_PAIR_INFO responder previously validated only the requested KeyPairID's own fields (capabilities, one-hot association when not shareable, algorithm consistency) and never checked whether a slot it was about to associate was already owned by a different, same-algorithm KeyPairID.

Before applying the change, iterate the other KeyPairIDs and reject the request if one of them has the same current asymmetric algorithm (traditional and PQC) as this request's effective algorithm and already associates a slot in DesiredAssocCertSlotMask. A slot shared across DIFFERENT algorithms remains allowed, since only one such key pair is active per connection.

DSP0274 does not define an error code for this conflict; ERROR=OperationFailed is used as a libspdm policy, consistent with the other association-conflict cases in the SET_KEY_PAIR_INFO error-handling clause.

- libspdm_rsp_set_key_pair_info_ack.c: add the cross-KeyPairID, per-algorithm slot-collision check before the write.
- test_spdm_responder/set_key_pair_info_ack.c: add case 7, which discovers two same-algorithm key pairs and asserts that associating a slot already owned by one of them with the other is rejected with OperationFailed.

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_set_key_pair_info_ack.c
+++ b/library/spdm_responder_lib/libspdm_rsp_set_key_pair_info_ack.c
@@ -312,6 +312,58 @@ libspdm_return_t libspdm_get_response_set_key_pair_info_ack(libspdm_context_t *s
         }
     }
 
+    /* Within one connection a certificate slot resolves to a single KeyPairID for the negotiated
+     * algorithm: DIGESTS returns exactly one KeyPairID per slot. Two different KeyPairIDs of the
+     * SAME asymmetric algorithm must therefore not both be associated with the same slot, otherwise
+     * that slot's KeyPairID would be ambiguous. Reject a request that would associate a slot which
+     * a different, same-algorithm KeyPairID already owns.
+     *
+     * DSP0274 does not define an error code for this conflict; OperationFailed is used here as a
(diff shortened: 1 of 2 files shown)


sharanu088 (approved):
looks fine

PR #3676, opened 2026-07-01
bound opaque_length against response size before opaque check

naruto-lgtm:
Repro: a responder returns a short CHALLENGE_AUTH or MEASUREMENTS response whose opaque_length field is set near SPDM_MAX_OPAQUE_DATA_SIZE while the actual trailing bytes are truncated.
Cause: libspdm_try_challenge and both branches of libspdm_try_get_measurement pass opaque_length into libspdm_process_general_opaque_data_check before validating it against spdm_response_size. The helper treats data_in_size as the buffer length and dereferences up to that many bytes, so the peer-declared length drives an out-of-bounds read of up to ~1KB past the received message. The unsigned MEASUREMENTS branch also left SPDM_NONCE_SIZE out of its size check, so the nonce and length fields were read past the buffer as well.
Fix: check opaque_length against the remaining response size before the opaque parse, the same guard the sibling key_exchange and psk_exchange parsers already apply, and add the missing nonce term to the unsigned MEASUREMENTS check.

### DIFF
--- a/library/spdm_requester_lib/libspdm_req_challenge.c
+++ b/library/spdm_requester_lib/libspdm_req_challenge.c
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2021-2025 DMTF. All rights reserved.
+ *  Copyright 2021-2026 DMTF. All rights reserved.
  *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
  **/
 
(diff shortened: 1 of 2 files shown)


naruto-lgtm:
Pushed a one-line fix to bump the copyright year to 2026 in libspdm_req_challenge.c, which is what the copyright check was flagging. The Windows job failed in the checkout step, not the build, so that one looks like CI infra rather than anything in the change.

jyao1:
please squash. We do not need one commit to only update copyright year.

naruto-lgtm:
Squashed. The copyright bump is folded into the fix commit now, so it's a single commit.

PR #3679, opened 2026-07-03
null digest_context_th after free in libspdm_append_message_k

naruto-lgtm:
Repro: on the responder, a KEY_EXCHANGE whose transcript hashing hits a crypto HAL failure. libspdm_append_message_k frees session_transcript.digest_context_th on the error path, the caller (libspdm_rsp_key_exchange.c) then tears the session down through libspdm_free_session_id, and libspdm_session_info_init frees the same context again.
Cause: the five hash-failure paths in append_message_k free digest_context_th but leave the dangling pointer in the session struct. The sibling transcript builders append_message_f/_m/_e all null it after freeing; only this one does not.
Fix: null digest_context_th after each free, matching the siblings.

### DIFF
--- a/library/spdm_common_lib/libspdm_com_context_data.c
+++ b/library/spdm_common_lib/libspdm_com_context_data.c
@@ -1948,6 +1948,7 @@ libspdm_return_t libspdm_append_message_k(libspdm_context_t *spdm_context,
             if (!result) {
                 libspdm_hash_free (spdm_context->connection_info.algorithm.base_hash_algo,
                                    spdm_session_info->session_transcript.digest_context_th);
+                spdm_session_info->session_transcript.digest_context_th = NULL;
                 return LIBSPDM_STATUS_CRYPTO_ERROR;
             }
             result = libspdm_hash_update (spdm_context->connection_info.algorithm.base_hash_algo,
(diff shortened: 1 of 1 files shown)


library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -1948,6 +1948,7 @@ libspdm_return_t libspdm_append_message_k(libspdm_context_t *spdm_context,
             if (!result) {
                 libspdm_hash_free (spdm_context->connection_info.algorithm.base_hash_algo,
                                    spdm_session_info->session_transcript.digest_context_th);
+                spdm_session_info->session_transcript.digest_context_th = NULL;
```
alistair23, line 1951: This seems very error prone, why not assign `hash_context = NULL` in `libspdm_hash_free()` instead. I understand the existing code does this already, but maybe a good time to fix it up :)

alistair23 (approved):

library/spdm_common_lib/libspdm_com_context_data.c
jyao1: > why not assign hash_context = NULL in libspdm_hash_free() instead

You can assign hash_context = NULL in libspdm_hash_free().
But that is meaningless, because what you NULLifed is the parameter on the stack or GPR, which is a copy of original data.
Finally, you still need the caller to NULLifiy the real pointer.

naruto-lgtm: Right, that was my read too. libspdm_hash_free takes the context as a void * by value, so nulling it inside the function only clears the local copy. Making the free clear the caller's pointer would mean changing the public API to void ** and touching every call site, which seemed out of scope for this fix. The caller-side null also matches what append_message_f/_m/_e already do.

PR #3694, opened 2026-07-15
validate size before reading storage secured message fields

naruto-lgtm:
Repro: a peer storage device returns a secured-message response shorter than the 4 reserved bytes plus the 4-byte session id (a 0-7 byte reply); with ASAN this is a heap-buffer-overflow READ of size 4 at libspdm_storage.c:369.

Cause: the response branch of libspdm_transport_storage_decode_message reads the session id at transport_message+4 and then derives the ciphertext length as transport_message_size - LIBSPDM_STORAGE_SECURED_MESSAGE_HEADER_RESERVED_BYTES with no size guard, so the read runs past the buffer and the subtraction underflows into the length handed to libspdm_decode_secured_message. The request path (libspdm_storage_decode_message) already validates this. The same shape exists in libspdm_storage_secured_message_decode, which reads num_descriptors and the 16-byte descriptor from the decrypted message before checking *message_size.

Fix: guard both reads with the minimum-size checks the request path already performs, before any field is dereferenced.

### DIFF
--- a/library/spdm_transport_storage_lib/libspdm_storage.c
+++ b/library/spdm_transport_storage_lib/libspdm_storage.c
@@ -91,7 +91,16 @@ static libspdm_return_t libspdm_storage_secured_message_decode(
     libspdm_error_struct_t spdm_error;
     uint32_t spdm_msg_offset;
     spdm_storage_secured_message_descriptor *descriptor;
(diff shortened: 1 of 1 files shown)


jyao1:
@twilfredo @alistair23 could you please double check

twilfredo (approved):
looks good. Thanks!

PR #3697, opened 2026-07-16
bound application_data_length by cipher header in secured decode

naruto-lgtm:
Repro: a session peer sends an ENC_MAC secured message whose decrypted application_data_length equals cipher_text_size.
Cause: libspdm_decode_secured_message bounds application_data_length against cipher_text_size, but the returned payload starts sizeof(spdm_secured_message_cipher_header_t) bytes into the plaintext, so it can run past the decrypted region, and past the plaintext buffer when the ciphertext fills it.
Fix: require application_data_length plus the cipher header to fit within cipher_text_size before returning the payload.

case13 forges such a record and expects INVALID_MSG_SIZE. The suite main was missing the config include that gates it behind LIBSPDM_AEAD_AES_256_GCM_SUPPORT, so none of these tests were compiling in; added the include so they run.

### DIFF
--- a/library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
+++ b/library/spdm_secured_message_lib/libspdm_secmes_encode_decode.c
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2021-2025 DMTF. All rights reserved.
+ *  Copyright 2021-2026 DMTF. All rights reserved.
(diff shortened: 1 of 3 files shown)


jyao1:
@naruto-lgtm Please fix CI failure - update year in copy right header.

jyao1 (approved):

naruto-lgtm:
Done, bumped the copyright years in the two headers and amended so it stays one commit.

PR #3698, opened 2026-07-17
Digests transcript window

### LINKED ISSUE #3692: Spec gap for DIGESTS in transcript
See latest SPDM 1.4.1

When a transcript hash includes the DIGESTS response, the first DIGESTS that immediately follows the VCA shall be
the DIGESTS response that is used for the remainder of the SPDM connection. **If the Requester does not send a
GET_DIGESTS immediately following the VCA, then the DIGESTS shall no longer be part of the transcript for the
remainder of the SPDM connection even if the Requester sends the request later.** Similarly, for mutual authentication
in a multi key session, if the first encapsulated response is a DIGESTS response in the session handshake phase,
then that encapsulated DIGESTS shall be included in the transcript hash for the corresponding session. **If the first
encapsulated response is not a DIGESTS response from the Requester in mutual authentication, then no
encapsulated DIGESTS response shall be part of the transcript hash for the corresponding session.** Furthermore, the
aforementioned rules do not apply to M1 or M2 ordering rules.

jyao1:
Fix https://github.com/DMTF/libspdm/issues/3692

### DIFF
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -155,6 +155,10 @@ typedef struct {
     /* multi-key negotiated result */
     bool multi_key_conn_req;
     bool multi_key_conn_rsp;
+
+    /* Sticky: set once a non-GET_DIGESTS request is processed after NEGOTIATED. A DIGESTS is added
+     * to message_d only when this is false (DIGESTS immediately follows the VCA). */
+    bool digest_transcript_window_closed;
 } libspdm_connection_info_t;
 
 typedef struct {
--- a/include/internal/libspdm_common_lib.h
+++ b/include/internal/libspdm_common_lib.h
@@ -446,6 +446,9 @@ typedef struct {
 
 typedef struct {
     libspdm_message_d_managed_buffer_t message_encap_d;
+    /* Sticky: set once a non-GET_DIGESTS encapsulated request is processed first. An encapsulated
+     * DIGESTS is added to message_encap_d only when this is false (it is the first encap response). */
+    bool encap_digest_window_closed;
 #if LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT
     libspdm_message_k_managed_buffer_t message_k;
     libspdm_message_f_managed_buffer_t message_f;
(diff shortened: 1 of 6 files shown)


sharanu088 (approved):
looks fine

include/internal/libspdm_common_lib.h
```diff
@@ -155,6 +155,10 @@ typedef struct {
     /* multi-key negotiated result */
     bool multi_key_conn_req;
     bool multi_key_conn_rsp;
+
+    /* Sticky: set once a non-GET_DIGESTS request is processed after NEGOTIATED. A DIGESTS is added
+     * to message_d only when this is false (DIGESTS immediately follows the VCA). */
+    bool digest_transcript_window_closed;
```
LeonLinAtAMI, line 161: Please correct me if wrong.

I am wondering why a new  bool digest_transcript_window_closed  instead of deriving it from  libspdm_get_managed_buffer_size(&message_d) > 0 ?

Because these two facts are not equivalent message_d  size only tells you "a DIGESTS has already been captured," never "the window in which one could be captured is still open." Both are 0 abd false at the exact same moment: right after VCA, before anything else happens. The bug this PR fixes is precisely that ambiguity — if a non‑GET_DIGESTS request arrives first,  message_d  stays empty (size 0), and without an explicit flag the next, late GET_DIGESTS would see  size == 0  and wrongly be accepted. Buffer size is an output of the append decision, not a valid input to it. A derived-state approach literally cannot represent "closed-but-still-empty," so an explicit sticky bit is the minimum necessary state, not merely a stylistic choice.

library/spdm_common_lib/libspdm_com_context_data.c
```diff
@@ -1500,6 +1511,27 @@ void libspdm_reset_message_buffer_via_request_code(void *context, void *session_
...
+    }
+    spdm_session_info = session_info;
+
+    /* Close the encapsulated DIGESTS window if the first encapsulated request is not GET_DIGESTS.
+     * Enforced here (not from a fixed local sequence) so it holds for whatever order the peer drives,
+     * for both the Requester and Responder encapsulated flows. */
+    if ((request_code != SPDM_GET_DIGESTS) &&
+        (libspdm_get_managed_buffer_size(
+             &spdm_session_info->session_transcript.message_encap_d) == 0)) {
+        spdm_session_info->session_transcript.encap_digest_window_closed = true;
```
LeonLinAtAMI, line 1531: Looks good to me. I believe all edge cases are for setting/unsetting the flag covered and all paths included CHUNK mechanism and RESPOND_IF_READY handling are considered.

unit_test/test_spdm_requester/encap_digests.c
```diff
@@ -370,6 +370,93 @@ static void req_encap_digests_case6(void **state)
     }
 }
 
+/**
+ * Test 7: encapsulated DIGESTS is added to message_encap_d only if it is the first encap response.
+ * Drives the real chokepoint for both MutAuthRequested paths:
+ *   - Sub Case 1 (bit 2, WITH_GET_DIGESTS): GET_DIGESTS first, window open, DIGESTS added.
+ *   - Sub Case 2 (bit 1, WITH_ENCAP_REQUEST): GET_CERTIFICATE first, window closed, DIGESTS skipped.
+ **/
+static void req_encap_digests_case7(void **state)
```
LeonLinAtAMI, line 379: Verified test cases all passed.

LeonLinAtAMI (approved):

include/internal/libspdm_common_lib.h
jyao1: right

PR #3717, opened 2026-07-28
Validate per-slot `DIGESTS` consistency

jyao1:
The per-slot SupportedSlotMask, KeyPairID, CertificateInfo and KeyUsageMask in DIGESTS are integrator-provisioned. Verify them in the DIGESTS path via libspdm_validate_supported_slot_mask and libspdm_validate_multi_key_slot_info before emitting.

Ref: https://github.com/DMTF/libspdm/issues/3638

### DIFF
--- a/include/internal/libspdm_responder_lib.h
+++ b/include/internal/libspdm_responder_lib.h
@@ -147,6 +147,30 @@ libspdm_return_t libspdm_get_response_algorithms(libspdm_context_t *spdm_context
 
 #if LIBSPDM_ENABLE_CAPABILITY_CERT_CAP
 /**
+ * Verify that the integrator-provisioned SupportedSlotMask is consistent with the populated
+ * certificate slots: every populated slot shall be indicated in SupportedSlotMask.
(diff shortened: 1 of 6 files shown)


steven-bellock:
Will have review done before meeting on 3-Aug-2026.

alistair23:
This feels pretty hacky. It looks like this has stemmed partly from me [pointing out that there are currently a range of consistency issues in libspdm around local_key_usage_bit_mask](https://github.com/DMTF/libspdm/pull/3668#discussion_r3640243029). But this seems like a hacky way to detect it.

Adding something like this to the core library, but only calling it from the unit tests seems wrong. Are users supposed to call these functions? Is that documented somewhere (besides the function comments)? Are users supposed to call these periodically to catch libspdm bugs?

As described here: https://github.com/DMTF/libspdm/pull/3668#discussion_r3776195318 I feel we should focus on removing this duplicate information, instead of adding a bunch of checks that waste CPU cycles to try and catch them

PR #3744, opened 2026-08-11
Support for building libspdm as a Zephyr OS module

edersondisouza:
Add support to build libspdm as a [Zephyr OS](https://docs.zephyrproject.org/latest/introduction/index.html) module. This PR can be divided in three parts:

 - Implementation of OS stubs using Zephyr API;
 - Zephyr OS module bits (cmake and Kconfig);
 - Zephyr applications samples using libspdm.

Support for libspdm being a Zephyr module allows Zephyr applications to naturally use libspdm. Other DMTF protocols library implementations are also supported on Zephyr, namely OpenBMC's libmctp and libpldm.

Development of the sample applications guided the development of some libspdm <-> Zephyr bits that seem to be common for any Zephyr application - even if such bits are not complete right now. Those live under `zephyr/src` directories. The samples themselves are pretty simple and also share code (`zephyr/samples/common`), but they are specific to the samples, not the port. Note that one of the samples is a loopback sample that runs on Qemu, and the others need real boards (and were tested on them).

Also part of this PR is a Github workflow that uses a Zephyr project provided container to build (and run) the samples.

Please review and let me know of any questions or comments!

### DIFF
--- a/os_stub/spdm_device_secret_lib_sample/key_pair.c
+++ b/os_stub/spdm_device_secret_lib_sample/key_pair.c
@@ -42,7 +42,9 @@ typedef struct {
 
 /* Up to (9 traditional + 3 ML-DSA) PRIMARY key pairs, each backing slots 0 and 1, plus one
  * SECONDARY key pair per algorithm backing slot 4 (the multi-key example) -- hence x2. */
+#ifndef LIBSPDM_MAX_KEY_PAIR_COUNT
 #define LIBSPDM_MAX_KEY_PAIR_COUNT ((9 + 3) * 2)
+#endif
 
 libspdm_key_pair_info_t m_key_pair_info[LIBSPDM_MAX_KEY_PAIR_COUNT];
 


alistair23:
Does this need to be in libspdm? Shouldn't most of this be somewhere Zephyr specific?

edersondisouza:
> Does this need to be in libspdm? Shouldn't most of this be somewhere Zephyr specific?

I´d say the `os_stub` bits are more natural to be here. For the things inside `zephyr`, the cmake and Kconfig too - it's the preferred way to support Zephyr. I mean, it's about a library supporting the OS. Finally, the samples and the "interfacing" bits may or may not make sense to be on libspdm - if the interfacing bits grow, it could make more sense to be on Zephyr side. But I wanted it here, at least for v1 of this PR, so it can also be referenced/reviewed.

include/industry_standard/spdm.h
```diff
@@ -1688,7 +1688,9 @@ typedef struct {
  * RFC9909
  * The PQC slhdsa public key info len is 13.
  **/
+#ifndef SPDM_MAX_PUBLIC_KEY_INFO_LEN
 #define SPDM_MAX_PUBLIC_KEY_INFO_LEN 65535
+#endif
```
jyao1: This is defined by standard. Please do NOT use ifndef here.

If you want to customize the length, I think the right way is to define LIBSPDM_MAX_PUBLIC_KEY_INFO_LEN and use it in right place.

edersondisouza:
v2:
  - Fix formatting with `format_nix.sh`;
  - Introduce `LIBSPDM_MAX_PUBLIC_KEY_INFO_LEN` to avoid touching `SPDM_MAX_PUBLIC_KEY_INFO_LEN`;
  - Fix twister workflow missing "nxp_hal".

jyao1:
I think there are 3 different category of feature.

1) Adding general support for other architecture, such as `LIBSPDM_MAX_PUBLIC_KEY_INFO_LEN`, in include and library dir.
2) Adding xxx_zephyr.c to extend the capability for zephyr OS in os_stub dir.
3) Adding Zephyr directory and zephyr.yaml for a full zephyr support.

Here is my thought:
1) I will definitely support 1) since it is beneficial to other integrators. Please submit another dedicated PR.
2) I am OK for 2) as well since it extends the capability. Please limit this PR scope to os_stub, then we can continue the discussion.
3) I do not think 3) is needed because we not add other board example. Please drop this part in the PR. I recommend you put the code to Zephry project. Then Libspdm could add a link to show people how to enable libspdm for Zephry. (like what we did for other projects in readme.md). 

For example, the sample glue code Zephyr directory in 3) can be in Zephyr project - a new one under https://github.com/zephyrproject-rtos. Then libspdm project can add a readme to link to Zephyr. E.g. "Support is included in Zephyr project \<link to Zephyr_libspdm glue\>".

@steven-bellock , what do you think?

steven-bellock:
>@steven-bellock, what do you think?

That all sounds reasonable.

edersondisouza:
v3:
  - Dropped any Zephyr glue, kept only os_stub and overrides.

edersondisouza:
Thanks for the feedback. As asked, reduced this PR to only contain the "non-controversial" bits, and created #3796 for the Zephyr module bit.

I guess we should continue the discussion there regarding that, but one thing I want to point here: current Zephyr project policy is to only add forks to https://github.com/zephyrproject-rtos if the module is being used inside the OS. For instance, if Zephyr starts to have a DMTF stack that uses libspdm, then it's fine (this is the long term goal, btw).

The `zephyr` directory can be a bit controversial, but on #3796 I made it very small, with no code, only build stuff. Other projects have accepted that (like [libpldm](https://github.com/openbmc/libpldm)) and you can get a sense of other external projects that have the Zephyr module support [here](https://docs.zephyrproject.org/latest/develop/manifest/index.html#external-projects-modules).

teburd:
I have a need for this as well and would like to not have to do weird things to use libspdm with Zephyr.

As @edersondisouza noted libpldm didn't seem to mind and neither have other projects (zenoh-pico for example), but I understand it looks a bit odd. It's really an embedded port layer for Zephyr after all I suppose. But Zephyr is far from unpopular so perhaps its ok?

jyao1:
libspdm is neutral for OS. I do not think it is a right way to have Zephyr dir.

edersondisouza:
v4: 
  - No more `LIBSPDM_MAX_PUBLIC_KEY_INFO_LEN` - instead, the `spdm_device_secret_lib_sample` lib now uses a smaller buffer - it never needed a big one.

os_stub/spdm_device_secret_lib_sample/key_pair.c
```diff
@@ -27,6 +27,28 @@
...
+ * fails the build as well.
+ */
+#if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)
+#define LIBSPDM_KEY_PAIR_STATIC_ASSERT(condition, message) _Static_assert(condition, message)
+#else
+#define LIBSPDM_KEY_PAIR_STATIC_ASSERT_NAME2(line) libspdm_key_pair_static_assert_ ## line
+#define LIBSPDM_KEY_PAIR_STATIC_ASSERT_NAME(line) LIBSPDM_KEY_PAIR_STATIC_ASSERT_NAME2(line)
+#define LIBSPDM_KEY_PAIR_STATIC_ASSERT(condition, message) \
+    enum { LIBSPDM_KEY_PAIR_STATIC_ASSERT_NAME(__LINE__) = 1 / (int)(!!(condition)) }
+#endif /* defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L) */
```
jyao1, line 50: I recommend we use _Static_assert() directly, since most modern compilers already support this macro. (VS 2019, GCC 4.6, Clang 3.0)

Static assert was introduced in 2011, and now it is 2026. Which old compiler we need support?

Or, if the concern is only missing `c11` flags, we can add it in CMake

edersondisouza: Yes, it is about `c11`. I'm happy to change that, but https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170#stdc11 indicates that MSVC2015 doesn't support it. This means support for it will dropped, is that fine?

edersondisouza:
v5:
 - Move to `C11`, so no more hacks for `_Static_assert()`.

jyao1 (approved):

PR #3753, opened 2026-08-19
Fix large_message_capacity unset on chunked GET path

### LINKED ISSUE #3731: Specula - `large_message_capacity` unset on chunked GET path
*Files:* `spdm_responder_lib/libspdm_rsp_receive_send.c:659-685` (`libspdm_build_response`); `libspdm_rsp_chunk_response.c:213` (cleanup `libspdm_zero_mem`).
The SEND path correctly assigns `large_message_capacity` (see `libspdm_rsp_chunk_send_ack.c:152`); the GET path does not, so the field stays at zero. The cleanup call `libspdm_zero_mem(ptr, 0)` then becomes a no-op, and the chunked-response buffer (which can have held certificate, measurement, or key-exchange material) is not scrubbed when chunking ends. Whether any leftover bytes ultimately escape depends on the surrounding allocator's reuse pattern, but the call is intended to be a scrub and is not currently performing one. Setting `large_message_capacity` correctly on the GET path matches the SEND path treatment.

LeonLinAtAMI:
Fixed #3731 as described in the issue and added assertions in original tests.

The CHUNK process is complicated for me. So I requested Claude Sonnet 5 for reviewing my fix and visualing the lifecycle of scratch buffer during CHUNK_GET as the figure below. 

chunk_get_sequence

Probably the sequence diagram helps a lot into realizing the problem statement.

Assisted-by: GitHub Copilot CLI 1.0.79:Claude Sonnet 5

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_chunk_send_ack.c
+++ b/library/spdm_responder_lib/libspdm_rsp_chunk_send_ack.c
@@ -246,12 +246,17 @@ libspdm_return_t libspdm_get_response_chunk_send(libspdm_context_t *spdm_context
          * the CHUNK_ACK with EarlyErrorDetected with a generic error. */
         status = LIBSPDM_STATUS_SUCCESS;
 
+        if (send_info->large_message != NULL) {
+            libspdm_zero_mem(send_info->large_message, send_info->large_message_capacity);
+        }
+
         send_info->chunk_in_use = false;
         send_info->chunk_handle = 0;
         send_info->chunk_seq_no = 0;
         send_info->chunk_bytes_transferred = 0;
         send_info->large_message = NULL;
         send_info->large_message_size = 0;
+        send_info->large_message_capacity = 0;
     } else if (send_info->chunk_bytes_transferred == send_info->large_message_size) {
         uint8_t opcode;
 
@@ -276,8 +281,12 @@ libspdm_return_t libspdm_get_response_chunk_send(libspdm_context_t *spdm_context
         send_info->chunk_handle = 0;
         send_info->chunk_seq_no = 0;
         send_info->chunk_bytes_transferred = 0;
+        if (send_info->large_message != NULL) {
+            libspdm_zero_mem(send_info->large_message, send_info->large_message_capacity);
+        }
         send_info->large_message = NULL;
         send_info->large_message_size = 0;
+        send_info->large_message_capacity = 0;
 
         *response_size = response_header_size + chunk_response_size;
     } else {
--- a/library/spdm_requester_lib/libspdm_req_send_receive.c
+++ b/library/spdm_requester_lib/libspdm_req_send_receive.c
@@ -609,6 +609,7 @@ static libspdm_return_t libspdm_handle_large_request(
         send_info->chunk_bytes_transferred = 0;
         send_info->large_message = NULL;
         send_info->large_message_size = 0;
+        send_info->large_message_capacity = 0;
     }
 
     return status;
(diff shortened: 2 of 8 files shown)


jyao1:
Other comment from AI:

There are three reset sites for chunk_context.get, but the PR only covers the normal-completion one. The GET-transfer‑interrupted‑by‑GET_VERSION reset in libspdm_rsp_receive_send.c#L570-L575 clears every field except large_message_capacity, and — more importantly — it does not zero the scratch buffer:
```
context->chunk_context.get.chunk_in_use = false;
context->chunk_context.get.chunk_handle++;
context->chunk_context.get.chunk_seq_no = 0;
context->chunk_context.get.large_message = NULL;      /* buffer never wiped here */
context->chunk_context.get.large_message_size = 0;
context->chunk_context.get.chunk_bytes_transferred = 0;
/* large_message_capacity left stale */
```

So when a GET_VERSION interrupts an in-flight CHUNK_GET, the sensitive large response stays in the scratch buffer un-wiped — the exact hygiene issue #3731 targets, just on a different path. To be consistent it should libspdm_zero_mem(large_message, large_message_capacity) before nulling, and set large_message_capacity = 0.

LeonLinAtAMI:
@jyao1  I reviewed the side paths again, and resolved all found related problems. Also added more assertions on the original unit tests. Please let me know if missing any edge case and any enhancement. Thanks a lot.

library/spdm_responder_lib/libspdm_rsp_chunk_send_ack.c
```diff
@@ -246,12 +246,15 @@ libspdm_return_t libspdm_get_response_chunk_send(libspdm_context_t *spdm_context
          * the CHUNK_ACK with EarlyErrorDetected with a generic error. */
         status = LIBSPDM_STATUS_SUCCESS;
 
+        libspdm_zero_mem(send_info->large_message, send_info->large_message_capacity);
```
steven-bellock: `send_info->large_message` is probably `NULL` when it reaches here. `libspdm_zero_mem` probably needs better documentation on legal values that libspdm passes to the Integrator.

LeonLinAtAMI: Thanks for your review. I checked libspdm_zero_mem does not support nullable pointer. So I would add NULL check before reset the scratch buffer.

steven-bellock: Filed https://github.com/DMTF/libspdm/issues/3785 to improve `memlib` documentation.

steven-bellock (approved):

library/spdm_responder_lib/libspdm_rsp_chunk_send_ack.c
steven-bellock: Look what Coverity came back with.
```
*** CID 913611:         Null pointer dereferences  (REVERSE_INULL)
/home/runner/work/libspdm/libspdm/library/spdm_responder_lib/libspdm_rsp_chunk_send_ack.c: 284             in libspdm_get_response_chunk_send()
278             }
279     
280             send_info->chunk_in_use = false;
281             send_info->chunk_handle = 0;
282             send_info->chunk_seq_no = 0;
283             send_info->chunk_bytes_transferred = 0;
>>>     CID 913611:         Null pointer dereferences  (REVERSE_INULL)
>>>     Null-checking "send_info->large_message" suggests that it may be null, but it has already been dereferenced on all paths leading to the check.
284             if (send_info->large_message != NULL) {
285                 libspdm_zero_mem(send_info->large_message, send_info->large_message_capacity);
286             }
287             send_info->large_message = NULL;
288             send_info->large_message_size = 0;
289             send_info->large_message_capacity = 0;
```

PR #3755, opened 2026-08-20
library/spdm_responder_lib: Convert array of functions to switch case

alistair23:
The C standard requires memcpy to be avaliable [1], [2], but it insn't in the case of some libspdm tests (test_size_of_spdm_responder) as libc isn't used and memcpy isn't provided.

This results in build failures like this (depending on compiler versions as it doesn't happen in the CI).

```
[100%] Linking C executable ../../../bin/test_size_of_spdm_responder
/usr/bin/ld.bfd: /tmp/cc9WR1GI.ltrans0.ltrans.o: in function `libspdm_get_response_func_via_request_code':
/scratch/alistair/software/tier4/SPDM-Utils/third-party/libspdm/library/spdm_responder_lib/libspdm_rsp_receive_send.c:19:(.text+0x7ae): undefined reference to `memcpy'
collect2: error: ld returned 1 exit status
make[2]: *** [unit_test/test_size/test_size_of_spdm_responder/CMakeFiles/test_size_of_spdm_responder.dir/build.make:145: bin/test_size_of_spdm_responder] Error 1
make[1]: *** [CMakeFiles/Makefile2:7358: unit_test/test_size/test_size_of_spdm_responder/CMakeFiles/test_size_of_spdm_responder.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

We can avoid the memcpy() by converting the get_response_struct array into a switch case.

This fixes build failures seem on Linux when using GCC 16.1.1.

1: https://gcc.gnu.org/pipermail/gcc-help/2025-May/144229.html
2: https://gcc.gnu.org/onlinedocs/gcc-15.1.0/gcc/Standards.html#C-Language

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_receive_send.c
+++ b/library/spdm_responder_lib/libspdm_rsp_receive_send.c
@@ -9,111 +9,99 @@
 
 libspdm_get_spdm_response_func libspdm_get_response_func_via_request_code(uint8_t request_code)
 {
-    size_t index;
+    switch (request_code) {
+    case SPDM_GET_VERSION: return libspdm_get_response_version;
+    case SPDM_GET_CAPABILITIES: return libspdm_get_response_capabilities;
+    case SPDM_NEGOTIATE_ALGORITHMS: return libspdm_get_response_algorithms;
 
-    typedef struct {
-        uint8_t request_response_code;
-        libspdm_get_spdm_response_func get_response_func;
-    } libspdm_get_response_struct_t;
+    #if LIBSPDM_ENABLE_CAPABILITY_CERT_CAP
+    case SPDM_GET_DIGESTS: return libspdm_get_response_digests;
+    case SPDM_GET_CERTIFICATE: return libspdm_get_response_certificate;
+    #endif /* LIBSPDM_ENABLE_CAPABILITY_CERT_CAP */
 
-    libspdm_get_response_struct_t get_response_struct[] = {
-        { SPDM_GET_VERSION, libspdm_get_response_version },
-        { SPDM_GET_CAPABILITIES, libspdm_get_response_capabilities },
-        { SPDM_NEGOTIATE_ALGORITHMS, libspdm_get_response_algorithms },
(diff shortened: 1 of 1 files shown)


library/spdm_responder_lib/libspdm_rsp_receive_send.c
```diff
@@ -9,109 +9,96 @@
...
-    libspdm_get_response_struct_t get_response_struct[] = {
-        { SPDM_GET_VERSION, libspdm_get_response_version },
-        { SPDM_GET_CAPABILITIES, libspdm_get_response_capabilities },
-        { SPDM_NEGOTIATE_ALGORITHMS, libspdm_get_response_algorithms },
+    switch (request_code) {
+    case SPDM_GET_VERSION: return libspdm_get_response_version;
+    case SPDM_GET_CAPABILITIES: return libspdm_get_response_capabilities;
+    case SPDM_NEGOTIATE_ALGORITHMS: return libspdm_get_response_algorithms;
 
         #if LIBSPDM_ENABLE_CAPABILITY_CERT_CAP
```
steven-bellock: The `#if`s and `#endif`s should align with the `case`s.

library/spdm_responder_lib/libspdm_rsp_receive_send.c
```diff
@@ -9,109 +9,96 @@
...
+    case SPDM_VENDOR_DEFINED_REQUEST: return libspdm_get_vendor_defined_response;
         #endif /*LIBSPDM_ENABLE_VENDOR_DEFINED_MESSAGES*/
-    };
-
-    for (index = 0; index < LIBSPDM_ARRAY_SIZE(get_response_struct); index++) {
-        if (request_code == get_response_struct[index].request_response_code) {
-            return get_response_struct[index].get_response_func;
-        }
     }
     return NULL;
```
steven-bellock: Can just have a `default: return NULL;`.

steven-bellock (changes requested):

steven-bellock (approved):

LeonLinAtAMI (approved):

PR #3776, opened 2026-08-30
Fix missing transcript resets on MEASUREMENTS

### LINKED ISSUE #3726: Specula Cluster B - Missing transcript reset on retry / error paths
libspdm uses transcript buffers (`message_b`, `message_f`, the measurement transcript, etc.) to feed signature and verification computations. The helper `libspdm_reset_message_buffer_via_request_code` (in `spdm_common_lib/libspdm_com_context_data.c:1447-1500`) is meant to clear the right buffer for each request that signals "previous attempt is invalid, start over." Three call sites that need a reset do not get one, causing transcripts to accumulate stale bytes across retries. Issue #524 already tracks this class; the three sites below appear to be additional instances.

**B.1 — `GET_CERTIFICATE` retry does not reset `message_b`**
*File:* `libspdm_com_context_data.c:1447-1500`.
The `switch` in `libspdm_reset_message_buffer_via_request_code` does not have a case for `SPDM_GET_CERTIFICATE`; the request code falls through to the default arm, which leaves `message_b` untouched. On a `GET_CERTIFICATE` retry the new request bytes are appended to whatever the failed attempt left behind. Any later `CHALLENGE_AUTH` whose signature covers the transcript is then signed over a polluted buffer. Adding the missing case calls `libspdm_reset_message_b` and matches the treatment of the other request codes already in the switch.

**B.2 — Responder `message_f` not reset between `FINISH` retries**
*File:* `spdm_responder_lib/libspdm_rsp_finish_rsp.c` (and the corresponding key-exchange flow).
On the responder side, when a `FINISH` retry occurs in the same session, `message_f` is not cleared between attempts. The HMAC over `message_f` for the second `FINISH` ends up covering bytes from the failed first attempt as well, which makes the legitimate `FINISH` look invalid. The symmetric reset to the one already performed on success is missing on the retry path.

**B.3 — `key_usage_bit_mask` error path skips measurement transcript reset**
*File:* `spdm_responder_lib/libspdm_rsp_measurements.c`.
When `GET_MEASUREMENTS` fails the `key_usage_bit_mask` validation, the early return does not reset the measurement transcript, even though the standard error paths do. A subsequent legitimate `GET_MEASUREMENTS` on the same context starts with stale transcript bytes, and the signature it later produces no longer matches what the requester computes from the wire.

The common shape of these three is "an error/retry path forgot to call the same reset that the success path uses." Each is a single-line fix: a missing case in the helper for B.1, a missing call before the retry returns for B.2, a missing call on the early-return for B.3.

LeonLinAtAMI:
Fixes #3726 on cases B.3.

About B.1 (message_b not reset on GET_CERTIFICATE) 
As issue #3726 literally suggests, would break legitimate multi-portion/multi-slot retrieval.
A correct fix requires distinguishing "retry of a failed request" from "next portion/slot request," which needs additional state (e.g., comparing against  last_spdm_request ) not present in the current codebase. 

2026/09/18 Edited: Moved fixes on B.1 case into another PR #3876.

2026/09/18 Edited: Moved fixes on B.2 case into another PR #3877.

Please let me know if any suggestion.


steven-bellock:
@LeonLinAtAMI please fix the merge conflicts.

jyao1 (changes requested):
I have doubt for B1. We need to discuss.

I recommend to split B1/B2/B3 and other related "reset missing" issues. (I assume 4 different issues totally)

Rule is 1 PR for 1 problem. Please do not mix them.

LeonLinAtAMI:
@jyao1 Moved fixes on B.1 case into another PR #3876 and PR #3877. 

@steven-bellock Already rebased and resolved conflicts.

steven-bellock (approved):

jyao1 (approved):

PR #3796, opened 2026-09-04
Zephyr RTOS support - libspdm as a module

edersondisouza:
Builds on top of #3744 - reuses its basic patches and adds a new one, which adds a `zephyr` directory with `module.yml` file, as well as `Kconfig` and `CMakeLists.txt`. This allows Zephyr applications to simply point [west](https://docs.zephyrproject.org/latest/develop/west/index.html) to it, as this is what is needed to make an external project a [Zephyr module](https://docs.zephyrproject.org/latest/develop/modules.html).

In this patch, there's no code - a Zephyr application still need to implement any bits that libspdm expects, such as wallclock and IO.

### DIFF
--- a/os_stub/spdm_device_secret_lib_sample/key_pair.c
+++ b/os_stub/spdm_device_secret_lib_sample/key_pair.c
@@ -42,7 +42,9 @@ typedef struct {
 
 /* Up to (9 traditional + 3 ML-DSA) PRIMARY key pairs, each backing slots 0 and 1, plus one
  * SECONDARY key pair per algorithm backing slot 4 (the multi-key example) -- hence x2. */
+#ifndef LIBSPDM_MAX_KEY_PAIR_COUNT
 #define LIBSPDM_MAX_KEY_PAIR_COUNT ((9 + 3) * 2)
+#endif
 
 libspdm_key_pair_info_t m_key_pair_info[LIBSPDM_MAX_KEY_PAIR_COUNT];
 


library/spdm_requester_lib/libspdm_req_get_key_pair_info.c
```diff
@@ -19,7 +19,7 @@ typedef struct {
     uint32_t current_asym_algo;
     uint16_t public_key_info_len;
     uint8_t assoc_cert_slot_mask;
-    uint8_t public_key_info[SPDM_MAX_PUBLIC_KEY_INFO_LEN];
+    uint8_t public_key_info[LIBSPDM_MAX_PUBLIC_KEY_INFO_LEN];
```
steven-bellock, line 22: I do not understand the need for this. Please [file an issue](https://github.com/DMTF/libspdm/issues).

jyao1:
I do not think we should add `zephyr` directory. Libspdm is an OS neutral project.

library/spdm_requester_lib/libspdm_req_get_key_pair_info.c
edersondisouza: [SPDM_MAX_PUBLIC_KEY_INFO_LEN](https://github.com/DMTF/libspdm/blob/main/include/industry_standard/spdm.h#L1691) is unreasonably large, at `64kb`. That's a lot for constrained devices, which may have little more than that of memory available. This change to use `LIBSPDM_MAX_PUBLIC_KEY_INFO_LEN`, so that applications can tune the size of this buffer to more reasonable sizes.
And a previous review stated that it should not be allowed to override `SPDM_MAX_PUBLIC_KEY_INFO_LEN` directly.

edersondisouza: Created https://github.com/DMTF/libspdm/issues/3824 - let me know if I can help in any way.

os_stub/spdm_device_secret_lib_sample/key_pair.c
```diff
@@ -37,12 +37,14 @@ typedef struct {
     uint32_t current_pqc_asym_algo;
     uint16_t public_key_info_len;
     uint8_t assoc_cert_slot_mask;
-    uint8_t public_key_info[SPDM_MAX_PUBLIC_KEY_INFO_LEN];
+    uint8_t public_key_info[LIBSPDM_MAX_PUBLIC_KEY_INFO_LEN];
```
edersondisouza: @steven-bellock This is the offending one actually - if I use the device secret lib sample.

jyao1 (changes requested):
Please remove `zephyr` directory, because libspdm is OS neutral project. We should not add specifical embed-OS under root.

edersondisouza:
v2: 
  - No more `LIBSPDM_MAX_PUBLIC_KEY_INFO_LEN` - instead, the `spdm_device_secret_lib_sample` lib now uses a smaller buffer - it never needed a big one.

edersondisouza:
> Please remove `zephyr` directory, because libspdm is OS neutral project. We should not add specifical embed-OS under root.

I know this can be a bit confusing, but #3744 is the PR that doesn't add the Zephyr build bits. My understanding was that I needed a new one with that for further discussions on that. I can close this one if it makes things simple and continue on #3744 (which has only the first three patches from here).

jyao1:
Yes, it is confusing to me that there are both #3744 and #3796. And the titles are very similar.

Anyway, my feedback is same as previous - We should not add `Zephyr` dir.

edersondisouza:
> Yes, it is confusing to me that there are both #3744 and #3796. And the titles are very similar.
> 
My bad. Will close this one then and keep #3744.

edersondisouza:
> Please remove `zephyr` directory, because libspdm is OS neutral project. We should not add specifical embed-OS under root.

Fair. Just note that adding a `zephyr` directory doesn't make a library not neutral. It only means that it supports Zephyr , not that it only works on Zephyr. For instance, [LVGL](https://github.com/lvgl/lvgl) works with Zephyr, but it's not Zephyr only.

Issue #3729, opened 2026-08-04
Specula Cluster D - Event handler input validation


steven-bellock:
Two further cases where event handlers accept inputs that the rest of libspdm would normally reject.

**D.1 — Encap `SEND_EVENT` skips version and capability checks**
*File:* `spdm_responder_lib/libspdm_rsp_encap_send_event.c:11-65`; dispatcher in `libspdm_rsp_encap_response.c:246-266`.
The non-encap event path validates spdm version and the relevant capability bits before processing; the encap path performs neither. A peer that has not advertised `EVENT_CAP` can still drive event handling through an encap channel, and a peer can send an event message of a version that was never negotiated. Mirroring the regular path's checks at the encap entry point fixes this.

**D.2 — `SUBSCRIBE_EVENT_TYPES` does not check against advertised list**
*File:* `spdm_responder_lib/libspdm_rsp_subscribe_event_types_ack.c:132`; encap variant `libspdm_req_encap_subscribe_event_types_ack.c:116`.
The handler forwards the requested event types to the HAL without checking them against `supported_event_groups_list`, which libspdm itself populated from `GET_SUPPORTED_EVENT_TYPES`. We acknowledge that "HAL is the policy enforcer" is a defensible layering position; we mention it because the data needed for a sanity check is already in libspdm's hands, and a one-line cross-check would catch malformed or stale subscriptions before they reach HAL.

PR #3800, opened 2026-09-08
Enhanced Responder encapsulated flow

### LINKED ISSUE #3775: Implement the new encapsulated flow
The new encapsulated flow is documented at https://github.com/DMTF/libspdm/blob/main/doc/internal/encapsulated_flow.md.

steven-bellock:
This implements the flow described in https://github.com/DMTF/libspdm/blob/main/doc/internal/encapsulated_flow.md.


library/spdm_responder_lib/libspdm_rsp_encap_send_event.c
```diff
@@ -9,36 +9,38 @@
...
+    if (session_info == NULL) {
+        return LIBSPDM_STATUS_INVALID_STATE_LOCAL;
+    }
+    session_state = libspdm_secured_message_get_session_state(
+        session_info->secured_message_context);
+    if (session_state != LIBSPDM_SESSION_STATE_ESTABLISHED) {
+        return LIBSPDM_STATUS_INVALID_STATE_LOCAL;
     }
 
     spdm_request->header.spdm_version = libspdm_get_connection_version(spdm_context);
```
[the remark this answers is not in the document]
steven-bellock: https://github.com/DMTF/libspdm/issues/3729 already exists for this issue, and it will not be fixed in this pull request.

steven-bellock:
This has been tested with `spdm-emu` at https://github.com/steven-bellock/spdm-emu/tree/encap-flow.

library/spdm_responder_lib/libspdm_rsp_encap_get_digests.c
```diff
@@ -32,7 +50,7 @@ libspdm_return_t libspdm_get_encap_request_get_digest(libspdm_context_t *spdm_co
 
     spdm_request = encap_request;
 
-    libspdm_reset_message_buffer_via_request_code(spdm_context, NULL,
+    libspdm_reset_message_buffer_via_request_code(spdm_context, session_info,
                                                   spdm_request->header.request_response_code);
```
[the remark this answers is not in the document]
steven-bellock: This is a real bug, but it points to a much larger issue in how transcripts are handled in the encapsulated flow. It will be resolved in this pull request.

steven-bellock:
Also once this goes in then I can start work on the Requester's enforcement of the Responder's encapsulated requests.

doc/encapsulated_flow_user_guide.md
```diff
@@ -0,0 +1,356 @@
...
+libspdm classifies encapsulated flows into three types:
+- Basic Mutual Authentication
+    - Triggered by the Responder through its `CHALLENGE_AUTH` response to the Requester.
+- Session-based Mutual Authentication
+    - Triggered by the Responder through its `KEY_EXCHANGE_RSP` response to the Requester.
+- Requester-initiated
+    - Triggered by the Requester when it sends the `GET_ENCAPSULATED_REQUEST` message to the
+      Responder.
+    - The Requester may send this message periodically, or the Responder may possess an out-of-band
+      (non-SPDM) method of signaling to the Requester to start this encapsulated flow.
```
jyao1, line 27: For the encap flow classification, IMHO, there are two perspective:
1) Initiator-based, e.g. requester-initiated, responder-initiated.
2) Function-based, e.g. BasicMutAuth, SessionMutAuth, KeyUpdate, Event, EndPointInfo.

The current 3 types (BasicMutAuth, SessionMutAuth, RequesterInitiated) seems a mix of them.

If we really care about the type, I suggest we should define a way to combine them.

For example, 
```
typedef struct {
    bool initiator; // 0: requester, 1: responder
    uint16_t function; // BasicMutAuth, SessionMutAuth, KeyUpdate, Event, EndPointInfo, VDM, FreeFunction
} libspdm_encap_flow_type_t;
```

As such, the restriction can be added based on function. E.g. No interrupt between KeyUpdate (UpdateKey or UpdateAllKeys, VerifyNewKey according to the SPDM spec "Upon the successful reception of the KEY_UPDATE_ACK, the sender shall transmit a KEY_UPDATE request with the VerifyNewKey operation using the new session keys.").
Or even VDM message, which can be initiated from either requester or responder.

I think more. I use "FreeFunction" (or use "Unknown") to indicate the state that the purpose of encap is unknown yet. Now, I feel that we can remove "initiator" at all.

Technically, we can treat the only initiator is responder, because it is responder that sends **indicator** (in band such as BasicMutAuth/SessionMutAuth, or out of band such as KeyUpdate, Event) to the requester, then the requester sends `GET_ENCAPSULATED_REQUEST` as the **next step**. Everything is initiated by responder.

As such, I feel "Requester-initiated" could be changed to "FreeFunction" or "UnknownFunction" in the 1st round. Then it will be known and updated by the responder on the real purpose, such as "KeyUpdate", "Event" for this specific encap flow. That type will be used to reject the out-of-sequence command, such as inject GetEndPointInfo between 2 KeyUpdate.

steven-bellock: I am fine renaming `Requester-initiated` / `LIBSPDM_ENCAP_FLOW_REQ_INITIATED` to something else. Probably `General Encapsulated Flow` / `LIBSPDM_ENCAP_FLOW_GENERAL`. Anything beyond that we should discuss it at the SPDM Working Group meeting tomorrow.

jyao1: Yes, my main concern is `libspdm_encap_flow_type_t`.
I am fine with `LIBSPDM_ENCAP_FLOW_GENERAL`.

steven-bellock: The rename has been applied.

jyao1:
LGTM. Could you squash them?

steven-bellock:
> LGTM. Could you squash them?

It is done.

PR #3808, opened 2026-09-08
Enhance `memlib` documentation and reference code

### LINKED ISSUE #3785: Enhance `memlib.h` documentation
https://github.com/DMTF/libspdm/blob/main/include/hal/library/memlib.h

For example, pointers shall never be `NULL`, lengths shall never be `0`.

steven-bellock:
Fix #3785.

### DIFF
--- a/include/hal/library/memlib.h
+++ b/include/hal/library/memlib.h
@@ -1,6 +1,6 @@
 /**
  *  Copyright Notice:
- *  Copyright 2021-2025 DMTF. All rights reserved.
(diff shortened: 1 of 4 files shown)


steven-bellock:
It looks like at least one test passed `NULL` to `libspdm_zero_mem()`.
```
[ RUN      ] libspdm_test_secured_message_encode_case12
LIBSPDM_ASSERT: /home/runner/work/libspdm/libspdm/os_stub/memlib/zero_mem.c(19): buffer != NULL
```

Issue #3733, opened 2026-08-04
Specula Cluster A - Incomplete atomicity in `UPDATE_ALL_KEYS`


steven-bellock:
`UPDATE_ALL_KEYS` is treated in libspdm as a two-phase commit: a backup pair is created first, then committed when the peer's ACK is received (or rolled back on error). Three independent gaps in this commit pattern can leave a session with desynchronized or unrecoverable key material if a step fails partway.

**A.1 — Backup not marked valid before HKDF can fail**
*File:* `spdm_secured_message_lib/libspdm_secmes_session.c:383-407` (`libspdm_create_update_session_data_key`).
The function derives the new request and response keys, copies them into the backup slot, and sets `backup_valid = TRUE` only after both derivations succeed. If the second `libspdm_hkdf_expand` fails (out of memory, HAL crypto failure, etc.), the backup slot has already been partially overwritten but `backup_valid` stays `FALSE`. The error path then returns without restoring the original keys, and a subsequent rollback request finds nothing to roll back to. The natural fix is to commit `backup_valid = TRUE` after a snapshot of the *old* keys is captured, before any new derivation, so a partial failure is recoverable.

**A.2 — Responder pre-commits before sending the ACK**
*File:* `spdm_responder_lib/libspdm_rsp_key_update_ack.c:179`.
On the responder side, `libspdm_activate_update_session_data_key(..., LIBSPDM_KEY_UPDATE_OPERATION_COMMIT_UPDATE_KEYS)` runs before the `KEY_UPDATE_ACK` is constructed and handed to the transport. If transmission of the ACK then fails, the responder has already committed to the new keys while the requester still holds the old ones and never sees the ACK — the session is permanently desynchronized. The two states the requester can be in (saw ACK / did not see ACK) are observationally identical on its side, so it cannot recover. Constructing and sending the ACK first, and only committing on successful transmission, restores the intended ordering.

**A.3 — Two `create_update` calls without a rollback between them**
*File:* `spdm_responder_lib/libspdm_rsp_key_update_ack.c:152` and `:165`.
[...]

PR #3816, opened 2026-09-10
Fix key update rollback for Specula cluster A.1 and A.3

alanhc:
Fixes two of the Specula cluster A findings from #3733: A.1 (backup not marked
valid before HKDF can fail) and A.3 (no rollback between the two
`create_update` calls).

The two commits are ordered, and A.3 depends on A.1: without the flag ordering
fixed, every rollback added in the second commit is a silent no-op.

## A.1 — mark the backup valid before deriving

`libspdm_create_update_session_data_key()` snapshots the current keys into the
backup slot and then derives the new ones, but only set
`requester_backup_valid` / `responder_backup_valid` after both
`libspdm_hkdf_expand()` and `libspdm_generate_aead_key_and_iv()` had
succeeded.

If either derivation fails, the function returns `false` with the backup slot
already holding a complete copy of the old keys, yet the valid flag still
clear. `libspdm_activate_update_session_data_key()` gates its restore path on
that flag, so a rollback with `use_new_key = false` silently does nothing and
the old keys are unrecoverable.

The fix moves the assignment to immediately after the snapshot completes,
before any derivation runs. At that point the backup is a full,
self-consistent copy of the pre-update keys, which is exactly what a rollback
needs. The success path is unchanged.

## A.3 — roll back a failed update instead of leaving keys half applied

`libspdm_create_update_session_data_key()` derives the new data secret **in
place**: `request_data_secret` is passed as both the PRK and the output buffer
of `libspdm_hkdf_expand()`. So by the time the function can fail, the live
secret has already been replaced. Any failure arm that returns without a
rollback leaves the session holding a key the peer never derived, and the peer
cannot detect it because it never receives the ACK. Nothing later reverts it:
the pending backup is only consumed by a subsequent `KEY_UPDATE`, which can no
longer be decrypted.

This makes it three sites in `libspdm_get_response_key_update()`, where #3733
describes one:

1. `UPDATE_KEY`, when the request-direction create fails.
2. `UPDATE_ALL_KEYS`, when the request-direction create fails — the first of
   the two calls; the report covers only the second.
3. `UPDATE_ALL_KEYS`, when the response-direction create fails. This is the
   one in the report, and it rolls back the request direction created just
   before it.

Sites 1 and 2 are single-direction, so they do not produce the "request side
committed and response side not" desynchronization the report describes. They
are still wrong for the reason above, so they are fixed the same way.

The failed response-direction *commit* arm is deliberately left alone.
`libspdm_activate_update_session_data_key()` cannot currently return false, so
a rollback there would be unreachable and untestable. Happy to add it if you
would rather the arm be defensively complete.

I kept these as minimal per-site fixes rather than the single transactional
rewrite suggested in #3733, so each is reviewable on its own. Glad to redo it
as a rewrite if you prefer, though that wants a clear owner for the
commit/rollback boundary and is a larger API discussion.

## Testing

x64 / GCC / Debug / mbedtls.

- Three new tests in `test_spdm_secured_message` (`key_update.c`) for A.1:
  rollback after a mid-update derivation failure in each direction, plus the
  success path.
- `rsp_key_update_ack_case28` in `test_spdm_responder` for A.3: a failed
  `UPDATE_ALL_KEYS` must leave both directions on the pre-update secrets with
  no backup pending.
- Each test was checked against its own bug: reverting a fix and re-running
  makes the corresponding tests fail, and they pass with the fix applied.
- No regressions — `test_spdm_responder` (527), `test_spdm_requester` (742),
  `test_spdm_common` (35), `test_spdm_secured_message` (20), `test_spdm_crypt`
  (11).
- Both changed source files are clean under `uncrustify 0.78` via
  `.uncrustify.cfg`.

To force a derivation failure without mocking the crypto layer, the tests set
`aead_key_size` above `255 * hash_size`, which HKDF-Expand rejects per
RFC 5869 Section 2.3. The check is on the requested length alone, so it fails
before anything is written to the output buffer. An invalid `base_hash_algo`
is not usable for this — that path is `LIBSPDM_ASSERT(false)` and aborts the
run.

### DIFF
--- a/library/spdm_responder_lib/libspdm_rsp_key_update_ack.c
+++ b/library/spdm_responder_lib/libspdm_rsp_key_update_ack.c
@@ -127,6 +127,9 @@ libspdm_return_t libspdm_get_response_key_update(libspdm_context_t *spdm_context
             session_info->secured_message_context,
             LIBSPDM_KEY_UPDATE_ACTION_REQUESTER);
         if (!result) {
+            libspdm_activate_update_session_data_key(
+                session_info->secured_message_context,
+                LIBSPDM_KEY_UPDATE_ACTION_REQUESTER, false);
             return LIBSPDM_STATUS_UNSUPPORTED_CAP;
         }
         libspdm_trigger_key_update_callback(
@@ -153,6 +156,9 @@ libspdm_return_t libspdm_get_response_key_update(libspdm_context_t *spdm_context
             session_info->secured_message_context,
             LIBSPDM_KEY_UPDATE_ACTION_REQUESTER);
         if (!result) {
+            libspdm_activate_update_session_data_key(
+                session_info->secured_message_context,
+                LIBSPDM_KEY_UPDATE_ACTION_REQUESTER, false);
             return LIBSPDM_STATUS_UNSUPPORTED_CAP;
         }
         libspdm_trigger_key_update_callback(
@@ -166,6 +172,9 @@ libspdm_return_t libspdm_get_response_key_update(libspdm_context_t *spdm_context
             session_info->secured_message_context,
             LIBSPDM_KEY_UPDATE_ACTION_RESPONDER);
         if (!result) {
+            libspdm_activate_update_session_data_key(
+                session_info->secured_message_context,
+                LIBSPDM_KEY_UPDATE_ACTION_REQUESTER, false);
             return LIBSPDM_STATUS_UNSUPPORTED_CAP;
         }
         libspdm_trigger_key_update_callback(
--- a/library/spdm_secured_message_lib/libspdm_secmes_session.c
+++ b/library/spdm_secured_message_lib/libspdm_secmes_session.c
@@ -380,6 +380,11 @@ bool libspdm_create_update_session_data_key(void *spdm_secured_message_context,
         .request_data_sequence_number =
             secured_message_context->application_secret.request_data_sequence_number;
 
+        /* The backup now holds a complete snapshot of the old keys, so mark it valid before
+         * deriving the new ones. If a derivation below fails the caller can still roll back
+         * via libspdm_activate_update_session_data_key(..., use_new_key = false). */
+        secured_message_context->requester_backup_valid = true;
+
         status = libspdm_hkdf_expand(
             secured_message_context->base_hash_algo,
             secured_message_context->application_secret.request_data_secret,
@@ -403,8 +408,6 @@ bool libspdm_create_update_session_data_key(void *spdm_secured_message_context,
             return status;
         }
         secured_message_context->application_secret.request_data_sequence_number = 0;
-
-        secured_message_context->requester_backup_valid = true;
     } else if (action == LIBSPDM_KEY_UPDATE_ACTION_RESPONDER) {
         libspdm_copy_mem(&secured_message_context->application_secret_backup
                          .response_data_secret,
@@ -431,6 +434,11 @@ bool libspdm_create_update_session_data_key(void *spdm_secured_message_context,
         .response_data_sequence_number =
             secured_message_context->application_secret.response_data_sequence_number;
 
+        /* The backup now holds a complete snapshot of the old keys, so mark it valid before
+         * deriving the new ones. If a derivation below fails the caller can still roll back
+         * via libspdm_activate_update_session_data_key(..., use_new_key = false). */
+        secured_message_context->responder_backup_valid = true;
+
         status = libspdm_hkdf_expand(
             secured_message_context->base_hash_algo,
             secured_message_context->application_secret.response_data_secret,
@@ -455,8 +463,6 @@ bool libspdm_create_update_session_data_key(void *spdm_secured_message_context,
             return status;
         }
         secured_message_context->application_secret.response_data_sequence_number = 0;
-
-        secured_message_context->responder_backup_valid = true;
     } else {
         return false;
     }
--- a/unit_test/test_spdm_responder/key_update_ack.c
+++ b/unit_test/test_spdm_responder/key_update_ack.c
@@ -2056,6 +2056,74 @@ static void rsp_key_update_ack_case27(void **state)
                         m_rsp_secret_buffer, secured_message_context->hash_size);
 }
 
+/**
+ * Test 28: UPDATE_ALL_KEYS must not leave one direction updated while the other
+ * is not.
+ * Expected Behavior: when the key update cannot be completed for both
+ * directions the responder reports an error and leaves both directions on the
+ * old keys, with no backup left pending for a later commit.
+ **/
+static void rsp_key_update_ack_case28(void **state)
+{
+    libspdm_return_t status;
+    libspdm_test_context_t            *spdm_test_context;
+    libspdm_context_t                 *spdm_context;
+    uint32_t session_id;
+    libspdm_session_info_t            *session_info;
+    libspdm_secured_message_context_t *secured_message_context;
+
+    size_t response_size;
+    uint8_t response[LIBSPDM_MAX_SPDM_MSG_SIZE];
+
+    uint8_t m_req_secret_buffer[LIBSPDM_MAX_HASH_SIZE];
+    uint8_t m_rsp_secret_buffer[LIBSPDM_MAX_HASH_SIZE];
+    size_t original_aead_key_size;
+
+    spdm_test_context = *state;
+    spdm_context = spdm_test_context->spdm_context;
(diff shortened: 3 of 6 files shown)


library/spdm_responder_lib/libspdm_rsp_key_update_ack.c
```diff
@@ -127,6 +127,10 @@ libspdm_return_t libspdm_get_response_key_update(libspdm_context_t *spdm_context
             session_info->secured_message_context,
             LIBSPDM_KEY_UPDATE_ACTION_REQUESTER);
         if (!result) {
+            /* A partially applied update must not be left behind. */
```
steven-bellock: Remove comment.

library/spdm_responder_lib/libspdm_rsp_key_update_ack.c
```diff
@@ -153,6 +157,10 @@ libspdm_return_t libspdm_get_response_key_update(libspdm_context_t *spdm_context
...
```
steven-bellock: Remove comment.

library/spdm_responder_lib/libspdm_rsp_key_update_ack.c
```diff
@@ -166,6 +174,11 @@ libspdm_return_t libspdm_get_response_key_update(libspdm_context_t *spdm_context
             session_info->secured_message_context,
             LIBSPDM_KEY_UPDATE_ACTION_RESPONDER);
         if (!result) {
+            /* Roll back the Requester update created above so both directions
```
steven-bellock: Remove comment.

steven-bellock (changes requested):

alanhc:
All three comments removed in 338d9f1, which amends the second commit. No code changes, so the two-commit structure and the A.1-before-A.3 ordering are unchanged.

Re-verified after the amend: builds clean, uncrustify 0.78 passes on the changed file, and the suites are unchanged at test_spdm_responder (527), test_spdm_requester (742), test_spdm_common (35), test_spdm_secured_message (20), test_spdm_crypt (11).

The two questions from the description are still open whenever you want to weigh in: whether to defensively handle the failed response-direction commit arm (currently unreachable, so a rollback there would be untestable), and whether you would prefer the single transactional rewrite over these per-site fixes.

steven-bellock:
>The two questions from the description are still open whenever you want to weigh in: whether to defensively handle the failed response-direction commit arm (currently unreachable, so a rollback there would be untestable), and whether you would prefer the single transactional rewrite over these per-site fixes.

No and no.

steven-bellock (approved):

library/spdm_responder_lib/libspdm_rsp_key_update_ack.c
```diff
@@ -166,6 +172,9 @@ libspdm_return_t libspdm_get_response_key_update(libspdm_context_t *spdm_context
             session_info->secured_message_context,
             LIBSPDM_KEY_UPDATE_ACTION_RESPONDER);
         if (!result) {
+            libspdm_activate_update_session_data_key(
+                session_info->secured_message_context,
+                LIBSPDM_KEY_UPDATE_ACTION_REQUESTER, false);
```
LeonLinAtAMI, line 177: Could you please confirm whether here require another explicit rollback on LIBSPDM_KEY_UPDATE_ACTION_RESPONDER together?

unit_test/test_spdm_responder/key_update_ack.c
```diff
@@ -2056,6 +2056,74 @@ static void rsp_key_update_ack_case27(void **state)
...
+        m_req_secret_buffer, (uint8_t)(0xEE));
+
+    /* Make the AEAD key derivation fail: HKDF-Expand rejects an output longer
+     * than 255 * hash_size per RFC 5869 Section 2.3, and rejects it on the
+     * requested length alone, before writing to the output buffer. */
+    original_aead_key_size = secured_message_context->aead_key_size;
+    secured_message_context->aead_key_size = 255 * secured_message_context->hash_size + 1;
+
+    response_size = sizeof(response);
+    status = libspdm_get_response_key_update(spdm_context,
```
LeonLinAtAMI, line 2105: Could you please confirm whether the test covers the block you added lines from 175 to 177 for SPDM_KEY_UPDATE_OPERATIONS_UPDATE_ALL_KEYS? Thanks.

LeonLinAtAMI (changes requested):
Please check again and let me know if I misunderstood the rollback function of updating all keys.

jyao1:
@alanhc , would you please take a look at issue from copilot?

library/spdm_responder_lib/libspdm_rsp_key_update_ack.c
alanhc: Yes, you were right. The response data secret can already be overwritten by then. Fixed in c586a8e, which now rolls back both directions.

unit_test/test_spdm_responder/key_update_ack.c
alanhc: Good catch, it did not. The failure hit the first create. case28 now fails the second create via the key update callback, so this block is covered.

library/spdm_secured_message_lib/libspdm_secmes_session.c
```diff
@@ -380,6 +380,11 @@ bool libspdm_create_update_session_data_key(void *spdm_secured_message_context,
         .request_data_sequence_number =
             secured_message_context->application_secret.request_data_sequence_number;
 
+        /* The backup now holds a complete snapshot of the old keys, so mark it valid before
+         * deriving the new ones. If a derivation below fails the caller can still roll back
+         * via libspdm_activate_update_session_data_key(..., use_new_key = false). */
+        secured_message_context->requester_backup_valid = true;
```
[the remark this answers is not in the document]
alanhc: These arms already left a partially derived live key before this PR, because the derivation happens in place. A.1 only changes whether a backup is left marked valid. On the decode side, that lets TRY_DISCARD_KEY_UPDATE fall back to the old key. Keeping this PR scoped to the responder KEY_UPDATE path; the requester and encap callers can be a follow-up.

alanhc:
Thanks for the ping, and sorry for the delay. Updated in c586a8e (amends the A.3 commit; A.1 is unchanged apart from the trailer):

- Copilot #1 / @LeonLinAtAMI: when the response-direction create fails in UPDATE_ALL_KEYS, the response direction is now rolled back as well as the request direction, and the callback gets a matching request-direction DISCARD_UPDATE.
- Copilot #3 / @LeonLinAtAMI: case28 did not reach that arm, since the injected failure hit the first create. It now fails the second create through the key update callback. The test fails without the fix and passes with it.
- Copilot #2: replied inline. The other create() callers already left a half-derived key before this PR, so I'd like to keep them out of scope here.

All unit test suites pass (responder 527, requester 742, common 35, secured_message 20, crypt 11).

unit_test/test_spdm_secured_message/key_update.c
```diff
@@ -0,0 +1,215 @@
+/**
+ *  Copyright Notice:
+ *  Copyright 2026 DMTF. All rights reserved.
+ *  License: BSD 3-Clause License. For full text see link: https://github.com/DMTF/libspdm/blob/main/LICENSE.md
+ **/
+
+#include "spdm_unit_test.h"
+#include "internal/libspdm_secured_message_lib.h"
+
+#if LIBSPDM_AEAD_AES_256_GCM_SUPPORT
```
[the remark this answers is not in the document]
alanhc: Fixed in 423c6f9 (folded into the A.1 commit): the test and its runner are now guarded by `LIBSPDM_AEAD_AES_256_GCM_SUPPORT && LIBSPDM_SHA256_SUPPORT`.

library/spdm_responder_lib/libspdm_rsp_key_update_ack.c
```diff
@@ -127,6 +127,9 @@ libspdm_return_t libspdm_get_response_key_update(libspdm_context_t *spdm_context
             session_info->secured_message_context,
             LIBSPDM_KEY_UPDATE_ACTION_REQUESTER);
         if (!result) {
+            libspdm_activate_update_session_data_key(
+                session_info->secured_message_context,
+                LIBSPDM_KEY_UPDATE_ACTION_REQUESTER, false);
```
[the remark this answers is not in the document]
alanhc: Added rsp_key_update_ack_case29 (UPDATE_KEY) and case30 (first create of UPDATE_ALL_KEYS), which cover this line and line 159. Each one fails if its rollback is removed.

PR #3842, opened 2026-09-14
Cast `rsp_session_id` to `uint32_t` before shift

### LINKED ISSUE #3841: Signed shift overflow in `libspdm_generate_session_id`
`libspdm_generate_session_id` computes `(rsp_session_id << 16) | req_session_id`, but `rsp_session_id` is a `uint16_t` that promotes to `int`, so the shift overflows signed `int` for any value with the high bit set.

steven-bellock:
Fix #3841.


jyao1:
> libspdm_generate_session_id computes `(rsp_session_id << 16) | req_session_id`, but `rsp_session_id` is a uint16_t that promotes to int, so the shift overflows signed int for any value with the high bit set.

I am not clear about the problem statement.
I agree that `rsp_session_id` is a uint16_t that promotes to int, but how does the shift overflow signed int for any value with the high bit set?

E.g. `rsp_session_id` is 0xFF00. The most significant bit is 1. Then  `(rsp_session_id << 16)` is 0xFF00_0000 with sign extension.
What is the impact for the final result?

steven-bellock:
>What is the impact for the final result?

On all target architectures that we care about the result is the same. It is just technically undefined behavior, and doing this makes UBSAN and Coverity happy.

>The result of E1 << E2 is E1 left-shifted E2 bit positions; vacated bits are filled with zeros. If E1 has an unsigned type, the value of the result is E1 × 2^E2, reduced modulo one more than the maximum value representable in the result type. If E1 has a signed type and nonnegative value, and E1 × 2^E2 is representable in the result type, then that is the resulting value; **otherwise, the behavior is undefined.**

jyao1:
> > What is the impact for the final result?
> 
> On all target architectures that we care about the result is the same. It is just technically undefined behavior, and doing this makes UBSAN and Coverity happy.
> 
> > The result of E1 << E2 is E1 left-shifted E2 bit positions; vacated bits are filled with zeros. If E1 has an unsigned type, the value of the result is E1 × 2^E2, reduced modulo one more than the maximum value representable in the result type. If E1 has a signed type and nonnegative value, and E1 × 2^E2 is representable in the result type, then that is the resulting value; **otherwise, the behavior is undefined.**

If this rule is followed, then there is nothing wrong in previous code.
uint16_t is promoted to int, it is **signed type** and **nonnegative value**, and **E1 x 2^E2 is representable** since no bit is lost. The behavior is defined.

Please help me understand why it becomes *undefined*.

steven-bellock:
>E1 x 2^E2 is representable since no bit is lost.

Let `E1 == 32_768 (0x8000)` and `E2 == 16`. Then `32_768 ⋅ 65_536 = 2_147_483_648`, and `2_147_483_648` is outside of a `int`'s maximum positive range, which is `2_147_483_647`. Hence, it is not representable and it is undefined behavior.

jyao1:
OK. that is fair.

PR #3846, opened 2026-09-15
TPM: Fall back to default NV chunk size when `NV_BUFFER_MAX` query fails

Jino26:
Some TPMs with older firmware do not report `TPM2_PT_NV_BUFFER_MAX` and fail the `TPM2_GetCapability` query. `libspdm_tpm_read_nv()` bailed out on that error before reaching the default-size fallback, so no certificate chain could be read from NV.

Use the default chunk size whenever the query fails or returns a different property, instead of failing the whole NV read.


manojkiraneda:
> Some TPMs do not report TPM2_PT_NV_BUFFER_MAX and fail the TPM2_GetCapability query with TPM_RC_VALUE. 

@Jino26  - I think this should be supported in the latest TPM hardware/firmware. what type of tpms are you talking about here ? did you observe this on a simulator or hardware with older TPM firmware ? can you be a bit more explicit ?

Jino26:
> > Some TPMs do not report TPM2_PT_NV_BUFFER_MAX and fail the TPM2_GetCapability query with TPM_RC_VALUE.
> 
> @Jino26 - I think this should be supported in the latest TPM hardware/firmware. what type of tpms are you talking about here ? did you observe this on a simulator or hardware with older TPM firmware ? can you be a bit more explicit ?

Yeah right, this property is not present on the simulator which is running an older TPM firmware and also doesn't appear in the results of `tpm2_getcap properties-fixed`.

Jino26:
> > > Some TPMs do not report TPM2_PT_NV_BUFFER_MAX and fail the TPM2_GetCapability query with TPM_RC_VALUE.
> > 
> > 
> > @Jino26 - I think this should be supported in the latest TPM hardware/firmware. what type of tpms are you talking about here ? did you observe this on a simulator or hardware with older TPM firmware ? can you be a bit more explicit ?
> 
> Yeah right, this property is not present on the simulator which is running an older TPM firmware and also doesn't appear in the results of `tpm2_getcap properties-fixed`.

Updated commit message also to highlight this.

itsManjeet (approved):

steven-bellock (approved):

PR #3861, opened 2026-09-16
Introduce continuous integration document

### LINKED ISSUE #3823: `DISABLE_TESTS=1` leg in `build.yml` has no effect
[`build.yml`](https://github.com/DMTF/libspdm/blob/main/.github/workflows/build.yml) passes `matrix.configurations` through `CFLAGS`, but `DISABLE_TESTS` is a CMake variable rather than a preprocessor macro, so the `-DDISABLE_TESTS=1` leg never reaches CMake.

steven-bellock:
This includes the document and the .yml file that implements Tier 1.


steven-bellock:
This targets Tier 1 testing. See https://github.com/DMTF/libspdm/actions/runs/35154795529 for an example of what things look like. This would eventually replace the `build.yml` file.

steven-bellock:
And the reproducer looks like
```bash
#!/usr/bin/env bash
# Reproduces: linux-x64-gcc-mbedtls-debug
# Run:        https://github.com/steven-bellock/libspdm/actions/runs/35152349689
# Runner:     ubuntu-latest (image ubuntu24 20260907.300.1)
# OS:         Ubuntu 24.04.5 LTS
# Kernel:     Linux 6.17.0-1022-azure x86_64
# Compiler:   GNU 13.3.0
# NOTE: this restores the source tree, not the toolchain. If your
#       compiler differs from the one above, a non-reproduction is
#       inconclusive rather than a refutation.
set -euo pipefail

git clone https://github.com/steven-bellock/libspdm.git libspdm-repro
cd libspdm-repro
git checkout d8319182fffd691d108421ef6b34ff6a9513e717
git submodule update --init --recursive
# pinned: os_stub/mbedtlslib/mbedtls e185d7fd85499c8ce5ca2a54f5cf8fe7dbe3f8df
# pinned: os_stub/openssllib/openssl 67b5686b4419b4cb8caa502711c41815f5279751
# pinned: unit_test/cmockalib/cmocka a01cc69ee9536f90e57c61a198f2d1944d3d4313

export CFLAGS='-DLIBSPDM_CHECK_MACRO=1 -DLIBSPDM_DEBUG_LIBSPDM_ASSERT_CONFIG=3 '
cmake -B build -DARCH=x64 -DTOOLCHAIN=GCC -DTARGET=Debug -DCRYPTO=mbedtls 
cmake --build build --target copy_sample_key
cmake --build build --parallel "$(getconf _NPROCESSORS_ONLN)"

cd build/bin
if [ -x ./test_spdm_requester ]; then ./test_spdm_requester; fi
if [ -x ./test_spdm_responder ]; then ./test_spdm_responder; fi
if [ -x ./test_spdm_crypt ]; then ./test_spdm_crypt; fi
if [ -x ./test_spdm_secured_message ]; then ./test_spdm_secured_message; fi
if [ -x ./test_spdm_common ]; then ./test_spdm_common; fi
if [ -x ./test_crypt ]; then ./test_crypt; fi
```

jyao1:
I notice that there is still `TBD` in the doc. Is current version draft or final?

steven-bellock:
It is a work-in-progress. I would like to get Tier 1 done, then Tier 2, etc. This pull request targets Tier 1.

doc/internal/continuous_integration.md
```diff
@@ -0,0 +1,175 @@
...
+alter coarse macros present in `spdm_lib_config.h` that greatly alter the size and behavior of the
+library. Examples of such macros include `LIBSPDM_FIPS_MODE` and
+`LIBSPDM_RECORD_TRANSCRIPT_DATA_SUPPORT`.
+
+All tier 1 tests must pass before a pull request that targets production code or tests is merged
+into `main` or any of the release branches.
+
+For execution time, when all tests pass, the total test time should be less than 30 minutes.
+
+### Tier 2 - Nightly Coverage
```
steven-bellock: `Nightly` should really be `Daily`, since my night might be @jyao1's day.

jyao1:
Please switch to draft if it is still WIP, and move the ready-for-review after it is done.

I do not think there is need to split Tier 1 or Tier 2. Eventually, we need get it done for all.

steven-bellock:
This pull request is now limited to Tier 1 testing.

PR #3862, opened 2026-09-16
Add message logging to more Requester messages

### LINKED ISSUE #1656: Add message logging for all (relevant) messages
Message logging was enabled for `VCA` and `GET_MEASUREMENTS / MEASUREMENTS`, its primary use case. But it should also be enabled for the rest of the (relevant) messages.

steven-bellock:
Fix #1656.


jyao1:
Is this complete?

steven-bellock:
> Is this complete?

Yes, this will close https://github.com/DMTF/libspdm/pull/3862.

PR #3869, opened 2026-09-17
Improve fuzzing harness and build

### LINKED ISSUE #3868: Fuzzing harness and build hide several classes of defects
The fuzzing harness allocates `libspdm_get_max_buffer_size()` bytes but hands the code under test a shorter length, and `libspdm_alignment_size` then rounds that declared length up, so a read past the end of a message lands inside the allocation and AddressSanitizer cannot see it. `libspdm_debug_assert` prints the failing expression to stdout, which is fully buffered when redirected, and libFuzzer's exit callback skips the stdio flush, so the assertion text is absent from the crash log. `LIBSPDM_DEBUG_LIBSPDM_ASSERT_CONFIG` defaults to `LIBSPDM_DEBUG_LIBSPDM_ASSERT_DEADLOOP`, so an assertion reached while fuzzing spins forever and is reported as a timeout rather than an assertion. The `LIBFUZZER` toolchain compiles with `-fsanitize=fuzzer,address` and never `-fsanitize=undefined`, so undefined behavior is not detected at all.

steven-bellock:
Fix #3868.


steven-bellock:
CIFuzz is failing due to https://github.com/DMTF/libspdm/pull/3871. So once that goes in then this pull request can progress.

PR #3875, opened 2026-09-17
Check core library print format strings

### LINKED ISSUE #2576: Enable checking of `printf` format and specifiers
From #2573.

`LIBSPDM_DEBUG` calls into `libspdm_debug_print` which can mask `printf` issues. Add a build to CI/CD that can check these issues.

steven-bellock:
Fix #2576.


steven-bellock:
Although the attribute is in `include/internal/libspdm_common_lib.h`, it looks like the unit tests bring that in as well.

steven-bellock:
@jyao1 let's discuss this at the 21-Sep-2026 meeting before merging.

PR #3880, opened 2026-09-18
Pass SPDM context and session ID to HAL functions

### LINKED ISSUE #3879: Update appropriate `hal/` functions with SPDM context or session ID
Implement https://github.com/DMTF/libspdm/issues/3340#issuecomment-5732644116.

steven-bellock:
Fix #3879.


include/hal/library/responder/setcertlib.h
```diff
@@ -16,11 +16,13 @@
...
  *
  * @param[in]  spdm_context  A pointer to the SPDM context.
+ * @param[in]  session_id    Indicates the session in which the request was received.
+ *                           If NULL then the request was not received in a session.
  *
  * @retval  true   It is in a trusted environment.
  * @retval  false  It is not in a trusted environment.
  **/
-extern bool libspdm_is_in_trusted_environment(void *spdm_context);
+extern bool libspdm_is_in_trusted_environment(void *spdm_context, const uint32_t *session_id);
```
jyao1, line 25: I am not sure why this need session_id.

The rule from SPDM spec is in trusted_env or in a session. A trusted env is irrelevant to a session.

steven-bellock: The philosophy is that if the `spdm_context` and/or `session_id` is available, then send it to the Integrator. They are free to ignore it.

jyao1: ok

PR #3883, opened 2026-09-18
Exempt chunk messages from mutual authentication enforcement

### LINKED ISSUE #3881: Responder rejects `CHUNK_GET` during mutual authentication
When the Responder requests mutual authentication, in `CHALLENGE_AUTH` or in `KEY_EXCHANGE_RSP`, it requires the next request to be `GET_ENCAPSULATED_REQUEST` or `FINISH` and exempts only `GET_VERSION`. If that response is larger than `DataTransferSize` it is delivered with `ERROR(LargeResponse)` followed by `CHUNK_GET`, which is then rejected with `ERROR(UnexpectedRequest)`, so the Requester can neither retrieve the response nor advance the encapsulated flow.

This was introduced by 0ebbf9aa and breaks the spdm-emu Chunking and PQC workflows, both of which fail with `LIBSPDM_STATUS_ERROR_PEER` during authentication. The per-channel encapsulated flow check added by the same commit already exempts `CHUNK_GET` and `CHUNK_SEND`, and the two mutual authentication checks should do the same.

steven-bellock:
Fix #3881.


steven-bellock:
This fixes failures in https://github.com/DMTF/spdm-emu/pull/523.

PR #3885, opened 2026-09-18
Apply legality checks to chunked requests

### LINKED ISSUE #3882: Ordering checks are not applied to requests delivered by `CHUNK_SEND`
`libspdm_build_response` rejects out-of-order requests before dispatching the response handler, but `libspdm_get_response_chunk_send` dispatches the assembled large request directly, so that request is never checked. The per-channel encapsulated flow check already exempts `CHUNK_SEND`, so a Requester can wrap a request that should have been rejected with `ERROR(RequestInFlight)` and have it processed instead.

steven-bellock:
Fix #3882.


jyao1:
Could you please make it clear, that which condition is being handled?
The description of https://github.com/DMTF/libspdm/issues/3882 is vague to me.

steven-bellock:
0ebbf9a introduced more request message checks in `libspdm_build_response()`, roughly: https://github.com/DMTF/libspdm/blob/08a155c3a3e1b246f96cfe840716d60c6285ce2a/library/spdm_responder_lib/libspdm_rsp_receive_send.c#L606-L750 However those checks get bypassed if the request message is sent via `CHUNK_SEND`: https://github.com/DMTF/libspdm/blob/08a155c3a3e1b246f96cfe840716d60c6285ce2a/library/spdm_responder_lib/libspdm_rsp_chunk_send_ack.c#L259-L277 This pull request fixes that.

jyao1:
Thanks.

But the term "out of order" is very confusing to me.
Do you want to support any arbitrary "out of order" request? or just handle mut-auth?
I feel that is latter. If so, the name "libspdm_is_request_out_of_order()" is misleading.

steven-bellock:
@jyao1 please re-review.

include/internal/libspdm_responder_lib.h
```diff
@@ -965,4 +965,22 @@ libspdm_return_t libspdm_get_response_endpoint_info(libspdm_context_t *spdm_cont
...
+ *                       this is the assembled request rather than CHUNK_SEND itself.
+ * @param  error_code    On return, the SPDM error code to respond with.
+ *
+ * @retval true   The request is out of order and error_code has been set.
+ * @retval false  The request may be dispatched.
+ **/
+bool libspdm_is_request_legal_flow(libspdm_context_t *spdm_context,
+                                   const uint32_t *session_id,
+                                   uint8_t request_code,
+                                   uint8_t *error_code);
```
jyao1: the name is still confusing, because libspdm_is_request_legal_flow():
* true (is legal) - send error code if it is legal ?
* false (is not legal) - dispatch if it is illegal ?

Maybe just call it:
libspdm_is_chunk_request_unexpected_for_mut_auth_encap()?

steven-bellock: >Maybe just call it:
libspdm_is_chunk_request_unexpected_for_mut_auth_encap()?

It is used for more than just chunked requests, but you are right that it is still confusing.
How about
1. `libspdm_is_request_unexpected_per_flow()`
2. `libspdm_is_request_illegal_per_flow()`

?

jyao1: I just think it should be limited to mutual authentication and encap.
I prefer `libspdm_is_request_unexpected_for_mut_auth_encap()`

steven-bellock: >I prefer libspdm_is_request_unexpected_for_mut_auth_encap()

It is done.

jyao1 (approved):

PR #3886, opened 2026-09-18
Cherry pick bug fixes from `main` / `4.0` to `3.8.3`

steven-bellock:


steven-bellock:
~~https://github.com/steven-bellock/libspdm/actions/runs/35397808389 shows these commits passing on the release branch.~~

Checks now work on https://github.com/DMTF/libspdm, so using those instead.

jyao1:
Question: What is criteria to cherry pick bug fix for a released branch?
Previously, I think it is only applicable to security bug or very important functional bug, but definitely not all bugs.

steven-bellock:
>What is criteria to cherry pick bug fix for a released branch?

Simple, straight forward bug fixes that our friends still using 3.8 would appreciate.

jyao1:
I suggest we discuss the criteria, get it approved, and document it.
I prefer we do the work based on priority / severity / importance, but not simplicity.

jyao1:
The recent regression issues (#3946, #3938, #3935) reminded me that it might be hard just cherry pick some of fixes. There might be later fix on top of that.

PR #3891, opened 2026-09-19
Sweep the hash algorithms in `test_spdm_crypt`

steven-bellock:
The hash, HMAC, and HKDF entry points in spdm_crypt_lib dispatch on base_hash_algo, but the unit tests only ever select SHA-256, SHA-384, or SHA-512, so the SHA3 and SM3 arms were never taken. libspdm_crypt_hash.c reported 54 of 198 lines covered and libspdm_crypt_hmac.c 4 of 148.

Sweep every algorithm the specification defines. A cryptlib may stub an algorithm that spdm_lib_config.h enables, as the Mbed TLS backend does for SM3, so the sweep does not demand success. It probes with the one-shot entry point and requires the incremental entry points to agree with it.


unit_test/test_spdm_crypt/test_spdm_crypt.c
```diff
@@ -1369,6 +1369,238 @@ static void libspdm_test_crypt_ecdsa_palindrome(void **state)
     }
 }
 
+/* The hash, HMAC, and HKDF entry points in spdm_crypt_lib are switch dispatchers over
+ * base_hash_algo. The other unit tests only ever select SHA-256, SHA-384, or SHA-512, so the
+ * SHA3 and SM3 arms are never taken. These tests sweep every algorithm the specification
+ * defines.
```
steven-bellock: This can be removed.

PR #3894, opened 2026-09-19
Do not start watchdog when `HeartbeatPeriod` is zero

### LINKED ISSUE #3015: What should be the session timeout value when HeartbeatPeriod is set to 0?
Spec:
1. `The Responder shall terminate the session if session traffic is not received in twice HeartbeatPeriod.`
2. `HeartbeatPeriod: The value of this field shall be zero if Heartbeat is not supported by one of the endpoints. Otherwise, the value shall be in units of seconds. Zero is a legal value if Heartbeat is supported, but means that a heartbeat is not desired on this session.`

What should be the correct handling for session timeout in this case?
Thanks!

steven-bellock:
Fix #3015.


jyao1:
Please resolve conflict!

jyao1 (approved):

PR #3900, opened 2026-09-21
Require `CERT_PARSE_SUPPORT` when `SEND_GET_CERTIFICATE_SUPPORT` is enabled

### LINKED ISSUE #3897: If `LIBSPDM_SEND_GET_CERTIFICATE_SUPPORT` is set then `LIBSPDM_CERT_PARSE_SUPPORT` must also be set
Else you get a 
```
C4013: 'libspdm_get_pqc_leaf_cert_public_key_from_cert_chain' undefined
```
This should be added to `LIBSPDM_CHECK_MACRO`.

steven-bellock:
Fix #3897.


jyao1:
Really? If it is NOT required before but required now, should we just fix the code?

PR #3915, opened 2026-09-22
Replace static array with switch statement

### LINKED ISSUE #3913: Replace `libspdm_get_code_str()` array with `switch / case`
Related issue / pull request: https://github.com/DMTF/libspdm/pull/3755.

`libspdm_get_code_str()` use a large and `static` array to return a string based on an integer index. This can be converted to a `switch / case` statement like in #3755.

https://github.com/DMTF/libspdm/blob/60ad78ff614e855ef76a54e7995a9c95c27e7fb8/library/spdm_common_lib/libspdm_com_support.c#L15-L23

steven-bellock:
Fix #3913.


LeonLinAtAMI (approved):
Looks good to me.

